PR #6, opened 2020-05-08
fix path to the neovim node executable

Petka17:
it might save some time for people who will decide to setup nvm node_host_prog


ChristianChiarulli:
lgtm

PR #19, opened 2020-06-01
Porting config over to Lua

mattcoker:


mattcoker:
I'm good with that, definitely take a look at the whichkey support, and how its configured in-line with the keybindings so that there's a single exposed API for binding keys and mapping them in whichkey. I want to make sure that's the way we want to go about it, and that we don't hit some crazy roadblocks later with this strategy.

From a lua standpoint, are there better ways of formatting/organizing the code? I have essentially no experience other than what I've seen in the last week on stackoverflow, and in the repo that I based the `lua/c` and `lua/l` directories on.

ChristianChiarulli:
Cool stuff, I'll test it out soon.

PR #21, opened 2020-06-02
Fix setup file airline.vim error

teepobharu:


ChristianChiarulli:
No idea how that got there, thanks!

PR #25, opened 2020-06-10
feat: can move line selection

AceMouty:
users can now move a single line or an entire block of text / code
while in visual mode

simply select what you need moved and then use
shift + k to move the selection up
shift + j to move the selection down


ChristianChiarulli:
🆒

PR #42, opened 2020-08-07
This will help macOS install not fail

vladdoster:
Needed these 2 packages for things to install via the install script

### DIFF
--- a/utils/install.sh
+++ b/utils/install.sh
@@ -4,7 +4,9 @@ set -o nounset    # error when referencing undefined variable
 set -o errexit    # exit when command fails
 
 installnodemac() { \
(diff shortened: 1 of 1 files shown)


utils/install.sh
```diff
@@ -4,7 +4,9 @@ set -o nounset    # error when referencing undefined variable
 set -o errexit    # exit when command fails
 
 installnodemac() { \
+  brew install lua
   brew install node
+  brew install yarm
```
ChristianChiarulli, line 9: wait, should this be yarn?

vladdoster: lmao, whoops. You're 100% right.

PR #53, opened 2020-09-08
fix not-existing Floaterm wintype option

sobelek:
Wintype type 'popup' is probably deprecated and no longer exists in Floaterm according to docs here:
https://github.com/voldikss/vim-floaterm/blob/master/doc/floaterm.txt


ChristianChiarulli (approved):
I noticed this stopped working but didn't know why. Thanks

sobelek:
No problem, also thanks for a great job with the whole project :)

PR #54, opened 2020-09-10
Fix Pynvim crashing on OSX

stealthanthrax:
Using using the script in the description, Pynvim install crashes on the default python config. 
Adding a simple `--user` flag fixes it.

`Add --user flag`


ChristianChiarulli (approved):
:cool:

PR #63, opened 2020-09-26
dedup & (fix ?) keybindings.json

Twiggeh:
## Deleted
Reason: Overrides `shift ctrl e` & `ctrl e`, which are defined right below the deleted portion
```
  {
    "key": "shift+ctrl+e",
    "command": "actions.findWithSelection"
  },
  {
    "key": "ctrl+e",
    "command": "-actions.findWithSelection"
  },
```

Reason: Duplicate
```
  {
    "key": "shit+tab",
    "command": "selectPrevSuggestion",
    "when": "editorTextFocus && suggestWidgetMultipleSuggestions && suggestWidgetVisible"
  },
```

### DIFF
--- a/utils/vscode_config/keybindings.json
+++ b/utils/vscode_config/keybindings.json
@@ -1,13 +1,5 @@
 [
   {
-    "key": "shift+ctrl+e",
-    "command": "actions.findWithSelection"
-  },
-  {
-    "key": "ctrl+e",
-    "command": "-actions.findWithSelection"
-  },
-  {
     "key": "ctrl+e",
     "command": "workbench.view.explorer"
   },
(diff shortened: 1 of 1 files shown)


ChristianChiarulli:
I could be wrong, I'm not super familiar with VSCode keybindings, but the removals of "ctrl+e" and "ctrl+shift+e" are just to make sure they aren't ever triggered. I'm pretty sure you can bind "ctrl+e" for instance to a lot of things depending on 
the context.

Twiggeh:
Why would you unbind them though ? 
`Ctrl shift e` opens the file explorer by default, but you are binding `find with selection` to it, which is `ctrl f` per default. 

Also in your whichkey config you actually just toggle the sidebar visibility, not the explorer itself :D
There is no reason to have identical keybindings ...

ChristianChiarulli:
fixed

PR #80, opened 2020-11-16
Update keybindings.json

haininhhoang94:
Fix the error that vim which key doesn't work at welcome screen (or anything that doesn't open by Neovim backend). This error was mostly cause by the way the code is constructed (whichkey.show is handle inside Neovim, not vscode)


ChristianChiarulli (approved):
:cool:  I wanted this

PR #83, opened 2020-11-27
faild to run npm as sudo

GabrielGCardoso:
Screenshot from 2020-11-25 23-41-39

I had this problem when I tried to install with nodejs from asdf, and when I removed sudo it worked normal.


GabrielGCardoso:
can be useful if someone else also uses asdf https://github.com/asdf-vm/asdf

ChristianChiarulli:
This whole will need to be updated since I am moving away from CoC in favor of native LSP

PR #87, opened 2020-12-01
run updates every week automatically on startup

danielfalbo:
From https://gist.github.com/kkoomen/68319b08ab843ce67cf7b282b0b2fd24 but also executes CocUpdate and other stuff, you may want to modify it a little bit.

### DIFF
--- a/vim-plug/plugins.vim
+++ b/vim-plug/plugins.vim
@@ -178,3 +178,26 @@ autocmd VimEnter *
   \  if len(filter(values(g:plugs), '!isdirectory(v:val.dir)'))
   \|   PlugInstall --sync | q
   \| endif
+
+" Run updates every week automatically on startup
+function! OnVimEnter() abort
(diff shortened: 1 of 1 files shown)


vim-plug/plugins.vim
```diff
@@ -178,3 +178,26 @@ autocmd VimEnter *
...
+    endif
+
+    let l:this_week = strftime('%Y_%V')
+    let l:contents = readfile(l:filename)
+    if index(l:contents, l:this_week) < 0
+      call execute('PlugUpgrade')
+      call execute('PlugUpdate')
+      call execute('CocUpdate')
+      call execute('UpdateRemotePlugins')
+      call execute('PlugClean')
```
danielfalbo, line 197: I do not know why but PlugClean throws errors in this scenario, better to remove line 197.

ChristianChiarulli:
removing CoC

PR #95, opened 2020-12-15
enable auto install coc-extensions without npm

codehia:
replace npm installation of coc-extensions with g:coc_global_extension
replace coc-python with coc-pyright


ChristianChiarulli:
removing coc in favor of native lsp

Issue #96, opened 2020-12-15
CocSearch not working


aniedzwiedz:
CocSearch command itself is working fine, 
problem is with:
`nnoremap <leader>? CocSearch <C-R>=expand("<cword>")<CR><CR>`

> space + '?' ends up with cutting word and =expand("<cword>") in 'commands' 

![nvim_CocSearch](https://user-images.githubusercontent.com/34684295/102245398-cbb1b080-3efd-11eb-87e9-1c80ff787d80.gif)

eddyekofo94:
Getting the same error @aniedzwiedz any solutions?

ChristianChiarulli:
No more CoC. I recommend Telescope for searching.

PR #111, opened 2021-02-07
add missing colon makes CocSearch working with which-key

aniedzwiedz:
fix for:
CocSearch not working #96


ChristianChiarulli:
🆒

PR #123, opened 2021-03-12
Added VimTeX plugin for LaTeX support

fanmih:
Intellisense for LaTeX currently does not work (because this plugin does not use the language server protocol? need to check it out).
Syntax highlighting, folding, latex compile, 2 way sync with pdf viewer (zathura) works.
Added also some basic Which-key mappings under the "L". Will add more in a future commit.

### DIFF
--- a/init.vim
+++ b/init.vim
@@ -47,6 +47,7 @@ else
   source ~/.config/nvim/plug-config/lightbulb.vim
   source ~/.config/nvim/lua/lsp-wrapper.vim
   source ~/.config/nvim/plug-config/lsp-config.vim
+  source ~/.config/nvim/plug-config/vimtex.vim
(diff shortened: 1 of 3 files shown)


ChristianChiarulli (approved):
🆒

arxra:
Suggestion, prompt the user during installation for zathura might lead to a better experience, or at least a notice

AngelBerihuete:
Hi @fanmih, 
Is it possible to set your plug-config/vimtex.vim configuration in terms of the LunarVim i.e., `.config/nvim/lua/lv-vimtex/init.lua` or something like this. I know nothing about lua language, and it is difficult to me understand how to tell vimtex to use zathura and other stuffs.

PR #128, opened 2021-03-13
Feature/extra settings

jordisantamaria:
I think this options are useful for everyone, how do you see?


jordisantamaria:
Sorry, didn't notice It's also overriding from the other PR, I wish It's fine...

PR #132, opened 2021-03-14
lua keybindings

martinsione:
I added a helper function for setting up the keybindings, `{noremap = true, silent = true}` are set by default so code is so much cleaner but you still can overwrite them and set them to false by passing them as a table like I did with `{expr = true}`.
You can now remove the block of code you copy for tab-completion in nv-compe and tab-completion will work just fine, I tested all the keybindings and they work.
I have my [nvim config](https://github.com/martinsione/dotfiles/tree/master/src/.config/nvim) almost completely in lua so if you wanna take a look you are more than welcome.
Hope you find it useful.


smikeblog:
maybe move which-key to this project 
`https://github.com/AckslD/nvim-whichkey-setup.lua.git`

PR #134, opened 2021-03-15
NVCode logo proposal

gauthierc:


tarkin88:
Can you share some screenshots on description to easy visualize?

PR #161, opened 2021-03-25
added install tree-sitter-cli

ivano9:
Added install tree-sitter-cli


ChristianChiarulli:
Not sure what I need this for, will have to do some research

ivano9:
When i installed it using the method with curl in a Ubuntu clean, and then open the nvim, i got an error telling that need that cli for run some stuff of tree sitter

ChristianChiarulli:
@ivano9 Ok, fair enough 🐨

PR #169, opened 2021-03-26
[LSP] clangd - Add support for C/C++ with clangd.

jameswalmsley:
I'm using this as my daily driver for C/C++ LSP.
It expects to find a compile-commands.json file in your root project folder.

It also picks up extra project specific settings from a .clangd file as described by the clangd docs.


ChristianChiarulli:
Cool 🌵

PR #170, opened 2021-03-26
[TELESCOPE] Support smart_send_to_qflist()

jameswalmsley:


jameswalmsley:
@ChristianChiarulli I add this, so that if you have a list in telescope, you can press C-q and it will put all items in the quicklist.
If you have selected some specific items (Pressing TAB on those items), then only those items will be put into the quickfix list.

ChristianChiarulli (approved):
Awesome 🪕

PR #175, opened 2021-03-27
Fix for the popup with max preload size is low

seocamo:
Fix for the popup with max file preload size is to low, i dig into the code of sumneko and you need to add maxPreload = 10000 to workspace and the number is in kb


ChristianChiarulli:
nice, thanks 🥍

PR #179, opened 2021-03-28
Add display name for key ' '(<Space>) in whichkey

chaeing:
Key for ' '(<Space>) is displayed on the panel as 'SPC' by default.
However, it is displayed as blank instead of 'SPC'
when assigning 'which_key_display_names' variable without specifiying it.


ChristianChiarulli (approved):
cool 🦒

PR #183, opened 2021-03-28
[LSP] latex - Add support for latex via LspInstall

fanmih:
Basic LaTeX support via LSP (installs TexLab)


ChristianChiarulli (approved):
nice 🏞️

PR #188, opened 2021-03-29
Define shell + php lsp + dart lsp + flutter + relative number

BiserStoilov:
Define shell on floaterm (bash, zsh, etc.)
Add php-lsp support
Add dart-lsp support
Add Flutter support
Ser relative number


BiserStoilov:
For Dart + Flutter add those plugins:
  use 'dart-lang/dart-vim-plugin'
  use 'thosakwe/vim-flutter'

ChristianChiarulli (approved):
great stuff 🏒

PR #190, opened 2021-03-29
Prototype global install of nvcode.

jameswalmsley:
I got a prototype of a global install config working.
This allows the user to use a standard nvim config under ~/.config/nvim when running `nvim`.

Or you can run nv and it will use the /usr/local/share/nvcode/init.lua config.

I managed to get this to install all packages into a separate folder, without affecting the users local config.

Just wanted to get your thoughts on this...

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,28 @@ LunarVim provides neovim configuration files that take advantage of
 tree-sitter and language server protocol. The configuration is written
 in lua.
 
+# Global Install
+
+To install globally:
+
+```
+git clone https://github.com/jameswalmsley/nvcode.git -b global-install nvcode.global
+cd nvcode.global/utils/bin
+sudo ./nvcode-upgrade.sh -g
+```
+
+This install a copy of the nvcode config into `/usr/local/share/nvcode/nvim`, and installs the nv,
+starter script to `/usr/local/bin`
+
+The script will use sudo to run nvim as the calling user to sync plugins to `~/.local/share/nvcode/data/nvim`.
+
+It leaves nvim open, simply exit nvim when finished. (Couldn't get headless nvim to not quit early on packer).
+
+Once completed you should be able to run nv without interfering with the main .config/nvim config.
+
+NOTE.. This global install will use `.config/nvcode/nvim`
+
+
 ## Why do I want tree-sitter and LSP?
 
 - Normally, an editor uses regular expression parsing for things like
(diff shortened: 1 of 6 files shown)


ChristianChiarulli:
Awesome! I will have to take a closer look at this during a stream.

smikeblog:
just to check try to fire nvim  as ``` nvim - u ~/Documents/init.vim ``` where init vim is basic settings from old code or onle set number as an example. It seems that you'll have dashboard and galaxylane frame.

jameswalmsley:
Hi Mike, I'll upload a readme on how to use this.
I also found this. I had to modify the init.lua to remove the default config and pack paths. Then it only picks up the global configuration.

There maybe some cases that aren't quite perfect.. still researching it a bit.

smikeblog:
I've solved this issue. If you're interested I can show my solution.

smikeblog:
Hi, James.
In your script to respect linux rules it's better to clone repo not in 
```
his install a copy of the nvcode config into `/usr/local/share/nvcode/nvim`, and installs the nv,
starter script `/usr/local/bin` 
```
but  in /opt dir  and make links to `/usr/bin` or `/usr/local/bin`. As an example look at firefox browser  with
`which firefox`.

jameswalmsley:
I've updated this quite a bit now.

In summary:

  * Default to /opt/nvcode install dir.
    - User can specify dir with e.g. PREFIX=/usr/local/share nvcode-upgrade.sh -g
  
  * PackerSync is not done during install (needs to be done on a per-user basis).
  
  * PackerSync is automatic: 
    - /opt/nvcode dir has a .packer_sync which is touched on install / upgrade.
    - If users local config .packer_sync is older than the installed version, then nv will detect to call PackerSync
    
  * Base configuration is installed:
    - Base nvcode-settings.lua is copied to $HOME/.config/nvcode 

  * Copied deps / extra package installation from original installer script.
    
  * nvcode-upgrade.sh (without -g) does a normal local install (basically the installer script).
  
  Issues / further work:
  
    * Config is under ~/.config/nvcode/nvim/nv-settings.lua (perhaps an optional include for config_dir/../nv-settings.lua??)
    * Current issue with galaxyline packer (see #244)
    * Turn on extra packages.
    * Test on other platforms, e.g. OSX.

ChristianChiarulli:
closing for now, still plan on adding this feature, but this may be stale at this point.

PR #191, opened 2021-03-30
add set shell=/bin/bash in init.lua

midnightexigent:
some people (myself included) use fish shell as their default. which doesn't really play well in neovim

in my old init.vim I used to have a `set shell=/bin/bash`

I don't know if there is an equivalent in pure lua, but the dirty `vim.cmd('set shell=/bin/bash')` worked

feel free to change it if there is a more "lua" way to do it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ eclipse.jdt.ls/
 .debuggers/
 spell/
 nv-settings.lua
+lv-settings.lua


ChristianChiarulli:
I have a new Option worked in for shell, could you possible find a way to use that?

midnightexigent:
Much cleaner, thanks. 

There is a `vim.o.shell = O.shell` that needs to go somewhere for the setting to actually take effect

seocamo:
shell is remove in nvim and do nothing, the shell that is used is the one you are running nvim from.

nvim doc following
:sh[ell]		Removed. |vim-differences|

midnightexigent:
> shell is remove in nvim and do nothing, the shell that is used is the one you are running nvim from.

That's not correct

seocamo:
this is correct, open neovim type :help shell see for your self
also if you really want to use fish(zsh can do the same thing) but then the right is to start nvim with a real shell like bash -e nvim
you can make a script for this name it nvim and add it to the .local/bin and use that. the shell=/bin/bash force all the users of he's config to use bash if it is in vi mode(old fallback)
there is a reason that Neovim remove this(outside of fallback)
next time, please check the manual before you call someone a lier here ok.
ps. /bin/sh need to be a posix compatible for must thing to work right, if you use it because DT joke that it want to use fish, then this was a joke, because he know that /bin/sh need to be a posix shell.

midnightexigent:
you're talking about the `shell` command, as in `:shell ...`

I am talking about the `sh[ell]` *option*. `:help options` to learn about options or `:set shell?` to see the current value. 

also `:help option-list` to see all possible options, if you scroll down, you'll see that `shell` is actually there, and it *is* used for running external commands. Its initial value seems to be sourced from`$SHELL` which is equal to whatever is in the `/etc/passwd` for the currently logged-in user. 

I just suggested setting the `shell` option to the `O.shell` nvcode setting via `vim.o.shell = O.shell` so idk what's upsetting you?


nExT tImE, pay more attention and understand what is being discussed before throwing shade, being rude, or giving any input.


> 
> 
> this is correct, open neovim type :help shell see for your self
> also if you really want to use fish(zsh can do the same thing) but then the right is to start nvim with a real shell like bash -e nvim
> you can make a script for this name it nvim and add it to the .local/bin and use that. the shell=/bin/bash force all the users of he's config to use bash if it is in vi mode(old fallback)
> there is a reason that Neovim remove this(outside of fallback)
> next time, please check the manual before you call someone a lier here ok.
> ps. /bin/sh need to be a posix compatible for must thing to work right, if you use it because DT joke that it want to use fish, then this was a joke, because he know that /bin/sh need to be a posix shell.

PR #210, opened 2021-04-01
Add more space between icon and description in dashboard

DominikMarcinowski:
Only small thing, but it looks nicer and less cluttered.

Before:
Before

After:
After


ChristianChiarulli (approved):
🥽

PR #211, opened 2021-04-01
Add scrolloff to nv-globals.lua and settings.lua

DominikMarcinowski:
Default option doesn't affect default behavior by setting it to `0` (which is the default).


jameswalmsley (approved):

luxus:
thanks for the pr
scroll off is working by now.. this can be closed

PR #212, opened 2021-04-01
Add bindings to quickly quit certain buffers

DominikMarcinowski:
I added bindings for `dashboard`, `lspinfo`, and `floaterm` to quickly quit out of them with `q` key.
The `<buffer>` thing scopes keybind to buffer of given filetype only.

This is inspired by `startify` plugin becouse these plugins doesn't have way to set bindings like that.

In floaterm you have to first hit `<Esc>` then `q`.


ChristianChiarulli (approved):
🆒

PR #213, opened 2021-04-01
add compe-tabnine

mjcc30:
https://github.com/tzachar/compe-tabnine
from example : https://github.com/JoseConseco/nvim_config/blob/926af57d9f6ec5ec7527eeceaea7fc81c78858a1/lua/plugins.lua


ChristianChiarulli:
This doesn't auto enable tabnine or anything right? It just gives compe bindings for if you enable it?

mjcc30:
Hello, i think this install tabnine withe install.sh then enable bindings for compe
https://github.com/tzachar/compe-tabnine/blob/main/install.sh

PR #214, opened 2021-04-01
Fix opening settings from dashboard

DominikMarcinowski:
Previously the tabline wouldn't show up after opening `Settings` from dashboard.


ChristianChiarulli (approved):
🍨

PR #215, opened 2021-04-02
Fix clang global vars nil value

nawfalmrouyan:
Latest change to clang didn't initialize the clang global vars.


ChristianChiarulli (approved):
thanks 🥼

PR #216, opened 2021-04-02
Make tsserver work on standalone files

DominikMarcinowski:
This patch allows tsserver to activate on standalone files, by adding `.`, as a `root_dir` marker.

This shouldn't break working inside projects, becouse now tsserver will behave like vscode's integrated server.


ChristianChiarulli (approved):
🆒

PR #221, opened 2021-04-02
load plugin configs via packer

heimer16:
example of loading specific plugin configs via packer
cleans up init.lua
allows for disabling plugin and its config by commenting out one line

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -8,7 +8,7 @@ require('nv-autocommands')
 require('settings')
 require('keymappings')
 require('colorscheme')
-require('nv-galaxyline')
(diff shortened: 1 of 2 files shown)


ChristianChiarulli (approved):
This is great

heimer16:
Works when plugins are already installed but was causing issues on fresh
install. When installing plugins it hangs on the first one with config.
Might have to tweak the local function or something, so maybe hold off for
now

heimer16:
https://github.com/wbthomason/packer.nvim/issues/242

working on a solution

PR #225, opened 2021-04-02
Setup persistent undo, swapfiles and backupfiles

alanjjenkins:
* Sets up persistent undo so that if you ever make a change and quit the
  buffer or nvim you can reopen the file and still undo the changes.
* Setup backups so that if your computer runs out of battery or neovim
  gets closed for another reason that you have a backup of changes.
* Sets up swapfiles properly so that they do not get saved to the git
  repos.


seocamo:
some servers got problems with the backupfiles part, maybe a switch to turn it off ??

michalens:
Can you double-check that this works?
For me only persistent undo only works like this
```
vim.bo.undofile = true
```

michalens:
Also, please make use of global variables from `nv-globals`
```lua
vim.o.backupdir = DATA_PATH .. '/backup'
```
Or `vim.bo`, I am not sure

PR #231, opened 2021-04-04
minor-installer-fix: remove install argument from pacman

Walkability:
I also mentioned adding `sudo` to `npm i -g`, but you actually would not want to do this. The reason I was getting the error is because I had installed nvm, but hadn't set it up on my fresh install yet.


ChristianChiarulli (approved):
cool thanks :octocat:

PR #257, opened 2021-04-15
replace lsp-kind for vim.lsp.protocol.CompletionItemKind native option

rafamadriz:
As you mentioned, trying to remove as much plugins as possible while keeping 95% of the functionality is what you want. This PR replaces lsp-kind plugin for the  vim.lsp.protocol.CompletionItemKind native option. I added the option to `lua/lsp/init.lua` and if you don't like the text and just want the icons I can remove it. Personally I found it to be really useful since I can't remember what each icon means.

This is an example of how it looks.
2021-Apr-15


lua/plugins.lua
```diff
@@ -22,84 +22,87 @@ local function require_plugin(plugin)
...
+
+return require("packer").startup(
+    function(use)
+        -- Packer can manage itself as an optional plugin
+        use "wbthomason/packer.nvim"
+
+        -- TODO refactor all of this (for now it works, but yes I know it could be wrapped in a simpler function)
+        use {"neovim/nvim-lspconfig", opt = true}
+        use {"glepnir/lspsaga.nvim", opt = true}
+        use {"onsails/lspkind-nvim", opt = true}
```
ChristianChiarulli, line 41: could you remove this as well? thank you for the PR

ChristianChiarulli (changes requested):

lua/plugins.lua
rafamadriz: You mean just removing lspkind right ? I just did that. And btw I guess I have a different formatter because when saved the file it replaced all the single quotes with double quotes, hope you don't mind it.

ChristianChiarulli (approved):
looks good 💇🏻

PR #258, opened 2021-04-15
Add lsp support for ruby

przbadu:
```sh
:LspInstall ruby
```


lua/lsp/ruby-ls.lua
```diff
@@ -0,0 +1,5 @@
+-- If you are using rvm, make sure to change below configuration
+require'lspconfig'.solargraph.setup {
+    cmd = { DATA_PATH .. "~/.rbenv/shims/solargraph", "--stdio" },
+    on_attach = require'lsp'.common_on_attach
+}
```
ChristianChiarulli, line 5: Consider adding the handler, you can find it in the python lsp module and then pass the ruby diagnostic values here

ChristianChiarulli, line 5: consider adding the filetypes field as well.

lua/lsp/ruby-ls.lua
```diff
@@ -0,0 +1,5 @@
+-- If you are using rvm, make sure to change below configuration
+require'lspconfig'.solargraph.setup {
+    cmd = { DATA_PATH .. "~/.rbenv/shims/solargraph", "--stdio" },
```
ChristianChiarulli, line 3: you could also make the solargraph path configurable and expose that in the settings for the user if they would like it to point to a virtuual env or binary found in the PATH

ChristianChiarulli (changes requested):
This is a good start, consider adding the handler as well for

lua/lsp/ruby-ls.lua
przbadu: Hey, thank you for reviewing my PR, I am totally new to nvim and lua, I might need to look into lua soon. :smile: 

can you please suggest me on how to make solargraph path configurable? I am trying to Fix these reviews, I really love what you have done with LunarVim. It is easily configurable for my need.

ChristianChiarulli: sure, I think what we have now is probably good enough, but you could just have a setting like:

 `O.ruby.solargraph_path="~/.rbenv/shims/solargraph"`

Then let the user override this in lv-settings

ChristianChiarulli (approved):

Issue #296, opened 2021-04-23
nvim default keybinding overriding plugins bindings


afridley:
When I hit **gd** I will go to definition in the current function
which you can see from neovim :help index
```
|gd|		gd		1  go to definition of word under the cursor
				   in current function
```
however :map shows
```
n  gd          * <Cmd>lua vim.lsp.buf.definition()<CR>
```

The weird thing is if I hit gd really fast twice I will get the lsp response and go to the definition in the file its defined in.
This is true for most of the lsp keybindings.

This is with the default installation and LspInstall typescript

rebuilt:
What's the output of
```bash 
verbose map gd
```

Do you have python lsp installed and running?  What's the output of:
```bash
:LspInfo
```
If I remember correctly a keymapping gets created and attached to the buffer when an lsp client is created.  They are not attached to the buffer if the current filetype has no defined lsp settings.  This doesn't explain why they would work sometimes and not others.  But check your lsp settings to make sure everything is in place first.

afridley:
verbose map gd: `n  gd          * <Cmd>lua vim.lsp.buf.definition()<CR>`
:LspInfo : 
```
Configured servers: html, jsonls, kotlin_language_server, tailwindls, emmet_ls, cssls, dartls
Neovim logs at: /home/x2/.cache/nvim/lsp.log

0 client(s) attached to this buffer: 

1 active client(s): 

  Client: tsserver (id 1)
  	root:      /home/x2/Sites/pinggolf-biz-webapp
  	filetypes: javascript, javascriptreact, javascript.jsx, typescript, typescriptreact, typescript.tsx
  	cmd:       /home/x2/.local/share/nvim/lspinstall/typescript/node_modules/.bin/typescript-language-server --stdio
  	


Clients that match the filetype typescript:
  
  Config: tailwindls
  	cmd:               node /home/x2/.local/share/nvim/lspinstall/tailwindcss/tailwindcss-intellisense/extension/dist/server/index.js --stdio
  	cmd is executable: True
  	identified root:   /home/x2/Sites/pinggolf-biz-webapp
  	custom handlers:   tailwindcss/getConfiguration
```

I dont have python installed but im using it for typescript. Its working for completion and I can type :LspDefinition which works fine.

Just when using keybindings it doesnt work

rebuilt:
[...]

Issue #286, opened 2021-04-21
I can't use Emmet


marcograhl:
How Can I enable Emmet < C - y > , is not doing much for me. Anybody uses Emmet with LunarVim and has a hint?

rebuilt:
Assuming you have everything installed correctly, got to ./lua/settings.lua and comment out line 35.  

```bash
-- vim.o.timeoutlen = 100
```

This is tip from the emmet project
```bash

FAQ
    Pressing ctrl+y+, doesn't work

    Probably you set timeoutlen=0. Most of Vim plugins which using key-mappings does not work with timeoutlen=0.
```

ChristianChiarulli:
I recommend emmet-ls, right now you need to globally install `npm i -g emmet-ls`

marcograhl:
Thx for the fast reply, I did install emmet-ls do I need to Install it with Packer too? I mean the use and require you did with the outer plugins? I sometimes get an autocompletion randomly, I don't know if I'm doing the ctrl-y +, right :). Sometimes there is an autocompletion with Lsp Emmet without me doing anything. Im just a little confused, the resulting snippit is wrongly indented as well. I will dig deeper , cause I love the work you put in, its nice to have "default nvim" this beautifull. Thx Chris

marcograhl:
Is there a way I can see if the keybind ctrl-y , is actually being used by emmet?

rebuilt:
```bash
    :nmap for normal mode mappings
    :vmap for visual mode mappings
    :imap for insert mode mappings
```

Try this instead
```bash
:verbose imap <C-y>
```

marcograhl:
Its not mapped to anything weird. I didn't change anything from the starter files. I got Lsp working and its beautifull => love the preview popup window omg. I installed emmet-ls globaly.. mhh is there a step I missed.

rebuilt:
I haven't set it up myself but according to[ the docs](https://www.npmjs.com/package/emmet-ls) you need something like this in your configs

```bash
local nvim_lsp = require'lspconfig'
local configs = require'lspconfig/configs'
local capabilities = vim.lsp.protocol.make_client_capabilities()
capabilities.textDocument.completion.completionItem.snippetSupport = true

configs.emmet_ls = {
  default_config = {
    cmd = {'emmet-ls', '--stdio'};
    filetypes = {'html', 'css'};
    root_dir = function()
      return vim.loop.cwd()
    end;
    settings = {};
  };
}

[...]

PR #298, opened 2021-04-25
allows enough time for keybindings to be attached to buffer

rebuilt:
fixes issue #296.  Though I think this setting depends on the speed of the underlying hardware.


rebuilt:
I should also say that I've changed my personal timeoutlen to 500 because I'm using tpope's surround.vim.  It's hard to type all the keystrokes with only a 200ms buffer.  The same was true for emmet.

ChristianChiarulli:
@rebuilt I've always had this conflict between timeoutlen being fast and the whichkey menu showing up. Btw have you tried emmet-ls?

rebuilt:
Yes.  Had a conversation about it [here](https://github.com/ChristianChiarulli/LunarVim/issues/286)  #286

I had my own personal neovim config and was able to use emmet-ls. I couldn't get it to work with compe.  I had to switch to completion-nvim because it supports chain completion.  I added the required steps in the other thread.  

Yesterday I switched to full time use of lunarvim but I haven't switched the completion plugin from compe to completion-nvim yet.

rebuilt:
Does emmet-ls work out of the box in lunarvim?

rebuilt:
I have changed the whichkey keybinding to \.  It works better for me and I dont have to see whichkey open up when I'm just commenting things out.

rebuilt:
Autocomplete for emmet is bugged.  I don't think it works with compe.  

lunar_emmet

rebuilt:
Looks like the autocomplete is bugged with emmet.  Its probably not compatible with compe; needs chain completion support.

lunar_emmet

rebuilt:
> @rebuilt I've always had this conflict between timeoutlen being fast and the whichkey menu showing up. Btw have you tried emmet-ls?

There will always be a conflict between whichkey speed and the proper functioning of plugin mappings if they are both bound to the leader key.  Whichkey should be on its own key.  If it's the only thing bound to a particular key(like backspace or a function key) ttimeout and ttimeout length will apply.  It's very important for there to be no other mappings, otherwise timeout and timeoutlen apply.  I think ttimeout is already at 50ms in LunarVim.   

https://www.johnhawthorn.com/2012/09/vi-escape-delays/

rebuilt:
To see what's already been set:
```vim
:verb set timeout?
:verb set ttimeout?
:verb set timeoutlen?
:verb set ttimeoutlen?
```

rebuilt:
Never mind.  I just tried it.  Changing the keymappings to '\\' work and whichkey loads instantly but none of the tables populate correctly.  I guess the plugin will only work with the leader key.

rebuilt:
I tested the latest changes you made to Lunarvim.   Looks like emmet works out of the box now.

rebuilt:
Maybe not.  It worked on first load but not anymore.

ChristianChiarulli:
@rebuilt the changes I made should activate snippets (which weren't active before) and emmet-ls should just work out of the box now (after a global install). 

I would love to find a way to decouple which key popup time from `timeoutlen`.

ChristianChiarulli:
@rebuilt the changes I made should activate snippets (which weren't active before) and emmet-ls should just work out of the box now (after a global install). 

I would love to find a way to decouple which key popup time from `timeoutlen`.

rebuilt:
No longer relevant since the switch to lua whichkeys

PR #308, opened 2021-04-27
Allows Emmet-ls completion to work

rebuilt:
Looks like compe works with emmet.  The only step missing for emmet-ls to work was requiring friendly-snippets.

lunar_emmet_success


rebuilt:
Chris has fixed Emmet with his last commits already.  This is no longer needed

Issue #307, opened 2021-04-27
Prettier error not showing on typescript project


mandaputtra:
## Step to reproduce

1. Follow the step on README.md to install `typescript` with `LspInstall:`
2. Changing `efm-general.lua` to this

```
require"lspconfig".efm.setup {
    -- init_options = {initializationOptions},
    cmd = {DATA_PATH .. "/lspinstall/efm/efm-langserver"},
    init_options = {documentFormatting = true, codeAction = false},
    filetypes = {"lua", "python", "typescript", "javascriptreact", "javascript", "sh", "html", "css", "json", "yaml", "markdown", "vue"},
    settings = {
        rootMarkers = {".git/"},
        languages = {
            python = python_arguments,
            lua = lua_arguments,
            sh = sh_arguments,
            javascript = tsserver_args,
            javascriptreact = tsserver_args,
            html = {prettier},
            css = {prettier},
            json = {prettier},
            yaml = {prettier},
            markdown = {markdownPandocFormat},
	    typescript = {prettier, eslint}
            -- javascriptreact = {prettier, eslint},
            -- javascript = {prettier, eslint},
            -- markdown = {markdownPandocFormat, markdownlint},
        }
    }
}

```

3. Open typescript repository that had eslint and prettier in it
4. Error on typescript side are shown but not on prettier side (indentation error format etc)

Actually a _dirty_ solution already provided by @rebuilt [here](https://github.com/ChristianChiarulli/LunarVim/issues/299#issuecomment-827296578) :+1: :bowing_man: . I create this issue so @rebuilt can reference it during pull request. 

## Some info

```
NVIM v0.5.0-dev+1273-g8402865cb
Ubuntu 20.10
```

mandaputtra:
I'm checking in again, I test it with Vue. It also requires the _dirty_ solution to make prettier working properly.

```lua
--  efm-general.lua
require"lspconfig".efm.setup {
    -- init_options = {initializationOptions},
    cmd = {DATA_PATH .. "/lspinstall/efm/efm-langserver"},
    init_options = {documentFormatting = true, codeAction = false},
    filetypes = {"lua", "python", "typescript", "javascriptreact", "javascript", "sh", "html", "css", "json", "yaml", "markdown", "vue"},
    settings = {
        rootMarkers = {".git/"},
        languages = {
            python = python_arguments,
            lua = lua_arguments,
            sh = sh_arguments,
            javascript = tsserver_args,
[...]

PR #311, opened 2021-04-27
Typescript

rebuilt:
Provides typescript formatting. Addresses issue #307

### DIFF
--- a/lua/lsp/efm-general-ls.lua
+++ b/lua/lsp/efm-general-ls.lua
@@ -92,7 +92,7 @@ require"lspconfig".efm.setup {
     -- init_options = {initializationOptions},
     cmd = {DATA_PATH .. "/lspinstall/efm/efm-langserver"},
     init_options = {documentFormatting = true, codeAction = false},
-    filetypes = {"lua", "python", "javascriptreact", "javascript", "sh", "html", "css", "json", "yaml", "markdown"},
+    filetypes = {"lua", "python", "javascriptreact", "javascript", "typescript","sh", "html", "css", "json", "yaml", "markdown"},
     settings = {
         rootMarkers = {".git/"},
         languages = {
@@ -101,6 +101,7 @@ require"lspconfig".efm.setup {
             sh = sh_arguments,
             javascript = tsserver_args,
             javascriptreact = tsserver_args,
+			typescript = tsserver_args,
             html = {prettier},
             css = {prettier},
             json = {prettier},
--- a/lua/lsp/efm-general-ls.lua
+++ b/lua/lsp/efm-general-ls.lua
@@ -92,7 +92,7 @@ require"lspconfig".efm.setup {
     -- init_options = {initializationOptions},
     cmd = {DATA_PATH .. "/lspinstall/efm/efm-langserver"},
     init_options = {documentFormatting = true, codeAction = false},
-    filetypes = {"lua", "python", "javascriptreact", "javascript", "typescript","sh", "html", "css", "json", "yaml", "markdown"},
+    filetypes = {"lua", "python", "javascriptreact", "javascript", "typescript","sh", "html", "css", "json", "yaml", "markdown", "vue"},
     settings = {
         rootMarkers = {".git/"},
         languages = {
--- a/lua/lsp/efm-general-ls.lua
+++ b/lua/lsp/efm-general-ls.lua
@@ -92,7 +92,7 @@ require"lspconfig".efm.setup {
     -- init_options = {initializationOptions},
     cmd = {DATA_PATH .. "/lspinstall/efm/efm-langserver"},
     init_options = {documentFormatting = true, codeAction = false},
-    filetypes = {"lua", "python", "javascriptreact", "javascript", "typescript","sh", "html", "css", "json", "yaml", "markdown", "vue"},
+    filetypes = {"lua", "python", "javascriptreact", "javascript", "typescript","typescriptreact","sh", "html", "css", "json", "yaml", "markdown", "vue"},
     settings = {
         rootMarkers = {".git/"},
         languages = {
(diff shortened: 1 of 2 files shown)


mandaputtra:
I tried your solution the prettier work, but it triggers endless loop. My machine freezes. Lol.  Step to reproduce are the same. Just edit file longer. I do think this is because the `auto_formatter` I can't debug because my machine freezes. I must undo the changes on `lv-autocommands/init.lua`   to stop nvim from making my machine freezes.

ChristianChiarulli:
@rebuilt @mandaputtra can we confirm that this behavior is due to this change? I don't wanna merge this until we know for sure.

mandaputtra:
I tried it 4 times, so I can confirm this behaviour happens. I don't know why is this happening.

rebuilt:
I tested on my desktop which works without errors but I think I should do additional testing in the cases where prettier and eslint have not been installed or not initialized or is not part of a git project.  I'll perform the tests on my laptop.  

Maybe it's also possible that only one or the other can be enabled at the same time?  Maybe prettier changes the file and eslint works on it.  Then eslint works on it.  Then prettier sees the file has changed AGAIN and the cycle continues.  This might be the case if the user has an autosave on file change autocommand set.

rebuilt:
I couldn't reproduce the bug.  I tested on my laptop and in a virtual machine on my desktop.  I performed variations of the steps below.  Maybe I'm making assumptions in my testing that I shouldn't be making?   I start off by removing the 'local' and 'cache' directories because sometimes leaving them in causes weird behavior.  The only error I could trigger was when I didn't initialize the git repo. 

```bash
sudo rm -R ~/.config/nvim
sudo rm -R ~/.cache/nvim/
sudo rm -R ~/.local/share/nvim/

bash <(curl -s https://raw.githubusercontent.com/ChristianChiarulli/lunarvim/master/utils/installer/install.sh)
git fetch origin pull/311/head:typescript
git checkout typescript

npm install -g --save-dev --save-exact prettier
echo {}> .prettierrc.json
npm install eslint --save-dev
npm init
npx eslint --init
git init

:LspInstall efm
:LspInstall typescript
```

afridley:
@rebuilt @ChristianChiarulli This should also include typescriptreact 

```lua
require"lspconfig".efm.setup {
    -- init_options = {initializationOptions},
    cmd = {DATA_PATH .. "/lspinstall/efm/efm-langserver"},
    init_options = {documentFormatting = true, codeAction = false},
    filetypes = {"lua", "python", "javascriptreact", "javascript", "typescript", "typescriptreact", "sh", "html", "css", "json", "yaml", "markdown"},
    settings = {
        rootMarkers = {".git/"},
        languages = {
            python = python_arguments,
            lua = lua_arguments,
            sh = sh_arguments,
            javascript = tsserver_args,
            javascriptreact = tsserver_args,
            html = {prettier},
            css = {prettier},
            json = {prettier},
            yaml = {prettier},
            markdown = {markdownPandocFormat},
            typescript = tsserver_args
            typescriptreact = tsserver_args
            -- javascriptreact = {prettier, eslint},
            -- javascript = {prettier, eslint},
            -- markdown = {markdownPandocFormat, markdownlint},
        }
    }
}
```

lv-autocommands
```lua
local javascript_autoformat = {'BufWritePre', '*.js', 'lua vim.lsp.buf.formatting_sync(nil, 1000)'}
local javascriptreact_autoformat = {'BufWritePre', '*.jsx', 'lua vim.lsp.buf.formatting_sync(nil, 1000)'}
local typescript_autoformat = {'BufWritePre', '*.ts', 'lua vim.lsp.buf.formatting_sync(nil, 1000)'}
local typescriptreact_autoformat = {'BufWritePre', '*.tsx', 'lua vim.lsp.buf.formatting_sync(nil, 1000)'}
if O.tsserver.autoformat then
    table.insert(auto_formatters, javascript_autoformat)
    table.insert(auto_formatters, javascriptreact_autoformat)
    table.insert(auto_formatters, typescript_autoformat)
    table.insert(auto_formatters, typescriptreact_autoformat)
end
```

mandaputtra:
Found the culprit _I think_ this is a problem on `efm` side or just my machine f*ck up. If I wrote : 

```lua
typescript = {prettier, eslint},
```

the endless loop that freezes my machine would happen but, if I wrote the settings like these

```lua
typescript = {eslint, prettier},
```

No freezes happen, The downside of this are prettier error won't show up on the editor (neovim) but formatting works! So I'm okay with this.

rebuilt:
@mandaputtra Sorry, I'm a little confused.  When you tested it, where you copy/pasting code from the issue: https://github.com/ChristianChiarulli/LunarVim/issues/307   

Or did you check out this branch with
```bash
git fetch origin pull/311/head:typescript
git checkout typescript
```

The code I posted on the issue doesn't match the code I'm committing in this branch,

mandaputtra:
@rebuilt I'm copy pasting from your comment, when I tried clean slate install it doesn't make my machine freeze. 

But prettier just doesn't show up the error message but the formatting are working properly, is the prettier error message  showing on yours?

Something like these

```
delete ...., must be 8 space 
```

rebuilt:
@mandaputtra 
If I understand the code correctly, you'll never see messages from prettier.  The efm language server passes your source file through stdin/stdout to prettier. Prettier returns a prettied file and the buffer gets updated with those changes.  

Maybe your thinking of error hints provided by the language server?  For that you need to have the typescript lsp installed.  Sometimes this breaks when you change the LunarVim build.  Make sure you reinstall the language server, :LspInstall typescript

Also make sure you enable eslint in lv-settings.lua.  If you don't have eslint installed and configured, you'll need to do that as well. 

```bash
O.tsserver.linter = 'eslint'
```

PR #314, opened 2021-04-28
moves whichkey keybinding from space to backslash. Also turns off tim…

rebuilt:
Moves whichkey keybinding from space to backslash. Also turns off timeoutlen setting for increased compatibility with plugins like, comments, endwise,toop, vim-surround, and other plugins that rely on longer key combination timeouts.

whichkey


rebuilt:
This might need additional work.  Since the two functions are now separate, keybindings like <space>f won't work since they are a  whichkey binding, not a general binding with space.

rebuilt:
The first part of the gif is me opening and closing whichkey quickly.  The second half shows that the comment plugin now works because timeoutlen isn't so low.

rebuilt:
If I duplicate all the keybindings, it will mean a user must ALSO change keybindings in two places if they don't like the defaults.

rebuilt:
No longer needed since the switch to lua version of whichkey.

Issue #313, opened 2021-04-28
AutoPairs not working in latest pull of lunarvim


ferran9908:
In the latest pull of LunarVim, autopairs don't seem to be working. It was working before I pulled the latest version.

I'm on a Mac, with the neovim --HEAD installed.

Looks like commit `102b46ca4a99522f` introduced this bug. Reverting the file to how it was earlier solves the problem, but the makes autocomplete experience obsolete.

ChristianChiarulli:
hmm, I need to look into this

ChristianChiarulli:
I think they should work now

hlmeerichs:
Not sure if its the same issue (also on mac), but after a fresh install I get:

E5113: Error while calling lua chunk: /Users/admin/.config/nvim/lua/lv-autopairs
/init.lua:81: module 'nvim-autopairs.rule' not found:
        no field package.preload['nvim-autopairs.rule']
        no file './nvim-autopairs/rule.lua'
        no file '/opt/homebrew/Cellar/luajit-openresty/20201229_1/share/luajit-2
.1.0-beta3/nvim-autopairs/rule.lua'
        no file '/usr/local/share/lua/5.1/nvim-autopairs/rule.lua'
        no file '/usr/local/share/lua/5.1/nvim-autopairs/rule/init.lua'
        no file '/opt/homebrew/Cellar/luajit-openresty/20201229_1/share/lua/5.1/
nvim-autopairs/rule.lua'
        no file '/opt/homebrew/Cellar/luajit-openresty/20201229_1/share/lua/5.1/
nvim-autopairs/rule/init.lua'
        no file './nvim-autopairs/rule.so'
        no file '/usr/local/lib/lua/5.1/nvim-autopairs/rule.so'
        no file '/opt/homebrew/Cellar/luajit-openresty/20201229_1/lib/lua/5.1/nv
im-autopairs/rule.so'
        no file '/usr/local/lib/lua/5.1/loadall.so'
        no file './nvim-autopairs.so'
        no file '/usr/local/lib/lua/5.1/nvim-autopairs.so'
        no file '/opt/homebrew/Cellar/luajit-openresty/20201229_1/lib/lua/5.1/nv
im-autopairs.so'
        no file '/usr/local/lib/lua/5.1/loadall.so'

soerenmartius:
same here

[...]

PR #315, opened 2021-04-28
Fix Autopairs Not working issue

ferran9908:
In the latest pull of LunarVim, the autopairs plugin doesn't work. (Issue #313)

Looks like it had a few rules missing.

Added them in this PR.

### DIFF
--- a/lua/lv-autopairs/init.lua
+++ b/lua/lv-autopairs/init.lua
@@ -78,6 +78,7 @@
 
 local remap = vim.api.nvim_set_keymap
 local npairs = require('nvim-autopairs')
+local Rule = require('nvim-autopairs.rule')
 
 -- skip it, if you use another global object
 _G.MUtils= {}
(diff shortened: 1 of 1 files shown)


rebuilt (commented):
Autopairs changes will also affect emmet behavior.  This is how this commit behaves with emmet.  It will not support more than one html element completion.
The full emmet syntax wont work.

```bash
div+div>p>span+em^bq
```

autopairs_emmet_conflict

rebuilt:
Actually, I don't think it's a problem with your commit.  The master branch also has this behavior.

ferran9908:
> Autopairs changes will also affect emmet behavior. This is how this commit behaves with emmet. It will not support more than one html element completion.
> The full emmet syntax wont work.
> 
> ```shell
> div+div>p>span+em^bq
> ```
> 
> autopairs_emmet_conflict

I think this is a separate issue on it's own, as this is the default behavior in the master branch as well.

rebuilt:
You're right.  Emmet has the same behavior as the main branch.

rebuilt:
Autopairs works correctly in javascript and typescript.

rebuilt:
Lua works as well.

ferran9908:
@rebuilt You mean in the current build or in this commit?

rebuilt:
Sorry to be vague, I'm testing your commit.  I can confirm that autopairs is broken in the main branch. I've tested your commit with lua, javascript and typescript.  I can confirm your commit fixes autopair problems with those languages.

ferran9908:
@rebuilt No problem. Thanks for verifying!

PR #328, opened 2021-04-30
Adds LSP for the R language

fanmih:
R language support via LSP.
For this to work one needs to install the 'languageserver' package on the R side:
`install.packages("languageserver")`
https://github.com/REditorSupport/languageserver


bartoszluka:
Does it really work? I tested it on my machine and the language server doesn't attach though it is recognized

fanmih:
@bartoszluka It does work in my case.
Do not forget to install the language server in R:
Start the R environment and run:
`install.packages("languageserver")`

See attached image on Neo vim. The r_language_server is attached
nvim

and from LspInfo:
LspInfo

bartoszluka:
r_language_server is here but as you can see it is not active
image
I installed languageserver packare in R and I run it manually but it does not help
image

rebuilt:
The cmd executable has to be available globally.  If 'which R' returns the pathe to the executable, it's available globally.  If not, the full path must be specified in the cmd property.

bartoszluka:
@rebuilt the `R` command is available globally and points to usr/bin/R. I tried to pass the full path but the result was the same

rebuilt:
What's the output of 
```bash
:LspStart r
```
If the error message gets cut off
```bash
:messages
```

bartoszluka:
`:LspStart r`
E5108: Error executing lua ...im/site/pack/packer/opt/nvim-lspconfig/lua/lspconfig.lua:27: attempt to index a nil value

rebuilt:
Are you in a git repository?

Also, looking at the code, none of the default settings are being altered.  It should be the same as
```bash
require'lspconfig'.r_language_server.setup{}
```
Try that and see if anything changes.

bartoszluka:
Yes, I am in a git repo.
I tried what you suggested but the results are exactly the same. No point in pasting the `:LspInfo` output as it is the same as the one above

ChristianChiarulli:
I will merge this when someone can confirm it is working. Thanks for adding this @fanmih

bartoszluka:
@ChristianChiarulli It somehow started working! I opened the same file in another git repo and it turned on the server. My lsp config is exactly the same as the one in the commit.
BTW thanks for adding this!
image
image

fanmih:
> I will merge this when someone can confirm it is working. Thanks for adding this @fanmih

@ChristianChiarulli I can confirm that this is working.

abzcoding:
because a lot has changed regarding the formatting structure since this PR, I'm closing this in favor of #965

PR #330, opened 2021-05-01
Implement fzy native to speed up telescope

afridley:
I think this is worth adding as a default. It will significantly speed up telescope

References
https://github.com/nvim-telescope/telescope-fzy-native.nvim
which uses
https://github.com/romgrk/fzy-lua-native

The main highlight is
```
Lines: 100000

Native:
 -> Total: 34130
 -> Time:  69.418 ms
Original:
 -> Total: 34130
 -> Time:  835.683 ms
```

### DIFF
--- a/lua/lv-telescope/init.lua
+++ b/lua/lv-telescope/init.lua
@@ -61,5 +61,11 @@ require('telescope').setup {
                 -- ["<C-i>"] = my_cool_custom_action,
             }
         }
+    },
+    extensions = {
+        fzy_native = {
+            override_generic_sorter = false,
+            override_file_sorter = true,
+        }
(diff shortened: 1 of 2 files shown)


tarkin88 (approved):
Looking nice 👍🏾

ChristianChiarulli:
Does the user need to have any extra programs installed? Will it fallback to normal without it? @afridley

afridley:
@ChristianChiarulli yeah it will fall back no extra programs needed other then the base telescope plugin.

If the user wants to remove telescope completely they should remove base telescope and this

ChristianChiarulli:
🆒

Issue #336, opened 2021-05-01
Some default bindings won't work anymore


OuterHeavenCC:
Some default bindings like "caw", "caW", "c0" won't work anymore, they just move the cursor without deleting and going into insert mode

rebuilt:
The 'c' keybind is being overwritten by a keybind set by lunarvim.  ~/.config/nvim/lua/lv-which-key/init.lua   I just submitted a pull request to be able to toggle whichkey on/off.

rebuilt:
In the meantime you can use 
'cf'  (change forward) 
'cF'  change backwords 
'cb' change backwords
'ct' change until
'cT' change backwards until

folke:
With WhichKey, default bindings on c should work. Is there anything still not working related to this?

Might be I fixed that yesterday.

rebuilt:
@folke Confirmed.  After :PackerSync these keybindings work as expected.

@OuterHeavenCC  You'll also have to remove the keybinding in ~/.config/nvim/lua/lsp/init.lua line 23. 'ca' is mapped to LspSaga code_action.

rebuilt:
@OuterHeavenCC  In whichkey, a codeaction is mapped to 
```bash
<space>la
```
So that might be a good keybind to use

OuterHeavenCC:
> @folke Confirmed. After :PackerSync these keybindings work as expected.
> 
> @OuterHeavenCC You'll also have to remove the keybinding in ~/.config/nvim/lua/lsp/init.lua line 23. 'ca' is mapped to LspSaga code_action.

Thanks, that worked!

sakshambedi:
The new Lua based which key is buggy. It takes time compared the the vim based whichkey and doesn't work sometimes.

folke:
@sakshambedi the time it takes before showing the popup is configured by `timeoutlen`. Please refer to the [documentation](https://github.com/folke/which-key.nvim#%EF%B8%8F-configuration) about this.

In what cases does it not work for you?

Would be great if you could submit a bug report.

Thanks!

PR #337, opened 2021-05-01
Adds the ability to toggle on/off whichkey and dashboard

rebuilt:
Adds the ability to toggle on/off whichkey and dashboard.  Leaves them on by default.  On my machine the startup time with both on is 166ms both off is 130ms.  Also, whichkey default keybinds overwrite the 'c' bind.  

Partially addresses #336


rebuilt:
If you're checking out this branch to test, you may have to re-run :PackerCompile, :PackerInstall since the packer_compiled.vim file won't match the changes you bring in through the checkout.

folke:
@rebuilt I've made some big improvements on performance when loading the plugin. Should take around 2ms to 5ms now.

rebuilt:
@folke 

Thanks for all your work on whichkey.  I've noticed how responsive you've been to your users requests.  

Honestly, I didn't like the old whichkey because it was incompatible with so many plugins (a lot of tpope plugins)  
But your version of whichkey is excellent and so far it's been compatible with all the plugins I've tested.

folke:
Thanks man. Glad you like it!

Issue #351, opened 2021-05-03
Autopairs not ignoring \


atropos112:
Hello everyone,

I've been switching from VS Code to nVim, initially trying to set things up from scratch but once I found LunarVim I was over the moon (Lunar, moon get it ; ) ). I installed VimTex as I want to write latex and installed LSP for it with `:LspInstall Latex ` and magic, it works. 

I have a problem however and I am unsure how to remedy this. I want to use a snipet `\(` which turns (via vsnip)  into `\( $1 \)` however when i write `\(` instead autopairs clicks in and gives `\()` at which point i can't trigger the snippet because  there is the ) part.

I noticed its the lv-autopairs closing the brackets and that if I look at https://github.com/jiangmiao/auto-pairs for example ,there is a "ignore when \ is used" feature. Is there any way I can turn it on here? That is can i somehow prevent autopairs from closing the bracket if the previous letter is \ ?

Thank you for your input !

rebuilt:
Maybe the autopairs project can help?  https://github.com/windwp/nvim-autopairs

rebuilt:
If you want to turn it off temporarily while you figure out the autopair rules, you can comment out 

```bash
        basic("(", ")")
            :with_pair(cond.check_is_bracket_line()),
```
From ~/.local/share/nvim/site/pack/packer/opt/nvim-autopairs/lua/nvim-autopairs/rules/basic.lua

atropos112:
> Maybe the autopairs project can help? https://github.com/windwp/nvim-autopairs

Thank you for your response, I must admit i looked at the wrong repo (ooh lord...), looking at it right now.

rebuilt:
The other rules already defined might be taking precedence.  I think there are two regexes that define any symbol is always bracketed

atropos112:
I see two rules already existing there but they have the "lua" tag on them surely they wouldn't affect .tex file types ?

rebuilt:
Autopairs also includes basic rules I think

rebuilt:
Also check the filetype 
```bash
:set filetype?
```

atropos112:
Im guessing u wanted me to run it on the file where I want to have that rule working, in that case `filetype=tex`

rebuilt:
Literally run the command with the question mark appended.  It gives me filetype 'plaintext'

rebuilt:
Now that you've set it, you might have to restart nvim for it to show the filetype detected on start

atropos112:
[...]

PR #357, opened 2021-05-04
Fix: makes autopairs filetype specific rules work

rebuilt:
In the current version, autopairs filetype specific rules don't work.   

```lua
npairs.add_rules({
-- This rule doesn't work
  Rule("(", ")",{"tex", "latex"})
    -- don't add a pair if  the previous character is xxx
    :with_pair(cond.not_before_regex_check("xxx", 3))
     },
-- This rule works
 Rule("(", "")
 -- don't add a pair if  the previous character is xxx
    :with_pair(cond.not_before_regex_check("xxx", 3))
)
```
The gif shows the way the current code is broken.  Rules only work if you don't add a filetype.  

autopairs_not_working

Changing the way LunarVim installs the two together in packer fixes the problem.  
I changed 
```bash
# This breaks autopairs.  Autopairs filetype rules will only work if I manually launch nvim-tree.  
use "kyazdani42/nvim-tree.lua"
``` 
to
```bash
# This code makes autopairs rules work regardless of whether nvim-tree has been launched.  
use {"kyazdani42/nvim-tree.lua", opt = true }
```
This is the longest I've ever spent adding two words to a codebase.

### DIFF
--- a/lua/plugins.lua
+++ b/lua/plugins.lua
@@ -59,7 +59,7 @@ return require("packer").startup(
         use {"windwp/nvim-ts-autotag", opt = true}
 
         -- Explorer
-        use "kyazdani42/nvim-tree.lua"
+        use {"kyazdani42/nvim-tree.lua", opt = true}
         -- TODO remove when open on dir is supported by nvimtree
         use "kevinhwang91/rnvimr"
 


rebuilt:
Addresses #351

rebuilt:
This commit changes the plugins.lua file so you'll need to recompile it with packer.  Additionally, it changes the opt setting for a plugin so you'll need to run a :PackerSync, and :PackerInstall so it moves the nvim-tree directory.    

The rule I tested is taken from the examples on the autopairs project.  If you want to test the rule, you can use this code:
```lua
local cond = require('nvim-autopairs.conds')
npairs.remove_rule("(")
-- press % => %% is only inside comment or string
npairs.add_rules({
  Rule("%", "%", "lua")
    :with_pair(ts_conds.is_ts_node({'string','comment'})),
  Rule("$", "$", "lua")
    :with_pair(ts_conds.is_not_ts_node({'function'})),
  Rule("(", ")",{"tex", "latex","plaintex"})
    -- don't add a pair if  the previous character is xxx
    :with_pair(cond.not_before_regex_check("xxx", 3))
})
```

ChristianChiarulli:
interesting, I had an issue with nvimtree being optional. I'll give this a shot.

PR #361, opened 2021-05-04
Improvements

john-yan:
Hello Christian,

Please take a look.

The first commit is to swap setting and lv-setting execute order. In my understanding, personal changes will go into lv-settings. Please correct me if I am wrong.

The second commit fixes the arrow key. In my case, I found the left key will move the middle separator to the right, and right key to the left, which is very weird.

Thanks.


tarkin88 (approved):
**LGTM**
👍🏾

sozforex:
Greetings.

Previously `lua/settings.lua` used some variables that are defined in `lv-settings.lua`, with current load order it can not.

`O.number`, `O.relative_number`, `O.hidden_files`, `O.wrap_lines`

afridley:
@ChristianChiarulli @john-yan moving lv-settings bellow settings breaks autoformat on save for efm atleast

john-yan:
sorry about the breakage, I am not using autoformats. Maybe a better solution would be to create a separate file for customization? what do you think?

PR #367, opened 2021-05-05
Fix: Nvimtree controls don't work if initialized after colorscheme

rebuilt:
Reordered the require statements so that nvimtree is required before colorscheme.  I'm not sure why adding a colorscheme breaks the keybindings.  Nvimtree will not have the lunar colorscheme applied to it.  

I confirmed that autopairs still works.


rebuilt:
Accidentally pulled in readme file changes.  Will resubmit without those

Issue #365, opened 2021-05-05
Some nvim-tree keybinding stopped working


soerenmartius:
After pulling in the latest updates, some keybindings stopped working:

- Can't expand / close nested directories anymore with the normal `h` and `l` navigation
- `l` won't load files anymore

ChristianChiarulli:
yea, I think thats because I made it optional @rebuilt.

rebuilt:
Probably.

rebuilt:
I wonder if making them both required, instead of both optional will work.

rebuilt:
I just don't know enough about how packer works.  Is it because these plugins are in different root directories?  Is it because optional plugins load after required plugins that causes an issue?  Hard to say.

rebuilt:
I'm looking into it now.

rebuilt:
@soerenmartius Did you run 
```bash
:PackerSync
:PackerCompile
```
The most recent changes move nvim-tree to the optional folder and those commands need to be run to perform the move.
This is only required if you pulled in recent changes, not if you installed a clean build.  Can you let us know how you got the new version?

rebuilt:
I have no idea why this works.  Place 'require('lv-nvimtree')' above 'require('colorscheme')'. 

Nvimtree will not have the colorscheme applied to it but the controls will work as expected.  
Somthing about setting the color on it breaks the controls when it's lazy loaded.

soerenmartius:
> I have no idea why this works. Place 'require('lv-nvimtree')' above 'require('colorscheme')'.
> 
> Nvimtree will not have the colorscheme applied to it but the controls will work as expected.
> Somthing about setting the color on it breaks the controls when it's lazy loaded.

Yip that works. Shall I open a PR?

rebuilt:
@soerenmartius  I wanted to make some comments about further steps on the PR.  So I submitted it myself.

PR #368, opened 2021-05-05
Fix: Nvimtree navigation re-enabled by requiring it before colorschemes

rebuilt:
Fix: Nvimtree navigation re-enabled by requiring it before colorschemes.

Also added explanatory comments to the init file informing users about the order of requiring those files.  
Are there other files that need to be ordered in a particular manner?  Perhaps those reasons should be documented in the code.

Also, I think this colorscheme thing needs further investigation.  Should we move it to last to avoid it breaking other keymappings or will it break something else to move it after the other plugin configurations?  I'm not sure.  

Fixes: #365


lua/complete/init.lua
```diff
@@ -0,0 +1,12 @@
+-- Use <Tab> and <S-Tab> to navigate through popup menu
```
ChristianChiarulli, line 1: What does this do?

rebuilt: Crap.  That shouldn't be in there.  I guess I missed this when I was resolving a merge conflict.  I started to experiment with replacing compe with completion-nvim to get emmet stuff working for myself.  It shouldn't be in this pull.  lua/complete/init.lua  can be safely deleted.  The code you see in there is from the example setup on the completion-nvim readme.

PR #374, opened 2021-05-06
Add LSP support for scala language

kpritam:


kdav5758:
Tested it all in my fork and it was working fine. Although I think you should have mentioned in the description that:

1. You are using tabs... Which is not the "default" in LunarVim
2. You were adding a new provider to Galaxyline

abzcoding:
Closed in favor of #986 986

Issue #346, opened 2021-05-02
Error while calling lua chunk


Dunrar:
I tried to install LunarVim on my Ubuntu server suing the one-command option, but after installation, when I open Neovim, I get `ES113: Error while calling lua chunk: cannot open /home/user/.local/share/nvim/site/pack/packer/opt/nvim-lspconfig/lua/lspconfig/tailwindls.lua: No such file or directory`.

Dunrar:
Also, dependencies like ranger are not being installed.

rebuilt:
#340

rebuilt:
I don't know what uname returns on ubuntu server but if it doesn't have 'Ubuntu' in the text the script won't detect it.  
```bash
	[ -n "$(uname -a | grep Ubuntu)" ] && installnodeubuntu
```

What's the output of uname on Ubuntu server?


If you want to install the dependencies manually:  

```bash
        sudo apt install ripgrep fzf ranger
	sudo apt install libjpeg8-dev zlib1g-dev python-dev python3-dev libxtst-dev
        sudo apt install nodejs
	sudo apt install npm
	sudo apt install python3-pip 
	pip3 install ueberzug
	pip3 install neovim-remote
	npm install -g tree-sitter-cli
	pip3 install pynvim --user
```

Dunrar:
For me, uname returned 'Linux'.

Dunrar:
Installed the dependencies by hand, and most things seem to be working now. Will open another issue for something else.

simonfontana:
Same for me. Using Ubuntu 20.04 but but uname -a only showed Linux for me. Could it be because I'm using my own compiled Linux kernel? How about using lsb_release in the check instead? That works for me.

```
$ lsb_release -d
Description:	Ubuntu 20.04.2 LTS
```

Installing dependencies manually also worked for me.

PR #381, opened 2021-05-09
Use os-release file to check Ubuntu distro

simonfontana:
Previous check uname -a does not work if user has changed kernel.
Use os-release for true operating system identification.

Reference #346


simonfontana:
I noticed Darwin has the same check but I didn't dare change those without verifying the contents of /etc/os-release. If someone can check then I'll change those as well.

PR #382, opened 2021-05-09
Fix packer and nvim already installed check

simonfontana:
It was not possible to install LunarVim if these programs were already
installed. The check is syntactically incorrect and would always return
non-0 status. The installer would always continue to install packer and
nvim even though they were already installed. This failed with an error:
"fatal: destination path '<PATH>/packer.nvim' already exists and is not
an empty directory."

Correct the syntax error by using test -e to check if the file exists.


tarkin88:
Make sense. Thx 👍

Issue #198, opened 2021-03-30
PackerInstall is not a editor command


jilvanx:
I had a issue:
I ran the command:
`bash <(curl -s https://raw.githubusercontent.com/ChristianChiarulli/nvim/master/utils/installer/install.sh)`

![2021-03-30_16-52](https://user-images.githubusercontent.com/5150533/113048478-f5bf7380-9178-11eb-82e3-c74d043f67fa.png)

After that I opened 'nvim' and I can't execute `:PackerInstall` 
![2021-03-30_16-53](https://user-images.githubusercontent.com/5150533/113048733-3c14d280-9179-11eb-82e1-1d563c32bf8e.png)

What's going on ?

ChristianChiarulli:
Not sure, is the packer repo in your start directory?

Khwarazmi:
I ran the script without any problems, how ever when i try packerinstall i'm getting the same error
Not an editor command: PackerInstall

smikeblog:
IMO is better to remove function installpacker() from install script and install it from /lua/plugins.lua file to avoid surprises

jilvanx:
> Not sure, is the packer repo in your start directory?

~/.local/share/nvim/site/pack/packer/start/packer.nvim

jilvanx:
> I ran the script without any problems, how ever when i try packerinstall i'm getting the same error
> Not an editor command: PackerInstall

weiss-d:
Had the same problem.
Have manually deleted `~/.cache/nvim` and `~/.local/share/nvim` folders which seemed to contain some stuff from previous config installation (SpaceVim). Then reinstalled Neovim and it worked. Cannot say that both steps are necessary, but I didn't check in between :]

ChristianChiarulli:
This will hopefully be fixed when the new install script lands. Bear with me on this for a little while.

smikeblog:
> This will hopefully be fixed when the new install script lands. Bear with me on this for a little while.

IMO not many users will agree to install smth using sudo command. Cn alternative maybt to use several scripts as in Arco Linux 001-installglobal-sudo-pkgs.sh; 002-compileNvimFromGit.sh; 003-installPacker&pkgs.sh ...

jameswalmsley:
> > This will hopefully be fixed when the new install script lands. Bear with me on this for a little while.
> 
[...]

PR #392, opened 2021-05-13
Improve commands in README

kidonng:
Sorry for the bad title, I can't think of a better one.

LunarVim is a great project! However I think the commands offered in the README can be improved:

- Remove unnecessary `sudo` usage
- Improve manual compile instructions (do it in a temporary directory)
- Deleting only `site` directory when troubleshooting (so users don't lose history/swap/etc.)
- `-f` is required for `rm` when deleting plugins since `.git` may contain read only files

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,13 +121,12 @@ sudo pacman -S base-devel cmake unzip ninja tree-sitter
 Download and compile Neovim
 
 ``` bash
-cd ~
-sudo rm -r neovim
-git clone https://github.com/neovim/neovim
+cd $(mktemp -d)
+git clone https://github.com/neovim/neovim --depth 1
 cd neovim
 sudo make CMAKE_BUILD_TYPE=Release install
-cd ~
-sudo rm -r neovim
+cd ..
+rm -rf neovim
 ```
 
 or if you are on Arch you can get it from the AUR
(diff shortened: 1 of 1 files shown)


rebuilt (commented):
Removing only the packer directory works if we are only cleaning up LunarVim files.  There are old issues stemming from conflicting files from other configurations (SpaceVim).   https://github.com/ChristianChiarulli/LunarVim/issues/198#issuecomment-811658134_   Lvim also puts the packer directory elsewhere.  It's easier just to remove the entire nvim directory than to troubleshoot incompatibilities with old installations.

rebuilt:
Everything else looks good.  I learned you don't need sudo when -f will suffice.

kidonng:
> Removing only the packer directory works if we are only cleaning up LunarVim files. There are old issues stemming from conflicting files from other configurations (SpaceVim). [#198 (comment)](https://github.com/ChristianChiarulli/LunarVim/issues/198#issuecomment-811658134)_ Lvim (a different set of configurations) also puts the packer directory elsewhere. It's easier just to remove the entire nvim directory than to troubleshoot incompatibilities with old configurations.

Then I guess removing the entire `~/.local/share/nvim/site` will be enough.

rebuilt:
Ya.  I think you're right.  Deleting the 'site' directory should work.  I tested switching between two different packer configurations.  The trouble was in the 'compile' file ~/.local/share/nvim/site/lua/_compiled.lua. Deleting that file enables packer to do its job correctly.   But I still think we should delete the entire 'site' directory to be safe.  I also tested switching between packer and plug configurations.  In theory, they should never conflict but I just wanted to sanity-check.  Plug creates ~/.local/share/nvim/plugged.  Looks like leaving it in place doesn't hurt anything.

PR #416, opened 2021-05-21
Improves startup time

rebuilt:
Reduces startuptime from 165ms to 75ms (on my machine).
Uses packer's lazy loading.

Deletes configuration files from init.lua. Moves loading of configurations to plugins.lua.

nvim-dap did not seem to have any configuration so I didn't add anything to it. nvim-ts-autotag has a one line configuration so I added it inline. This enables auto-pairing of html brackets.

Pros: Reduces startup time. Eliminates conflicts like nvimtree needing to be loaded before colorschemes or else break its controls. (Fixes colorscheme for nvimtree). Finer control of when plugins start.

Cons: Complicates code for requiring plugins.

This pull request modifies the loading strategy of all plugins so it will need testing to make sure no plugin incompatibilities are introduced by this change.


ChristianChiarulli:
thanks for doing this @rebuilt, looks good so far.

rebuilt:
I've configured lazy loading for all plugins except the optional ones.  

I only have one problem left to fix but I'm stuck.  Rnvimr does not replace netrw when you launch it with 'nvim .'   I'm not sure how to set it up.  Even when I match the rnvimr config with config in master, it still doesn't open up ranger with a directory listing.  I've disabled netrw in lv-globals:  nvim_tree_disable_netrw = 1  
It will open ranger if you open all files and directories with 'nvim *'

rebuilt:
Everything is working now. I didn't touch the extra plugins except for markdownpreview.  I added an optional run command since the one in the original code didn't work for me. 

The original boot time with all the new plugins installed was about 190ms on my desktop.  Now it's 110.  On my laptop it went from 220 to 145.

luxus:
Thanks @rebuilt, awesome work

luxus:
does it got reverted? I cannot see the changes in master

rebuilt:
> does it got reverted? I cannot see the changes in master

I think he force pushed a later commit which overwrote the init and plugin files

rebuilt:
@luxus  After consideration, I think it was my fault. The last thing I should have done is rebase before marking it ready for merging.

PR #429, opened 2021-05-25
Add treesitter indicator to status line

tafryn:
It can be difficult to determine if treesitter is functioning in the current buffer. This indicator provides a quick signal to show treesitter status.


ChristianChiarulli:
@tafryn thanks!

PR #430, opened 2021-05-25
Enable the fzy file sorter for telescope

tafryn:
The current telescope configuration doesn't appear to actually use the fzy file sorter even though it intends to (override_file_sorter = true).

### DIFF
--- a/lua/lv-telescope/init.lua
+++ b/lua/lv-telescope/init.lua
@@ -15,7 +15,7 @@ require('telescope').setup {
         sorting_strategy = "descending",
         layout_strategy = "horizontal",
         layout_defaults = {horizontal = {mirror = false}, vertical = {mirror = false}},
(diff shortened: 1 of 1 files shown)


ChristianChiarulli:
@tafryn do you know if fzy would be mandatory for this to work? Or will it just default to something if the fzy binary is unavailable?

tafryn:
I don't think so. Removing `telescope-fzy-native.nvim` from `plugins.lua` and running PackerClean doesn't break anything. Also, `get_fzy_sorter` is defined in `telescope.nvim` and not `telescope-fzy-native.nvim`.

tafryn:
I think another option would be to add
```
require('telescope').load_extension('fzy_native')
```
to the end of telescope's `init.lua` and require the plugin in `plugins.lua`. But that creates more of a dependency on `telescope-fzy-native.nvim`.

PR #443, opened 2021-05-30
Enable auto format on save with vim api autocmd.

meowjesty:
This change enables auto formatting on save by using the vim api and creating an autocmd for the lsp (I have only tested it for rust).

I'm not really sure on how to allow the user to configure auto formatting for only certain file types.

### DIFF
--- a/lua/lsp/init.lua
+++ b/lua/lsp/init.lua
@@ -73,12 +73,8 @@ vim.lsp.protocol.CompletionItemKind = {
     "   (TypeParameter)"
 }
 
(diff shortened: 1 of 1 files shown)


rebuilt:
This functionality is already in lunarvim.  
Look at lv-autocommands/init.lua line 68.  

To use it with this version of lunarvim, enable rust language support and turn on efm language server.  There is also another setting to enable rust autoformatting but it's true by default so you don't need to change it.  

```
:LspInstall efm
```

PR #446, opened 2021-05-31
Add Sourcekit support

DanielOberg:
XCode uses LSP internally so all developers with XCode have an LSP server installed. It's also already included in [neovim/nvim-lspconfig](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/sourcekit.lua).


abzcoding:
Closed in favor of #1002

PR #448, opened 2021-06-01
added rust-tools

bob3000:


satu:
It looks like this patch doesn't make the vim commands available e.g. `RustRunnables`. Calling `:lua require('rust-tools.runnables').runnables()` works though. I've debugged this and the problem seems similar tp https://github.com/simrat39/rust-tools.nvim/issues/13 , and it's because the rust-tools executes its own registration to the lsp and registers the commands as part of it.

So I commented out `lsp/rust-ls.lua` and moved its logic in the initialisation of rust-tools.

```diff
diff --git a/lua/lsp/rust-ls.lua b/lua/lsp/rust-ls.lua
index 377a31e..db884ca 100644
--- a/lua/lsp/rust-ls.lua
+++ b/lua/lsp/rust-ls.lua
@@ -1,5 +1,5 @@
-require'lspconfig'.rust_analyzer.setup{
-    cmd = {DATA_PATH .. "/lspinstall/rust/rust-analyzer"},
-    on_attach = require'lsp'.common_on_attach
-}
+-- require'lspconfig'.rust_analyzer.setup{
+--     cmd = {DATA_PATH .. "/lspinstall/rust/rust-analyzer"},
+--     on_attach = require'lsp'.common_on_attach
+-- }

diff --git a/lua/lv-rust-tools/init.lua b/lua/lv-rust-tools/init.lua
index 58b3a2a..2a3904f 100644
--- a/lua/lv-rust-tools/init.lua
+++ b/lua/lv-rust-tools/init.lua
@@ -65,7 +65,10 @@ local opts = {
     -- all the opts to send to nvim-lspconfig
     -- these override the defaults set by rust-tools.nvim
     -- see https://github.com/neovim/nvim-lspconfig/blob/master/CONFIG.md#rust_analyzer
-    server = {}, -- rust-analyer options
+    server = {
+        cmd = {DATA_PATH .. "/lspinstall/rust/rust-analyzer"},
+        on_attach = require'lsp'.common_on_attach
+    }, -- rust-analyser options
 }

 require('rust-tools').setup(opts)
```

bob3000:
@satu Thank you! I applied your changes and also added some mappings for the commands which seemed useful to me. I wasn't completely sure which keybindings to assign or if it would be better to leave the keybindings entirely to everyone's personal customization. If the mappings do not comply with the project's guidelines I'm happy to remove or change them.

init.lua
```diff
@@ -60,7 +61,6 @@ require('lsp.lua-ls')
 require('lsp.php-ls')
 require('lsp.python-ls')
 require('lsp.ruby-ls')
-require('lsp.rust-ls')
```
satu, line 63: maybe leave it commented and leave an explanation why it's commented out.

satu (approved):

lua/lv-rust-tools/init.lua
```diff
@@ -0,0 +1,79 @@
...
+    -- see https://github.com/neovim/nvim-lspconfig/blob/master/CONFIG.md#rust_analyzer
+    server = {
+        cmd = {DATA_PATH .. "/lspinstall/rust/rust-analyzer"},
+        on_attach = require'lsp'.common_on_attach
+    }, -- rust-analyser options
+}
+
+require('rust-tools').setup(opts)
+
+vim.api.nvim_set_keymap("n", "<leader>lm", "<cmd>RustExpandMacro<CR>", {noremap = true, silent =  true})
```
abzcoding, line 76: Might i suggest adding this, to make sure it only shows in rust files:
```lua
if vim.bo.filetype == 'rust' then
-- your keymappings
end
```

bob3000: I like the idea, thanks. This only has one major drawback which is that the keybindings will only show up if the first file you open is a rust file since nvim reads the whole configuration only once when it starts. If you just start the editor without giving it a file path and then open a rust file there won't be any keybindings. I guess we probably would have to fix that with an auto command.

abzcoding: Yeah i agree, looking forward for this merge, thank you for your work :pray:

abzcoding: Is this any good?
```lua
vim.api.nvim_exec(
    [[
    autocmd Filetype rust nnoremap <leader>lm <Cmd>RustExpandMacro<CR>
    autocmd Filetype rust nnoremap <leader>lH <Cmd>RustToggleInlayHints<CR>
    autocmd Filetype rust nnoremap <leader>le <Cmd>RustRunnables<CR>
    autocmd Filetype rust nnoremap <leader>lh <Cmd>RustHoverActions<CR>
    ]], true)

```

bob3000: Good job! I added that. Maybe at some point the project could introduce a more global solution to dynamically add and remove key bindings depending on the file type but for the time being this solution is probably the best we can do. Thanks!

abzcoding:
@bob3000 i think we can close this in favor of #544

bob3000:
closed in favour of #544

PR #451, opened 2021-06-02
Add symbols outline plugin.

satu:


satu:
Having a nice outline is one of the thing I miss the most from VSCode. I feel that this plugin is appropriate for LunarVim

PR #453, opened 2021-06-02
Enable autoformatting in Rust.

satu:


satu:
Actually https://github.com/ChristianChiarulli/LunarVim/pull/448 supersede this request.

PR #462, opened 2021-06-05
Change Workspace Symbols binding to Telescope DynamicFinder

fabriziosestito:
This PR changes the which-key binding `<leader>lS` to the new Telescope DynamicFinder (`Telescope lsp_dynamic_workspace_symbols`), fixing https://github.com/ChristianChiarulli/LunarVim/issues/425 as suggested in the issue comments.


jameswalmsley (approved):
This provides a better behaviour especially for language servers that don't accept empty queries.

PR #472, opened 2021-06-08
gf-remote-fix: fix remote paths with netrw gf command while keeping b…

rmagillxyz:
…rowser gx commands working with remote urls

* better fix for netrw's gx command which does not break gf for remote paths


rmagillxyz:
I'm going to cancel this and submit a pr with the tiny tiny plugin to deal with more use cases.

Issue #465, opened 2021-06-06
`gf` not working when paths aren't relative


talefe:
https://user-images.githubusercontent.com/31940907/120911221-f2fb6600-c674-11eb-8e79-60e754011775.mov

talefe:
for context: `'os/application'` is a path within `src`

talefe:
@rebuilt can you help with this?

rebuilt:
You can add ** to your path:

```
set path+=**
```
This way it will find every file recursively based on your current directory. But apparently it's [not recommended.](https://twitter.com/tpope/status/612991667751264256)

If you want to add other paths, you can just add them as follow:

```
set path+=/path/to/your/lib
set path+=c:/include
```

Answer found [here](https://vi.stackexchange.com/questions/7453/access-a-file-under-subdirectories-of-a-path-through-gf-command)

I tested this in a react project.  It seems to work.  

Also read through [this](https://damien.pobel.fr/post/configure-neovim-vim-gf-javascript-import/)

rebuilt:
https://github.com/ChristianChiarulli/LunarVim/pull/472

rmagillxyz:
@talefe I just created a pull request which should fix this. The issue is that the `gx` command was not working correctly and I found a hack which fixed it, but didn't realize it broke `gf` for remote paths. Let me know if there are any other issues with netrw's `g` commands as I've been trying to fix this for awhile now and want to get this working correctly! 
Here is the pull request, there are only a few small changes: https://github.com/ChristianChiarulli/LunarVim/pull/472

rebuilt:
@rmagillxyz 

Thanks! Good work.

rmagillxyz:
@rebuilt There is a small plugin if any other issues arise. It may be more of a full-proof fix, but I haven't tested anything but my pr with `gx` and `gf`. There is a thread that goes back several years with people frustrated with this issue and people digging into netrw commits from over 20 years ago! 

[plugin](https://github.com/felipec/vim-sanegx.git)
[from issue](https://github.com/vim/vim/issues/4738#issuecomment-856925080)

rmagillxyz:
hmm. `gd`, `gf`, and `gx` all seem to be working properly for me (finally!). Did you also remove `vim.g.loaded_netrwPlugin = 1` from `lua/settings.lua`? 

[...]

PR #479, opened 2021-06-09
vim-sanegx: tiny plugin which deals with more use cases to the netrw_…

rmagillxyz:
…gx bug

Very tiny plugin to deal with the `netrw_gx` bug. Both `gx` and `gf` are working with remote paths for me.  

[plugin](https://github.com/felipec/vim-sanegx.git)
[from issue](https://github.com/vim/vim/issues/4738#issuecomment-856925080)


rmagillxyz:
#465

talefe:
I am using macOS Catalina. I still can't get it to work...

Issue #478, opened 2021-06-09
Update plugins.lua


pard68:
I started working on a branch to update plugins.lua, but decided I shouldn't put much work into this without some feedback from others.

It seems to me that the plugins installation method is way to complicated. It requires three separate lines in two different files (sometimes) to get a plugin working. I need a `use` line to install with packer. I need a `require_plugin` line to do... something... and then I need to `require` the plugin's own config file in `init.lua`. In my own neovim config that I was working on before LSP broke me and I just switched to LunarVim, I wrote my own `use` function, `Use` which called `use` and then called `require` if there was a dir/file matching the plugin's name, thus requiring only one line to add a plugin and not three. I wanted to add something like this for LunarVim. Thoughts?

Also, why does `require_plugin` exist? I understand what it does, but I cannot figure out why it exists.

rmagillxyz:
From my understanding the `require_plugin` is used to load that plugin, giving people the option to not load the plugin on initial load but to still have it on hand if needed. That is what the `opt = true` is for; if you look in `~/.local/share/nvim/site/pack/packer` you'll see `opt` and `start`, the plugins in `start` start without a `require_plugin` on initial load. Chris was trying to move away from other projects like `oh my zsh` which have tons of bloat that many people will not use when first starting out. I'm assuming that's why he added the `extras` section too. You have lots of ways stripping things up or down. I remember him saying his last config became a little too opinionated and thick, or something like that. As for your other questions, that wouldn't be up to me, but I hope this helps some.  @pard68

pard68:
Awesome, thanks for that info. That was enough to reason out where in the loading order it needs to be called. I'm a WIP PR to better understand my idea for loading plugins.

rebuilt:
You don't have to use require_plugin if you don't want.  You can just set your plugin to required and it will work.  

For an example of how to take advantage of packer's lazy loading, look here: https://github.com/ChristianChiarulli/LunarVim/pull/416/files

rebuilt:
I messed something up when I submitted it.  I think I should have rebased before marking it ready for merging.  A later commit reverted my changes.   But I'm not 100% sure if this was intentional or not.  I didn't get any feedback about it from Chris.  I'm not sure if he even knows it happened.  I've been trying to figure out if I broke something with that commit.  Someone else used it and said they changed compe to load at launch because the load time for it was noticeable on switching into insert mode.   but other than that, I think it works fine and it improves load time.

[...]

PR #480, opened 2021-06-09
WIP: Added basic plugin use/load/require function

pard68:
`use_maker` takes Packer's `use` and returns a function which takes
Packer's `plugin_settings` table. This function is then used to call
Packer's own `use`, then determines if the plugin is lazy loaded and if
it has a config that needs to be `require`'d. Due to execution order, we
don't call `require_plugin` or `require` immediately, but add the plugin
to a table. This table is then called at the end of `plugins.lua` and
will first "pre-load" the plugin if it's being loaded lazily (ie `opt =
true`) and it will call `require` if it was determined that this plugin
does in fact have a configuration that needs to be required.

The one major issue at the moment is that the logic used to dertermine
if `require` should be called is not very robust. It only is going to
look for directories in our `./lua` dir that are named after the
plugin, sans any extensions (ie for `nvim-treesitter/nvim-treesitter` it
looks for the `lv-nvim-treesitter` dir and for `folke/trouble.nvim` it
looks for the `lv-trouble` dir). Either more robust logic can be added
to account for a variety of potential directory names or a single
directory naming scheme would need to be adapted.

Additional work that probably needs to be done prior to a merge:

- removing `require` statements from `init.lua` for all plugins, but this
  cannot be done until the above issue is resolved
- move helper functions into a `utils` file to clean up the
  `plugins.lua` file, which in my opinion should ideally just be the
  minimum boilerplate needed to `Use` a plugin
- I'd love to make use of the Packer optionals like `config` and
  `setup`, however at the moment it seems to be an issue due to load
  order, some plugins need other plugins, not sure if this is even an
  option, but it might be worth looking into

### DIFF
--- a/lua/plugins.lua
+++ b/lua/plugins.lua
@@ -8,6 +8,10 @@ if fn.empty(fn.glob(install_path)) > 0 then
     execute "packadd packer.nvim"
 end
 
+local plugin_count = 1
+local load_index = {}
+local load_list = {}
+
 --- Check if a file or directory exists in this path
 local function require_plugin(plugin)
     local plugin_prefix = fn.stdpath("data") .. "/site/pack/packer/opt/"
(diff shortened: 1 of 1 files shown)


pard68:
Marking this as WIP. There are two or three things that need to be addressed and I do not want to make the decision on them as it's a bit of a major change and this isn't my project. This PR corresponds with #478

pard68:
It looks like #416 might be a better choice. Seems that it was merged last week and a subsequent merge/commit might have reverted it. I cannot find the merge in the commit history at all though, so I am unsure what happened there. Since I can't tell how/why it isn't in the code base currently I am going to keep this open, but I do think that PR offers a far more robust solution.

pard68:
It looks like #416 might be a better choice. Seems that it was merged last week and a subsequent merge/commit might have reverted it. I cannot find the merge in the commit history at all though, so I am unsure what happened there. Since I can't tell how/why it isn't in the code base currently I am going to keep this open, but I do think that PR offers a far more robust solution.

luxus:
@pard68 can we close this?

rebuilt:
Thank you for your work.  Things have changed radically since this pr was made so this won't be needed.

PR #485, opened 2021-06-12
Fix: Correctly require FTerm in plugins.lua

OliverNChalk:
It looks like `master` is not correctly loading `FTerm`, this is not a major issue because `FTerm` is not used in the base config. However, if people start to dip into `extras` they will get caught by this like I did


maxime50:
I just found out that nvim-spectre is also incorrectly loaded. Should I make a separate PR for it?

Tastyep:
IMO this should belong to the same commit.
While at it, could we be consistent with the use of ' and " for the plugins file.

OliverNChalk:
I'm happy to draft a commit that resolves all of the extra deps correctly, or you can @maxime50 , if you're happy for me to do it, can you link me your commit where you fix `nvim-spectre`?

PR #519, opened 2021-06-21
WhichKey: Make <leader>h clear search instead of toggling highlight

OliverNChalk:
This may be controversial because it changes the functionality from "toggle highlighting forever" to "clear highlighting / current search"

I think the latter is the more natural behaviour, however, feel free to reject this change. There is also the potential to bind the clear search to `<leader>n` to keep it in line with `n/N` keybinds for next/prev search.


tafryn:
I also prefer this style of clearing the current search.

rebuilt:
Another option is to make escape clear search and spelling highlights
```
:nohls | :setlocal nospell<ESC>
```

PR #523, opened 2021-06-22
Fix barbar integration with nvim-tree

kylo252:
- Add an offset to display NvimTree buffer in barbar.
_It won't work correctly if you enable "vim.g.nvim_tree_quit_on_open"._
- Add which-key support to barbar to enable missing mappings.

### DIFF
--- a/lua/lv-which-key/init.lua
+++ b/lua/lv-which-key/init.lua
@@ -51,7 +51,7 @@ vim.g.mapleader = ' '
 vim.api.nvim_set_keymap('n', '<Leader>h', ':set hlsearch!<CR>', {noremap = true, silent = true})
 
 -- explorer
-vim.api.nvim_set_keymap('n', '<Leader>e', ':NvimTreeToggle<CR>', {noremap = true, silent = true})
+vim.api.nvim_set_keymap('n', '<Leader>e', ":lua require'lv-nvimtree'.toggle_tree()<CR>", {noremap = true, silent = true})
 
 -- telescope
 vim.api.nvim_set_keymap('n', '<Leader>f', ':Telescope find_files<CR>', {noremap = true, silent = true})
@@ -78,6 +78,16 @@ local mappings = {
     ["f"] = "Find File",
     ["h"] = "No Highlight",
     ["p"] = "Projects",
+    b = {
+      name = "+Buffers",
+      j = {"<cmd>BufferPick<cr>", "jump to buffer"},
+      w = {"<cmd>BufferWipeout<cr>", "wipeout buffer"},
+      e = {"<cmd>BufferCloseAllButCurrent<cr>", "close all but current buffer"},
+      h = {"<cmd>BufferCloseBuffersLeft<cr>", "close all buffers to the left"},
+      l = {"<cmd>BufferCloseBuffersRight<cr>", "close all BufferLines to the right"},
+      D = {"<cmd>BufferOrderByDirectory<cr>", "sort BufferLines automatically by directory"},
+      L = {"<cmd>BufferOrderByLanguage<cr>", "sort BufferLines automatically by language"},
+    },
     d = {
         name = "+Diagnostics",
         t = {"<cmd>TroubleToggle<cr>", "trouble"},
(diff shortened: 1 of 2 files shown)


lua/lv-barbar/init.lua
```diff
@@ -1,3 +1,28 @@
...
+local wk = require('which-key')
+local opts = {
+    mode = "n", -- normal mode
+    buffer = nil, -- global mappings. specify a buffer number for buffer local mappings
+    silent = true, -- use `silent` when creating keymaps
+    noremap = true, -- use `noremap` when creating keymaps
+    nowait = false -- use `nowait` when creating keymaps
+}
+
+local mappings = {
```
Tastyep, line 10: IMO, either we centralize mappings in which-key or we put them in the specific plugin, but a mix of both is bad.

Centralizing in which key offers a 'all in one place' solution, which is nice because you only have one file to maintain.
Centralizing in the plugin file makes it easier to load/unload the plugin because the which-key file doesn't need to be updated.

I personally prefer the first option.

Tastyep (commented):
It would be nice to have someone with more seniority on the project comment on the mapping matter

lua/lv-barbar/init.lua
kylo252: I think there are merits to both approaches, when you're defining custom functions then you might want to have the mappings closer since you're more likely to make more modifications. For default mappings, it doesn't matter much I guess.

I didn't notice that the project is now following a more centralized layout, so I'll adjust the commit to keep the pattern :)

lua/lv-which-key/init.lua
```diff
@@ -74,10 +71,24 @@ vim.api.nvim_set_keymap('n', '<leader>p', ":lua require'telescope'.extensions.pr
...
     ["/"] = "Comment",
     ["c"] = "Close Buffer",
-    ["e"] = "Explorer",
+    ["e"] = {"<cmd>lua require\"lv-nvimtree\".toggle_tree()<CR>", "Explorer"},
     ["f"] = "Find File",
     ["h"] = "No Highlight",
     ["p"] = "Projects",
+    ['<TAB>']   =  {'<cmd>BufferNext<cr>', 'jump to next buffer'},
+    ['<S-TAB>'] =  {'<cmd>BufferPrevious<cr>' , 'jump to prev buffer'},
+    ['<S-x>']   =  {'<cmd>BufferClose<cr>', 'close buffer'},
```
Tastyep, line 80: There is already a buffer close binding, see ["c"]

lua/lv-which-key/init.lua
```diff
@@ -74,10 +71,24 @@ vim.api.nvim_set_keymap('n', '<leader>p', ":lua require'telescope'.extensions.pr
 local mappings = {
     ["/"] = "Comment",
     ["c"] = "Close Buffer",
-    ["e"] = "Explorer",
+    ["e"] = {"<cmd>lua require\"lv-nvimtree\".toggle_tree()<CR>", "Explorer"},
```
Tastyep, line 74: Would it also work with single quotes instead of doing \\"\\" ?

Tastyep (changes requested):

lua/lv-which-key/init.lua
```diff
@@ -74,10 +71,24 @@ vim.api.nvim_set_keymap('n', '<leader>p', ":lua require'telescope'.extensions.pr
...
     ["f"] = "Find File",
     ["h"] = "No Highlight",
     ["p"] = "Projects",
+    ['<TAB>']   =  {'<cmd>BufferNext<cr>', 'jump to next buffer'},
+    ['<S-TAB>'] =  {'<cmd>BufferPrevious<cr>' , 'jump to prev buffer'},
```
Tastyep, line 79: I would also argue that these two are not necessary since already present in [keymappings.lua](https://github.com/ChristianChiarulli/LunarVim/blob/master/lua/keymappings.lua)

lua/lv-which-key/init.lua
kylo252: Interesting! I basically kept the same bindings from [lua/lv-barbar/init.lua](https://github.com/ChristianChiarulli/LunarVim/blob/19bd48ac809cf7e17ac5f56acf0b4153f311e7e1/lua/lv-barbar/init.lua). So I guess it was a duplicate all along. Not sure which one people use more often tho.

lua/lv-which-key/init.lua
kylo252: Those are not `barbar` commands btw, and I'm also surprised that we have yet another dupe 😞 

My idea was to add some extra functionality and not really refactor anything in particular. So now the latest patch contains only a minimal change that provides some missing keybindings, while not really bothering with unifying/moving anything else.

PR #535, opened 2021-06-27
Update Whichkey normal, visual and extra keymapings

williamgoulois:
Hi ChristianChiarulli,
I was in your youtube stream yesterday and would like to give my first contribution to LunarVim (also my first contribution to OpenSource !)

I updated the whichkey keymappings and require lspsaga keymappings if extras :

- [X] Cleanup normal mappings
- [X] Add visual mappings
- [X] Update extra mappings according to your last commit
- [X] Add lspsaga keymappings if extras


williamgoulois:
During my experimentation i saw a broken keybinding due to the removal of plugins in your last commit.
A solution could be to `require` specific keymapping if `O.extras` is `true`.

In the end to make LunarVim completely modular we could group the require plugin logic with plugin keymappings and plugin whichkey entries !

What do you think of that ?

lua/lv-which-key/init.lua
```diff
@@ -52,39 +52,24 @@ local opts = {
...
-    ["c"] = "Close Buffer",
-    ["e"] = "Explorer",
-    ["f"] = "Find File",
-    ["h"] = "No Highlight",
-    ["p"] = "Projects",
+    ["/"] = {"<cmd>CommentToggle<cr>", "Comment"},
+    ["c"] = {"<cmd>BufferClose<cr>", "Close Buffer"},
+    ["e"] = {"<cmd>NvimTreeToggle<cr>", "Explorer"},
+    ["f"] = {"<cmd>Telescope find_files<cr>", "Find File"},
+    ["h"] = {"<cmd>set hlsearch!<cr>", "No Highlight"},
```
Tastyep, line 72: The command should be ":let @/=''\<CR>"

lua/lv-which-key/init.lua
```diff
@@ -52,39 +52,24 @@ local opts = {
...
 
-    ["/"] = "Comment",
-    ["c"] = "Close Buffer",
-    ["e"] = "Explorer",
-    ["f"] = "Find File",
-    ["h"] = "No Highlight",
-    ["p"] = "Projects",
+    ["/"] = {"<cmd>CommentToggle<cr>", "Comment"},
+    ["c"] = {"<cmd>BufferClose<cr>", "Close Buffer"},
+    ["e"] = {"<cmd>NvimTreeToggle<cr>", "Explorer"},
```
Tastyep, line 70: The command should be ":lua require'lv-nvimtree'.toggle_tree()\<CR>"

lua/lv-which-key/init.lua
```diff
@@ -198,27 +178,26 @@ if O.extras then
...
+        f = {"<cmd>lua require('spectre').open_file_search()<cr>", "Current File"},
+        p = {"<cmd>lua require('spectre').open()<cr>", "Project"}
+    }
+    normalMappings["z"] = {
+        name = "Zen",
+        s = {"<cmd>TZBottom<cr>", "toggle status line"},
+        t = {"<cmd>TZTop<cr>", "toggle tab bar"},
+        z = {"<cmd>TZAtaraxis<cr>", "toggle zen"}
+    }
+    normalMappings[";"] = {"<cmd>Dashboard<cr>", "Dashboard"}
```
Tastyep, line 193: The dashboard is not an extra plugin, group the mapping with the other ones above

lua/lv-which-key/init.lua
```diff
@@ -198,27 +178,26 @@ if O.extras then
         t = {"<cmd>VimtexTocToggle<cr>", "Toggle Table Of Content"},
         v = {"<cmd>VimtexView<cr>", "View PDF"}
     }
+    normalMappings["p"] = {"<cmd>lua require'telescope'.extensions.project.project{}<cr>", "Projects"}
```
Tastyep, line 181: Telescope is not an extra dependency, please group with the ones above.

lua/lv-which-key/init.lua
```diff
@@ -198,27 +178,26 @@ if O.extras then
...
+    normalMappings["r"] = {
+        name = "Replace",
+        f = {"<cmd>lua require('spectre').open_file_search()<cr>", "Current File"},
+        p = {"<cmd>lua require('spectre').open()<cr>", "Project"}
+    }
+    normalMappings["z"] = {
```
Tastyep, line 187: I will stop commenting for every plugin that's not part of extra, please review your work again and only configure extra plugins in this condition.

lua/lv-which-key/init.lua
```diff
@@ -179,17 +134,42 @@ local mappings = {
...
-    z = {
-        name = "Zen",
-        s = {"<cmd>TZBottom<cr>", "toggle status line"},
-        t = {"<cmd>TZTop<cr>", "toggle tab bar"},
-        z = {"<cmd>TZAtaraxis<cr>", "toggle zen"}
-    }
+}
+
+local visualMappings = {
+    ["/"] = {":'<,'>CommentToggle<cr>", "Comment"},
```
Tastyep, line 140: This is indeed the solution, replacing \<cmd> with :, but did you understand the rational ?
You only need to do ":CommentToggle<cr>"

lua/lv-which-key/init.lua
```diff
@@ -179,17 +134,42 @@ local mappings = {
...
-    -- extras
-    z = {
-        name = "Zen",
-        s = {"<cmd>TZBottom<cr>", "toggle status line"},
-        t = {"<cmd>TZTop<cr>", "toggle tab bar"},
-        z = {"<cmd>TZAtaraxis<cr>", "toggle zen"}
-    }
+}
+
+local visualMappings = {
```
Tastyep, line 139: Please add a note above this which says:
-- NOTE: Prefer using : over \<cmd> in the command expression as the later avoids going back in normal-mode.
-- see https://neovim.io/doc/user/map.html#:map-cmd

lua/lv-which-key/init.lua
```diff
@@ -198,27 +178,26 @@ if O.extras then
...
+    normalMappings["z"] = {
+        name = "Zen",
+        s = {"<cmd>TZBottom<cr>", "toggle status line"},
+        t = {"<cmd>TZTop<cr>", "toggle tab bar"},
+        z = {"<cmd>TZAtaraxis<cr>", "toggle zen"}
+    }
+    normalMappings[";"] = {"<cmd>Dashboard<cr>", "Dashboard"}
+    visualMappings["r"] = {
+        name = "Replace",
+        f = {"<cmd>lua require('spectre').open_visual({path = vim.fn.expand('%')})<cr>", "File"},
```
Tastyep, line 196: Spectre will have the same problem as CommentToggle, replace \<cmd> with : so that it catches the selection.

lua/lv-which-key/init.lua
```diff
@@ -138,28 +105,16 @@ local mappings = {
     },
     l = {
         name = "LSP",
-        a = {"<cmd>Lspsaga code_action<cr>", "Code Action"},
```
Tastyep, line 141: You removed entries here, but then completely duplicated the entry below, this needs to be reworked.

Tastyep (commented):
Hello, I was following up on [#520](https://github.com/ChristianChiarulli/LunarVim/pull/520), but when opening the PR I saw you did try first. I left some comments, there are some things that should be fixed first before this could get merged.

Also, I understood you tried to split configurations for extra plugins, but you mixed in together many plugins that are not part of the extras, did I miss something ?

lua/lv-which-key/init.lua
williamgoulois: I merged the new keys with the already defined ones !

williamgoulois:
Hi @Tastyep thanks for reviewing !

> Also, I understood you tried to split configurations for extra plugins, but you mixed in together many plugins that are not part of the extras, did I miss something ?

Yes i think you did not see this commit https://github.com/ChristianChiarulli/LunarVim/commit/aa78accb6fbe76b8e257909ab301b672f613002d 
It removed :
- dashbord 
- trouble
- lspsaga
- telescope-project (i kept other telescope outside of extras)

During my edits i used some commands that were outdated (they were in comments and i assumed they were correct) and will use the ones you reviewed !

> but did you understand the rational ?

I think i do not understand the question ! 😄

lua/lv-which-key/init.lua
williamgoulois: it is actually (see comment below)

lua/lv-which-key/init.lua
williamgoulois: see comment below telescope-project is extra

williamgoulois:
I just rebased my branch from upstream

lua/lv-which-key/init.lua
```diff
@@ -52,43 +52,28 @@ local opts = {
...
-    ["/"] = "Comment",
-    ["c"] = "Close Buffer",
-    ["e"] = "Explorer",
-    ["f"] = "Find File",
-    ["h"] = "No Highlight",
-    ["p"] = "Projects",
-    ["z"] = "Zen",
+    ["/"] = {"<cmd>CommentToggle<cr>", "Comment"},
+    ["c"] = {"<cmd>BufferClose<cr>", "Close Buffer"},
+    ["e"] = {":NvimTreeToggle<cr>", "Explorer"},
```
Tastyep, line 73: For consistency here we should use <cmd> (sorry, while the commands needed to be updated, I didn't see you used <cmd> everywhere else)

Tastyep:
Hello [**@williamgoulois**](https://github.com/williamgoulois), I'm happy to help, even though for this one I had an outdated version of the code while reviewing which lead to inaccurate comments.
Sorry about that and good work !

williamgoulois:
Seems like Christian is making a lot of changes at the moment so i'll wait a little bit to rebase and make changes 😄

PR #556, opened 2021-06-30
Fixes broken installs due to colorscheme changes

rebuilt:
The colorscheme is set in lv-config.lua before plugins are installed.  This only works if the plugin has already been installed.  All new users are getting errors because the config is trying to set the colorscheme before the colorscheme has been installed.  

This request moves the colorscheme setting back to it's own file and requires it after plugins are installed.


rebuilt:
https://github.com/ChristianChiarulli/LunarVim/issues/554

PR #564, opened 2021-06-30
lazyload telescope

kylo252:
Use lazy-loading for telescope to gain some performance.


rebuilt:
I tested this pull request.  Lazy loading telescope breaks telescope-project.nvim.  You'll need to load telescope before telescope-project.nvim.  

<space>p is the shortcut for telescope-project.  It only works after the first time telescope is used.  Otherwise it fails.

rebuilt:
I'm not sure how to verify telescope-fzy-native.nvim is working correctly

rebuilt:
This seems to prevent telescope-project from blowing up.
```
        after = "telescope.nvim",
```

ChristianChiarulli:
@rebuilt with any luck fzy is just ignored if not present, @kylo252 can you add the after field for telescope project so we can merge this?

kylo252:
> @rebuilt with any luck fzy is just ignored if not present, @kylo252 can you add the after field for telescope project so we can merge this?

Sorry about not getting back to you in time. I was away all day yesterday and I didn't have access to my computer. 

I don't use fzy but I didn't need to make any adjustments to `telescope-fzf-native.nvim` and it appears to work fine.

PR #565, opened 2021-06-30
nvimtree new keymap configuration

PBird:
there is a new update on nvim-tree that has changed nvim_tree_bindings object structure. I fixed.


ChristianChiarulli:
I think I updated this at the end of the stream

PR #573, opened 2021-07-01
the colorscheme part that you talked about in the stream

pulme:
:-) thanks for your work!! on this


rebuilt (commented):
You can't set the colorscheme before plugins because on fresh installs, the colorscheme does not yet exist.   

These changes will break new installs
See this issue: https://github.com/ChristianChiarulli/LunarVim/issues/554

rebuilt:
To test it yourself, delete the plugins to simulate starting from scratch

```
# Delete the plugins
rm -Rf ~/.local/share/nvim
```

And then try to install plugins.

pulme:
Could we write this in init.lua?

PR #578, opened 2021-07-01
Change back to TrueZen.nvim

### LINKED ISSUE #568: "new zen mode, other one didn't work with lazy loading" - Why can't you lazy load TrueZen.nvim?
Hello again. I'm a little bit concerned by this because in my case lazy loading the plugin does work as shown below:

https://user-images.githubusercontent.com/58336662/124066700-c4b44080-d9fe-11eb-9a42-1daa31a032aa.mp4

As you can see I'm lazy loading TrueZen by setting `cmd  = {"TZAtaraxis", "TZMinimalist", "TZFocus"}`. After any of those commands gets executed the config gets loaded, which should print `ran -------------------`. If the plugin wasn't lazy loaded then `ran -------------------` would have appeared after starting nvim rather than after running `TZAtaraxis`.

Did you try [TrueZen.nvim](https://github.com/Pocco81/TrueZen.nvim) with the latest commits? If so, was the problem still there?

(Btw, the plugin was refactored completely. I'd recommend you try it out)

Edit: I just tried lazy loading the "older version" of TrueZen and it works just fine.

abzcoding:
Fixes #568


rebuilt:
Thanks for your effort on this.  We are going in a different direction with plugin installation.  LunarVim will only include a small set of predefined plugins.  Everything else will be loaded from lv-config.lua using the custom_plugins table.  

```
# lv-config.lua
# Other, user-defined plugins will be another entry in this table
```

What we'll do is add a plugin section on the wiki with example packer configurations for individual plugins.  We'll use the configuration you came up with for TrueZen.   Those wiki pages have not yet been created.  

I'll also talk to Chris about populating the lv-config.example.lua file with those configurations as well.

abzcoding:
@rebuilt sure, feel free to close this issue after integrating it in the wiki

rebuilt:
This is now in the wiki.  https://github.com/ChristianChiarulli/LunarVim/wiki/Example-plugin-configurations

rebuilt:
This is now in the wiki.  https://github.com/ChristianChiarulli/LunarVim/wiki/Example-plugin-configurations

abzcoding:
@rebuilt  may i suggest adding
"```lua"
to make the code blocks look better

rebuilt:
I added the 'lua' but it didn't make a difference so I manually formatted the sections with '='

PR #580, opened 2021-07-02
feat: add custom hooks for user settings

mgutz:
This project is a wonderful base. I can finally stop over obsessing with my personal configuration :) 

I see there are todo comments about adding user customizability. This is a quick stab it. A user defines a set of hook functions that are called at various points during startup. The script is `lua/lv--custom/init.lua` which is named for sorting purposes and make it clear it's not  a maintained folder.

There are likely other places that require hooks integration, but these are the ones I've needed thus far.


luxus:
custom hooks are now available in rolling.
maybe this can be closed now.
greetings

PR #582, opened 2021-07-02
Add snap as an optional alternative to telescope

camspiers:
Worth testing it first of course, but this is a first pass at adding snap as an optional plugin, if enabled it will use the same bindings as Telescope does, at least for the a subset of Telescopes features. I can follow up with the other features that Snap supports and more, like grepping under cursor, grepping selection, etc.

It works for me locally, but please test.

### DIFF
--- a/lua/default-config.lua
+++ b/lua/default-config.lua
@@ -50,6 +50,7 @@ O = {
         codi = {active = false},
         telescope_fzy = {active = false},
         sanegx = {active = false},
+        snap = {active = false},
         ranger = {active = false},
         todo_comments = {active = false},
         lsp_colors = {active = false},
--- a/lua/lv-snap/init.lua
+++ b/lua/lv-snap/init.lua
@@ -1,0 +1,14 @@
+local M = {}
+
+M.config = function()
+    local snap = require "snap"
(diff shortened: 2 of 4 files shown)


kylo252:
I've been using `snap` alongside `telescope` for a while now and it's absolutely an amazing plugin so thanks for that!

A few comments: 

1- I had no idea about this new `snap.register.command` and how easily it can replace/replicate a lot of the functionalities of `telecope`. I do however have a minor complaint that it's hard to grasp and it doesn't help that there's no mention of it whatsoever in the main README.

2- I'm not sure if it's worth it to replace the bindings for `telescope` but rather add another one like `<leader>s` instead to invoke `snap`. That way we can preserve `telescope` just in case we want to compare the output of the two, or maybe in case one of them breaks. I do understand the point of avoiding redundancy though.

3- It would be really nice to setup lazy-loading with `packer`. Since we have the rare opportunity to have the dev personally see if there are any issues with that. I have tried it like this and I think it's fine (?)
```lua
  use {
    "camspiers/snap",
    cmd = "Snap",
    config = [[require('config.snap')]]
  }
```

Issue #62, opened 2020-09-26
Cannot use vim-surround


fairulazmin:
Hi Chist,

Love your work. I'm really new to nvim. I coundn't use the vim-surround. The PlugStatus is OK. Then I copy tpope/vim-surround.vim into plug-config folder and source it in init.vim but still not working. Please advise. Thanks.

ChristianChiarulli:
I don't think I have vim surround working right now. I'll leave this open until I fix it.

riolly:
Me too have this issue. Is it because I using WSL2?

pviniciusfm:
@fairulazmin  In my case, it was caused by this property: 
https://github.com/ChristianChiarulli/nvim/blob/master/general/settings.vim#L35 after I increased the timeout for me everything worked fine.

fairulazmin:
Thanks @pviniciusfm. I changed the timeoutlen to 1000 and it works perfectly fine.

sundy-li:
@fairulazmin  Yes, it works, but `space leader key` cause too much time to respond.

ChristianChiarulli:
tracking solution to issue here: https://github.com/liuchengxu/vim-which-key/issues/158

rafamadriz:
Closing because of inactivity.

rosborn:
Encountered this same problem. The fix that worked for me was setting `keys` properly when loading the plugin. For some reason [this](https://www.lunarvim.org/plugins/02-extra-plugins.html#vim-surround) page suggests the wrong mapping for `keys`. Here is the correct mapping which should give you the behavior you are looking for.
```
 keys = {
  { 'n', 'ds' },
  { 'n', 'cs' },
  { 'n', 'cS' },
  { 'n', 'ys' },
  { 'n', 'yS' },
  { 'n', 'yss' },
  { 'n', 'ySs' },
  { 'n', 'ySS' },
  { 'x', 'S' },
  { 'x', 'gS' },
  { 'i', '<C-S>' },
  { 'i', '<C-G>s' },
  { 'i', '<C-G>S' }
}
```

saimonmoore:
Has anyone got this to work? I've always used vim-surround in vim and badly miss it. I'm trying out lunarvim but there's no way to get it to work properly.

I tried out your suggestion @rosborn . Didn't have any effect. (I'm trying to do `cs"'`)

[...]

PR #590, opened 2021-07-02
Add surround.nvim

colinaaa:
default to use sandwich mapping which will override the mapping to hop's mapping.

maybe should use surround mapping. see https://github.com/blackCauldron7/surround.nvim#configuration

re #62

### DIFF
--- a/lua/default-config.lua
+++ b/lua/default-config.lua
@@ -62,6 +62,7 @@ O = {
         diffview = {active = false},
         bracey = {active = false},
         telescope_project = {active = false},
+        surround = {active = false},
         dap_install = {active = false}
 
     },
--- a/lua/plugins.lua
+++ b/lua/plugins.lua
@@ -406,6 +406,15 @@ return require("packer").startup(function(use)
         event = "BufRead",
         disable = not O.plugin.dap_install.active
     }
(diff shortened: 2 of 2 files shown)


IndianBoy42:
Does surround.nvim support custom surroundings to the extent that vim-sandwhich has? the latex addons for vim-sandwhich (like surround with environment) are invaluable to me

colinaaa:
> Does surround.nvim support custom surroundings to the extent that vim-sandwhich has?

I don't think so.

IndianBoy42:
So what do we prefer? lua plugins or vim-sandwhich for more features. I can give a PR for vim-sandwhich with the latex addons

rebuilt:
Thanks for your work on this. We are going in a different direction with plugin installation. LunarVim will only include a small set of predefined plugins. Everything else will be loaded from lv-config.lua using the custom_plugins table.

# lv-config.lua
# Other, user-defined plugins will be another entry in this table

What we'll do is add a plugin section on the wiki with example packer configurations for individual plugins. We'll use the configuration you came up with for Surround. Those wiki pages have not yet been created.

I'll also talk to Chris about populating the lv-config.example.lua file with those configurations as well.

Issue #193, opened 2021-03-30
Add Deno LSP


hemedani:
It's added to [LspInstall](https://github.com/kabouzeid/nvim-lspinstall) so it can easily installed.
add some configuration to avoid conflict with tsserver something like vim `coc-settings.json` or vscode `settings.json` file with these details :
```
{
  "deno.enable": true,
  "deno.lint": true,
  "deno.unstable": true
}
```
and maybe add some `which-key` for `DenolsCache`. [this is the denols configuration](https://github.com/neovim/nvim-lspconfig/blob/master/CONFIG.md#denols)

ChristianChiarulli:
Hopefully someone considers putting a PR in soon, I'll gladly accept it.

eduardjbotha:
Hey @ChristianChiarulli, this is my first PR attempt on Github, so please bear with me.
From what I can gather, the steps are

1. clone the repo
2. create a new branch
3. make the change
4. push the branch
5. click "Create Pull Request" in the browser

I'm stuck at point 4 with a 403. Is this the correct approach?
If so, could you grant me the necessary rights, so that I can create the PR?

abzcoding:
you can use `deno` through `null-ls`
```lua
lvim.lang.javascript.formatters = {
  {
   -- can be prettier or prettierd or eslint or eslint_d or 
    exe = "deno_fmt",
    args = {},
  },
}
```

jackmac92:
@abzcoding is that full deno lsp or just the formatting?

abzcoding:
> @abzcoding is that full deno lsp or just the formatting?

just formatting i assume, but you can use the deno ls as well, wait a bit ...

abzcoding:
ok so you need sth like this in your `config.lua`
```lua
lvim.builtin.lspinstall.on_config_done = function()
  lvim.builtin.rooter.active = false
  lvim.lang.javascript.lsp = {
    provider = "denols",
    setup = {
      cmd = {
        DATA_PATH .. "/lspinstall/deno/bin/deno",
        "lsp",
      },
      filetypes = { "javascript", "javascriptreact", "javascript.jsx", "typescript", "typescriptreact", "typescript.tsx" },
      init_options = {
        enable = true,
        lint = false,
        unstable = false,
      },
      root_dir = require("lspconfig").util.root_pattern("package.json", "tsconfig.json", ".git"),
      on_attach = require("lsp").common_on_attach,
      on_init = require("lsp").common_on_init,
[...]

PR #595, opened 2021-07-02
Added denols configuration and a which-key entry.

eduardjbotha:
#193 
If the default-config has been changed to activate deno, the which-key option is available. I'm not sure if it's possible to have an optional which-key entry, based on the active LSP.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -10,3 +10,4 @@ require('lv-treesitter')
 require('lv-which-key')
 require('lsp')
 if O.lang.emmet.active then require('lsp.emmet-ls') end
+if O.lang.deno.active then require('lsp.deno-ls') end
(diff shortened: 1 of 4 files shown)


ChristianChiarulli:
this is somewhat out of date with the new way of doing things. which ftplugin file should this go in?

eduardjbotha:
Ok, so I've managed to from from an out-of-date version of your repo. I see what you mean with the new way of doing things.
I've gone through the new setup and tested it locally by addin the config to a new file in /lua/lsp. 

What I find confusing is the ftplugin directory, with regard to javascript and typescript (normal and react). Those four files seem to contain the same configuration. My first reaction is to attempt configuring deno in these files as well, since it only applies to these file types. I did, however, manage to get it to work by adding the /lua/lsp configuration and adding another property to O.lang.

I've also read your wiki, trying to understand the relationship/difference between /lua/ftplugin and lua/lsp, but couldn't find an explanation. Could you give me a brief explanation?

Then I'll close this pull request, refork from the current state and open a new pull request.

PR #597, opened 2021-07-02
Add option to use treesitter for folds

tafryn:
Along with `set foldmethod=expr` this option will enable treesitter driven folds.


abzcoding:
Closed in favor of #978

PR #599, opened 2021-07-02
[WHICH-KEY] Add SymbolsOutline mapping.

jameswalmsley:
What it says on the tin!

### DIFF
--- a/lua/lv-which-key/init.lua
+++ b/lua/lv-which-key/init.lua
@@ -213,7 +213,8 @@ local mappings = {
         r = {"<cmd>Lspsaga rename<cr>", "Rename"},
         t = {"<cmd>LspTypeDefinition<cr>", "Type Definition"},
         x = {"<cmd>cclose<cr>", "Close Quickfix"},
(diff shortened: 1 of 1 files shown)


ChristianChiarulli:
can you add a check to see if symboloutline plugin is active? There should be decent examples at the bottom of the file.

abzcoding:
something like this maybe?
```lua
o = {O.plugin.symbol_outline.active and "<cmd>SymbolsOutline<cr>" or "<cmd>Telescope lsp_document_symbols<cr>", "Symbols Outline"},
```
or maybe even change the `s` itself

```lua
s = {O.plugin.symbol_outline.active and "<cmd>SymbolsOutline<cr>" or "<cmd>Telescope lsp_document_symbols<cr>", "Document Symbols"},
```

jameswalmsley:
@abzcoding I favour your suggestion of switching to using SymbolsOutline for document symbols if its enabled... otherwise Telescope.

I've updated the PR.

PR #602, opened 2021-07-02
[LSP] key-bind - ca is the nvim default for "change around".

jameswalmsley:
`ca` is the default motion for change around...
Could we bind code-action to something else?

My muscle memory just used <leader>-l-a for it...


ChristianChiarulli:
Code actions can be done from <space>la I'm pretty sure anyway.

PR #609, opened 2021-07-03
Cleaner and Faster `settings.lua` File

zahin-muhaimeen:
Before, the `settings.lua` file was quite messy and was hard to understand. It also used a mixture of `vim.o`, `vim.wo`, `vim.bo`, and `vim.opt`, when everything could just be placed with `vim.opt`.

This (for my slow and crispy computer) had shaved off 14 milliseconds (average of100 takes of all vim.opt `-` average of 100 takes of current). But even if this doesn't make much of a performance difference, it at least makes the `settings.lua` file easier to work with.

ALL SETTINGS WERE KEPT INTACT (mostly)

### DIFF
--- a/lua/default-config.lua
+++ b/lua/default-config.lua
@@ -1,6 +1,7 @@
 CONFIG_PATH = vim.fn.stdpath('config')
 DATA_PATH = vim.fn.stdpath('data')
 CACHE_PATH = vim.fn.stdpath('cache')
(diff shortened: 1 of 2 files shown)


rebuilt (approved):
Looks like this also changes the gui font from Fira code to monospace.  I think this is a sensible change.

rebuilt:
Good work overall.  Much cleaner

jameswalmsley (approved):
Really nice.

Issue #615, opened 2021-07-03
REQUEST: add option for transparent bg


Pasqat:
I'm wondering if it could be pollible add the option in lv-config.lua to toggle the background transparency.
Maybe it could be just a cmd "hi Normal guibg=none"

ChristianChiarulli:
I could probably add a setting for that.

williamgoulois:
I just opened a PR to add this option

Pasqat:
Thank you all.
I opened a PR to add `hi SignColumn guibg=none ...`.
I think that this issue could be closed

PR #617, opened 2021-07-03
feat(615): transparent background option

williamgoulois:
Option for transparent background #615


Pasqat:
I was starting to add it right now, but you have done what I had in mind (I was unsure where to put the cmd, likely in settings.lua.

Issue #563, opened 2021-06-30
Indentation issue in cpp files. 


mayankdutta:
i have indentation issue while editing `cpp` file 

1. with `autopairs` and `treesitter indent` enabled
```
        if (something) {|}


        if (something) {
        |
        }

        /* expected */
        if (something) {
             |
        }
```
2. With `treesitter indent off` and enabling `cindent` and `autoindent`
    I get the expected behaviour. 
    
```
         if (something) {
               |
         }
```

`Question` Is there option to include `clang-formatter`, same as `tsserver using prettier`? so that it can format my file after saving.  if yes, How to enable it?

rebuilt:
No, there isn't a clang-formatter yet.  

Do you have the language server on?
If so, can you try formatting by pressing 

```
<space>lf   
```
(That's a lowercase L)

If that works, you can create an autocommand to format on save with 

```
BufWritePre *.cpp lua vim.lsp.buf.formatting_sync(nil,1000)
```

ChristianChiarulli:
we may need to remove cpp from the list of autoindents with treesitter then. Also do you have a link to the clang formatter?

mayankdutta:
> we may need to remove cpp from the list of autoindents with treesitter then. Also do you have a link to the clang formatter?

dont know much about that, [saw this online](https://clang.llvm.org/docs/ClangFormat.html#vim-integration), tried to integrate this but couldn't.

mayankdutta:
> No, there isn't a clang-formatter yet.
> 
> Do you have the language server on?
> If so, can you try formatting by pressing
> 
> ```
> <space>lf   
> ```
> 
> (That's a lowercase L)
> 
> If that works, you can create an autocommand to format on save with
> 
> ```
> BufWritePre *.cpp lua vim.lsp.buf.formatting_sync(nil,1000)
> ```

Thanks, also i want to point out that even after disabling `treesitter indent`, I also had to disable `autopairs` to get correct indentation. I tried to disable them individually but only case in which `both disabled` gave desired indentation.

rebuilt:
https://github.com/windwp/nvim-autopairs/issues/66

ChristianChiarulli:
Can this b closed after merging #569

mayankdutta:
Thanks for quick support.

mayankdutta:
I am using current version of LunarVim and I observed few bugs in indentation. 

## Case-1, In nested if else
`Current Result` <br>
[...]

PR #621, opened 2021-07-03
added Clangd format_on_save

mayankdutta:
I always have to manually add code, this would be better if this gets merge, though by default `save on format` and `clangd` both are `false`.

### DIFF
--- a/utils/installer/lv-config.example.lua
+++ b/utils/installer/lv-config.example.lua
@@ -84,8 +84,8 @@ O.lang.go.autoformat = true
 O.lang.rust.autoformat = true
 
 -- clang
-O.lang.clang.active = false -- make this true to enable clang support
-O.lang.clang.autoformat = false -- make this true to enable auto complete in C, C++ files
+O.lang.clang.active = false -- Set to true to enable clang support in C/C++ files.
+O.lang.clang.autoformat = false -- Set to true to enable auto-format in C/C++ files.
 
 
 
--- a/utils/installer/lv-config.example.lua
+++ b/utils/installer/lv-config.example.lua
@@ -84,7 +84,6 @@ O.lang.go.autoformat = true
 O.lang.rust.autoformat = true
 
 -- clang
-O.lang.clang.active = false -- Set to true to enable clang support in C/C++ files.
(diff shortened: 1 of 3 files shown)


utils/installer/lv-config.example.lua
```diff
@@ -82,6 +82,13 @@ O.lang.go.autoformat = true
 
 -- rust
 O.lang.rust.autoformat = true
+
+-- clang
+O.lang.clang.active = false -- make this true to enable clang support
+O.lang.clang.autoformat = false -- make this true to enable auto complete in C, C++ files
```
Tastyep, line 88: auto complete in -> auto formatting for ?

mayankdutta: sorry, if this caused some misunderstanding, by `auto formatting`. I meant indentation stuff on saving code. I had terrible time with `treesitter indentation` with `autopairs` as mentioned in #563 .

utils/installer/lv-config.example.lua
```diff
@@ -82,6 +82,13 @@ O.lang.go.autoformat = true
 
 -- rust
 O.lang.rust.autoformat = true
+
+-- clang
+O.lang.clang.active = false -- make this true to enable clang support
```
jameswalmsley, line 87: You no longer need to set active here..
It will only activate if you:

  1. open a c/cpp file.
  2. Have installed cpp support (LspInstall cpp)

  If you remove this, I'll be happy to approve.

utils/installer/lv-config.example.lua
jameswalmsley, line 88: Change the comment to:

```
-- Set to true to enable auto-format in C/C++ files.
```

jameswalmsley (changes requested):

utils/installer/lv-config.example.lua
mayankdutta: Comment updated.

ChristianChiarulli:
is this good? @Tastyep @jameswalmsley

jameswalmsley (approved):

jameswalmsley:
@ChristianChiarulli This is good now :)

PR #625, opened 2021-07-03
Update nv script folder

henrypalacios:
The path used by default in install.sh is [~/.config/nvim](https://github.com/ChristianChiarulli/LunarVim/blob/0096e0baa93f26097c1ce26a9e0a72bed0c5a468/utils/installer/install.sh#L103)


luxus:
does this file get used anywhere? u did u just tried to start nvim with nv?

PR #626, opened 2021-07-03
Define a form for LSP issues

Tastyep:
Follow up of [618](https://github.com/ChristianChiarulli/LunarVim/pull/618)

Tell me if I forgot anything.

lsp-issue-form


ChristianChiarulli:
also mention to check ftplugin for your lang, also maybe look at `echo &ft` to tell us your filetype.

PR #640, opened 2021-07-04
Adding tabnine for compe

### LINKED ISSUE #631: Add tabnine option for compe
there is a setting with compe

abzcoding:
Fixed #631 

For activation, you need to add the following line to your `lv-config.lua`
```lua
O.plugin.tabnine.active = true
```
Due to this [issue](https://github.com/tzachar/compe-tabnine#packer-issues), sometimes you might need to do this after `:PackerSync`
```bash
cd ~/.local/share/nvim/site/pack/packer/start/compe-tabnine
./install.sh
```

### DIFF
--- a/lua/default-config.lua
+++ b/lua/default-config.lua
@@ -68,7 +68,8 @@ O = {
         diffview = {active = false},
         bracey = {active = false},
         telescope_project = {active = false},
-        dap_install = {active = false}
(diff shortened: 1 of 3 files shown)


abzcoding:
I can add `event = "InsertEnter"` to the plugin as well for lazy loading but I'm not sure if it is necessary or not

PR #643, opened 2021-07-04
Add custom plugin file 

lavixta:
I think it is better for users to add their plugins beside your default plugins. So I create a file named "CustomPlugin.lua" where users can put their additional plugins into it and put it in gitignore. Also, make a small config in plugins.lua to use the plugins from CustomPlugin.lua . So everytime users pull your git and update your git their custom plugins will not disappear

### DIFF
--- a/lua/CustomPlugin.lua
+++ b/lua/CustomPlugin.lua
@@ -1,0 +1,5 @@
+-- Add this file to gitignore
+customPlugin = {
+    {"windwp/nvim-ts-autotag"}
+    -- put your plugin here
+}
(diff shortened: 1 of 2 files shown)


colutti:
This follows the same ideas of #653 

I think it would a better idea to think about what would be the ideal pattern and do it for every config file.

ChristianChiarulli:
@lavixta thank you I mentioned this in stream but implemented it a little different.

lavixta:
@ChristianChiarulli Ah, I saw it, I think it is better to put in lv-config like you did

PR #646, opened 2021-07-04
Hook to run PackerCompile during install phase

roobert:
Added a hook to run `PackerCompile` after install - I think this should help prevent some common problems for new users.

This hook could also be used to run `LspInstall lua` which would mean users could use the dashboard to edit their configs without getting errors out of the box, what do you think?


IndianBoy42:
Is there a reason to only have this on the first run? automatically compiling after installing/syncing seems like a good idea in general isn't it?

roobert:
@IndianBoy42 - I did consider this but I didn't want to change existing behaviour / confuse users that have expectations around how Packer functions

luxus:
this should be fixed in the latest rolling branch

roobert:
@luxus l

PR #648, opened 2021-07-04
[UTILS] Global Install for LunarVim

jameswalmsley:
I'm updating the global-install feature for LunarVim.


ChristianChiarulli:
I'll have to test this in a stream before I merge but if we could get this done we could easily have a package in the AUR.

jameswalmsley:
Nice, I'll have to take a look at the latest changes in the install script,
and update anything that has come in.

My thoughts were that we can call the install scripts... install.sh or lunarvim-install.sh

And running lv / nv with an upgrade flag like will upgrade the global install..

```
lv --upgrade                # Pull in latest stable branch.
lv --upgrade --master       # Bleeding edge!
```

PR #649, opened 2021-07-04
Add tool to profile startuptime for plugins

rebuilt:
Adds a tool to profile startup time
Run with 

```
# Default is 10 runs
:StartupTime <num_of_runs>
:StartupTime 100
```

Activate with 
```
O.plugin.startuptime = true
```


rebuilt:
Irrelevant now

PR #651, opened 2021-07-04
Install script refactor

roobert:
Tidied up the script and got rid of linting errors, feel free to reject this PR if this is not to your taste.

Two things to note:
1. Switched shebang to `/usr/bin/env bash` - I'm not sure of any systems that don't have `bash` these days so think this should be OK but I can switch back to `sh` and make posix complaint again if you'd prefer.
2. I can't accurately test replacing `expr` with `[[]]` since I'm not sure what  the `uname` string for Windows looks like - if you could let me know I can double check this works..

### DIFF
--- a/utils/installer/install.sh
+++ b/utils/installer/install.sh
@@ -1,90 +1,151 @@
-#!/bin/sh
+#!/usr/bin/env bash
 
-set -o nounset # error when referencing undefined variable
-set -o errexit # exit when command fails
+set -o nounset  # error when referencing undefined variable
+set -o errexit  # exit when command fails
+set -o pipefail # prevent errors in pipelines from being masked
 
-installnodemac() {
-	brew install lua
-	brew install node
-	brew install yarn
+main() {
+	echo 'Installing LunarVim'
+	move_old_nvim
+	ask_to_install_pip
+	ask_to_install_node
+	install_pynvim
+	install_packer
+	lunarvim_clone_config
+	echo
+	echo "It is recommended to install and activate a font from here: https://github.com/ryanoasis/nerd-fonts"
+	echo
 }
 
-installnodeubuntu() {
-	sudo apt install nodejs
-	sudo apt install npm
-}
-
-moveoldnvim() {
+# Handle existing nvim installations
+move_old_nvim() {
+	[ -d "$HOME/.config/nvim" ] && return
 	echo "Not installing LunarVim"
 	echo "Please move your ~/.config/nvim folder before installing"
 	exit
 }
 
-installnodearch() {
+#
+# User Interaction
+#
+ask_to_install_node() {
+	if which node >/dev/null; then
+		echo "node installed, moving on..."
+		return
+	fi
+
+	echo "Node not found"
+	echo -n "Would you like to install node now (y/n)? "
+	read -r answer
+	[ "$answer" != "${answer#[Yy]}" ] && install_node
+}
+
+ask_to_install_pip() {
+	if which pip3 >/dev/null; then
+		echo "pip installed, moving on..."
+		return
+	fi
+
+	# echo "pip not found"
+	# echo -n "Would you like to install pip now (y/n)? "
+	# read answer
+	# [ "$answer" != "${answer#[Yy]}" ] && install_pip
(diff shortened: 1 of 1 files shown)


rebuilt:
Have you already tested in a virtual machine?  If so, on which distros?

roobert:
I originally tested this natively with OS-X and I have just tested this in parallels with Ubuntu and Fedora - all work fine.

I have no easy way to test these changes with Ubuntu or Arch so would appreciate it if someone else may be able to do the following to validate the changes:

```
# Warning: this will destroy your existing nvim configuration!
rm -rf ~/.config/nvim ~/.cache/nvim/ ~/.local/share/nvim/

bash <(curl -s https://raw.githubusercontent.com/roobert/LunarVim/install-script-refactor/utils/installer/install.sh)
```

Thanks!

roobert:
@rebuilt if you could let me know if this PR is OK in principle and no-one else has time to test, I'll put some effort into building Arch and Gentoo installs in parallels this week, thanks!

rebuilt:
I can confirm it works in arch.  

@jameswalmsley  I think this needs your input.  You know more about shell scripting than I do.  
* I think we should stick to the posix compliant way of doing things to ensure compatibility to the maximum set of architectures and users.  But I'll defer to your opinion.  
* The functions in the installer are renamed and I know the pull request you're working on uses these function names as well.

barun108:
I am using it on Arch 

I installed it with latest script , removing old nvim folder

roobert:
@jameswalmsley - be good to get your opinion on this as I may have time tonight to get Gentoo running if this is provisionally approved, thanks!

jameswalmsley:
Hey @rebuilt @roobert thanks for your work looking at this.

I'm currently working on. A global installer patch that will allow a system wide install of lunarvim alongside a normal user nvim config without any interference.

In principle I think POSIX compliant shell scripts are laudable.. however for all practical purposes if you target bash syntax that works everywhere, especially if that keeps the scripts clean and maintainable that's good.

I think I'll create a branch for the global install that we could work on..

See lv.global-install on my repo..

That way I can base my work on-top if your script and we can work towards the same goal.

My initial work is the original install script updated to support global install.

I think we should have a lunarvim-install.sh (or just install.sh) that supports initial install, global or local..

The once the global version is installed you'd be able to run:

sudo lv --upgrade

Or 

sudo lv --upgrade --master 

To update the global installation.

jameswalmsley (commented):
Your current set of changes look fine.
Take a look at my lv.global-install branch on my repo, and we can combine are efforts.

roobert:
@jameswalmsley - that sounds great! I've had a look at your changes and propose the following: we get this merged into master and then merge master into your branch, once merged it should be easy to apply the changes you've made to `lunarvim-upgrade.sh` to a new version of `lunarvim-upgrade.sh` that is based off this refactored `install.sh`, does that make sense? Once stuff is in your branch I'd be happy to help merge the changes or add features, if you like.

Regarding POSIX: I would be amazed if anyone wanted to use this on a system that didn't have bash (I can't think of any that don't have it these days) but if someone does open up a bug I will A) eat my hat. and B) revert this to use `/bin/sh` and make POSIX compatible! 😄 

Even systems that don't ship with bash for licensing reasons (like the BSDs) almost always have bash installed since so many things depend on it!

utils/installer/install.sh
```diff
@@ -94,114 +160,104 @@ installpynvim() {
...
-else
-	# clone config down
-	cloneconfig
 	# echo 'export PATH=$HOME/.config/nvim/utils/bin:$PATH' >>~/.zshrc
 	# echo 'export PATH=$HOME/.config/lunarvim/utils/bin:$PATH' >>~/.bashrc
-fi
 
-echo "I recommend you also install and activate a font from here: https://github.com/ryanoasis/nerd-fonts"
+	echo "Cloning LunarVim configuration"
+	git clone https://github.com/ChristianChiarulli/lunarvim.git ~/.config/nvim
```
kylo252, line 248: Add `--depth 1` by default. 
Experienced users can always do a `git fetch --deepen`

kylo252 (changes requested):

PR #653, opened 2021-07-04
Allow specifying custom path for `lv-config.lua`

roobert:
The idea here is to allow people to store configuration outside of the `nvim` directory.

This would allow people to easily update LunarVim with a `git pull` or by re-running `install.sh` and would mean configuration stored in someones `dotfiles` that may already be checked out/exist in their home directory would be automatically used.


rebuilt:
I'm not sure what the plan is for how to implement this.  Chris mentioned it was on the radar: https://github.com/ChristianChiarulli/LunarVim/issues/614

The current way to modify some settings is to place your personal configuration in lv-config.lua.  This file will not be over-written when you update.  

That file gets loaded at the beginning of init.lua so it's not enough.  Adding an 'overrides' file at the end of init.lua could be a good place to put configs that address: 
    key mappings
    plugin configurations
    install extra plugins

roobert:
@rebuilt - the other important point I was  trying to raise here is the ability to have config outside of the `nvim` directory which is important for those people who store their config in, for example, a`dotfiles` repository.

rebuilt:
Ah, ok.

colutti:
I think this is a great idea. In fact, we could do the same for other files such as the plugins.lua, so people would also have a personal plugin installer script. keymaps.lua also needs that and many others.

roobert:
@colutti you could simply require  these from `~/.nvim-local/init.lua`!

colutti:
Yes, but that breaks modularity. I think it would be nicer to have different user files just like we have different configs for mappings, plugins, etc.

roobert:
Could you explain a bit about how it'd break modularity? I'm still very new to Lua!

My original intent with this PR was to create something extremely simple that'd add immediate extensibility to LunarVim so I could adopt LunarVim but add my custom plugins straight away.

I think if you were to reuse this pattern multiple times for different files then you may risk users requiring their customisation file from `~/.nvim-local/init.lua` which would result in files being required multiple times. This could happen if the behaviour of the `.nvim-local` tree would be made to work differently from your standard `.nvim` directory, which could cause confusion?

colutti:
I am just saying that we could for example add to every config script (keymap, plugins) a call to an user-defined script (keymap-user.lua, plugins-user.lua) if such scripts exists (if they dont, dont raise an exception). Then we could add those user-defined scripts to the .gitignore so they would not be tracked. 

This way the keymap.lua would call keymap-user.lua, the plugins.lua would call a plugins-user.lua. Both files added to .gitignore. 

This allows the user to separate the logic in different scripts, to no mix up keymaps and plugins in the same file (just like Lunar Vim does)

I hope it makes sense

roobert:
@colutti ah right I understand, perhaps being able to specify a `NVIM_LOCAL` path through an environment variable would be nice - that way by default you could have it set to `${HOME}/.config/nvim`, and therefore load files in paths under that tree, but for those of us who prefer to keep our customizations outside of `.config/nvim`, we could specify a different path, does that make sense?

colutti:
It makes a lot of sense actually.

roobert:
My main concern here is being able to checkout my `dotfiles` to something like `~/.config/nvim-local`, clone/install `LunarVim` and not need to take any action to have things "just work", i.e: no configuration of files in `LunarVim` or symlinking, or anything, this is why I suggested potentially using an environment variable to modify the default behaviour.. what do you think?

colutti:
Yes, I agree. I am just saying that for people who does not want to use env variables the default should be a folder.

roobert:
I have updated the code to default to `CONFIG_PATH/init-user.lua` but allow overriding the directory path through an environment variable.

I think for now it makes sense to keep this simple by loading additional config from a single point that could optionally require other files rather than trying to permit overriding every file.

luxus:
I like @colutti idea with the folder, I saw something similar in lvim .. https://github.com/lvim-tech/lvim#customize-plugins
maybe at a later stage we need more then just 1 file.

roobert:
This could easily be adapted to require all files in some path, perhaps: `.nvim/local` or `.nvim/user`?

I think this probably needs some input from the maintainers about what is likely to be accepted and what is not, then we can iterate on this until we have an acceptable solution.

colutti (approved):

luxus:
Can you please add the changes to .gitignore

roobert:
@luxus - done!

luxus:
maybe we should explain it in the wiki too?

roobert:
Chris has added support for defining custom plugins from `lv-config.lua` - since this is the case I've updated this PR to simply permit specifying a custom path to the `lv-config.lua` file.

roobert:
This PR is now incomplete due to multiple references in the codebase referencing `lv-config.lua`. I will close this in favour of a new PR that addresses specifying paths to `lv-config.lua` and `lv-user-config` using environment variables in a more global way.

rebuilt:
You can now accomplish this using the technique here: https://github.com/rebuilt/lv-user-config

roobert:
Hi @rebuilt, nice idea but it doesn't solve one of the original main points of this issue - that of being able to checkout an untainted LunarVim nvim dir and just have everything work, without symlinks or changing the directory structure of the checkout.

I think what you've done is good but for it to be truly flexible we need to be able to specify custom paths to the `lv-user-config` dir and `lv-config.lua` path through environment variables.

I have some code to do this but Chris has said this kind of thing will come later.

For the sake of discussion, something like this:
```-- allow specifying custom path to lv-config.lua with an environment variable
-- defaults to ~/.config/nvim/lv-config.lua
local function lv_config_path()
    local _lv_config_path = os.getenv("LV_CONFIG_PATH")
    local path

    if _lv_config_path == nil then
        path = CONFIG_PATH .. "/lv-config.lua"
    else
        path = _lv_config_path
    end

    return path
end

LV_CONFIG = lv_config_path()

if vim.fn.filereadable(LV_CONFIG) == 1 then
    -- ordering here is important so we don't accidentally load some other
    -- file called `lv-config.lua`
    package.path = LV_CONFIG .. ';' .. package.path
    vim.cmd("luafile " .. LV_CONFIG)
end

-- allow specifying custom path to lv-user with an environment variable
-- defaults to ~/.config/nvim/lua/lv-user
local function lv_user_path()
    local _lv_user_path = os.getenv("LV_USER_PATH")
    local path

    if _lv_user_path == nil then
        path = CONFIG_PATH .. "/lua/lv-user"
    else
        path = _lv_user_path
    end

    return path
end

LV_USER = lv_user_path()

if vim.fn.filereadable(LV_USER) == 1 then
    package.path = package.path .. ';' .. LV_USER
end```

Then anywhere `lv-config.lua` is referenced (such as in the reload function) we can use `LV_CONFIG_PATH`, and `LV_USER_PATH` (or `LV_USER_CONFIG_PATH` can be used to specify the location of your custom configs, which in my case, are checked out elsewhere in my filesystem because they're part of my dotfiles.

I wrote this before you added `lv-user-config` - but presumably this is the same as the`lv-user` dir which Chris added?

Be good to hear your thoughts!

rebuilt:
I haven't had any conversations with Chris about it.  All I know is what he's said publicly so I don't want to speak for him.  

I agree with his public statement that getting to release 1.0 is the priority.  To do that, we focus on speed improvements and bug fixes.  This means we have to put additional features on the backburner.  I don't think this is the end of the conversation about configurable paths for userland settings.  But for the time being, adding any code opens us up to new bugs or performance hits or time for developers to read and understand the codebase.  The pace of change has increased rapidly and I'm having a hard time keeping up with all the changes.  And we still have to document a lot of the changes before the 1.0 release.  

PS. 

You don't have to use submodules.  I changed the readme on https://github.com/rebuilt/lv-user-config  to default to using just a standard git repository in ~/.config/lunarvim.

roobert:
@rebuilt - nice one! Once 1.0 has been released it'd definitely be good to discuss env vars again since it'll help towards that seamless experience of having things Just Work for people.

nathanielks:
Just voicing support for environment variables 🙏 I too use a dotfiles repo and not being able to symlink to a config there is a minor annoyance. It'd be helpful to be able to point directly to that file or use a symlink 😄 Thank you so much for all of this, it's such an improvement to my previous config!

PR #655, opened 2021-07-04
Adds tree-sitter text objects

rebuilt:
Adds the plugin treesitter-text-objects as a default and enabled plugin.  
Adds text-objects for: functions and classes
Enables keybinds like the following

```
af = function outer
if = function inner
ac = class outer
ic = class inner

yaf = yank around function
vif = select inside function
etc... 
```


rebuilt:
Closing because a new pull request is a better implementation of this.  

https://github.com/ChristianChiarulli/LunarVim/pull/700

PR #656, opened 2021-07-04
Pressing escape also clears search and spelling highlights

rebuilt:
In addition to normal function, pressing Escape also clears search and spelling highlights

To enable spelling or search highlights:

```
:set spell
:set hlsearch
```

### DIFF
--- a/lua/keymappings.lua
+++ b/lua/keymappings.lua
@@ -48,6 +48,10 @@ vim.cmd('inoremap <expr> <c-k> (\"\\<C-p>\")')
 
 vim.cmd('vnoremap p "0p')
 vim.cmd('vnoremap P "0P')
+
+-- Escape key clears search and spelling highlights
+vim.api.nvim_set_keymap('n', '<ESC>', ":nohls | :setlocal nospell<ESC>", {silent = true} )
+
 -- vim.api.nvim_set_keymap('v', 'p', '"0p', {silent = true})
 -- vim.api.nvim_set_keymap('v', 'P', '"0P', {silent = true})
 


rebuilt:
I didn't touch the jk, jj escape mappings because those are done from insert mode, not normal mode.

jameswalmsley (approved):

colutti:
I would not do that. I had a lot of issues in the past when I tried to bind sometthing to the ESC key. In addition, there is already a binding to clear the search highlight.

jameswalmsley:
@colutti I've had this same mapping in my config for nearly a year, and I've had no problems.
@rebuilt I actually really like this, as it kind of "resets" normal mode.

I think @rebuilt 's mapping is fine, its only active in normal mode, so pressing Esc would be a NOP usually..

jameswalmsley:
I suppose we could even map this to a LunarResetNormal function which would do the above.. but you can then extend it to reset other items that have been activated in the future.

I take @colutti 's point that there is a mapping for this already, this just feels natural to me.

jameswalmsley:
@colutti Out of interest what is the existing mapping?

colutti:
@jameswalmsley  <space>h
@rebuilt I will try it and see how it goes, thanks.

OliverNChalk:
I think this change is a little too opinionated to be default behaviour, especially given we have a bind `<leader>h` that clears the search term

rebuilt:
Closing this since it will be outdated by https://github.com/ChristianChiarulli/LunarVim/pull/753 .  And it seems controversial.

PR #659, opened 2021-07-04
[KEYBINDINGS] Add Alt-jk etc for move line up/down.

jameswalmsley:
Matches behaviour in vscode.

Moves current line up or down, or a selection up and down.


antoineco:
I've been looking where these mappings actually come from and traced their addition back to this PR (thanks!).

For posterity, the explanation is here: https://vim.fandom.com/wiki/Moving_lines_up_or_down

PR #660, opened 2021-07-04
[KEYBINDINGS] Navigate quick-fix list with [q and ]q

jameswalmsley:
As seen on ThePrimeagen

This one is awesome.

### DIFF
--- a/lua/keymappings.lua
+++ b/lua/keymappings.lua
@@ -50,6 +50,10 @@ vim.api.nvim_set_keymap('i', '<A-k>', '<Esc>:m .-2<CR>==gi', {noremap = true, si
 vim.api.nvim_set_keymap('x', '<A-j>', ':m \'>+1<CR>gv-gv', {noremap = true, silent = true})
 vim.api.nvim_set_keymap('x', '<A-k>', ':m \'<-2<CR>gv-gv', {noremap = true, silent = true})
 
+-- QuickFix
+vim.api.nvim_set_keymap('n', ']q', ':cnext<CR>', {noremap = true, silent = true})
+vim.api.nvim_set_keymap('n', '[q', ':cprev<CR>', {noremap = true, silent = true})
+
 -- Better nav for omnicomplete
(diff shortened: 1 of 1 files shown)


rebuilt:
Doesn't this override the window movement mappings?

```
vim.api.nvim_set_keymap('n', '<C-h>', '<C-w>h', {silent = true})
vim.api.nvim_set_keymap('n', '<C-j>', '<C-w>j', {silent = true})
vim.api.nvim_set_keymap('n', '<C-k>', '<C-w>k', {silent = true})
vim.api.nvim_set_keymap('n', '<C-l>', '<C-w>l', {silent = true})
```

jameswalmsley:
@rebuilt you are right. In my config I've disabled the window nav bindings.

I'd like to get the qf navigate upstreamed.. but I'd be happy to change my personal binding.

Any suggestions for an alternative (I'll fix the PR it's git another commit mixed in).

jameswalmsley:
Ok I've changed my binding to C-A-j C-A-k
I actually like that.

J

rebuilt:
I think that's a natural binding.  

If you were trying to stick to vim conventions [q would also be reasonable.  A lot of the square bracket mappings already have this convention of navigating to the next/previous item.  

```
|[(|		[(		1  cursor N times back to unmatched '('
|[[|		[[		1  cursor N sections backward
|[s|		[s		1  move to the previous misspelled word
|[z|		[z		1  move to start of open fold
|[/|		[/		1  cursor to N previous start of a C comment

... etc
```

rebuilt (approved):

jameswalmsley:
That sounds like a good idea.. I'll look into how you do that.. I think the C-A-jk is good natural binding..

Might be nice to map to the [q for more advanced vim users too.

jameswalmsley:
Fixed up PR conflicts.

jameswalmsley (commented):
Actually C-A-k doesnt work over ssh for some reason..
I'm looking to to the [q binding instead!

jameswalmsley:
Ok I'm going the whole way with [q ]q 
Its much better :) thanks @rebuilt

PR #667, opened 2021-07-04
Fix Lint CodeActions + Better dev experience on JS/TS

maxime50:
This makes CodeAction autofixes working

https://user-images.githubusercontent.com/51096534/124395931-e9493a80-dcd4-11eb-9b19-13cb85963f00.mp4

I changed efm to [null-ls](https://github.com/jose-elias-alvarez/null-ls.nvim) on js/ts files only for now as it's less bloated than efm and it uses built-in neovim sources to provide out-of-the-box functionality. efm's codeActions also doesn't work.

Also added [nvim-lsp-ts-utils](https://github.com/jose-elias-alvarez/nvim-lsp-ts-utils) plugin to get usefull utils like Import all missing imports (exposed as :TSLspImportAll)

### DIFF
--- a/lua/lsp/init.lua
+++ b/lua/lsp/init.lua
@@ -106,8 +106,48 @@ if O.document_highlight then
 end
 
 function lsp_config.tsserver_on_attach(client, bufnr)
-    lsp_config.common_on_attach(client, bufnr)
+    -- lsp_config.common_on_attach(client, bufnr)
     client.resolved_capabilities.document_formatting = false
+
+    local ts_utils = require("nvim-lsp-ts-utils")
+
+    -- defaults
+    ts_utils.setup {
+        debug = false,
+        disable_commands = false,
+        enable_import_on_completion = false,
+        import_all_timeout = 5000, -- ms
+
+        -- eslint
+        eslint_enable_code_actions = true,
+        eslint_enable_disable_comments = true,
+        eslint_bin = O.lang.tsserver.linter,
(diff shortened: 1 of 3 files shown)


ChristianChiarulli:
Looks great 🌵

yassinebridi:
Hey @maxime50, apparently this pull request has been reverted right?
it's such a shame since this makes TypeScript dev experience similar to what we used to have with coc.nvim

abzcoding:
@yassinebridi you can install `https://github.com/jose-elias-alvarez/nvim-lsp-ts-utils/` yourself and get the same experience

yassinebridi:
@abzcoding I thought i wouldn't work with the latest changes, thank you, you have been very helpful.

abzcoding:
> @abzcoding I thought i wouldn't work with the latest changes, thank you, you have been very helpful.

in case you wanted help adding it to your `lvim`
https://github.com/abzcoding/lvim/commit/c654785e2eccece5b1690d3e510288c3f63a212b

maxime50:
@abzcoding thanks, you're the best :)

I think this is indispensable for Typescript developers. ^^

They also moved the `nvim-lsp-ts-utils` TypeScript codeAction right inside `null-ls` bultins.

So it could be interesting for lunarvim use it from there also. (if we just want the codeaction for example.)

ChristianChiarulli:
we could probably support the code action natively if it's built into null-ls

PR #669, opened 2021-07-04
Cleanup

masylum:
This commit tries to clean some of the code. I've removed configurations
that are currently not used and may be confusing to people (such as the
`lv-lspinstall` and the `lv-spectre`).

I also took the chance to clean a bit the indentation which was a bit
random.


init.lua
```diff
@@ -3,7 +3,6 @@ require('keymappings')
 vim.cmd('luafile ' .. CONFIG_PATH .. '/lv-config.lua')
 require('settings')
 require('plugins')
-require('lv-utils')
```
masylum, line 6: this is not needed since it just exposes a function

lua/default-config.lua
```diff
@@ -82,7 +75,7 @@ O = {
             autoformat = false,
             isort = false,
             diagnostics = {
-                virtual_text = {spacing = 0, prefix = ""},
+                virtual_text = {spacing = 0, prefix = "⎨"},
```
masylum, line 78: ooops, this was not intended to be in this commit... I can remove it if you prefer the previous default

lua/default-config.lua
```diff
@@ -191,10 +184,10 @@ O = {
             },
             autoformat = false,
             diagnostics = {
-                virtual_text = {spacing = 0, prefix = ""},
+                virtual_text = {spacing = 0, prefix = "⎨"},
```
masylum, line 187: should we DRY all those diagnostics configuration?

IndianBoy42:
Was the indentation done with a formatter? eventually it should as thats the only practical way to get consistency, but I think the formatting should be separated from the rest of the changes in this PR, because the 'official' formatter (stylua/lua-format/lua-fmt) hasn't been chosen yet

masylum:
I've just used `=` to force the 4 spaces instead of 2. I can split it into smaller PRs if need be.

luxus:
there is some "hidden" discussion about cleaning the Lua files in https://github.com/ChristianChiarulli/LunarVim/issues/652

luxus:
@masylum you should run stylua script in extras on your branch. The changed shouldn't be that much after that

PR #672, opened 2021-07-04
feat(lsp_signature): Add function args suggestions plugin

nkoturovic:
# Info about Pull Request

⚠️ There is alternative PR #726 that has some additional features and it is focused to using `lsp_signature` as `builtin_plugin`. 
When decision to go for `lsp_signature` as `builtin_plugin` is final, that is probably the right PR for it.

ℹ️  Basically after decision to transfer this plugin from builtin to `custom_plugin` this whole PR became just a comment inside `lv-config.example.lua`, but I wanted to put it there and leave this small PR open because of potential to bubble up into builtin and since there are plenty of newcomers to LunarVim that want this feature.

## How to use it

To enable plugin uncomment it in `O.custom_plugins` inside `lv-config.lua`
```lua
O.custom_plugins = {
     -- {"windwp/nvim-ts-autotag"},
     {
         "ray-x/lsp_signature.nvim",
         config = function() require"lsp_signature".on_attach() end,
         event = "BufRead"
     }
 }
```
Then run `:PackerInstall`, `:PackerCompile`

## Info about Plugin

- Plugin works (almost perfectly), but there is but
- Compe has one bug currently: Accept completion with '('
- Please read compe bug (issue) description: https://github.com/ChristianChiarulli/LunarVim/issues/620#issuecomment-873664016_
- When that issue hrsh7th/nvim-compe#436 in compe repo is resolved it should work perfectly

## Plugin functionality 
- Able to highlight current parameter on the fly
 - **It works with compe (now)**
 - It should work seamlessly with compe in future
     after compe bugfix: hrsh7th/nvim-compe#436
 - It is disabled by default in default-config
 - It is disabled by default in sample lv-config.example.lua

 Plugin repo: [https://github.com/ray-x/lsp_signature.nvim](https://github.com/ray-x/lsp_signature.nvim)


lua/lv-lsp_signature/init.lua
```diff
@@ -0,0 +1,18 @@
+local M = {}
```
williamgoulois, line 1: I think you should move this file under lsp folder

nkoturovic: > I think you should move this file under lsp folder

Thanks for feedback, will do it.

nkoturovic: > I think you should move this file under lsp folder
Do you mean to move whole folder or renamed init.lua file?

nkoturovic: Also didn't put it there initially because it is plugin whose name contains `lsp`, and I saw `lv-lsp-rooter` which is located in separate folder as well.

nkoturovic:
I moved: `lua/lv-lsp_signature/init.lua` -> `lsp_signature.lua` as @williamgoulois  suggested (as I understood his suggestion).

lua/lv-lsp_signature/init.lua
nkoturovic: I moved: `lua/lv-lsp_signature/init.lua` -> `lsp/lsp_signature.lua` in as suggested.

abzcoding:
After recent changes to the plugins, I think we can suggest users to add this as a custom plugin
inside `lv-config.lua`:
```lua
O.custom_plugins = {
      {
          "ray-x/lsp_signature.nvim",
          config = function() require"lsp_signature".on_attach() end,
          event = "BufRead"
      }
}
```

nkoturovic:
> After recent changes to the plugins, I think we can suggest users to add this as a custom plugin
> inside `lv-config.lua`:
> 
> ```lua
> O.custom_plugins = {
>       {
>           "ray-x/lsp_signature.nvim",
>           config = function() require"lsp_signature".on_attach() end,
>           event = "BufRead"
>       }
> }
> ```

- Thanks for suggestion, I think that `custom_plugins` is right place for now.
- Also there is that little `compe` bug that slightly ruins whole experience by forcing us to use `<CR>` instead of `(` 
- On the other hand, feature like this may be something that we would like to work oob??
- Maybe to put it in `custom_plugins` with possibility to bubble up into `builtin` afterwards?

ChristianChiarulli:
I think this could be builtin when the bugs have been fixed

nkoturovic:
Closed: Relevant PR for this feature is now #726

PR #678, opened 2021-07-05
fixes condition to use local prettier

bob3000:
`vim.fn.glob("node_modules/.bin/prettier")` returns an empty string if the path does not exist which is a truthy value in lua. The effect is that the path to the global prettier is replaced by the potentially non existent prettier from the project level.


andrei-scripcaru:
Did not fixed it for me, I had to symlink local `node_modules/.bin/prettier` to global to have it working. For the record, local prettier is necessary to automatically pick-up all the plugins and configurations already existing in the project.

PR #685, opened 2021-07-05
Added editorconfig plugin

juanlopez4691:
Added support for basic code format `.editorconfig` files, using official plugin editorconfig/editorconfig. See https://editorconfig.org/

### DIFF
--- a/lua/default-config.lua
+++ b/lua/default-config.lua
@@ -70,6 +70,7 @@ O = {
     telescope_project = { active = false },
     dap_install = { active = false },
     tabnine = { active = false },
+    editorconfig = {active = false},
   },
 
   lang = {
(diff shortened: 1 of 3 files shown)


rebuilt:
Packer doesn't require you to use the opt keyword if some other lazy loading strategy is already being used.  Here, you're already using the 'BufRead' event to specify lazy loading so opt is redundant.

I'm not sure if Chris will approve this right now since he's focusing things that improve the speed or fix errors.

juanlopez4691:
> Packer doesn't require you to use the opt keyword if some other lazy loading strategy is already being used. Here, you're already using the 'BufRead' event to specify lazy loading so opt is redundant.
> 
> I'm not sure if Chris will approve this right now since he's focusing things that improve the speed or fix errors.

Thanks a lot for your comment! I will add a fix and hope for Chris to find the time to have a look 😄

PR #686, opened 2021-07-05
Add borders to lsp popups

### LINKED ISSUE #677: Borders for lsp popups

kjssad:
Attempt to fix #677.

### DIFF
--- a/lua/default-config.lua
+++ b/lua/default-config.lua
@@ -31,6 +31,10 @@ O = {
     rainbow = { enabled = false },
   },
 
+  lsp = {
+    popup_border = "single"
+  },
+
   database = { save_location = "~/.config/nvcode_db", auto_execute = 1 },
 
   plugin = {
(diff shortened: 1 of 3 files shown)


rebuilt:
borders

rebuilt:
borders2

rebuilt:
Border option set to 'double'

border-double

rebuilt:
Border option set to 'shadow'

border-shadow

IndianBoy42:
Feels like this should be configurable, maybe `O.lsp_border` or `O.popups` with more styling options

rebuilt:
Probably have to narrow the scope if you're making it configurable. 

```
O.lsp.border
```

kjssad:
This could be configurable but the current implementation is limited. There's no way to change the border highlights other than changing `FloatBorder`. Tho we can technically define our own border styles with specific highlights.

rebuilt (approved):

PR #687, opened 2021-07-05
[REFACTOR] move plugins' config to single files

kylo252:
Use a single file with the same name of the plugin's config folder.
This makes it easier to find and manipulate the config files. It would
also help tidying up the overall structure.


rebuilt:
I like this refactor.  I think it's great going forward.  

I also think it's going to create a big headache for everyone who is currently keeping up to date by merging commits into their personal branches.

kylo252:
> I also think it's going to create a big headache for everyone who is currently keeping up to date by merging commits into their personal branches.

Yeah sadly it will be a breaking-change for a lot of people, which is why I opted to limit the number of initial plugins just to show the potential first. 

We can also think of this as some sort of "technical debt" I guess 😄

kylo252:
oops, that didn't go to well 😢

kylo252:
Okay, I think that looks correct now. 

Here's a demo of how to override options

Y3eEf7YB8W

gnmearacaun:
Hi @kylo252 
I wonder if you would consider naming the folder something lv-related as the plugins are part of core. I was hoping to reserve lua/config as a convenient place for user configs (if approved) as they are accessed often and should be easiest to find as referenced [here](https://github.com/ChristianChiarulli/LunarVim/issues/461#issuecomment-877627955)

kylo252:
> Hi @kylo252
> I wonder if you would consider naming the folder something lv-related as the plugins are part of core. I was hoping to reserve lua/config as a convenient place for user configs (if approved) as they are accessed often and should be easiest to find as referenced [here](https://github.com/ChristianChiarulli/LunarVim/issues/461#issuecomment-877627955)

I actually had the same exact thought just a few hours ago. I think I can go with "lv-core" instead and strip out the "lv-" from the files themselves.

kylo252:
Adapted and merged in 2423fcdc904ec1413edaaac1fd8b9c0971f4a869.

Issue #689, opened 2021-07-05
module 'dap-install' not found


binnz:
when i open python file, error like this,
```
Error detected while processing /home/binnz/.config/nvim/ftplugin/python.lua:
E5113: Error while calling lua chunk: /home/binnz/.config/nvim/ftplugin/python.lua:79:
        no field package.preload['dap-install']
        no file './dap-install.lua'
        no file '/tmp/tmp.q5dutKvy2i/neovim/.deps/usr/share/luajit-2.1.0-beta3/dap-ins
        no file '/usr/local/share/lua/5.1/dap-install.lua'
        no file '/usr/local/share/lua/5.1/dap-install/init.lua'
        no file '/tmp/tmp.q5dutKvy2i/neovim/.deps/usr/share/lua/5.1/dap-install.lua'
        no file '/tmp/tmp.q5dutKvy2i/neovim/.deps/usr/share/lua/5.1/dap-install/init.l
        no file '/home/binnz/.cache/nvim/packer_hererocks/2.1.0-beta3/share/lua/5.1/da
        no file '/home/binnz/.cache/nvim/packer_hererocks/2.1.0-beta3/share/lua/5.1/da
        no file '/home/binnz/.cache/nvim/packer_hererocks/2.1.0-beta3/lib/luarocks/roc
        no file '/home/binnz/.cache/nvim/packer_hererocks/2.1.0-beta3/lib/luarocks/roc
        no file './dap-install.so'
        no file '/usr/local/lib/lua/5.1/dap-install.so'
        no file '/tmp/tmp.q5dutKvy2i/neovim/.deps/usr/lib/lua/5.1/dap-install.so'
        no file '/usr/local/lib/lua/5.1/loadall.so'
        no file '/home/binnz/.cache/nvim/packer_hererocks/2.1.0-beta3/lib/lua/5.1/dap-
cmd ["/home/binnz/.local/share/nvim/lspinstall/efm/efm-langserver"] is not executable.
Error detected while processing FileType Autocommands for "*":
Error while processing file: /home/binnz/.config/nvim/ftplugin/python.lua
/home/binnz/.config/nvim/ftplugin/python.lua:79: module 'dap-install' not found:
^Ino field package.preload['dap-install']
^Ino file './dap-install.lua'
```

then i active  dap_install plugin in default-config.lua, and PluginInstall the dap_install
```
plugin = {
    dap_install = { active = false},
}
to 
plugin = {
    dap_install = { active = true},
}
[...]

PR #691, opened 2021-07-05
only call dap_install.config if plugins are installed

IndianBoy42:
namely dap_install and nvim-dap.

Or else you can't open python files


Tastyep (approved):

IndianBoy42:
Fixes #689

rebuilt (approved):
Tested.  It works.  Thanks for your contribution.

PR #692, opened 2021-07-05
lv-vimtex exists but wasn't called, no config was called.

atropos112:
Vimtex wasn't loading the config `lua/lv-vimtex/init.lua `. I'm not sure if this is the best way to do this but it does work.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -4,7 +4,6 @@ vim.cmd("luafile " .. CONFIG_PATH .. "/lv-config.lua")
 require "settings"
 require "plugins"
 require "lv-utils"
-require "lv-vimtex"
 require "lv-galaxyline"
 require "lv-treesitter"
 require "lv-which-key"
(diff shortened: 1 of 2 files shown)


rebuilt:
Let's keep init.lua clean and tidy by doing all this in plugins.lua

```
  use { "lervag/vimtex", ft = "tex", config = function() require "lv-vimtex" end }
```

rebuilt (changes requested):
Don't add to init.lua

atropos112:
You are absolutely correct @rebuilt , I tried something similar and forgot to :PackerCompile and :PackerInstall and convinced myself that it didnt work , ill create a pull request for the suggested change in a moment

rebuilt:
Pretty soon, these plugins will be moved out of plugins.lua and into lv-config.lua.  This is already hooked up in master.  So, once it's removed vimtex will need to be added to the custom_plugins table in lv-config.lua

```
```

PR #693, opened 2021-07-05
Fish shell script support

IndianBoy42:
Highlighting and comment correctly.

tree-sitter-fish is already supported, just need to set the filetype

I thought about using `ftdetect.fish.lua`, but that would just be calling `vim.cmd([[]])` anyway. What is the performance implications of that?

### DIFF
--- a/ftdetect/fish.vim
+++ b/ftdetect/fish.vim
@@ -1,0 +1,1 @@
+au BufRead,BufNewFile *.fish setlocal filetype=fish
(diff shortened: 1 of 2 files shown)


ChristianChiarulli:
@IndianBoy42 I wanna say ftdetect is still probably the best place for this scopewise.

brandoncc:
+1 for this!

abzcoding:
Closed in favor of #968

PR #695, opened 2021-07-05
Add lightspeed.nvim

IndianBoy42:
This is a great plugin, which subsumes sneak, hop, easymotion and clever-f

https://github.com/ggandor/lightspeed.nvim.git


jameswalmsley:
I think I tried this before and thought it was nice...
I'd be happy for this to come in..

jameswalmsley (approved):
I think you need to run the formatter on one of your files.

Otherwise it's a thumbs up from me.

IndianBoy42:
> I think you need to run the formatter on one of your files.

Sorry switching branches keeps clearing the `.stylua.toml`

jameswalmsley (approved):

jameswalmsley:
Awesome.. all approved from my end.

ChristianChiarulli:
does anyone think this plugin is fundemental or should it be up to the user to define?

IndianBoy42:
I think it is very useful and fast for navigating code, idk about fundamental, but I'm getting about as much use of this as treesitter text objects. It's fairly lightweight and unintrusive, how many people use the 's' key in vim anyway

lua/lv-lightspeed/init.lua
```diff
@@ -0,0 +1,20 @@
+local M = {}
+
+M.config = function()
+	require("lightspeed").setup({
+		jump_to_first_match = true,
+		jump_on_partial_input_safety_timeout = 400,
+		highlight_unique_chars = true,
```
ggandor: Author here: `highlight_unique_chars` is unfortunately _extremely_ slow if the window area is big (the bottleneck seems to be the native `search()` itself, cannot really speed it up), so I deliberately turned it off by default, and I would recommend doing so here.

jameswalmsley:
Is it possible to lazy load this one on the s key being used?

I think this is a really great thing for people getting started with vim at least.

Advanced users can always disable it if they really don't want it bound to their s key..

lua/lv-lightspeed/init.lua
jameswalmsley: Thanks for the suggestion.

ggandor:
> Is it possible to lazy load this one on the s key being used?

FWIW, I personally find it ultra-annoying when a motion plugin (like targets.vim) loads on the first relevant keypress, with a very noticeable pause :)

luxus:
that's a very useful feature, I would love to have it as a core plugin

rebuilt:
To test this pull request:

```
git fetch origin pull/695/head:lightspeed
git checkout lightspeed

# When done. Return to master branch
git checkout master
```

Add this line to lv-config.lua
```
O.plugin.lightspeed.active = true
:PackerSync
```

rebuilt:
I've added this plugin configuration to the Wiki https://github.com/ChristianChiarulli/LunarVim/wiki/Example-plugin-configurations#lightspeed

It feels slower than hop for me but maybe I have to get used to it.  Hop has more visual distraction but lightspeed  has more mental overhead.  Also, hop is easier when navigating to one of many instances of the same word.

Lightspeed
```
First, make a decision to go up or go down and press a different key to activate lightspeed.
Start typing the word.  
But if the word is not unique, switch context and read the hints
```

Hop
```
Activate hop
Read and type the navigation hints
```

ggandor:
I hope you'll excuse me for getting involved here, but I've been spending _an awful lot of time_ thinking about the most efficient interface for such a plugin, so obviously I'm eager to discuss these issues, and very curious about different people's opinions.

I'm always surprised when someone says that searching forwards or backwards requires some kind of "decision". It's totally automatic, you see where the cursor is, and you see the target. (See also: https://github.com/ggandor/lightspeed.nvim#bi-directional-search on why there is no bi-directional mode.)

When using "word" search (or any kind of "hardcoded" target) instead of the universal 2-char search:
- Then you have to make a decision that you will use that specific search method in the first place. The advantage of the Sneak/Lightspeed philosophy is that you don't have to think about such things at all. When you want to jump to somewhere distant in the window, it's always `s/S`, simple as that. See also the [railway vs jetpack analogy](https://github.com/ggandor/lightspeed.nvim#a-short-pitch), and the introductory paragraph [here](https://github.com/ggandor/lightspeed.nvim#-an-in-depth-introduction-of-the-key-features) (if you're interested).
- If it's not a very short word, you have to move your gaze to the beginning, instead of keeping your focus on the section your eyes first caught - this is also disruptive.
- And obviously, you have to wait for and process the hints appearing, that is the most serious disruption. Searching for two (or more) characters is the only method where it is physically possible to label the targets ahead of time.

And as for 2-char search, Lightspeed's solution is clearly superior to Hop's (current) implementation, because of the above mentioned ahead-of-time labeling - this exactly the main selling point of this new plugin, after all. You don't have to "switch context and read the hints" - after the first input, you always know what the next keystroke should be. If the word is unique, it is highlighted differently, so you know that with the next keystroke you'll be done. But if it's not, then the hint is already visible _while_ you are entering the second character of the pattern, so (ideally) you will be ready when you should press the hint key.

rebuilt:
@ggandor I'm sorry for coming off as definitive.  I was just giving my initial impressions.  I'm sure with time, I can improve my speed with it.  

I do have to think about my keypress for forward or backwards.  Given time, it probably will become an automatic process.  My brain knows the mapping but my fingers still need to learn it. :)

Thank you for your explanation.  After reading the philosophy behind it, I think I missed a lot of this plugins potential.  

For context, I tested on files that had a large number of similar matches.

rebuilt:
@ggandor  And I will be spending more time with it, at least until it because automatic to use it.

ggandor:
> I'm sorry for coming off as definitive. I was just giving my initial impressions.

Oh, not at all, I didn't intend to "lecture" above, I'm just enthusiastic, and like to explain my thought process, debate it back and forth, and get feedback, that can lead to further improvements.

luxus:
its added as a custom plugin in the wiki, can this maybe get closed?

https://github.com/ChristianChiarulli/LunarVim/wiki/Example-plugin-configurations#lightspeed

ggandor:
I cannot edit the wiki, but
Hop: "...jump anywhere in the document" -> "anywhere in the window"
Lightspeed: "...file navigation" -> "on-screen navigation" or something like that. (Also, why just operator-pending mode mappings are added to which-key in the snippet?)

abzcoding:
@ggandor thank you for the wonderful i've updated the wiki, feel free to comment on this or reach out to me on discord if you had any other suggestions.

sidenote: we're planning to create something called `LunarModules` to help people discover and install plugins easily in LV, so we might come back to `lightspeed`

ggandor:
@abzcoding Thanks for the kind words! Actually, Lightspeed does not need any configuration at all, not even calling `setup`  (as of now, at least), so _if_ you think you want to include it among the wiki examples, then it might be better to just copy the setup snippet from the Readme as it is, showing the default settings.

PR #696, opened 2021-07-05
kmonad syntax support

IndianBoy42:
Support for syntax highlighting of https://github.com/kmonad/kmonad

Requires a plugin, Do you want a flag to disable installing it? its already lazy loaded by filetype

### DIFF
--- a/ftdetect/kmonad.vim
+++ b/ftdetect/kmonad.vim
@@ -1,0 +1,1 @@
+au BufWinEnter,BufRead,BufNewFile *.kbd setlocal filetype=kmonad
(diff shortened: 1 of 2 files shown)


rebuilt (approved):

ChristianChiarulli:
Can someone explain this to me?

IndianBoy42:
Kmonad is a software for remapping keyboards, it has a lot of features like macros, layers and tap-hold (different behavior when you tap vs hold)

The configs are written in its own lispy language. So this gives highlighting and some basic features for that

PR #697, opened 2021-07-05
vim-quick-scope plugin

IndianBoy42:
Highlights a character in each word that is unique to that line.

Helps with jumping quickly as you can just eyeball that word and see what you should press after 'f'

This seems quite useful for beginners

### DIFF
--- a/lua/default-config.lua
+++ b/lua/default-config.lua
@@ -70,8 +70,10 @@ O = {
     telescope_project = { active = false },
     dap_install = { active = false },
     tabnine = { active = false },
-    quickscope = { active = false, 
-        on_keys = {'f', 'F', 't', 'T'} -- Comment this line to have it always visible
+    quickscope = {
+      active = false,
+      event = "BufRead", -- Comment out this line to only load quickscope when keys are pressed. Then run :PackerSync
+      on_keys = { "f", "F", "t", "T" }, -- Comment this line to have it always visible
     },
   },
 
--- a/lua/default-config.lua
+++ b/lua/default-config.lua
@@ -70,7 +70,9 @@ O = {
     telescope_project = { active = false },
     dap_install = { active = false },
     tabnine = { active = false },
-    quickscope = { active = false },
+    quickscope = { active = false, 
+        on_keys = {'f', 'F', 't', 'T'} -- Comment this line to have it always visible
+    },
   },
 
   lang = {
--- a/lua/default-config.lua
+++ b/lua/default-config.lua
@@ -70,6 +70,7 @@ O = {
     telescope_project = { active = false },
     dap_install = { active = false },
     tabnine = { active = false },
+    quickscope = { active = false },
   },
 
   lang = {
--- a/lua/plugins.lua
+++ b/lua/plugins.lua
@@ -464,10 +464,11 @@ return require("packer").startup(function(use)
   use {
     "unblevable/quick-scope",
     keys = O.plugin.quickscope.on_keys,
-    event = (O.plugin.quickscope.on_keys ~= nil) and "BufRead" or nil,
+    event = O.plugin.quickscope.event,
     config = function()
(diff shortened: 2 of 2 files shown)


rebuilt (changes requested):
Can you stick to the convention of the 'disable' key being the last key in the list please? 

Another option for lazy loading is to enable it only on keypresses

```
    keys = {"f", "F", "t", "T"},
```

This means the highlighting won't be enabled by default when you open a file and will only activate after one of the keys is pressed.  This is a matter of preference.  It's how I have it set up, but I can see how the default behavior would be better for most.

rebuilt:
I'm not sure if Chris will approve this right now since he's focusing on things that improve the performance or fix bugs.

IndianBoy42:
> Can you stick to the convention of the 'disable' key being the last key in the list please?
> 
> Another option for lazy loading is to enable it only on keypresses
> 
> ```
>     keys = {"f", "F", "t", "T"},
> ```
> 
> This means the highlighting won't be enabled by default when you open a file and will only activate after one of the keys is pressed. This is a matter of preference. It's how I have it set up, but I can see how the default behavior would be better for most.

Fixed the first part.

For the second I do see the argument in that. My friend uses quick scope in that way, but for me the point of this plugin is too see what key i should press before even hitting 'f/t'. otherwise I would have to hit 'f', then look at which letter lit up and then hit that key. Its a tiny bit slower imo. 

I can make a config to allow the user to set which behaviour he wants

rebuilt:
I think it's ok to just define a second key for this.  I think the default should load with "BufRead"

```
  use {
    "unblevable/quick-scope",
    keys = O.plugin.quickscope.on_keys,
    event = O.plugin.quickscope.event,
    config = function()
        vim.g.qs_highlight_on_keys = O.plugin.quickscope.on_keys
    end,
    disable = not O.plugin.quickscope.active,
  }
```

And then in default-config.lua
```
    quickscope = { active = true, 
        event = "BufRead", -- Comment out this line to only load quickscope when keys are pressed. Then run :PackerSync
        on_keys = {'f', 'F', 't', 'T'} 
    },

```

rebuilt:
You've duplicated a line

```
         on_keys = {'f', 'F', 't', 'T'} -- Comment this line to have it always visible
```

The config fails to start now.

rebuilt (approved):
Looks good.  Thanks for your work on this.

rebuilt:
I think I mentioned it in another request of yours, but Chris is prioritizing requests that fix bugs or improve performance at the moment.  Not sure if he'll approve this yet.

IndianBoy42:
No worries, happy to help

Get Outlook for Android<https://aka.ms/AAb9ysg>

________________________________

jameswalmsley (approved):
I think this is a nice addition.
I'd like to see this functionality in a lua plugin, or maybe as an option added to lightspeed.

Maybe I'll add an issue on lightspeed..

jameswalmsley:
I had a conversation with the lightspeed author about quickscope. Please see their response:

https://github.com/ggandor/lightspeed.nvim/issues/23#issuecomment-874339534

Personally I think they are right..  and we should stick to lightspeed and not provide quick-scope as a recommendation/suggestion.

It's much better to learn how to move the lightspeed way.

rebuilt:
Hey James.  

Chris had a live-stream earlier and changed what the plugin architecture will look like going forward.
I've updated the wiki with the explanation https://github.com/ChristianChiarulli/LunarVim/wiki#install-your-own-plugins

It looks like we won't be approving any more plugins unless they are absolutely fundamental.  
All other plugins must be installed by the users in the lv-config.lua file

The process for adding a new plugin is dead simple.  Now we just need to build a set of example configurations to go in the Wiki.

I also think they should be commented out entries in the example.lv-config.lua file.

ChristianChiarulli:
This seems like it will belong to the user defined plugins. The TS text object stuff was really cool tho @IndianBoy42

PR #698, opened 2021-07-05
evenly resize split windows

bob3000:


rebuilt (approved):
I tested.  It's very nice.  I want this in LunarVim.

PR #699, opened 2021-07-05
reload buffer on external file change

bob3000:


rebuilt:
Isn't this already the behavior of neovim?  Or am I misunderstanding what this is supposed to do?

bob3000:
Hm, that's weird. I tested it again and you seem to be right. Actually I added these lines to my config a couple of days ago because it was in fact not reloading the buffer when the file was changed by a different application :thinking:  But now it also works for me without these lines. I guess I'll close this PR then. Thanks for reviewing my changes :pray:

PR #700, opened 2021-07-05
Add Treesitter extensions for better textobjects

IndianBoy42:
This adds three plugins that use treesitters parsing to provide smarter text objects

[ts_textobjects](https://github.com/nvim-treesitter/nvim-treesitter-textobjects) lets you define text objects for things like functions and classes

[ts_textsubjects](https://github.com/RRethy/nvim-treesitter-textsubjects.git) defines a smart text object that selects a semantic block around the cursor

[ts_hintobjects](https://github.com/mfussenegger/nvim-ts-hint-textobject) uses easymotion style hint labeling to show all the surrounding semantic blocks, letting you choose the correct one

I have also added which-key entries for all the above

### DIFF
--- a/lua/default-config.lua
+++ b/lua/default-config.lua
@@ -29,6 +29,29 @@ O = {
     ignore_install = { "haskell" },
     highlight = { enabled = true },
     rainbow = { enabled = false },
+    -- The below are for treesitter-textobjects plugin
+    textobj_prefixes = {
+      goto_next = "]", -- Go to next
+      goto_previous = "[", -- Go to previous
+      inner = "i", -- Select inside
+      outer = "a", -- Selct around
+      swap = "<leader>a", -- Swap with next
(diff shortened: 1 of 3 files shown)


rebuilt:
If it were up to me I'd say text objects can be enabled by default. They are pretty fundamental.  But that's a decision for @ChristianChiarulli 

I'm not going to have a chance to test today.  I'll pull it down tomorrow if Chris hasn't already  merged this.

jameswalmsley:
@rebuilt I'd agree with you on that.. its up to Chris.. but fundamentally this extends the current objects in Vim..

If we want to encourage people to use neovim and help them learn.. then it's nice if some of these fundamentals are the same everywhere..

jameswalmsley (approved):
I really like this.. I hope it can be a default..

PR #702, opened 2021-07-05
switch to absolute line numbers in insert mode

bob3000:
This switches to absolute line numbers in insert mode and back to relative line numbers when insert mode is left. if `relative_number` is set to `false` line numbers will stay absolute no matter which mode is active.


jameswalmsley (approved):
Awesome 👌

PR #711, opened 2021-07-05
Add csharp ls setup for cs filetype

PasiBergman:
Configuring csharp language server (omnisharp) as ftplugin for .cs files.


PasiBergman:
I need to fix the quotes.

PR #714, opened 2021-07-05
add autoformat for elixir - default to false

tiberiuc:
added autoformat for elixir, default to false, can be activated with 
```
O.elixir.autoformat = true
```
in lv-config.lua


rafamadriz:
Formatting with LSP is no longer going to be used, instead neoformat is going to be the default option. For reference see #715

PR #715, opened 2021-07-05
(feat) use neoformat as the defalt format option

rafamadriz:
* Get rid of all unnecessary complexity for setting up efm for formatting

* No need for efm server since it's no going to be used for formatting
  (some languages need it for lint, e.g python, if those cases the efm
  server was left and not completely removed)

*  Add `forma_on_save` option and remove all unnecessary format options for specific languages

### Benefits

1. Users can toggle format_on_save for all languages, and not have an option for every single language as it was before.
2. We don't need to worry about setting up formatters with efm anymore (which can be a pain to do)
3. Simplify the config, and get rid of a ton of unnecessary settings for specific languages. We don't have to parse x formatter that an user want to efm or whatever. Just having the formatter you want installed and in your path will work out of the box.

Hey @ChristianChiarulli this is done, and tested in my machine, all working as expected. Take a look a let me know if you want any change (also test this to be 100% sure before merging)

### DIFF
--- a/ftplugin/c.lua
+++ b/ftplugin/c.lua
@@ -18,14 +18,3 @@ require("lspconfig").clangd.setup {
     }),
   },
 }
-
-if O.lang.clang.autoformat then
-  require("lv-utils").define_augroups {
-    _clang_autoformat = {
-      { "BufWritePre *.c lua vim.lsp.buf.formatting_sync(nil,1000)" },
-      { "BufWritePre *.h lua vim.lsp.buf.formatting_sync(nil,1000)" },
-      { "BufWritePre *.cpp lua vim.lsp.buf.formatting_sync(nil,1000)" },
-      { "BufWritePre *.hpp lua vim.lsp.buf.formatting_sync(nil,1000)" },
-    },
-  }
-end
--- a/ftplugin/go.lua
+++ b/ftplugin/go.lua
@@ -6,17 +6,7 @@ require("lspconfig").gopls.setup {
   on_attach = require("lsp").common_on_attach,
 }
 
-if O.lang.go.autoformat then
-  require("lv-utils").define_augroups {
-    _go_format = {
-      { "BufWritePre", "*.go", "lua vim.lsp.buf.formatting_sync(nil,1000)" },
-    },
-    _go = {
-      -- Go generally requires Tabs instead of spaces.
-      { "FileType", "go", "setlocal tabstop=4" },
-      { "FileType", "go", "setlocal shiftwidth=4" },
-      { "FileType", "go", "setlocal softtabstop=4" },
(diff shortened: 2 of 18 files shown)


ChristianChiarulli:
Thanks, works great 🎃

soerenmartius:
This doesn't seem to work per default for Terraform anymore? How am I supposed to set autoformat for `*.tf` files on save?

rafamadriz:
> This doesn't seem to work per default for Terraform anymore? How am I supposed to set autoformat for `*.tf` files on save?

Neoformat supports terraform, (see [here](https://github.com/sbdchd/neoformat#supported-filetypes)). You just need to have terraform installed and in your $PATH.

Here how to [install terraform](https://www.terraform.io/downloads.html)

soerenmartius:
@rafamadriz I have terraform installed in my path and I've seen that Neoformat supports terraform also - thanks for that. Anyways, if I save a `*.tf` file, it won't get formated. If I run `:Neoformat! terraform` manually, it works.

soerenmartius:
just to make sure, I have `O.format_on_save = true` set in my config.

rafamadriz:
> just to make sure, I have `O.format_on_save = true` set in my config. Autoformat seems to work on other filetypes such as `*.ts` though.

Okay, I think that I know the problem, can you open a terraform file and do `:set filetype`, what's the output ?

rafamadriz:
hey @soerenmartius I've been testing this and it seems like the autoformatting for terraform files doesn't work correctly. I also tried to set an autocommand for filetype ft, but that didn't work either. I'm not sure why this happens only for terraform.

soerenmartius:
> :set filetype

```
filetype=tf
```

soerenmartius:
> hey @soerenmartius I've been testing this and it seems like the autoformatting for terraform files doesn't work correctly. I also tried to set an autocommand for filetype ft, but that didn't work either. I'm not sure why this happens only for terraform.

btw, when I run `:Neoformat! terraform` it changes the syntax highlighting also

soerenmartius:
Shall I open a new issue for this?

rafamadriz:
> Shall I open a new issue for this?

If you have the time open one in the Neoformat repo, there's nothing wrong in our end. I think the issue is how Neoformat is setting the formatter for terraform files. From what I saw it seems like they are setting literally `terraform`, instead of `ft` which is the output you get when doing `set filetype` in a terraform file. But I'm not sure tho, I may be wrong.

soerenmartius:
Hi, I've opened #782 for further discussion. `nvim-treesitter` uses `hcl` as a filetype which is technically spoken also correct. Neoformat used `terraform` as outlined by you. 

You can see that treesitter also applies the right syntax highlighting when running `:set filetype=hcl`

PR #716, opened 2021-07-05
Fix vue lsp root_dir configuration

PasiBergman:
More specific root_dir configuration for most (typical) vue projects. The "." does not work in certain cases.


nathanielks:
Thank you for this! 🙏

PR #717, opened 2021-07-05
Replace LazyGit Plugin with FTerm Instance

jonathan-raines:
The LazyGit plugin is redundant when using the FTerm plugin. You can issue commands to create a variety of floating window programs. This takes the default key mappings used in the FTerm readme, can easily be changed to something else. I like this way better as it's one less plugin to worry about and it functions virtually the same


rebuilt (changes requested):
Hey, I hope your ruby configuration is working.  Thanks for putting in this pull request.  It looks neat.

The 'gg' binding doesn't work for me 

```
<Leader>gg
```

The alt-l binding works but I think mapping the alt key won't work over an ssh connection.  I don't think that matters as long as you get the leader-gg binding working.  Just fix the gg binding and I'll approve it so Chris can look at it.

rebuilt:
For anyone else reading this: To test this pull request:

```
git fetch origin pull/717/head:lazygit
git checkout lazygit

# return to master
git checkout master
```

jonathan-raines:
I'll get the gg keybind fixed tomorrow and I'll ping you afterwards

rebuilt (approved):
To enable, floatterm.active must be set to true

```
O.plugin.floatterm.active = true
```

ChristianChiarulli:
Is there any benefit to the lazygit plugin vs this?

rebuilt:
This FTerm method fails silently if lazygit is not installed.

Here's how lazygit checks for the executable

```
local function is_lazygit_available()
    return fn.executable("lazygit") == 1
end

 if is_lazygit_available() ~= true then
        print("Please install lazygit. Check documentation for more information")
        return
    end
```

ChristianChiarulli:
@rebuilt if we can get that in this pr I'll approve it, then we can add options for node, htop, python, ncdu, lazydocker etc...

rebuilt:
@jonathan-raines  I pushed some changes to your repo.  When you're done reviewing those changes, push them here.

rebuilt (changes requested):
@jonathan-raines I pushed some changes to your repo. When you're done reviewing those changes, push them here.

kdheepak:
Well, I'm biased but I'm sad LazyGit.nvim didn't make the cut ;) 

But in all seriousness, looking forward to where this project is heading! Great work everyone!

PR #719, opened 2021-07-06
telescope project and packer had the same keybinding

abzcoding:
Just changed the telescope project keybinding to `<leader>P` instead of `<leader>p`

### DIFF
--- a/lua/lv-which-key/init.lua
+++ b/lua/lv-which-key/init.lua
@@ -295,7 +295,7 @@ if O.plugin.telescope_project.active then
   -- open projects
   vim.api.nvim_set_keymap(
     "n",
-    "<leader>p",
+    "<leader>P",
     ":lua require'telescope'.extensions.project.project{}<CR>",
     { noremap = true, silent = true }
   )


rebuilt (approved):

fabriziosestito:
I was about to comment on this but it was merged too fast.

I would have actually chosen `<leader>P` for packer and leave the old `<leader>p` for projects, this mainly because the project keybinding is going to be used more than the packer one.
Also, for what it's worth, Doom and Spacemacs refer to project using the mnemonic lowercase `p`.

abzcoding:
@rebuilt, I do agree with @fabriziosestito, @ChristianChiarulli which one do you prefer for project
```vim
<leader>p
```
or
```vim
<leader>P
```

rebuilt:
I don't have a preference for this.  The bindings in this commit are what I saw Chris use on stream so I figured they would be fine.

rebuilt:
Incidentally, the telescope_project plugin only loads correctly for me if I don't lazy load it.  Are you guys lazy loading this plugin without errors?

abzcoding:
Can confirm that i get errors as well

rebuilt:
Can you replace the current telescope-project config with this and confirm it works correctly please?  Then if you want, submit a pull request with this update.  

```
  -- Use project for telescope
  use {
    "nvim-telescope/telescope-project.nvim",
    event = "BufRead",
    setup = function() vim.cmd[[packadd telescope.nvim]] end,
    disable = not O.plugin.telescope_project.active,
  }
```

abzcoding:
@rebuilt 
added the PR

barun108:
there are more keybindings conflicts, 
I enabled few plugins ( don't remember exactly which one ) 
and, then 'jk' stopped working ( inoremap 'jk' <Esc> )

rebuilt:
Run this command to see where the mapping is set

```
:verbose imap jj
```

PR #723, opened 2021-07-06
fix telescope project issue

abzcoding:
all the credit goes to @rebuilt,
when you run the `<leader>P`, you will no longer get the following error:
Screenshot-20210706131559-2408x2394


rebuilt (approved):

rebuilt:
telescope-project used to load by using the 'after' keyword but broke when we added the 'disable' keyword.  This makes sure telescope is loaded before project.

PR #727, opened 2021-07-06
Fix error in ftplugin/php.lua when opening a PHP file.

juanlopez4691:
Trying to open/edit a PHP file throws an error:
```
Error detected while processing /Users/xxx/.config/nvim/ftplugin/php.lua:
E5113: Error while calling lua chunk: /Users/xxx/.config/nvim/ftplugin/php.lua:16: attempt to index field 'format' (a nil value)
Error detected while processing FileType Autocommands for "*":
Error while processing file: /Users/xxx/.config/nvim/ftplugin/php.lua
/Users/joanlopez/.config/nvim/ftplugin/php.lua:16: attempt to index field 'format' (a nil value)
Press ENTER or type command to continue
```

### DIFF
--- a/utils/installer/lv-config.example.lua
+++ b/utils/installer/lv-config.example.lua
@@ -65,6 +65,12 @@ O.lang.php.environment.php_version = "7.4"
 O.lang.php.diagnostics.signs = true
 O.lang.php.diagnostics.underline = true
 O.lang.php.filetypes = {"php", "phtml"}
+O.lang.php.format = {
+  format = {
+    default = "psr12"
+  }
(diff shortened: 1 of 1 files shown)


PZ31k0nauT:
That is strange. I did set this setting in default-config.lua. But it seems it got lost along the way.

I really think it would be better to set it there again.

Also I'm suprised it works because in php.lua you can find this:
```lau
format = {
        braces = O.lang.php.format.braces,
},

Hmm. But yeah I'm completely new to lua.

I will test it later. When it works it works.

PR #755, opened 2021-07-07
fix a mistake in get_lsp_client to get the right lsp client

iweizime:


ChristianChiarulli:
why is -1 necessary?

iweizime:
> why is -1 necessary?

When I open a cpp file with clangd as the lsp server, the status line shows 'LSP Inactive' instead of 'clangd'.
After adding a few `print`, I think I found the bug.

`vim.fn.index(list, 'x')` returns `-1` when `x` is not in the list. So I think there should be `-1`.

abzcoding:
this has been already fixed in `rolling` inside `lua/core/galaxyline.lua`

PR #756, opened 2021-07-07
we don't need all the neovim git history to install the latest version

abzcoding:


rebuilt (approved):
Yeah, that's much faster

rebuilt:
No need for changes in the wiki.  The manual install instructions already include this flag.

jameswalmsley (approved):
:+1:

PR #760, opened 2021-07-07
Add empty buffer option to dashboard

maash3r:
An empty buffer would be helpful I believe. [vim-startify](https://github.com/mhinz/vim-startify) has something like it too.
Something like this:
image


luxus:
thanks @maash3r, nice idea. in the rolling branch you can customize the dashboard in your own user config.
there will be some more refactoring (maybe tomorrow) and it will be hopefully in master.

maash3r:
Sure @luxus , saw it on stream, will close this.

PR #762, opened 2021-07-07
fix lazygit esc issue

### LINKED ISSUE #759: [Bug]: Escape does not work on lazzygit when opening it inside neovim
### Problem description

If I open lazzygit inside neovim, the esc does not work on lazzygit but in VIM.

The esc key is used on lazzygit as a <GO BACK> action, but since it is opened on nvim the esc behaves just like it behaves on buffers. 

The same would be applied to other apps opened inside floatterm.

### LunarVim version

Master

### Neovim version (>= 0.5)

NVIM v0.6.0-dev+1488-ga6cdfa27d

### Relevant log output

_No response_

### Screenshots

_No response_

### I have

- [X] Read the readme
- [X] Read the wiki

abzcoding:
Fixes #759 
While using `Lazygit` inside `FTerm`, `esc` key will act as quit instead of going to normal mode

### DIFF
--- a/lua/lv-floatterm/init.lua
+++ b/lua/lv-floatterm/init.lua
@@ -40,6 +40,17 @@ M.config = function()
     end
     lazy:toggle()
   end
+
+  -- Map esc to exit inside lazygit
+  vim.api.nvim_exec([[
+  function LazyGitNativation()
+    echom &filetype
+    if &filetype ==# 'FTerm'
+      tnoremap <Esc> q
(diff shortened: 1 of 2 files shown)


abzcoding:
@ChristianChiarulli what do you think about this, you seem to use `lazygit` a lot, what should the `esc` key do?

ChristianChiarulli:
i'll have to test this out, but I think this behavior is ideal.

PR #765, opened 2021-07-07
[REFACTOR] allow customizing efm configuration

kylo252:
- efm now follows the user-config in "~/.config/efm-langserver/config.yaml".
- Add an optional "efm-config.yaml" with support for a lot of popular
  linters/formmatters that works as a *fallback* in case the user-config
  file is completely missing. This gives the user the ability to copy the provided template
  and override whatever option they'd like.
- Using efm is now opt-in and can be selectively turned on for each
  language separately. For example: "O.lang.lua.efm.active = true"

### DIFF
--- a/ftplugin/c.lua
+++ b/ftplugin/c.lua
@@ -22,3 +22,7 @@ require("lspconfig").clangd.setup {
     }),
   },
 }
+
+if O.lang.clang.efm.active == true then
+  require("lsp.efm-ls").generic_setup({"c", "cpp"})
+end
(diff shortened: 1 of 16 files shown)


ChristianChiarulli:
this looks like a great idea since it abstracts the config, I'll just need a little bit to wrap my head around it

michaelslec:
I love this. This is a really great idea for a refactor!

In my opinion, this definitely brings up some questions about what configuration should look like in general for Lunarvim, as I think splitting config across too many files would be troublesome. One idea might be having opt-in packages instead of configuration with the lv-config.lua file. Something like how DWM patches code

abzcoding:
@kylo252 we kind of removed efm, do you think this PR is still relevant?

kylo252:
> @kylo252 we kind of removed efm, do you think this PR is still relevant?

Yeah we can always bring it back if necessary.

PR #766, opened 2021-07-07
doas support for the installer

luis-07:


luis-07:
there is a bug which breaks the installer 
DO NOT COMMIT THIS CODE UNTIL I FIX THE BUG

luis-07:
I fixed the installer

kylo252:
The installer doesn't require `sudo` elevation anymore. Feel free to open another PR if you feel something is not right when installing on Gentoo.

PR #767, opened 2021-07-07
show virtualenv inside galaxyline

### LINKED ISSUE #732: Python virtual env in galaxyline

abzcoding:
Fixes #732 
@ChristianChiarulli  did you have something like this in mind?
venv

### DIFF
--- a/lua/lv-galaxyline/init.lua
+++ b/lua/lv-galaxyline/init.lua
@@ -125,6 +125,47 @@ table.insert(gls.left, {
     highlight = "StatusLineGitDelete",
   },
 })
+-- get output from shell command
+function os.capture(cmd, raw)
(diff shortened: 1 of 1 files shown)


ChristianChiarulli:
yes, this is very cool

abzcoding:
so only activate it when we have python virtualenv or you want to have something similar for like rvm and others as well?

ChristianChiarulli:
probably good to stick to python as a first pass, also does this support conda?

abzcoding:
@ChristianChiarulli  now it does, can you test it on your end and see if you like it or not?

ChristianChiarulli:
Just tested, it looks great, do you think this is ready?

abzcoding:
yep I think it's ready

PR #773, opened 2021-07-07
Option to rename rust parameter hints & cleanup

Henrik-N:
- Added the ability to change parameter hint prefix for Rust in the user config file
- Reordered the language list to be in alphabetical order


Henrik-N:
Made a mistake, ignore this

PR #780, opened 2021-07-08
Use which-key for spelling hints

roobert:
This is a nice default to have which doesn't require any additional plugins or configuration but simply replaces the built-in extremely basic spelling hint list with the which-key interface.

To test, simply:
```
:PackerCompile
set spell -- enable spell checking
z= -- press this on a misspelled word that is underlined
```

I haven't included the following,  but this is a nice addition to enable spell checking by default for certain filetypes:
```
vim.cmd [[autocmd FileType text,latex,tex,md,markdown setlocal spell]]
```

Alternatively, having a keybinding to toggle spell checking could be an option?

### DIFF
--- a/lua/lv-which-key/init.lua
+++ b/lua/lv-which-key/init.lua
@@ -21,6 +21,7 @@ which_key.setup {
       z = true, -- bindings for folds, spelling and others prefixed with z
       g = true, -- bindings for prefixed with g
     },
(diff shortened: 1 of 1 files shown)


roobert:
Example screenshot:
Screenshot 2021-07-08 at 08 14 13

jameswalmsley (approved):

PR #785, opened 2021-07-08
Fix rust lsp performance issue

hulufei:
For some context, I raised a question in discord:

```
rust_analyzer lsp is really slow, I have to wait like 10s for every new opened rust file to 
response with lsp queries
```

Finally, I got the answer, LSP setup should not be located in `ftplugin/rust.lua`, because every time open a rust file, it will load this `ftplugin/rust.lua` language file, which causes rust_analyzer re-setup and re-index for current rust file, so it is slow.

So I'm wondering if we should put lsp setup code out of language files.


Uzaaft:
Just out of curiosity, how big is the reduction in wait time? Never thought about the 10s wait time whenever I opened a rust file, but now that I've noticed it, I can't ignore it xD

ChristianChiarulli:
I think we should leave them in lang files, but we may need a check to see the current lsp is active. And if so just return.

hulufei:
@ChristianChiarulli Make sense, I'll try to add the check.

hulufei:
@ChristianChiarulli Maybe another PR to add this check for all language files that have lsp setup code?

ChristianChiarulli:
@hulufei have you tested that this works? I tried something similar and it didn't seem to work for me.

hulufei:
@ChristianChiarulli I'd like to do it.

ChristianChiarulli:
closed by #790

PR #786, opened 2021-07-08
Replace deprecated value in  lv-telescope/init.lua

Uzaaft:
Fix issue with deprecated config value in telescope. Ref: telescope.changelog-839.

### DIFF
--- a/lua/lv-telescope/init.lua
+++ b/lua/lv-telescope/init.lua
@@ -36,7 +36,7 @@ telescope.setup {
     file_sorter = require("telescope.sorters").get_fzy_sorter,
     file_ignore_patterns = {},
     generic_sorter = require("telescope.sorters").get_generic_fuzzy_sorter,
-    path_display = "shorten,
+    path_display = {"shorten_path"},
     winblend = 0,
     border = {},
     borderchars = { "─", "│", "─", "│", "╭", "╮", "╯", "╰" },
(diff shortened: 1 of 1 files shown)


lua/lv-telescope/init.lua
```diff
@@ -36,7 +36,7 @@ telescope.setup {
     file_sorter = require("telescope.sorters").get_fzy_sorter,
     file_ignore_patterns = {},
     generic_sorter = require("telescope.sorters").get_generic_fuzzy_sorter,
-    shorten_path = true,
+    path_display = {"shorten_path"},
```
abzcoding, line 39: shouldn't this be `shorten` instead of `shorten_path` ? as described [here](https://github.com/nvim-telescope/telescope.nvim/wiki/Path-Display-Configuration#settings)

Uzaaft: The original issue let me to running the command: :help telescope.changelog-839.
There, they say in the examples:
"...
You can also still pass this to a single builtin call:
```lua
require("telescope.builtin").find_files{
path_display = {"shorten_path}
}
```
"

But I see now that they also mention above the examples:
"
The new way to configure to configure path displays is with:
    `path_display`: It is a table and accepts multiple values:
        - "hidden"    hide file names
        - "tail"      only display the file name, and not the path
        - "absolute"  display absolute paths
        - "shorten"   only display the first character of each directory in
                      the path
"
@abzcoding  I'll rewrite it to shorten.

abzcoding (approved):
Seems good to me, thank you for your contribution.

rebuilt:
Thanks for the fix!  That was quick.  I only noticed it a few minutes ago.  

I tested fix locally before merging.

Uzaaft:
No problem! I like keeping my stuff updated, so I often pull and update LunarVim multiple times a day(or even multiple times every hour xD)

Uzaaft:
@rebuilt  @ChristianChiarulli Should I open a similar PR for the stable branch. This issue is still present there.

ChristianChiarulli:
stable branch is being removed soon

PR #791, opened 2021-07-08
Scrolloff option in lv-config

Gruak:
A neat little feature that I find quite useful

> 	Minimal number of screen lines to keep above and below the cursor.
	This will make some context visible around where you are working.  If
	you set it to a very large value (999) the cursor line will always be
	in the middle of the window (except at the start or end of the file or
	when long lines wrap).


abzcoding (approved):
thank you for your contribution

PR #796, opened 2021-07-08
Split plugin loading logic from the configuration

Tastyep:
This splits the plugin configuration from the loading logic.
The plugins.lua file would now contain only the plugin's configurations.
The plugin-loader.lua file contains the logic for loading plugins + handling packer's initialization.

I went for an object-oriented approach for the package-loader as it provides a clear interface.
I think it's better to have an object with clear responsibilities that can be injected as a dependency instead of doing require("mymodule").myfunction everywhere in the code. IMO the code would be more structured and easier to maintain in the future.
Also, it will be possible to have a state and pass the package-loader to other parts of the application (as I said dependency injection). For example, we could inject it into the which-key module (probably at the cost of not lazy loading it though) to fetch registered modules with a lv-config file and retrieve the plugin's mapping from there.

Let's keep things simple for now and changes incremental, let me hear what you think of it.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -4,9 +4,14 @@ if not status_ok then
   print "something is wrong with your lv-config"
   print(error)
 end
+
 require "keymappings"
-require "plugins"
+
+local plugins = require "plugins"
+local plugin_loader = require("plugin-loader").init()
+plugin_loader:load { plugins, O.user_plugins }
 vim.g.colors_name = O.colorscheme -- Colorscheme must get called after plugins are loaded or it will break new installs.
+
 require "settings"
 require "lv-utils"
 
--- a/lua/plugin-loader.lua
+++ b/lua/plugin-loader.lua
@@ -1,0 +1,46 @@
+local plugin_loader = {}
+
+function plugin_loader:init()
+  local execute = vim.api.nvim_command
+  local fn = vim.fn
+
+  local install_path = fn.stdpath "data" .. "/site/pack/packer/start/packer.nvim"
+  if fn.empty(fn.glob(install_path)) > 0 then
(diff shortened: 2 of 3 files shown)


rebuilt (approved):
I tested to make sure it works and it does.
I tested the performance impact with 1000 runs of this pr and 1000 runs of master.  There's no difference.  I wouldn't expect any difference since it's probably just loading the same packer_compiled file anyway.
I like the approach.  It really cleans things up and separates responsibilities.  

Great work on this pr!

rebuilt:
This might be too much change right before a major release.  Not sure Chris will approve it.

abzcoding (approved):
:100:

abzcoding:
@ChristianChiarulli, both me and rebuilt have approved this, but since it is a pretty big change, we're waiting for you to review and then merge it. 😃

ChristianChiarulli:
hot reload works with this?

Tastyep:
> hot reload works with this?

No, hot reload would require much more work (not undoable, though).
However, this provides the capability for reloading plugins at different points in time.
Making a call to plugin_loader::load allows to reload a new list of plugins

Tastyep:
Up ? The scope of this PR is very minimalist, it only splits the logic in plugin-loader and keeps the configuration in plugins.

abzcoding:
> Up ? The scope of this PR is very minimalist, it only splits the logic in plugin-loader and keeps the configuration in plugins.

@ChristianChiarulli LGTM :rocket:

abzcoding:
@Tastyep is it ready to merge?

Tastyep:
Yep, I just rebased it

PR #803, opened 2021-07-09
Add support for FreeBSD

gregf:
Simple patch that lets LunarVim work on FreeBSD. Only been using it the last 10 minutes, but seems to be working for me.

### DIFF
--- a/utils/bin/install-latest-neovim
+++ b/utils/bin/install-latest-neovim
@@ -1,4 +1,4 @@
-!#/bin/bash
+!#/usr/bin/env bash
 cd ~
 sudo rm -r neovim
 git clone --branch master --depth 1 https://github.com/neovim/neovim
--- a/utils/installer/install.sh
+++ b/utils/installer/install.sh
@@ -1,7 +1,15 @@
-#!/bin/sh
+#!/usr/bin/env bash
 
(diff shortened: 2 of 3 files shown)


abzcoding:
added #807 to make sure the installation process doesn't  break after changes

utils/bin/install-latest-neovim
```diff
@@ -1,4 +1,4 @@
-!#/bin/bash
+!#/usr/bin/env bash
```
Tastyep, line 1: The shebang should be `#!/usr/bin/env bash` like for the other files

utils/installer/install.sh
```diff
@@ -89,6 +103,8 @@ installpynvim() {
     if [ -f "/etc/gentoo-release" ]; then
         echo "Installing using Portage"
         sudo emerge -avn dev-python/pynvim
+    elif [ "$(uname -s)" == "FreeBSD" ]; then
+        sudo pkg install py38-pynvim
```
Tastyep, line 107: If I understand correctly, the pip3 symlink is not created when installing the package.
Would it be possible to make one in a directory listed in $PATH ? This way we could remove the || conditions below

Tastyep (changes requested):

PR #807, opened 2021-07-09
Github workflows

abzcoding:
We should have some sort of testing, to make sure nothing breaks LunarVim, this is just an attempt to do so,
I'm open to suggestions

@ChristianChiarulli  @rebuilt  @rafamadriz

### DIFF
--- a/.github/workflows/install.yaml
+++ b/.github/workflows/install.yaml
@@ -1,0 +1,71 @@
+name: install
+on:
+  push:
+    branches: '**'
+  pull_request:
+    branches:
+      - 'master'
+      - 'release-[0-9]+.[0-9]+'
+
+jobs:
+  unixish:
+    name: ${{ matrix.os }} ${{ matrix.runner }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - runner: ubuntu-20.04
+            os: linux
+          - runner: macos-10.15
+            os: osx
+    runs-on: ${{ matrix.runner }}
+    if: github.event.pull_request.draft == false
+    steps:
+      - uses: actions/checkout@v2
+      
+      - name: Install dependencies for Linux
+        if: matrix.os == 'linux'
+        run: |
+          sudo apt-get install -y gettext libtool libtool-bin autoconf automake cmake g++ pkg-config unzip build-essential
+          sudo bash ./utils/bin/install-latest-neovim
+
+      - name: Installl dependencies for OSX
+        if: matrix.os == 'osx'
+        run: |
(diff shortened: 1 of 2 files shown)


Tastyep:
Great idea !

.github/workflows/install.yaml
```diff
@@ -0,0 +1,71 @@
...
+    runs-on: ${{ matrix.runner }}
+    if: github.event.pull_request.draft == false
+    steps:
+      - uses: actions/checkout@v2
+      
+      - name: Install dependencies for Linux
+        if: matrix.os == 'linux'
+        run: |
+          sudo apt-get install -y gettext libtool libtool-bin autoconf automake cmake g++ pkg-config unzip build-essential
+          sudo bash ./utils/bin/install-latest-neovim
```
Tastyep, line 30: We should not specify bash, if bash is required to run the script, then we should update the shebang

Tastyep (changes requested):

Tastyep:
While at it, could we add a formatting check ? `stylua --check` should do the job.

.github/workflows/install.yaml
abzcoding: Yeah, I think we should update it since we have `bash` in the readme for installation as well.

Tastyep (approved):

abzcoding:
> While at it, could we add a formatting check ? `stylua --check` should do the job.

ok why do we have [this much](https://github.com/abzcoding/LunarVim/runs/3028225889?check_suite_focus=true) formatting errors :sob:

abzcoding:
> > While at it, could we add a formatting check ? `stylua --check` should do the job.
> 
> ok why do we have [this much](https://github.com/abzcoding/LunarVim/runs/3028225889?check_suite_focus=true) formatting errors sob

#811 will address this :feelsgood:

.github/workflows/install.yaml
abzcoding: #803 has addressed this already

Tastyep: Then we only need to do `sudo ./utils/bin/install-latest-neovim`, right ?

Tastyep: While reviewing #803 I noticed a typo in the shebang of the installer, which is why you need to specify bash here.

ChristianChiarulli:
@abzcoding do you think this is ready?

abzcoding:
> @abzcoding do you think this is ready?

I think so, we should add new test cases later on

luxus:
@ChristianChiarulli can this be merged?

PR #818, opened 2021-07-09
make shiftwidth and tabstop configurable

abzcoding:
I need my 4 char wide tab stop :)


ChristianChiarulli:
@abzcoding everything at my work is formatted for 2, also indentline doesn't play nice with 4 right now, as far as I can tell.

PR #827, opened 2021-07-10
Feature/nvimtree configuration

krsntn:
allow user to configure nvimtree setting in lv-config.lua file


utils/installer/lv-config.example.lua
```diff
@@ -28,6 +28,12 @@ O.plugin.indent_line.active = false
 O.plugin.zen.active = false
 O.plugin.zen.window.height = 0.90
 
+-- nvimtree
+O.nvimtree.side = "left"
```
abzcoding, line 32: I think it is better to comment these lines out. to show that you can change them
cause that would [change](https://github.com/ChristianChiarulli/LunarVim/blob/rolling/utils/installer/install.sh#L115) the behavior for new installs.

krsntn: Cool! got it. I have made new commits to comment it out.

abzcoding (approved):
thank you for your contribution

abzcoding:
nvim-tree is already user-configurable through `lua/core/nvimtree.lua`->`M.config` function

```lua
-- lv-config.lua

O.plugin.nvimtree.side = "right"
-- or any other config you want under O.plugin.nvimtree.
```

tree

PR #834, opened 2021-07-10
nvim-tree settings now read from the default lv-settings file

arxra:
The default setting for auto-closing wasn't used in the settings


abzcoding:
#827 will address this

abzcoding (approved):
nice catch :eyes:

PR #840, opened 2021-07-10
Vimtex : Autosave and ignore errors options

atropos112:
I find myself using autosave (save &compile on focus change) on latex a lot as it compiles stuff in the background as I do something else, a lot of people i know have something similar within their IDEs, I believe it is an option that gets used by a lot of people, some hate it though so I thought it might be a good idea to give the option.

For ignore errors, sometimes latex gives errors the user may not care about and this results in them being bothered every time they compile.

I am unsure if lv-config.lua is the right place for this? I wanted to use `O.vimtex.autosave` and `O.vimtex.ignore_errors` but was getting some weird `lv-utils` error, not sure if this matters.

### DIFF
--- a/lua/lv-vimtex/init.lua
+++ b/lua/lv-vimtex/init.lua
@@ -1,7 +1,7 @@
 vim.g.vimtex_compiler_method = "latexmk"
 vim.g.vimtex_view_method = "zathura"
 vim.g.vimtex_fold_enabled = 0
-vim.g.vimtex_quickfix_ignore_filters = O.vimtex_ignore_errors
+vim.g.vimtex_quickfix_ignore_filters = O.lang.latex.ignore_errors
 O.user_which_key["L"] = {
   name = "+Latex",
   c = { "<cmd>VimtexCompile<cr>", "Toggle Compilation Mode" },
@@ -22,7 +22,7 @@ vim.api.nvim_exec(
     ]],
   false
 )
-if (O.vimtex_autosave)
+if (O.lang.latex.auto_save)
 then
   vim.api.nvim_exec([[au FocusLost * :wa]],false)
 end
--- a/lua/lv-vimtex/init.lua
+++ b/lua/lv-vimtex/init.lua
@@ -2,8 +2,16 @@ vim.g.vimtex_compiler_method = "latexmk"
 vim.g.vimtex_view_method = "zathura"
 vim.g.vimtex_fold_enabled = 0
 vim.g.vimtex_quickfix_ignore_filters = O.vimtex_ignore_errors
+O.user_which_key["L"] = {
+  name = "+Latex",
+  c = { "<cmd>VimtexCompile<cr>", "Toggle Compilation Mode" },
+  f = { "<cmd>call vimtex#fzf#run()<cr>", "Fzf Find" },
+  i = { "<cmd>VimtexInfo<cr>", "Project Information" },
+  s = { "<cmd>VimtexStop<cr>", "Stop Project Compilation" },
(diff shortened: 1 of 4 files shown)


utils/installer/lv-config.example.lua
```diff
@@ -44,6 +44,10 @@ O.lang.python.analysis.use_library_code_types = true
 -- javascript
 O.lang.tsserver.linter = nil
 
+-- latex 
+O.vimtex_autosave = false -- saves and compiles on focus change
```
abzcoding, line 48: Please comment these lines out, to show users that they can change it, but not actually forcing it on all users
because this file is [installed](https://github.com/ChristianChiarulli/LunarVim/blob/rolling/utils/installer/install.sh#L116) for all users

utils/installer/lv-config.example-no-ts.lua
```diff
@@ -44,6 +44,10 @@ O.lang.python.analysis.use_library_code_types = true
 -- javascript
 O.lang.tsserver.linter = nil
 
+-- latex 
+O.vimtex_autosave = false -- saves and compiles on focus change
```
abzcoding, line 48: I would appreciate it if you comment out these lines as well

lua/lv-vimtex/init.lua
```diff
@@ -23,3 +22,7 @@ vim.api.nvim_exec(
     ]],
   false
 )
+if (O.vimtex_autosave)
```
abzcoding, line 25: can you add a `auto_save = false` inside [lua/default-config.lua](https://github.com/ChristianChiarulli/LunarVim/blob/rolling/lua/default-config.lua#L151) 
and use `O.lang.latex.auto_save` instead of `O.vimtex_autosave`

atropos112: Thank you for your comments, I think its much cleaner now.

abzcoding (dismissed):
thank you for your contribution :+1: 

sidenote: can you fix the conflicts :sweat_smile:

atropos112:
> thank you for your contribution +1
> 
> sidenote: can you fix the conflicts sweat_smile

I was pushing to wrong branch it seems, changed to master (which is what I thought i initially did?)

abzcoding:
> > thank you for your contribution +1
> > sidenote: can you fix the conflicts sweat_smile
> 
> I was pushing to wrong branch it seems, changed to master (which is what I thought i initially did?)

please change it to `rolling` , 
`master` branch is reserved for stable changes (after being tested on rolling)

atropos112:
`rolling` doesnt even have vimtex in it and its plugin.lua vimtex content is behind the `master` branch. It seems `master` is the actual rolling and `rolling` is just behind. Its more work than I thought, I'll work on this later today.

abzcoding:
> `rolling` doesnt even have vimtex in it and its plugin.lua vimtex content is behind the `master` branch. It seems `master` is the actual rolling and `rolling` is just behind. Its more work than I thought, I'll work on this later today.

Nah we removed a lot of plugins from LunarVim, so might have to close this PR
rolling

atropos112:
> > `rolling` doesnt even have vimtex in it and its plugin.lua vimtex content is behind the `master` branch. It seems `master` is the actual rolling and `rolling` is just behind. Its more work than I thought, I'll work on this later today.
> 
> Nah we removed a lot of plugins from LunarVim, so might have to close this PR
> rolling

With respect to vimtex only changes I can see is that all vimtex config is removed, this is just worse.

abzcoding:
@atropos112  can you create an issue and explain the logic for `vimtex` stuff, we might bring it back

PR #841, opened 2021-07-10
Feature/goto preview

MatthiasX86:
Thought this would be a great addition. Thanks to such great work by https://github.com/rmagatti

https://github.com/rmagatti/goto-preview


abzcoding:
I think this is not a core feature and users can add them to their `user_plugins` if they want to as explained [here](https://github.com/ChristianChiarulli/LunarVim/wiki/Installing-Plugins#user-defined-plugins).

abzcoding:
Since the other PR got merged, I'm guessing this will not be merged , but just in case added this to [wiki](https://github.com/ChristianChiarulli/LunarVim/wiki/Example-plugin-configurations#goto-preview)

PR #845, opened 2021-07-10
Fallback Formatters

abzcoding:
to change the formatter for each lang:
```lua
-- lv-config.lua file

O.lang.python.formatter = { Formatter.black, Formatter.yapf }
```

to declare and use a new formatter:
```lua
-- lv-config.lua file
Formatter.sth = {
  exe = "stuff",
  args = {},
}

O.lang.python.formatter = {Formatter.sth, Formatter.black}
```


abzcoding:
@ChristianChiarulli I think this ready now

lua/lv-formatter/init.lua
```diff
@@ -11,8 +11,8 @@ if O.format_on_save then
   }
 end
 
--- check if formatter has been defined for the language or not
-local function formatter_exists(lang_formatter)
+-- check if formatter configuration is available inside `Formatter` in default-config.lua or not
+local function fmt_conf_present(lang_formatter)
```
Tastyep, line 15: Since basically this is `return lang_formatter != nil`, remove the function and replace the calls with the condition

lua/lv-formatter/init.lua
```diff
@@ -22,27 +22,38 @@ local function formatter_exists(lang_formatter)
...
-local function formatter_return(lang_formatter)
-  return {
-    exe = lang_formatter.exe,
-    args = lang_formatter.args,
-    stdin = not (lang_formatter.stdin ~= nil),
-  }
+-- returns default formatter for given language. reads from `O.lang.*.formatter` in default-config.lua
+local function default_formatter(formatter_list)
+  for _, selected_formatter in pairs(formatter_list) do
+    if fmt_conf_present(selected_formatter) then
```
Tastyep, line 28: (Just a detail from a style nazi) I like to fail early instead of wrapping the whole code in the condition, it makes code easier to read since it reduces imbrications.
```
for loop_condition do
  if condition do
    code logic
  end
end

Becomes

for loop_condition do
  if not condition do
    return or continue
  end
  code logic
end
```

Tastyep (changes requested):
I've not yet tested it, I'll give it a shot later ;)

lua/lv-formatter/init.lua
abzcoding: it also checks for `exe` and `args` to be filled as well

lua/lv-formatter/init.lua
abzcoding: i'm going through a couple of formatters and return the first available one
if none of theme where available, then i want to `return {}`
but updated this part a bit

abzcoding:
@Tastyep thank you for reviewing this, but chris want to do it another way so we won't merge it probably

PR #846, opened 2021-07-10
Now, cssls and autocomplete now works with css scss sass files

kubilaykandemir:
But for me bar at the bottom, i think it is galaxy line, it still says it is not working. But when I check lsp info. it says it is working and auto complete is working.


kubilaykandemir:
I am having treesitter issus now. 
it is saying /usr/share/nvim/runtime/lua/vim/treesitter/highlighter.lua:259: end_col value outside range.

kubilaykandemir:
Last commit I deleted the files, because of code repetition and I smylinked the scss, sass, less files to css.lua.

PR #847, opened 2021-07-10
Add QuickFix and Location list which-key mapping

PasiBergman:
With e.g. Telescope, we can export the search result to quick fix list. With these key mappings it is easy to browse the quick fix and location list items.

### DIFF
--- a/lua/lv-which-key/config.lua
+++ b/lua/lv-which-key/config.lua
@@ -116,7 +116,26 @@ O.plugin.which_key = {
         "Workspace Symbols",
       },
     },
-
+    Q = {
+      name = "+QuickFix",
+      o = { "<cmd>copen<cr>", "Open" },
+      c = { "<cmd>cclose<cr>", "Close" },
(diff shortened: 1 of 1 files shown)


jameswalmsley:
Looks nice... note that:

<C-q> toggles the quickfix menu already.
You can move between items with `[q` and `]q` also.

PasiBergman:
@jameswalmsley Good point. I use *vim with Mac's Skandinavian keyboard layout and `[` and `]` chars are not easy to produce as they require option key (option+8, option+9).

They may not be very useful for everyone, but I made a PR also to kind of promote the use of quick fix list, especially with Telescope's <c-q> -> quick fix list export.

I will leave the PR as-is for now, so that Chris can close it after considering the usefulness.

jameswalmsley (approved):

jameswalmsley:
Ah cool... I always forget that all these mappings are very English centric!

I've approved, I think making quickfix more discoverable to newer users is a bonus..  :)

PR #851, opened 2021-07-10
[DISTRIB] Support custom forks/distributions without clashing core LV.

jameswalmsley:
I want to be able to keep my own "distribution" with a set of base packages that I can extend the base config with..
and still allow the user to update their config in the standard lunarvim way.

I've created an lv-distrib folder, that won't be used by the base lunarvim, but provides a nice place to add stuff without getting git conflicts with base.

See: https://github.com/jameswalmsley/LunarVim/tree/jw for an example of how this can be used.

Note, I've updated how the which-key overrides work...
By calling wk.register() instead of looping over and setting the mappings, this means you can "extend" the mappings and not just override them.

This helps if you want to add an item to the Lsp menu without overriding everything else under 'L',
I may make that a separate PR.


ChristianChiarulli:
should be handled by lvim dir now

PR #855, opened 2021-07-10
Line wrap cursor movement option

roobert:
Added an option around this config option since it changes the default behaviour of vim and I can't seem to override it from `lv-user` or `lv-config.lua`.


lua/default-config.lua
```diff
@@ -7,6 +7,7 @@ USER = vim.fn.expand "$USER"
 O = {
   leader_key = "space",
   colorscheme = "spacegray",
+  line_wrap_cursor_movement = false,
```
abzcoding, line 10: can you set the default config to `true`, since this is the current configuration?

roobert: I have done this, although I think it encourages poor practice in VIM - some people (https://www.youtube.com/watch?v=E-ZbrtoSuzw) even disable `hjkl` to train themselves to use more efficient movement!

abzcoding (approved):

lua/default-config.lua
abzcoding: yeah I agree but a lot of LunarVim users are just starting to migrate to neovim, so we don't want to scare them away :sweat_smile:

PR #862, opened 2021-07-11
[WHICH-KEY] Add search for key-maps and commands.

jameswalmsley:
Add search for Commands and Keymaps.


abzcoding (approved):
nice one :100:

Issue #853, opened 2021-07-10
[Bug]: <Leader>f not working!


jameswalmsley:
### Problem description

Seems like <Leader>f (set in the telescope config) is set before leader as "space" is set..
Not sure why, but this seems to be an issue when dash-board is enabled..

Pressing '\'-f triggers the find files correctly.

### LunarVim version

master

### Neovim version (>= 0.5)

0.5.0

### I have

- [X] Read the readme
- [X] Read the wiki
- [X] Searched for similar issues

jameswalmsley:
My bad, its some issue with my config!

PR #863, opened 2021-07-11
[SETTINGS] Ensure <Leader> is mapped before plugin configs are sourced.

jameswalmsley:
# Description

Fix for leader key not being mapped before some plugins are configured.

Fixes #(issue)
#853

### DIFF
--- a/lua/lv-which-key/init.lua
+++ b/lua/lv-which-key/init.lua
@@ -8,15 +8,6 @@ end
 
 which_key.setup(O.plugin.which_key.setup)
 
--- Set leader
-if O.leader_key == " " or O.leader_key == "space" then
-  vim.api.nvim_set_keymap("n", "<Space>", "<NOP>", { noremap = true, silent = true })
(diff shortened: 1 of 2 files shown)


lua/lv-which-key/init.lua
```diff
@@ -8,15 +8,6 @@ end
 
 which_key.setup(O.plugin.which_key.setup)
 
--- Set leader
-if O.leader_key == " " or O.leader_key == "space" then
-  vim.api.nvim_set_keymap("n", "<Space>", "<NOP>", { noremap = true, silent = true })
-  vim.g.mapleader = " "
-else
-  vim.api.nvim_set_keymap("n", O.leader_key, "<NOP>", { noremap = true, silent = true })
-  vim.g.mapleader = O.leader_key
```
abzcoding, line 17: what if users want to set a custom `leader_key` inside `lv-config.lua` ?

abzcoding (approved):

PR #865, opened 2021-07-11
Add feature for local project specific config files

sethigeet:
Add feature for local project specific config files

# Description

Now you can create a `.lv-config.lua` file in any project directory and those settings will be read when opening neovim from that directory only.
In this way users can have some project specific settings that they would not want to have as global settings

## Type Of Change

Please check the relevant options.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?

You can go to any directory in which you have a specific project and make a `.lv-config.lua` file restart neovim.
Now you can have settings that will only be present in this project and will not affect other projects of yours.
Changes made to this config are also hot reloaded and hence neovim need not be restarted

## Checklist:

- [x] I read the [contributing](https://github.com/ChristianChiarulli/LunarVim/blob/rolling/CONTRIBUTING.md) guide
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings


kylo252:
Interesting idea! I do however see some potential issues with it: 

- What is the scope for this project-config? is it only concerned with plugins or is it supposed to be able to support overriding all the options?
- What if you want to open a nested folder? or jump to a file from Telescope or something similar? 

I have a partial solution for the second issue. if you use something like `git rev-parse --show-toplevel`. 

I do think that this is a really complicated problem with many intricacies that can really frustrate the end-user. Even the mighty MS are having problems solving it, take a look at one of the oldest standing issues https://github.com/microsoft/vscode/issues/40233.

There's a native vim method for this but it has caveats with security `:h exrc`, take a look at https://github.com/neovim/neovim/issues/13501.

sethigeet:
* tldr; I think it should support all the settings that are supported in the default config!
I think that this would be mainly useful for language specific and some settings like formatters, etc as when you are contributing to other projects you do not control which formatter to use or which tool to use for linting. So if you generally use some other tool but you have to use one specific one for a particular project, that particular setting won't interfere with other projects which might be using some other tool. For example, some people like using `black` for formatting while some prefer `yapf`, `autopep8` in python.
Also there are some projects that use different `tab widths` than what you may use.

* For the second one, using something like `telescope.nvim` or `snap` would not be an issue as they do not change the current working directory. So as long as you open `neovim` from the correct directory, everything should be fine. Still we could have a setting that specifies the max number of levels above the current directory to search for the local config file and then even if the user opens `neovim` from a nested directory we could search for the config file!

toimelin:
I would personally prefer to have this as a setting that is turned off by default. There are security concerns with reading files that are potentially not authored by the user.
There is a plugin that does something similar to this https://github.com/windwp/nvim-projectconfig. That plugin has project specific files defined elsewhere. Snippet from their readme:

> current directory is /home/abcde/projects/awesome/. you open vim in awesome directory.
> 
> It will load a config file from ~/.config/nvim/projects/awesome.lua or ~/.config/nvim/projects/awesome.vim

This way you need to actively author the configuration file for a project, and therefore there should be no security concerns. But that also means that you would have to symlink or something similar to have the configuration be read from the actual project folder.

Another workaround would be to query the user the first time the project specific config file is read and have them accept it being read, then persist that file path somewhere so it doesn't ask again (maybe even have a checksum so you have to accept again if it changes)

sethigeet:
I'll add the logic to ask the user first if they trust the local config file before loading the file. I'll also add an option in the global `lv-config` to set an option to always trust local `lv-config` files!

abzcoding:
can we have profiles instead? 

- you can select your profiles by a keybinding or a file at project_root, for example:
  - a profile for golang backend dev
  - another for front end dev
  - kernel development env

- **upon selecting the profile**:
  - all plugins that are not defined in this profile wouldn't load ( still installed but not loading) 
  - all plugins specific to this profile will get loaded
  - keybindings change according to your profile (reload of keybindings which should be easy)
  - custom settings that is set inside `O.default_options` or `O.user_autocmmands` gets loaded

- you can have a file inside a folder that declares which profile to use for this folder (e.g)
```lua
-- .lv-config.lua
project_root = somewhere
profile = sth
```

devtoi:
I think profiles may be scoping outside of the intention of this PR. It is however an interesting thing that imo should be solved with a plugin (although perhaps well integrated with LunarVim in the future). I have started on a plugin for loading project configuration files and may extend it to profiles in the future. It is my first neovim/vim plugin and I have limited time and lua experience, but may be possible to use that in the future.
<details>
  <summary>Plugin</summary>
  https://gitlab.com/devtoi/init_more.nvim
</details>

sethigeet:
I think I agree with you @abzcoding that profiles are better and as @devtoi pointed out, such a functionality is better to go in its own plugin rather that being implemented in the `lunarvim` core.

But anyway we will still require a local config file(`.lv-config.lua`) to set the profile and also it is usually much more easier to just change a few settings in a project specific config file rather than creating a profile for it as even for a single type of dev (eg. frontend dev, golang dev), there are times when you have to change just a specific setting for a particular project when working with other collaborators!

ftplugin/lua.lua
```diff
@@ -1,4 +1,4 @@
 require("lang.lua").format()
 require("lang.lua").lint()
 require("lang.lua").lsp()
-require("lang.lua").dap()
+require("lang.lua").dap()
```
Israel-Laguan, line 4: NIT: add end of file line

init.lua
```diff
@@ -1,19 +1,18 @@
...
+local lv_utils = require "lv-utils"
+-- Load the user config
+lv_utils.load_user_config()
 
 require "keymappings"
 
 local plugins = require "plugins"
 local plugin_loader = require("plugin-loader").init()
 plugin_loader:load { plugins, O.user_plugins }
+
```
Israel-Laguan, line 12: Nit: extra line here?

lua/lv-utils/init.lua
```diff
@@ -1,5 +1,100 @@
...
+          end
+        )
+      )
+    then
+      return
+    end
+
+    local local_status_ok, _ = pcall(vim.cmd, "luafile " .. local_config_path)
+    if not local_status_ok then
+      print "something is wrong with your local lv-config (./.lv-config.lua)"
```
Israel-Laguan, line 82: NIT/question: what about using `local_config_path` instead of `./.lv-config.lua`?

lua/lv-utils/init.lua
```diff
@@ -1,5 +1,100 @@
...
+
+function lv_utils.reload_lv_config()
+  vim.cmd "source ~/.config/nvim/lua/keymappings.lua"
+  lv_utils.load_user_config()  
+  vim.cmd "source ~/.config/nvim/lua/plugins.lua"
+  vim.cmd "source ~/.config/nvim/lua/settings.lua"
+  vim.cmd "source ~/.config/nvim/lua/core/formatter.lua"
+  vim.cmd ":PackerCompile"
+  vim.cmd ":PackerInstall"
+  -- vim.cmd ":PackerClean"
```
Israel-Laguan, line 95: NIT: please remove comment if not used

chase:
As of nvim 0.9, this can be accomplished with the exrc option.


To quote [exrc.nvim](https://github.com/MunifTanjim/exrc.nvim)'s README:

>
> Neovim v0.9.0 onwards supports secure `.exrc`, `.nvimrc` and `.nvim.lua` files. You don't need this plugin anymore.
>
> Just enable the `'exrc'` option:
>
> ```lua
> vim.o.exrc = true
> ```
>
> For more information, check:
> - `:help 'exrc'`
> - `:help exrc`


In the interim we could use:
https://github.com/MunifTanjim/exrc.nvim

PR #861, opened 2021-07-11
User configurable latex lsp#2

PZ31k0nauT:
### Description

This PR makes the latex lsp more flexible and easier to configure. Almost every setting can be made in lv-config.lua. With those the user can choose which build arguments and which preview options are to be used by lunarvim.

My hope is that this is a desired PR in the light of #638.

Disclaimer: This is a copy of #671, which has been closed by me. I've messed up big time I think. The main difference is that this PR is now based on rolling and has a cleaner commit history. I'm still learning sorry for the inconvenience.

#### Type Of Change

Please check the relevant options.

- Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality
- Breaking Change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update (maybe)

#### How has this been tested

- I have tested these changes with opening .tex and .bib files and checking for linting and lsp-autocomplete. It's working as expected.
- I built files with the command `:TexlabBuild` successfully.
- I also previewed .tex files with the command `:TexlabForward`. Only tested Zathura.
- I switched the settings for build and preview on save. They work as desired.

#### Checklist:

- [x] I read the contributing guide
- [x] My code follows the style guidelines of the project (at least I hope so)
- [x] I have performed a self-review of my own code
- I have commented my code, particularly in hard-to-understand areas (code is pretty straightforward)
- [x] I have made corresponding changes to the documentation (I would but I don't know where)
- [x] My changes generate no warnings

PR #873, opened 2021-07-11
Vimtex/Latex : Added autosave on focus loss option and list of errors to be ignored.

atropos112:
# Description
This is a repeat PR of #840 but in the new rolling set out.

I find myself using autosave (save &compile on focus change) on latex a lot as it compiles stuff in the background as I do something else, a lot of people i know have something similar within their IDEs, I believe it is an option that gets used by a lot of people, some hate it though so I thought it might be a good idea to give the option.

For ignore errors, sometimes latex gives errors the user may not care about and this results in them being bothered every time they compile.

There is another latex PR #861 , this doesn't interfere with it so no problem there.

## Type Of Change

Please check the relevant options.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?
I ran it with my configs and with default configs.

## Checklist:

- [x] I read the [contributing](https://github.com/ChristianChiarulli/LunarVim/blob/rolling/CONTRIBUTING.md) guide
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented on my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation - (I would, but I don't know where)
- [x] My changes generate no new warnings

### DIFF
--- a/ftplugin/tex.lua
+++ b/ftplugin/tex.lua
@@ -10,6 +10,7 @@ require("lspconfig").texlab.setup {
 vim.g.vimtex_compiler_method = "latexmk"
 vim.g.vimtex_view_method = "zathura"
 vim.g.vimtex_fold_enabled = 0
+vim.g.vimtex_quickfix_ignore_filters = O.lang.latex.ignore_errors
 
 O.plugin.which_key.mappings["L"] = {
   name = "+Latex",
@@ -32,3 +33,7 @@ vim.api.nvim_exec(
     ]],
   false
 )
+if (O.lang.latex.auto_save)
+then
+  vim.api.nvim_exec([[au FocusLost * :wa]],false)
+end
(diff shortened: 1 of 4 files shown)


abzcoding (approved):
👍

barun108:
I use Latex for teaching math online to my secondary level students, 
auto save and then auto compile and live preview helps me a lot

earlier I was using NeoTex plugin ( donRaphaco /neotex ) which exactly does this

now I moved to LunarVim , 
and 
I want to use extra plugins as less as I could

so, auto save feature will be very helpful indeed

PR #883, opened 2021-07-11
add function to preview definitions

rafamadriz:
Added function to preview definitions without having to jump to the file. It accepts a "context" which is the amount of lines that is going to show. It is set to 15 which I've found to be just right. I've map it to `gp` in normal mode, for preview or peek. 

https://user-images.githubusercontent.com/67771985/125206318-2766d100-e276-11eb-8de6-5f1944c7e888.mp4


rafamadriz:
@jameswalmsley @rebuilt @ChristianChiarulli let me know what you think before merging.

MatthiasX86:
This is a good idea. But, if we should consider having this I should make note that this request is essentially the same one that I just made in my PR #841 . 

Though admittedly, I already have an answer from @abzcoding about a feature like this to only have this as a user added plugin

If we this will be an issue that we will revisit there might be another question here: which would be better, using the built in PeekDefinition from the lsp or by way of a more feature-rich plugin that I suggested initially in my PR

jameswalmsley:
@rafamadriz nice work.
I think that this should probably be a separate plugin which can be enabled.

@MatthiasX86 thanks for your input too.. 
I'll have to try your goto-preview plugin.. it looks really nice.

abzcoding:
> This is a good idea. But, if we should consider having this I should make note that this request is essentially the same one that I just made in my PR #841 .
> 
> Though admittedly, I already have an answer from @abzcoding about a feature like this to only have this as a user added plugin
> 
> If we this will be an issue that we will revisit there might be another question here: which would be better, using the built in PeekDefinition from the lsp or by way of a more feature-rich plugin that I suggested initially in my PR

IMO, anything that requires **installing a new plugin** needs to be really useful to be merged, but anything that adds a feature without introducing new plugins is fine by me.

rafamadriz:
> This is a good idea. But, if we should consider having this I should make note that this request is essentially the same one that I just made in my PR #841 .
> 
> Though admittedly, I already have an answer from @abzcoding about a feature like this to only have this as a user added plugin
> 
> If we this will be an issue that we will revisit there might be another question here: which would be better, using the built in PeekDefinition from the lsp or by way of a more feature-rich plugin that I suggested initially in my PR

Oops sorry, didn't know about that PR!

@jameswalmsley can you elaborate on why you think this should be a separate plugin ? This is just a function using the builtin LSP. No need for plugins (although I know that the option suggested by @MatthiasX86 has more functionality than what I added). But my point being that one of the focuses of this config is adding features without adding 100 plugins and using builtin functionality

abzcoding:
@rafamadriz friendly reminder to use `rolling` branch for adding new features. 😃

rafamadriz:
> @rafamadriz friendly reminder to use `rolling` branch for adding new features. smiley

My bad, thanks.

abzcoding:
@rafamadriz can you change `lua/default-config.lua` to look like this:
```lua
  lsp = {
    document_highlight = true,
    popup_border = "single",
    preview_definition = false,
  },
```
and then inside `lua/lsp/init.lua`
we can have:
```lua
if O.lsp.preview_definition then
  vim.cmd "nnoremap <silent> gp <cmd>lua require'lsp'.PeekDefinition()<CR>"
end
```
so that it is disabled by default, but users can enable and use it.
@jameswalmsley  what do you think about this approach?

abzcoding:
> This is a good idea. But, if we should consider having this I should make note that this request is essentially the same one that I just made in my PR #841 .
> 
> Though admittedly, I already have an answer from @abzcoding about a feature like this to only have this as a user added plugin
> 
> If we this will be an issue that we will revisit there might be another question here: which would be better, using the built in PeekDefinition from the lsp or by way of a more feature-rich plugin that I suggested initially in my PR

ultimately @ChristianChiarulli  should decide between #883 and #841

jameswalmsley:
It's a nice feature.. and I'm not against adding it.
I'm just not sure it's something LunarVim itself should start adding.

It seems LunarVim is powerful because it can focus on creating a really decent config, and setting up packer, lsp formatting etc.

However bigger functionality should probably be provided by plugins IMO.

Under the hood..  there isn't a lot of difference to Vim if its a plugin or just a lua file in your config. They are both in the runtime path and have to be loaded.

If we start adding this kind of thing, it becomes a lunarvim only development that must be maintained.. and people will want to extend it and make it more powerful..

Probably the whole neovim community is better served by e.g. a plugin like goto-preview.

Anyway its a neat snippet.. I just think LunarVim should keep a tight focus.

jameswalmsley:
On the approach.. it looks good.. I guess Chris defaults to most things being disabled by default.

ChristianChiarulli:
idk, I really do like this, kinda nice not to need a plugin for this. I'll test it out later and then we'll see. As long as it stays stable while I try breaking it I think it might be nice to have this builtin.

rafamadriz:
> @rafamadriz can you change `lua/default-config.lua` to look like this:
> 
> ```lua
>   lsp = {
>     document_highlight = true,
>     popup_border = "single",
>     preview_definition = false,
>   },
> ```
> 
> and then inside `lua/lsp/init.lua`
> we can have:
> 
> ```lua
> if O.lsp.preview_definition then
>   vim.cmd "nnoremap <silent> gp <cmd>lua require'lsp'.PeekDefinition()<CR>"
> end
> ```
> 
> we can wrap those functions inside the same if statement as well.
> so that it is disabled by default, but users can enable and use it.
> @jameswalmsley what do you think about this approach?

I don't like that, it just to much overthinking and adding unnecessary options to people, were gonna keep adding features down the road, how is the config file going to look like if we start adding options to enable or disable ***individual*** mappings ? It would be better to add an option to disable all the mappings that this config comes with, so people define their own. But we shouldn't add options to enable or disable a ***single*** mapping.

jameswalmsley:
@rafamadriz I agree with you on what you are saying about the config for different options and mappings..

It's best to bundle that functionality in to a collection of useful and related features that you can enable or disable as a group depending on your needs. (Which is what a plugin does).

Ultimately its just about the approach.. if we are going to start bringing in more and more features direct to the lunar config, that's fine.. however if we don't plan on doing that then maybe we should be consistent.

I do think the neovim ecosystem is particularly successful and powerful precisely because they have encouraged a community of modular and decoupled plugins that the whole community can use and mix as they like, just like Lunar massively benefits from too.

rafamadriz:
> @rafamadriz I agree with you on what you are saying about the config for different options and mappings..
> 
> It's best to bundle that functionality in to a collection of useful and related features that you can enable or disable as a group depending on your needs. (Which is what a plugin does).
> 
> Ultimately its just about the approach.. if we are going to start bringing in more and more features direct to the lunar config, that's fine.. however if we don't plan on doing that then maybe we should be consistent.
> 
> I do think the neovim ecosystem is particularly successful and powerful precisely because they have encouraged a community of modular and decoupled plugins that the whole community can use and mix as they like, just like Lunar massively benefits from too.

I agree with you. But I believe that there's still value on having something like this built in without the need for plugins, and it is not something that needs to be maintain or anything, I've had something like that in my config for months now, I just copy pasted it and forgot about it, whenever I need it, there it is, without needing plugins. 

I totally see your point and agree 100%, as you said, being a separate plugin or just code in your config, it all gets loaded in runtime at the end of the day.

MatthiasX86:
@rafamadriz @ChristianChiarulli

It looks like at least we are all agreed the ability to peek/preview a definition is a worth feature to have. I for one didn't realize PeekDefinition was built into LSP.

I have been a SpaceVime user for quite some time (please don't hit me...). Though SpaceVim was feature-rich, many of their notable, builtin features were pegged to their long release cycles. This was particularly a problem if you had bugs

Personally I like the idea of offloading a feature like this to a plugin to take advantage of updates on their own release cycle 

Some of the other things that would be nice to peeking/preview would be:

- Implementations (included in goto-preview)
- Type definitions
- References

Even though the last two are not in goto-preview it may just be a matter of time before it becomes a request. So how far do you see this feature of LunarVim going? Should it extend to some of those other ones I mentioned? Would it be better implement all of these as builtin features in the future?

MatthiasX86:
@rafamadriz @ChristianChiarulli

It looks like at least we are all agreed the ability to peek/preview a definition is a worth feature to have.

I for one didn't realize PeekDefinition was built into LSP. I have been a SpaceVime user for quite some time (please don't hit me...). Though SpaceVim was feature-rich, many of their notable, builtin features were pegged to their long release cycles. This was particularly a problem if you had bugs

Personally I like the idea of offloading a feature like this to a plugin to take advantage of updates on their own release cycle 

Some of the other things that would be nice to peeking/preview would be:

- Implementations (included in goto-preview)
- Type definitions
- References

Even though the last two are not in goto-preview it may just be a matter of time before it becomes a request. So how far do you see this feature of LunarVim going? Should it extend to some of those other ones I mentioned? Would it be better implement all of these as builtin features in the future?

rafamadriz:
> Some of the other things that would be nice to peeking/preview would be:
> 
>     * Implementations (included in goto-preview)
> 
>     * Type definitions
> 
>     * References
> 
> 
> Even though the last two are not in goto-preview it may just be a matter of time before it becomes a request. So how far do you see this feature of LunarVim going? Should it extend to some of those other ones I mentioned? Would it be better implement all of these as builtin features in the future?

That's actually a pretty good idea. I can add that, the problem is that you can get multiple implementations, type definitions and so on... This would mean that you would only get one implementation, one type, one reference and so on. And also Telescope has builtin helpers for that (e.g. `Telescope lsp_implementations` and others). I'm going to add it though, kind cool to being able to do a quick preview of type, implementations and definitions. And thanks for the kind words.

rafamadriz:
Hey @ChristianChiarulli with the suggestion from @MatthiasX86  I added another functions to preview typeDefinition and implementations. Although I didn't map them to any key, gonna let you decide on that one. To test them you have call it with `:lua require'lsp'.PeekTypeDefinition()` and `:lua PeekImplementation()`

Here's an example of my test:

First I do a preview of the implementation and then of typeDefinition 

https://user-images.githubusercontent.com/67771985/125212376-89d2c800-e29c-11eb-8513-99311c78907b.mp4

ChristianChiarulli:
@rafamadriz very cool, is there a way to get a border on the floating window?

rafamadriz:
> @rafamadriz very cool, is there a way to get a border on the floating window?

@ChristianChiarulli  done, it's using the option from the config file `O.lsp.popup_borders`

PR #893, opened 2021-07-12
[Refactor] implement packer plugin interface

Tastyep:
# Description

I converted all lv-plugins to implement the same [Packer interface](https://github.com/wbthomason/packer.nvim#specifying-plugins)
[example](https://github.com/ChristianChiarulli/LunarVim/pull/893/files#diff-165bca18055194aae5dd77bf3f748849f4e40aa67a0ef1eace305e44a1987a08)

This allows to greatly reduce the size of plugins.lua, make it more readable, as well as having a consistent format for structuring our plugins.

## How Has This Been Tested?

I just printed the config names being loaded and then played with a few to see if it kept working.

### DIFF
--- a/lua/core/autopairs.lua
+++ b/lua/core/autopairs.lua
@@ -1,7 +1,11 @@
-local m = {
-  "hrsh7th/nvim-compe",
-  -- event = "InsertEnter",
-}
+local m = {}
+
+m.packer_config = function()
+  return require("lv-utils").adapt_plugin(m, {
+    "windwp/nvim-autopairs",
+    after = { "telescope.nvim" },
+  })
+end
 
 m.config = function()
   -- if not package.loaded['nvim-autopairs'] then
--- a/lua/core/autopairs.lua
+++ b/lua/core/autopairs.lua
@@ -1,51 +1,60 @@
--- if not package.loaded['nvim-autopairs'] then
---   return
--- end
-local status_ok, autopairs = pcall(require, "nvim-autopairs")
-if not status_ok then
-  return
-end
-local npairs = require "nvim-autopairs"
-local Rule = require "nvim-autopairs.rule"
+local m = {
+  "hrsh7th/nvim-compe",
+  -- event = "InsertEnter",
+}
 
--- skip it, if you use another global object
-_G.MUtils = {}
-
-vim.g.completion_confirm_key = ""
-MUtils.completion_confirm = function()
-  if vim.fn.pumvisible() ~= 0 then
-    if vim.fn.complete_info()["selected"] ~= -1 then
-      return vim.fn["compe#confirm"](npairs.esc "<cr>")
-    else
-      return npairs.esc "<cr>"
-    end
-  else
-    return npairs.autopairs_cr()
+m.config = function()
+  -- if not package.loaded['nvim-autopairs'] then
+  --   return
+  -- end
+  local status_ok, autopairs = pcall(require, "nvim-autopairs")
+  if not status_ok then
+    return
   end
-end
+  local npairs = require "nvim-autopairs"
+  local Rule = require "nvim-autopairs.rule"
 
-if package.loaded["compe"] then
-  require("nvim-autopairs.completion.compe").setup {
-    map_cr = true, --  map <CR> on insert mode
-    map_complete = true, -- it will auto insert `(` after select function or method item
+  -- skip it, if you use another global object
+  _G.MUtils = {}
+
+  vim.g.completion_confirm_key = ""
+  MUtils.completion_confirm = function()
+    if vim.fn.pumvisible() ~= 0 then
+      if vim.fn.complete_info()["selected"] ~= -1 then
+        return vim.fn["compe#confirm"](npairs.esc "<cr>")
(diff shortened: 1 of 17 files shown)


kylo252:
Interesting approach! My main concern is that this will obscure the "credit" to the plugins' publishers. 

I can see a few side-effects for that: 
- now the user needs to go to every single config file to get the name of the plugin. Sure we can make a readme list, but that will require maintenance to keep up-to-date. 
- some will start thinking it's black magic or some custom plugin, and then LunarVim will have to absorb and resolve all the plugins-related issues.
- some advanced users will probably not bother helping out the project if they see/assume it's using custom plugins.

Tastyep:
This provides structures for specifying plugins to Packer, it doesn't define new plugins.

> My main concern is that this will obscure the "credit" to the plugins' publishers.

I don't understand this, the code is not a place for the user to be, so in that regard, it is already obscure.
As an example, you wouldn't expect a user to browse the source code of the backend of an application to be able to know what he can or cannot do. He will use a front-end interface that itself communicates with the API.

Your other points relate to the same issue, the user is lacking an interface to pick the core plugins.
Also related: [PR-796](https://github.com/ChristianChiarulli/LunarVim/pull/796)

kylo252:
I guess it depends on how you define the scope of this project. Is it trying to make configuring neovim _and its plugins_  more accessible. Or is it some kinda new editor built on top of neovim, where the user treats it as a black-box more or less?

Do you think this project should be responsible for the behavior of XYZ plugin? or just the bootstrap process?

rafamadriz:
> This provides structures for specifying plugins to Packer, it doesn't define new plugins.
> 
> > My main concern is that this will obscure the "credit" to the plugins' publishers.
> 
> I don't understand this, the code is not a place for the user to be, so in that regard, it is already obscure.
> As an example, you wouldn't expect a user to browse the source code of the backend of an application to be able to know what he can or cannot do. He will use a front-end interface that itself communicates with the API.
> 
> Your other points relate to the same issue, the user is lacking an interface to pick the core plugins.
> Also related: [PR-796](https://github.com/ChristianChiarulli/LunarVim/pull/796)

@Tastyep  I agree with you.

>  I guess it depends on how you define the scope of this project. Is it trying to make configuring neovim and its plugins more accessible. Or is it some kinda new editor built on top of neovim, where the user treats it as a black-box more or less?

@kylo252  Not sure what you mean. New editor on top on neovim ? what ? And also, this config is already obscure enough for some people. Just look at the language servers. They all have their own file even though you can set up all servers with a boilerplate in just one file.  

> Do you think this project should be responsible for the behavior of XYZ plugin? or just the bootstrap process?

Why splitting up the plugin list would make us responsible of the behavior of X plugin ?

kylo252:
@rafamadriz, I'll try to give an example to get my point across. 

Let's take a look at `{"lv-indent"},` or `{"lv-comment"},`, from the plugins' list. We already know that `Packer` has intended that interface to be used for _local_ plugins. In case of a normal user, it doesn't really matter if they decide to clone/fork some repo or just define the source of the plugin later in the actual file, because the end-user is purposefully aware whether it's actually a local plugin or just a some lazy-loading trick. 

Whereas if a new comer to a project of this size would take a look at that list, they'd _probably_ assume it's one of the following: 
1- LunarVim defines its own custom "indent" plugin.
2- LunarVim is doing some weird hacks to get "indent" plugin to work.

Case (1): Now if they would take the first assumption, that might mean that when "indent", _btw it's actually "indent-lines"_, doesn't work on COBOL or that it doesn't render correctly in their terminal. Where do you think they'd open an issue?

Do you wanna take the chance that they know what the line actually means? that they need to go to "lua/lv-indent" and be able to understand from that indent-lines actually is done by "lukas-reineke/indent-blankline.nvim", and that LunarVim doesn't do anything special to it..

```lua
local m = {
  "lukas-reineke/indent-blankline.nvim",
  event = "BufRead",
  disable = not O.plugin.indent_line.active,
}
```

Case (2): Users that are comfortable with Packer might dismiss the project if they glance at the that list and see all of sorts of custom/local plugins. I know I would probably would. 


Sorry for the long rant. For me this can be summed-up as cost vs benefits/leverage. And I'm still not sure what that benefit is in regards to performance.

Tastyep:
I will answer this the same as I answered your previous comment.

The user should not browse code in order to configure the project.

I have no problem accepting criticism, but you are giving me arguments that I would find relevant only if the project was composed of only two files (plugins.lua and keymappings.lua).
The goal is to build a layer on top of nvim with lua to provide IDE functionalities, this is not achievable with 2 files.

Again, what you complain about is the lack of interface for the user to configure LunarVim.
Before he had to modify the code to apply changes, now there is O.user_plugins to configure plugins, and keymappings are easier to deal with. I agree that this is still not enough.

What I try to do on LunarVim is to implement interfaces, so that the project can scale. Without interfaces, the project will keep being a big bag of data that the user will have to open to modify.

Tastyep:
[PR-1011](https://github.com/ChristianChiarulli/LunarVim/pull/1011) More or less replaces this.

Issue #895, opened 2021-07-12
[Feature]: Customize LSP Keybinds


zethra:
### Feature motivation

Users should be able to change the lsp keybinds defined in `lua/lsp/init.lua`

### Describe the expected solution

There should be a setting to change these keybinds

### Describe the alternatives you've considered

There could be a single setting to disable the defaults, then allow the user to define their own.

abzcoding:
#896

PR #896, opened 2021-07-12
[Feature]: Add setting to enable/disable the default lsp keybinds

zethra:
# Description
This PR addresses issue #895 it the simplest way possible. It may be
preferable to implement a more complex solution to disable individual
keybinds or allow the user to override them in some other way.

I decided to submit this as a PR first because I didn't want to try
making larger changes without feedback.

## How Has This Been Tested?

I toggle the config option and it does enable and disable the default keybinds.

### DIFF
--- a/lua/default-config.lua
+++ b/lua/default-config.lua
@@ -70,6 +70,7 @@ O = {
     },
     document_highlight = true,
     popup_border = "single",
+    default_keybinds = true,
   },
 
   disabled_built_ins = {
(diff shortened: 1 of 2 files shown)


abzcoding:
#988 is merged, so this is possible to do rn

zethra:
@abzcoding am I missing something or are the key bindings set in lua/lsp/init.lua still set unconditionally?

abzcoding:
🤦 my bad, you are 💯 right

zethra:
Thx 👍

zethra:
@abzcoding Thx 👍

PR #898, opened 2021-07-12
[Feature]: Add save without formatting

### LINKED ISSUE #897: [Feature]: Add save without formatting
### Feature motivation

The steps involved in disabling format on save is already pretty slick as `O.format_on_save` applies immediately. This could still be improved for those cases where you want format on save, except for saving a single file – whether it be a forked project or some file you simply don't want formatted.

With some basic testing, this function seems to do the trick:

```lua
function SaveWithoutFormatting()
    -- Temporarily disable format on save
    vim.cmd([[if exists('#autoformat#BufWritePost')
    :autocmd! autoformat
    endif]])

    -- Save file
    vim.cmd(':w')

    -- Re-enable format on save if previously enabled
    if O.format_on_save then
        require('lv-utils').define_augroups({
            autoformat = {
                {
                    'BufWritePost',
                    '*',
                    ':silent FormatWrite',
                },
            },
        })
    end
end
```

Then we could add to which-key mappings:

```lua
l = {
    ...
    F = { "<cmd>lua SaveWithoutFormatting<cr>", "Save Without Formatting" },
    ...
}
```

I've considered just not saving files. Who needs em' anyways

### Additional context

Pending any objections, I'm happy to make a PR soon.

mvllow:
# Description

Add save without formatting via `<leader>lF`

Fixes #897

## How Has This Been Tested?

Setting `O.format_on_save = true` and ensuring formatting works as usual. Then create a scenario where formatting would normally be ran, but instead of saving the file, do `<leader>lF` and you should see the file is saved but not formatted. Saving again would properly format on save, assuming your config still has that set to true.


abzcoding:
side note: you can do this as well
```lua
-- lv-config.lua
O.format_on_save = false

-- Keybindings
vim.api.nvim_set_keymap("n", "<leader>lF", '<CMD>silent Format<CR>', { noremap = true, silent = true })
```
this way you can disable formatting everywhere and format without saving if you press <leader>lF

abzcoding:
@mvllow can you fix the formatting with `stylua` :pleading_face:

abzcoding (approved):

mvllow:
I think on stream Chris mentioned wanting to do a different binding since `lf` is technically there because of lsp formatting. Maybe `bf`/`bF` would be good for "Buffer Format" but that should be a different PR imo, just noting that I'm fine with the binding changing in the future :)

abzcoding:
@mvllow what do you think about #1153 ?

mvllow:
That PR looks nice. I'm okay dropping this one for that. Only concern is overwriting default `:w` etc. in case something goes wrong then maybe you would be unable to save at all.

abzcoding:
> That PR looks nice. I'm okay dropping this one for that if it's preferred. Only concern is overwriting default `:w` etc. in case something goes wrong then maybe you would be unable to save at all. Not sure why changing the built-ins is necessary but that could just be my lack of understanding

actually I do agree with you :sweat_smile:

kylo252:
Seeing as this outdated now, please feel free to open a new one based on #1937

PR #908, opened 2021-07-13
Added default formatters

abzcoding:
# Description

We need to have some default formatters for each lang
Also, there was a bug that wouldn't set the formatter config.

## How Has This Been Tested?

- open a lua file
- run `:Format`
- run `:lua print(vim.inspect(O.formatters.filetype['lua']))`

### DIFF
--- a/ftplugin/c.lua
+++ b/ftplugin/c.lua
@@ -1,17 +1,14 @@
-local clang_formatter = {
+O.formatters.filetype["c"] = {
   function()
     return {
-      exe = "clang-format",
-      --  TODO: append to this for args don't overwrite
-      args = {},
-      stdin = true,
+      exe = O.lang.c.formatter.exe,
+      args = O.lang.c.formatter.args,
+      stdin = not (O.lang.c.formatter.stdin ~= nil),
     }
   end,
 }
+O.formatters.filetype["cpp"] = O.formatters.filetype["c"]
 
-O.formatters.filetype["c"] = clang_formatter
-O.formatters.filetype["cpp"] = clang_formatter
-O.formatters.filetype["obj"] = clang_formatter
 require("formatter.config").set_defaults {
   logging = false,
   filetype = O.formatters.filetype,
--- a/ftplugin/c.lua
+++ b/ftplugin/c.lua
@@ -4,14 +4,18 @@ local clang_formatter = {
       exe = "clang-format",
       --  TODO: append to this for args don't overwrite
       args = {},
+      stdin = true,
     }
   end,
 }
 
 O.formatters.filetype["c"] = clang_formatter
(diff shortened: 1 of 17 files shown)


lua/core/formatter.lua
```diff
@@ -53,7 +53,7 @@ if not status_ok then
   return
 end
 
-formatter.setup {}
+formatter.setup { formatter.config }
```
abzcoding, line 56: @ChristianChiarulli take a look at this PR and let me know if you like it or not
also without this line, the formatters wouldn't have been set

ftplugin/c.lua
```diff
@@ -1,3 +1,18 @@
...
+      exe = "clang-format",
+      --  TODO: append to this for args don't overwrite
+      args = {},
+    }
+  end,
+}
+
+O.formatters.filetype["c"] = clang_formatter
+O.formatters.filetype["cpp"] = clang_formatter
+O.formatters.filetype["cmake"] = clang_formatter
```
Tastyep, line 13: I don't think clang-format handles cmake formatting, however, I use [cmake-format](https://github.com/cheshirekow/cmake_format) and it works great

Tastyep (changes requested):
I have tried for lua, golang, python and cpp but only got it working for lua.
Do I need to configure something or should this work out of the box ?

ftplugin/c.lua
abzcoding: nice catch 👀 i'm adding correct changes rn

abzcoding:
> I have tried for lua, golang, python, shell, and cpp but only got it working for lua.
> Do I need to configure something or should this work out of the box ?
> 
> When I run :Format I get: no formatter defined for xxx file

@Tastyep can you test it again to see if it works 🥺

Tastyep (approved):
Great work !
Looking forward to seeing this merged

abzcoding:
the issue with this is that users cannot override the default formatter, i'm working on it atm.

Tastyep:
Could you also please configure a formatter for javascript languages ?

abzcoding:
@Tastyep can you test if it works for `javascript` files as well,
also it first looks for local prettier, if not found then uses global one ( or the one that you define)

mvllow (approved):

ChristianChiarulli:
before we merge I'd like to test a few things on stream

mvllow:
Possibly a slick suggestion for allowing the user to override prettier args:

```lua
local prettier_args = { "--stdin-filepath", vim.api.nvim_buf_get_name(0) }

for k, v in pairs(O.lang.tsserver.formatter.args) do
	prettier_args[k] = v
end
```

Then in `lv-config.lua` if you have `O.lang.tsserver.formatters.args = { "--single-quote" }` it will be added :)

PR #913, opened 2021-07-13
[Doc] fix typo in install.sh not --> now

ichigo-gyuunyuu:


ChristianChiarulli:
I think that's actually correct

PR #914, opened 2021-07-13
[Feature] Added Deno Language Server

eduardjbotha:
# Description

Add deno to as a language server for javascript/typescript (and their react equivalents) files. 
Add a which-key entry when deno has been activated.

# How to activate

Since deno is an alternative to tsserver, their use should be mutually exclusive.
Add the following to the lv-config.lua
`O.lang.tsserver.active = false`
`O.lang.deno.active = true`

# Notes/Questions for the reviewers

1. The which-key entry will be visible whenever `O.lang.deno.active = true`. I would've liked to have it visible only when the deno-ls is running. Is this possible?
2. The commands that come with the language server (`DenolsDefinition` and `DenolsReferences`) should be used when going to the definition/references of a method/variable. They do work, but the default behaviour when pressing `gd` or `gr` is broken on  deno's files. According to the documentation these handlers can be overridden (see below), but my lack of lua knowledge prevents me from implementing `<function 1>`. How can I implement this function to execute the abovementioned neovim commands?

> handlers = {
      ["textDocument/definition"] = <function 1>,
      ["textDocument/references"] = <function 1>
    }

### DIFF
--- a/ftplugin/javascript.lua
+++ b/ftplugin/javascript.lua
@@ -1,7 +1,7 @@
 if O.lang.tsserver.active then
-  require "lsp.tsserver-ls"
+	require("lsp.tsserver-ls")
 elseif O.lang.deno.active then
-  require "lsp.deno-ls"
+	require("lsp.deno-ls")
 end
 
-vim.cmd "setl ts=2 sw=2"
+vim.cmd("setl ts=2 sw=2")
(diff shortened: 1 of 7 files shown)


eduardjbotha:
Closing due to the updates on 2021/08/04 in #193

PR #915, opened 2021-07-13
[WHICH-KEY] Use wk.register() to handle the overrides.

jameswalmsley:
The current O.user_which_key allows user to add top-level mappings, but its difficult to e.g. change a single item in a sub-menu..
Thats because the for() loop which iterates over each k,v pair replaces the root of each key entirely.

Simply giving the user-defined mappings to wk.register() overrides things which much finer control.


abzcoding (approved):
great :100:

PR #919, opened 2021-07-14
LunarVim 0.4.8

ChristianChiarulli:
- New (auto)formatting strategy introduced (local prettier works) 🥳 
- removed non "core" plugins 
- better java support
- full python debugging support
- fix nvimtree bug that borks UI
- Generic UI improvements
- update install script (overwrite option introduced BE CAREFUL)
- improve colorscheme support
- transparency


chuckmasucci:
I just got used to using Tab for `:BufferNext`. Any reason why I shouldn't add it back in on my local install?

barun108:
How do I enable prettier ?

PR #921, opened 2021-07-14
Allow override tsserver formatter args

hulufei:


ChristianChiarulli:
wheres the default?

hulufei:
@ChristianChiarulli `default-config.lua`

ChristianChiarulli:
oh, didn't realize we already had it there.

PR #929, opened 2021-07-14
Autopairs event trigger

rebuilt:
# Description

The 'after' keyword has no effect.  Since the dependencies are autoloaded, the 'after' setting is not being applied.  Packer can only manage plugins in the 'opt' folder with the 'after' keyword.

The 'after' keyword was needed when we were lazy loading compe and tree-sitter but now it does nothing.  

As described here https://github.com/wbthomason/packer.nvim/issues/435

The listing in the packer_compiled file shows an empty table for 'after'

```lua
  ["nvim-autopairs"] = {
    config = { "\27LJ\2\n.\0\0\3\0\2\0\0046\0\0\0'\2\1\0B\0\2\1K\0\1\0\19core.autopairs\frequire\0" },
    load_after = {},
    loaded = true,
    needs_bufread = false,
    path = "/home/nelson/.local/share/nvim/site/pack/packer/opt/nvim-autopairs"
  },
```

## How Has This Been Tested?

Opened lua file to make sure autopairs works and that indenting works as well. 

- ...

### DIFF
--- a/lua/plugins.lua
+++ b/lua/plugins.lua
@@ -60,7 +60,6 @@ return require("packer").startup(function(use)
   use {
     "windwp/nvim-autopairs",
     -- event = "InsertEnter",
-    after = { "nvim-treesitter" },
     config = function()
       require "core.autopairs"
     end,
--- a/lua/plugins.lua
+++ b/lua/plugins.lua
@@ -60,7 +60,7 @@ return require("packer").startup(function(use)
   use {
     "windwp/nvim-autopairs",
     -- event = "InsertEnter",
-    after = { "telescope.nvim" },
+    after = { "nvim-treesitter" },
     config = function()
       require "core.autopairs"
     end,


abzcoding:
nice catch :eyes: 
can we do sth like this, or is it useless?
```lua
  use {
    "hrsh7th/nvim-compe",
    event = "VimEnter",
    config = function()
      vim.cmd [[packadd telescope.nvim]]
      require("core.compe").setup()
    end,
  }
```

rebuilt:
The packadd line isn't needed.  'Telescope' is not the right plugin to load here. Compe has no telescope dependencies.  

```lua
   vim.cmd [[packadd telescope.nvim]]
```

The reason I started looking into this was autopairs needs to get loaded after treesitter(not telescope) and compe. 

I think the reason we made compe autostart is that there were indentation problems in c and cpp files.  There was a weird interaction between compe and autopairs.  Making them both required fixes things.  Setting compe to lazy load on 'VimEnter' and setting autopairs to load after compe also seems to work.  But I'm not sure if it will work for every language.

abzcoding:
> The packadd line isn't needed. 'Telescope' is not the right plugin to load here. Compe has no telescope dependencies.
> 
> ```lua
>    vim.cmd [[packadd telescope.nvim]]
> ```
> 
> The reason I started looking into this was autopairs needs to get loaded after treesitter(not telescope) and compe.
> 
> I think the reason we made compe autostart is that there were indentation problems in c and cpp files. There was a weird interaction between compe and autopairs. Making them both required fixes things. Setting compe to lazy load on 'VimEnter' and setting autopairs to load after compe also seems to work. But I'm not sure if it will work for every language.

ah I see :facepalm: my bad

PR #931, opened 2021-07-14
[Bugfix] Make if statement POSIX compliant

xdavidel:
# Description

Using `#!/bin/sh` can mean an interpreter that does not support bashism like `[[ ]]`.
instead using something like grep will allow more POSIX compliant code.

## How Has This Been Tested?

Using a shell interpreter like `dash` to run the installation script.

### DIFF
--- a/utils/installer/install.sh
+++ b/utils/installer/install.sh
@@ -182,7 +182,7 @@ installextrapackages() {
 # Welcome
 echo 'Installing LunarVim'
 
-if [[ $* == *--overwrite* ]]; then
+if [ -n "$(echo "$@" | grep "\-\-overwrite")" ]; then
   echo '!!Warning!! -> Removing all nvim related config because of the --overwrite flag'
   rm -rf "$HOME/.config/nvim"
   rm -rf "$HOME/.cache/nvim"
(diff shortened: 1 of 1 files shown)


utils/installer/install.sh
```diff
@@ -182,7 +182,7 @@ installextrapackages() {
 # Welcome
 echo 'Installing LunarVim'
 
-if [[ $* == *--overwrite* ]]; then
+if [ -n "$(echo "$@" | grep "\-\-overwrite")" ]; then
```
Tastyep, line 185: Nice solution, but there is an even better way, [parameters expansion](https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_06_02).
Check [this](https://stackoverflow.com/questions/2829613/how-do-you-tell-if-a-string-contains-another-string-in-posix-sh) post for example

Tastyep (changes requested):

utils/installer/install.sh
xdavidel: I like it but for some reason it didn't work using '!=' sign in dash on $@.
I'm using the case solution though - it is more extendable. thanks.

Tastyep (approved):

Tastyep (approved):

abzcoding:
Closed by #1026

PR #932, opened 2021-07-14
[Bugfix] Add root_dir configuration to pyright

PasiBergman:
# Description

Adds `root_dir` configuration to `pyright` language server config.

Fixes #(issue)

## How Has This Been Tested?

Opened a Python project, which had a requirements.txt file in the root directory -> pyright recognized project root correctly.

### DIFF
--- a/lua/lsp/python.lua
+++ b/lua/lsp/python.lua
@@ -74,6 +74,7 @@ M.lsp = function()
         update_in_insert = true,
       }),
     },
+    root_dir = require("lspconfig").util.root_pattern(".git/", "requirements.txt"),
     settings = {
       python = {
         analysis = {


lua/lsp/python.lua
```diff
@@ -74,6 +74,7 @@ M.lsp = function()
         update_in_insert = true,
       }),
     },
+    root_dir = require("lspconfig").util.root_pattern(".git/", "requirements.txt"),
```
abzcoding, line 77: the default value of this is (as described [here](https://github.com/neovim/nvim-lspconfig/blob/master/CONFIG.md#pyright)):
```lua
root_dir = function(filename)
    return util.root_pattern(unpack(root_files))(filename) or util.path.dirname(filename)
end,
```
that doesn't work for you?

PasiBergman: Good point @abzcoding ! I did not consider the default value.
The default value seems to be enough on pyright. The project root (i.e. root_dir) was resolved correctly without explicitly specifying the root_dir. Before I made the PR there was some issue with one of my Python projects project root detection, but I was not able to reproduce it. It must not been related to the root_dir after all. 

I will close the PR now as the default value for root_dir seems to be fine.

PasiBergman:
As @abzcoding  menetioned, the defaults are fine and there is no need to specify root_dir on pyright lsp.

PR #935, opened 2021-07-14
Fix rust root pattern

midnightexigent:
rust projects don't have a `Carrust.toml` file at the root. it's `Cargo.toml`

Edit:

Maybe this should go straight to `master` branch as rust projects are simply broken due to this ?

### DIFF
--- a/ftplugin/rust.lua
+++ b/ftplugin/rust.lua
@@ -95,7 +95,7 @@ else
     cmd = { DATA_PATH .. "/lspinstall/rust/rust-analyzer" },
     on_attach = require("lsp").common_on_attach,
     filetypes = { "rust" },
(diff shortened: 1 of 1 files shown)


ChristianChiarulli:
I would talk to the folks at lspconfig about this

mjlbach:
I don't know why this was overridden originally, but we have a much more reliable root detection mechanism in lspconfig (sans the typo)

https://github.com/neovim/nvim-lspconfig/blob/59bc16aa2c9cfc6f8c2a67ae01635eeea1d43943/lua/lspconfig/rust_analyzer.lua#L19-L34

The offending change was introduced in https://github.com/ChristianChiarulli/LunarVim/pull/908/files

PR #938, opened 2021-07-14
[Feature] hot reload configurations

Tastyep:
# Description

This work is based on [796](https://github.com/ChristianChiarulli/LunarVim/pull/796) and will stay as a draft until it gets merged.
This pr provides an object for observing file updates, I tested observing multiple files at the same time and it works.
At present, I only observe lv-config, when the config is updated I reload plugins and configurations.

#### Updates
[0] There are still some problems with this, mainly caused by code that is interpreted when the module is loaded, a better approach would be to wrap everything into functions. 

[1] I've reviewed the current problems, and I think [893](https://github.com/ChristianChiarulli/LunarVim/pull/893) would fix them.

## How Has This Been Tested?

- Open lv-config
- Add a plugin to user_plugins
- Observe that it's loaded
- Remove it
- Type y (asked by PackerClean, not sure is this can be automatized)
- Observe that the plugin has been removed

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -1,19 +1,38 @@
 require "default-config"
-local status_ok, error = pcall(vim.cmd, "luafile " .. CONFIG_PATH .. "/lv-config.lua")
-if not status_ok then
-  print "something is wrong with your lv-config"
-  print(error)
-end
-
 require "keymappings"
 
-local plugins = require "plugins"
 local plugin_loader = require("plugin-loader").init()
-plugin_loader:load { plugins, O.user_plugins }
-vim.g.colors_name = O.colorscheme -- Colorscheme must get called after plugins are loaded or it will break new installs.
+
+local function load_config()
+  require "default-config"
+  local status_ok, error = pcall(vim.cmd, "luafile " .. CONFIG_PATH .. "/lv-config.lua")
+  if not status_ok then
+    print "something is wrong with your lv-config"
+    print(error)
+  end
+end
+
+local function load_plugins()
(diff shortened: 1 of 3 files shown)


kylo252:
Not sure how related it is to this, but I noticed that a broken config will sometimes brick Packer until you actually delete `plugin/packer_compiled.lua` first and then you can actually use Packer again.

I think it might be worth implementing a mechanism to detect such cases and remove it first, if it's feasible. Or maybe just always perform a "clean" compile. Although I'm not sure how safe it is 😄

Tastyep:
> Not sure how related it is to this, but I noticed that a broken config will sometimes brick Packer until you actually delete `plugin/packer_compiled.lua` first and then you can actually use Packer again.
> 
> I think it might be worth implementing a mechanism to detect such cases and remove it first, if it's feasible. Or maybe just always perform a "clean" compile. Although I'm not sure how safe it is smile

We could simply make a function called on one whichkey mapping, the entry for packer already exists.

rebuilt:
Great work on 796.  This one looks like it will also move the project in a good direction.

abzcoding:
@Tastyep  can we have this plz :pleading_face:

Tastyep:
Fine, I'll work on it today ^^

chase:
@Tastyep is this still necessary with how things are currently setup? Any features not implemented yet?

Tastyep:
> @Tastyep is this still necessary with how things are currently setup? Any features not implemented yet?

No, I'll close it

PR #941, opened 2021-07-14
[Feature] Add option to disable fancy icons

mvllow:
# Description

Add option to disable all (current) ui icons. This should allow a decent experience without the devicons package and without nerd fonts. This is something I personally use and in the distant future could be a good solution to fallback on if no nerd fonts are detected on install.

This may be an unpopular setting, however it is marked experimental and completely opt-in.

This feature:
- Conditionally disables `nvim-web-devicons` (again, this is opt-in)
- Adds fallback icons for `nvim-tree`
- Adds fallback icons for `bufferline` (barbar)

### DIFF
--- a/lua/core/bufferline.lua
+++ b/lua/core/bufferline.lua
@@ -26,3 +26,19 @@ O.plugin.which_key.mappings["b"] = {
     "sort BufferLines automatically by language",
   },
 }
+
+if O.experimental.disable_fancy_icons then
+  -- TODO: How to do this initialise `g` lua?
+  vim.cmd [[let bufferline = get(g:, 'bufferline', {})]]
+
+  -- TODO: Setting these in lua doesn't seem to work
+  -- vim.g.bufferline = {
+  --   icon_close_tab = '×',
+  --   icon_close_tab_modified = '●',
+  --   icons = false
+  -- }
+  vim.cmd [[let bufferline.icon_close_tab = '×']]
+  vim.cmd [[let bufferline.icon_close_tab_modified = '●']]
+  vim.cmd [[let bufferline.icons = v:false]]
+end
+
--- a/lua/core/nvimtree.lua
+++ b/lua/core/nvimtree.lua
@@ -56,6 +56,42 @@ M.setup = function()
       symlink = "",
     },
   }
+
+  if O.experimental.disable_fancy_icons then
+    g.nvim_tree_show_icons = {
+      git = 1,
+      folders = 1,
+      files = 0,
+      folder_arrows = 0,
+    }
+    vim.g.nvim_tree_icons = {
+      default = '',
+      symlink = '',
+      git = {
+        unstaged = '×',
+        staged = '✓',
+        unmerged = '◇',
+        renamed = '➜',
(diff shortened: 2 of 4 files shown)


rebuilt:
It's not just nvim-tree that requires devicons.  Devicons are also used by galaxyline, bufferline, barbar, trouble, telescope.  And number of other plugins a user might install.  What happens if you just disable devicons?  Do those other plugins stop working?  If they have graceful fallbacks, we can just add an option to disable devicons.

mvllow:
Disabling devicons won't disable the usage of nerd font icons, though. For example, nvim-tree uses nerd font icons for folders. If someone doesn't have a nerd font those icons don't show up. This option provides fallback unicode options. Even without supporting every plugin, this experience will still be a step up from having broken symbols everywhere.

I am fully aware this may be out of the scope for LunarVim but hopefully the experimental label and being opt-in may let slide. Just seems like a nice-to-have for people who don't want to bother with nerd fonts or have devicons installed. Feel free to close, however, if this adds an unnecessary burden on maintenance.

rebuilt:
I pulled this to test.  Yes, this fixes the failing fonts for nvimtree.  The user still sees failing fonts in galaxyline, telescope, gitsigns, lsp errors, and dashboard.  They're probably also going to experience the same if they install some other plugin that relies on fancy fonts.  

I think the best approach would provide a mechanism for injectable configurations for the core plugins.  That seems like it would give you a solution to do what you want to do and also add flexibility to the core plugins.

rebuilt:
I just tested it with the following code:

```
  use {
    "glepnir/galaxyline.nvim",
    config = function()
      if(not O.plugin.galaxyline.config) then
      require "core.galaxyline"
      else 
      O.plugin.galaxyline.config()
      end
    end,
    event = "BufWinEnter",
    disable = not O.plugin.galaxyline.active,
  }

```

This code was just to make sure injectable configs were possible.  The condition should probably reside in core.galaxyline .

rebuilt:
Thank you, but this feature is no longer needed.  You can set the icons you like with `lvim.builtin.nvimtree.icons`

Below is the table of options

```
  lvim.builtin.nvimtree = {
    side = "left",
    width = 30,
    show_icons = {
      git = 1,
      folders = 1,
      files = 1,
      folder_arrows = 1,
      tree_width = 30,
    },
    ignore = { ".git", "node_modules", ".cache" },
    auto_open = 1,
    auto_close = 1,
    quit_on_open = 0,
    follow = 1,
    hide_dotfiles = 1,
    git_hl = 1,
    root_folder_modifier = ":t",
    tab_open = 0,
    allow_resize = 1,
    lsp_diagnostics = 1,
    auto_ignore_ft = { "startify", "dashboard" },
    icons = {
      default = "",
      symlink = "",
      git = {
        unstaged = "",
        staged = "S",
        unmerged = "",
        renamed = "➜",
        deleted = "",
        untracked = "U",
        ignored = "◌",
      },
      folder = {
        default = "",
        open = "",
        empty = "",
        empty_open = "",
        symlink = "",
      },
    },


```

PR #943, opened 2021-07-14
[Feature] Allow using custom paths for language servers

kylo252:
# Description

**_Breaking change: "O.lang.clang" has been changed to "O.lang.cpp"_**

Added the ability to use manually installed language servers. This can be done using:
`O.lang.rust.custom_languageserver = "/my/custom/path/rust-analyzer"`

## How Has This Been Tested?

Tested with `.lua` and `.c` files.

### DIFF
--- a/ftplugin/c.lua
+++ b/ftplugin/c.lua
@@ -1,4 +1,4 @@
-require("lang.clang").format()
-require("lang.clang").lint()
-require("lang.clang").lsp()
-require("lang.clang").dap()
+require("lang.cpp").format()
+require("lang.cpp").lint()
+require("lang.cpp").lsp()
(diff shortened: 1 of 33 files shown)


jameswalmsley (approved):

abzcoding:
@kylo252 check [this](https://github.com/abzcoding/LunarVim/blob/feature/clang-formatter/lua/lsp/clang.lua#L3) file. regarding #947 , we can probably add path to the `M.config` function, and gradually move all the language-specific configurations out of `default-config` and in their own file. and after that rename `lua/lsp` to `lua/languages`

abzcoding:
sorry mate, forget about this, closed by #1043

PR #948, opened 2021-07-14
Allow users to set disabled vim builtins

### LINKED ISSUE #946: [Feature]: How to quickly open file sibling which is out of current nvimtree root
### Feature motivation

cd to `~/.config/nvim` and use nevim to open `lv-config`

Use `:e ~/.bashrc` to open it

Try to open `~/.aliases` (which happens tto be sibling of currently open file)

## Using nvim_tree
Using `<space>e` open nvim_tree, go up two directories and find there the `.aliases file`

This requires at least 5 keystrokes to get to the user home dir.

## Using `:Explore`
(assuming `netrw` is not intentionally disabled) call `:Ex` what opens netrw listing of user home directory (using the directory where is current file open).

This takes 4 keystrokes to get to the user home dir listing.

## Using `tpope/vim-vinegar`
Press `-` and have the netrw listing of user home directory open.

This takes 1 keystroke to get to the user home dir listing.

Do not disable `netrw` (and whatever other builtin plugin needed to run Explore command)

Allow to install `tpope/vim-vinegar` plugin (either as standard LunarVim part or as user defined).

I was researching nvim_tree to see, if there is some simple option to open the tree where the current file resides when the file is out of current nvim_tree root. I did not find an option.

### Additional context

I understand, disabling some builtin plugins speeds up start up time.

But for day to day work, speed at opening sibling file is more important as it happens more often.

abzcoding:
# Description

Some users might want to enable and use `netrw`
Fixes #946


Tastyep (approved):
LGTM

PR #955, opened 2021-07-15
[Feature] Hide formatter.nvim tempfiles

mvllow:
# Description

Prefix formatter.nvim's tempfile with ".formatter" so they should be hidden by default. This was added to any lang that mentions `stdin = false`.


abzcoding:
@mvllow thank you for your contribution, can you fix the formatting using stylua:

```bash
# go to the root of project
cp utils/.stylua.toml .
rm -rf plugin  # Failed to read ./plugin/packer_compiled.lua: stream did not contain valid UTF-8
stylua .
```

PR #959, opened 2021-07-15
Update install.sh to include pip installation

Nikhil0504:
# Description

Self-explanatory.

Fixes #(issue)

## How Has This Been Tested?

I tested in google collab using the script so it should work in the Unix systems.

### DIFF
--- a/utils/installer/install.sh
+++ b/utils/installer/install.sh
@@ -125,12 +125,12 @@ asktoinstallnode() {
 }
 
 asktoinstallpip() {
-    # echo "pip not found"
-    # echo -n "Would you like to install pip now (y/n)? "
-    # read answer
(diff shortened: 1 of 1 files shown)


utils/installer/install.sh
```diff
@@ -60,15 +60,15 @@ installpiponmac() {
 }
 
 installpiponubuntu() {
-    sudo apt install python3-pip >/dev/null
+    sudo apt install python3-pip
```
abzcoding, line 63: it might need a `-y` to install it non-interactively

Shahryarmohammadie (approved):

kylo252 (commented):
You don't actually need `sudo` to install `pip`. There are multiple ways to handle this more gracefully.

We're also basically dropping the installation of system-wide apps in #1052 due to security concerns.

PR #963, opened 2021-07-15
[DEFAULT-CONFIG] Add sidescrolloff

UltimateOmega:
# Description
Add sidescrolloff to default-config


rebuilt:
Sidescroll:  The minimal number of screen columns to keep to the left and to the
	right of the cursor if 'nowrap' is set.



sidescroll

rebuilt (approved):
I think this is a more 'sane' default to pair with 'nowrap'

PR #964, opened 2021-07-15
[WHICH-KEY] Add new tab

UltimateOmega:
# Description
Add new tab to which key


ChristianChiarulli:
<leader>tab might be good for this

PR #969, opened 2021-07-15
Ruby debugging support

abzcoding:
# Description

Added Ruby debugging support

to make it work you need to install the following:
```bash
gem install readapt ruby-debug-ide
```
and obviously you need LSP and dap
```vim
:LspInstall ruby
:DIInstall ruby_vsc_dbg
```


ChristianChiarulli:
can you add this as a comment on top of the dap method? `gem install readapt ruby-debug-ide`

PR #971, opened 2021-07-15
Overrideable configs for autopairs

rebuilt:
# Description
Allows full configuration of autopairs by the end-user.  A user can inject their own configuration by adding a declaration in lv-config.lua

```
O.plugin.autopairs.config = function()
      YOUR_CUSTOM_CONFIG_HERE
end
```

Fixes #(issue)
https://github.com/ChristianChiarulli/LunarVim/issues/960

## How Has This Been Tested?
1. Launching with the default config to make sure autopairs still works
2. Launching with the added config to lv-config to make sure a user can add their configuration

### DIFF
--- a/lua/core/autopairs.lua
+++ b/lua/core/autopairs.lua
@@ -5,63 +5,63 @@ local M = {}
 
 M.config = function()
   O.plugin.autopairs = {
-    config = false,
+    config = nil,
   }
 end
 
 M.setup = function()
   if O.plugin.autopairs.config then
-    O.plugin.autopairs.config()
-  else
-    local status_ok, autopairs = pcall(require, "nvim-autopairs")
-    if not status_ok then
-      return false
-    end
-    local npairs = require "nvim-autopairs"
-    local Rule = require "nvim-autopairs.rule"
+    return O.plugin.autopairs.config()
+  end
 
-    -- skip it, if you use another global object
-    _G.MUtils = {}
+  local status_ok, autopairs = pcall(require, "nvim-autopairs")
+  if not status_ok then
+    return false
+  end
+  local npairs = require "nvim-autopairs"
+  local Rule = require "nvim-autopairs.rule"
 
-    vim.g.completion_confirm_key = ""
-    MUtils.completion_confirm = function()
-      if vim.fn.pumvisible() ~= 0 then
-        if vim.fn.complete_info()["selected"] ~= -1 then
-          return vim.fn["compe#confirm"](npairs.esc "<cr>")
-        else
-          return npairs.esc "<cr>"
-        end
+  -- skip it, if you use another global object
+  _G.MUtils = {}
+
+  vim.g.completion_confirm_key = ""
+  MUtils.completion_confirm = function()
+    if vim.fn.pumvisible() ~= 0 then
+      if vim.fn.complete_info()["selected"] ~= -1 then
+        return vim.fn["compe#confirm"](npairs.esc "<cr>")
       else
-        return npairs.autopairs_cr()
+        return npairs.esc "<cr>"
       end
+    else
+      return npairs.autopairs_cr()
     end
+  end
 
-    if package.loaded["compe"] then
-      require("nvim-autopairs.completion.compe").setup {
-        map_cr = true, --  map <CR> on insert mode
-        map_complete = true, -- it will auto insert `(` after select function or method item
-      }
-    end
-
-    npairs.setup {
-      check_ts = true,
-      ts_config = {
-        lua = { "string" }, -- it will not add pair on that treesitter node
-        javascript = { "template_string" },
-        java = false, -- don't check treesitter on java
-      },
-    }
-
-    require("nvim-treesitter.configs").setup { autopairs = { enable = true } }
-
-    local ts_conds = require "nvim-autopairs.ts-conds"
-
-    -- press % => %% is only inside comment or string
-    npairs.add_rules {
-      Rule("%", "%", "lua"):with_pair(ts_conds.is_ts_node { "string", "comment" }),
-      Rule("$", "$", "lua"):with_pair(ts_conds.is_not_ts_node { "function" }),
+  if package.loaded["compe"] then
+    require("nvim-autopairs.completion.compe").setup {
+      map_cr = true, --  map <CR> on insert mode
+      map_complete = true, -- it will auto insert `(` after select function or method item
     }
   end
+
+  npairs.setup {
+    check_ts = true,
+    ts_config = {
+      lua = { "string" }, -- it will not add pair on that treesitter node
+      javascript = { "template_string" },
+      java = false, -- don't check treesitter on java
+    },
+  }
+
+  require("nvim-treesitter.configs").setup { autopairs = { enable = true } }
+
+  local ts_conds = require "nvim-autopairs.ts-conds"
+
+  -- press % => %% is only inside comment or string
+  npairs.add_rules {
+    Rule("%", "%", "lua"):with_pair(ts_conds.is_ts_node { "string", "comment" }),
+    Rule("$", "$", "lua"):with_pair(ts_conds.is_not_ts_node { "function" }),
+  }
 end
 
 return M
--- a/lua/core/autopairs.lua
+++ b/lua/core/autopairs.lua
@@ -10,7 +10,6 @@ M.config = function()
 end
 
 M.setup = function()
-  -- if not O.plugin.autopairs.config then
   if O.plugin.autopairs.config then
     return O.plugin.autopairs.config
   end
--- a/lua/core/autopairs.lua
+++ b/lua/core/autopairs.lua
@@ -10,54 +10,58 @@ M.config = function()
 end
 
 M.setup = function()
-  local status_ok, autopairs = pcall(require, "nvim-autopairs")
-  if not status_ok then
-    return false
-  end
-  local npairs = require "nvim-autopairs"
-  local Rule = require "nvim-autopairs.rule"
-
-  -- skip it, if you use another global object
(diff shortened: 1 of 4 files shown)


lua/plugins.lua
```diff
@@ -59,10 +59,7 @@ return require("packer").startup(function(use)
   -- Autopairs
   use {
     "windwp/nvim-autopairs",
-    -- event = "InsertEnter",
-    config = function()
-      require "core.autopairs"
-    end,
+    config = require("core.autopairs").setup(),
```
kylo252, line 62: AFAIK, this needs to be a either a string or keep the `function() .. end` wrapper.

lua/core/autopairs.lua
```diff
@@ -1,51 +1,67 @@
...
+  if O.plugin.autopairs.config then
+    return O.plugin.autopairs.config
+  end
+
+  local status_ok, autopairs = pcall(require, "nvim-autopairs")
+  if not status_ok then
+    return
+  end
+  local npairs = require "nvim-autopairs"
+  local Rule = require "nvim-autopairs.rule"
```
kylo252, line 22: This is from the old code, but aren't `npairs` and `autopairs` basically the same variable?

lua/core/autopairs.lua
```diff
@@ -1,51 +1,67 @@
...
-vim.g.completion_confirm_key = ""
-MUtils.completion_confirm = function()
-  if vim.fn.pumvisible() ~= 0 then
-    if vim.fn.complete_info()["selected"] ~= -1 then
-      return vim.fn["compe#confirm"](npairs.esc "<cr>")
+  vim.g.completion_confirm_key = ""
+  MUtils.completion_confirm = function()
+    if vim.fn.pumvisible() ~= 0 then
+      if vim.fn.complete_info()["selected"] ~= -1 then
+        return vim.fn["compe#confirm"](npairs.esc "<cr>")
```
kylo252, line 31: Old code again, shouldn't we check if `compe` is loaded first?

lua/plugins.lua
```diff
@@ -59,10 +59,7 @@ return require("packer").startup(function(use)
   -- Autopairs
   use {
     "windwp/nvim-autopairs",
```
kylo252, line 61: I vote for adding a `disable` check and to ~~lazyload it on "InsertEnter *" similar to `compe`~~ or maybe that's not the case anymore (?).

lua/core/autopairs.lua
rebuilt: The check is happening on line 42.  Line 42 will get called before line 33 because completion_confirm only gets called once the user actually enters insert mode and starts editing.

lua/plugins.lua
Tastyep: Nah, it should be `require("core.autopairs").setup` here it's called directly.
Also, config = setup triggers me quick a bit (not related to your work)

lua/plugins.lua
rebuilt: We used to lazyload both compe and autopairs until users reported problems so we switched to requiring them.  Honestly I think it's possible to lazy load both but that's out of scope for this change.  If that change causes problems, I want to be able to revert a single commit.

lua/plugins.lua
rebuilt: I tested.  This style works, but I'll wrap it in a function just to be safe.  You're right, I'm not sure if this will cause problems reloading a config.  I'll also have to change what I return in autopairs.lua.

lua/core/autopairs.lua
```diff
@@ -1,51 +1,67 @@
...
--- skip it, if you use another global object
-_G.MUtils = {}
+M.setup = function()
+  if O.plugin.autopairs.config then
+    return O.plugin.autopairs.config
+  end
+
+  local status_ok, autopairs = pcall(require, "nvim-autopairs")
+  if not status_ok then
+    return
```
Tastyep, line 19: Shouldn't this return `false`  ?

lua/plugins.lua
Tastyep: I don't understand how it can work the way you expect it to, you assign to config the return value of the function, yes it is called, but not from packer once the plugin is loaded, since config is probably = nil.

lua/core/autopairs.lua
rebuilt: It looks like a mistake but Chris added it last week.  It's possible this is related to hot reloading of the config.  I'll have to ask about it.  But if it needs to change, that will be a separate pr.

Tastyep (changes requested):

lua/plugins.lua
rebuilt: I'm just going to add an override mechanism.  The disable check should have a different pr so the decision can be made separately.

lua/plugins.lua
rebuilt: I tried  ``` require("core.autopairs").setup ``` without calling it directly but it didn't work for the user config.

lua/core/autopairs.lua
rebuilt: I didn't change those lines.  They just got shifted down because I added code above it.  The return just makes sure the code below it stops executing.  There's nothing listening to the return value of the setup() function so you could return anything and it wouldn't change any behavior.

rebuilt:
I broke something when I wrapped the config definition in a function call.  I've got to fix it.

lua/plugins.lua
kylo252: You should try when it's lazyloaded. I thought about it that `Packer` needs to copy that line verbatim in that case, and it's not actually calling inline. I'll do some testing with a few plugins that I have and see if there's any truth to it 😄

kylo252: Okay so according to the docs: 
>     config = string or function, -- Specifies code to run after this plugin is loaded.
So my guess is that it didn't quite work for us when it was a normal `require('some-config')` since that does not necessarily return a function callback. 

Use whichever you think fits, I don't have a preference either way as long as it works!

lua/core/autopairs.lua
Tastyep: Sure, I just pointed out that the return value of the function made no sens

lua/plugins.lua
rebuilt: So these ways work, but require you to recompile any time you change from the user config to the default config.

```
config = require("core.autopairs").setup
config = function() require("core.autopairs").setup end
```

This way works, and does not require the user to recompile

```
config = require("core.autopairs").setup(),
```

lua/core/autopairs.lua
rebuilt: Tastyep, I don't know of any good place to put this, but I can tell by your pr's and this comment that you know your stuff.

lua/plugins.lua
Tastyep: I agree with what you said above, but the problem is that by using the second option, the function is called when plugins.lua is interpreted.
What you want is to only call it when the plugin is loaded (packer takes care of this by calling the config callback).

So for instance, If I don't have it installed yet (first time user for instance), First the call to setup will fail because of `pcall(require, "nvim-autopairs")`, then after running PackerInstall, because config = nil `  if not status_ok then return end` it will stay unconfigured until the editor is restarted.

rebuilt: Aha!  That's why those guard clauses are there; for the case where it hasn't been installed yet.  :)  Thanks for clarifying that.  Ok.  I'll pass a single callback to packer.

lua/core/autopairs.lua
rebuilt: I changed it to return false

lua/plugins.lua
rebuilt: It's more verbose now but it works.  And it fulfills the packer contract.   I tried to have the logic happen in the autopairs file but if I put it there, PackerCompile would fail to link properly to the correct callback function.  

```
  -- Autopairs
  use {
    "windwp/nvim-autopairs",
    config = function()
      if O.plugin.autopairs.config then
        O.plugin.autopairs.config()
      else
        require("core.autopairs").setup()
      end
    end,
  }

```

rebuilt: Alright, I think I got it now.  Thanks for bearing with me

lua/core/autopairs.lua
```diff
@@ -1,51 +1,67 @@
...
-end
+local M = {}
 
-if package.loaded["compe"] then
-  require("nvim-autopairs.completion.compe").setup {
-    map_cr = true, --  map <CR> on insert mode
-    map_complete = true, -- it will auto insert `(` after select function or method item
+M.config = function()
+  O.plugin.autopairs = {
+    config = false,
```
Tastyep, line 8: I'm being picky now, but since you expect config to be assigned a function, I think it's preferable to set it to nil instead of false. The conditions below would still work :)

lua/core/autopairs.lua
```diff
@@ -1,51 +1,67 @@
...
-  check_ts = true,
-  ts_config = {
-    lua = { "string" }, -- it will not add pair on that treesitter node
-    javascript = { "template_string" },
-    java = false, -- don't check treesitter on java
-  },
-}
+M.setup = function()
+  if O.plugin.autopairs.config then
+    O.plugin.autopairs.config()
```
Tastyep, line 14: Again me being picky (C++ code linters really make you behave :laughing: ), you could return here and remove the else + extra indentation

Tastyep (changes requested):
Looks good to me !
Please address the first comment, the second one is more opinionated so I don't mind :)

Tastyep (approved):

rebuilt:
Now that the general structure for this has been flushed out, I'll apply the pattern to the other plugins and resubmit

PR #973, opened 2021-07-15
[WHICH-KEY] Add Terminal

UltimateOmega:
# Description
This PR will add Teminal shortcut to Which-key


kylo252:
I was able to achieve the same effect by using: 

```lua
O.plugin.which_key.mappings["t"] = { "<cmd>ToggleTerm<CR>", "Terminal" }
```

There's also the problem that this binding only works for normal mode. So this is not enough to actually toggle it _off_ by pressing the keybinding again. You need another bind for "terminal" mode. And we don't actually have a table yet for it in which-key.

ChristianChiarulli:
no way to toggle it off, I think for toggling the terminal the builtin ctrl-t works best.

kylo252:
This works for me btw to toggle it from `term` mode

```lua
vim.api.nvim_set_keymap("t",  "<leader>t", "<cmd>ToggleTerm<CR>" , { noremap = true, silent = true })
```

PR #976, opened 2021-07-15
[Bugfix] Use error instead of print + prevent shadowing

Tastyep:
# Description

What's in the title + the error now sticks better instead of disappearing like a print would.

## How Has This Been Tested?

lv-config: O.toto.tata = true, opening / closing

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -1,9 +1,8 @@
 require "default-config"
 require "keymappings"
-local status_ok, error = pcall(vim.cmd, "luafile " .. CONFIG_PATH .. "/lv-config.lua")
-if not status_ok then
-  print "something is wrong with your lv-config"
-  print(error)
+local ok, err = pcall(vim.cmd, "luafile " .. CONFIG_PATH .. "/lv-config.lua")
+if not ok then
+  error("Configuration error: " .. err)
 end
 require "plugins"
 vim.g.colors_name = O.colorscheme -- Colorscheme must get called after plugins are loaded or it will break new installs.
--- a/init.lua
+++ b/init.lua
@@ -2,7 +2,7 @@ require "default-config"
 require "keymappings"
 local ok, err = pcall(vim.cmd, "luafile " .. CONFIG_PATH .. "/lv-config.lua")
 if not ok then
-  error("Configuration error: " .. err)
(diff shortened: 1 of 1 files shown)


kylo252 (commented):
Nice idea!

I tried it just now by adding an erroneous value in lv-config:
```lua
O.lang.cp.linter = "clangd"
``` 
But the error message is looking pretty confusing:
```log
Error detected while processing /home/user1/.config/nvim/init.lua:
E5113: Error while calling lua chunk: /home/user1/.config/nvim/init.lua:5: attempt to call local 'error' (a string value)
Press ENTER or type command to continue
```

Maybe we need to parse it slightly differently.

Tastyep:
I guess you reproduced the work instead of checking out the branch, check the error in your log.

kylo252:
> I guess you reproduced the work instead of checking out the branch, check the error in your log.

Yes that was it! I missed the other variable rename `error` -> `err`.
Much better now:
```log
Error detected while processing /home/user1/.cache/LunarVim/init.lua:
E5113: Error while calling lua chunk: /home/user1/.config/nvim/init.lua:5: Configuration error: Vim(luafile):E5113: Error while calling lua chunk: /home/user1/.config/nvim/lv-config.lua:59: attempt to index field 'cp' (a nil value)
```

There's a side-effect though that could be considered a breaking change.  When I used to have such errors before, it was still possible so use neovim relatively fine, but now it looks like it's stopping the loading of a lot plugins completely until you figure out where the issue, assuming you actually know how to resolve it. 

So seeing how `lv-config-example.lua` is not really bullet-proof, maybe should lower the requirement from an error to a warning instead? But generally I think you should define what is considered a "critical" error.

I messed around a bit with `vim.notify` but I couldn't get anything useful from it.

ChristianChiarulli:
let's lower to warning and merge this

Tastyep:
The error() function exits, so the rest of the code wasn't loaded, I agree that it's a too much.
I used notify with the ERROR flag as it asks to press ENTER and allows the user to actually see the error message.
The WARN flag behaves like a print, see [source code](https://github.com/neovim/neovim/blob/master/src/nvim/lua/vim.lua#L416)

ChristianChiarulli:
I feel like the print is still more communicative when testing I will have to do some testing on stream I think.

PR #977, opened 2021-07-15
[Feature] Add an installer for Stylua

kylo252:
# Description

Add an installer for Stylua mainly to be used for the CI as to avoid downloading an entire rust toolchain.

## How Has This Been Tested?

Only tested locally

```bash
bash ./utils/install/install_stylua.sh
```

### DIFF
--- a/.github/workflows/format.yaml
+++ b/.github/workflows/format.yaml
@@ -16,13 +16,10 @@ jobs:
 
       - name: Prepare dependencies
         run: |
-          sudo apt update
-          sudo apt install -y curl
-          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
-          cargo install stylua
-      
+          sudo apt install -y curl unzip --no-install-recommends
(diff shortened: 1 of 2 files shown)


abzcoding:
```
/home/runner/work/_temp/61e5d9b2-372a-48c3-9ef3-f8de49184e35.sh: line 1: stylua: command not found
```

.github/workflows/format.yaml
```diff
@@ -16,13 +16,10 @@ jobs:
...
-          cargo install stylua
-      
+          sudo apt install -y curl unzip --no-install-recommends
+          bash ./utils/installer/install_stylua.sh
+
       - name: Check formatting
         run: |
-          cp ./utils/.stylua.toml .
-          stylua -c .
+          stylua --config-path ./utils/.stylua.toml -c .
```
abzcoding, line 24: stylua is not in the path
```
./stylua
```

kylo252: It was actually `./utils/stylua`, and looks like it's working!

PR #982, opened 2021-07-16
[Feature] add linter support

### LINKED ISSUE #797: Potentially replace efm with nvim-lint
https://github.com/mfussenegger/nvim-lint

Tastyep:
# Description

Add [nvim-lint](https://github.com/mfussenegger/nvim-lint) dependency and configurations for the [supported](https://github.com/mfussenegger/nvim-lint#available-linters) linters.
While still in early development, the work looks promising and extensible (I might try to add clang-tidy).

Fixes #797 

## How Has This Been Tested?

Open a file and save.

I've tested all languages except HTML and Ruby, so if someone can take care of it and confirm it works, that would be great :+1: 
Also I removed some of the efm configurations since they seemed redundant with the work done by the formatter and linter plugins, if someone can double-check that it would be nice too.

### DIFF
--- a/lua/core/linter.lua
+++ b/lua/core/linter.lua
@@ -1,0 +1,15 @@
+local M = {}
+
+M.setup = function()
+  require("lv-utils").define_augroups {
+    autolint = {
+      {
+        "BufWritePost",
+        "<buffer>",
+        ":lua require('lint').try_lint()",
+      },
+    },
+  }
+end
+
+return M
--- a/lua/core/linter.lua
+++ b/lua/core/linter.lua
@@ -1,15 +1,28 @@
 local M = {}
 
 M.setup = function()
-  require("lv-utils").define_augroups {
-    autolint = {
-      {
-        "BufWritePost",
-        "<buffer>",
-        ":lua require('lint').try_lint()",
+  if O.lint_on_save then
+    require("lv-utils").define_augroups {
+      autolint = {
+        {
+          "BufWritePost",
+          "<buffer>",
+          ":lua require('lint').try_lint()",
+        },
       },
-    },
-  }
+    }
+  end
+end
+
+local status_ok, linter = pcall(require, "lint")
+if not status_ok then
+  return
+end
+
+if not O.lint_on_save then
+  vim.cmd [[if exists('#autolint#BufWritePost')
+	:autocmd! autolint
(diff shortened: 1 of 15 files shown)


ChristianChiarulli:
is it possible to lint without saving?

Tastyep:
> is it possible to lint without saving?

Probably, it would be a good idea to also run it when opening the file.
These are the events on which the command can be run
https://neovim.io/doc/user/autocmd.html  search "5. Events"

However, I'm afraid that for some languages (C++ for instance), continuous linting would be too demanding.

abzcoding:
if there are multiple linter defined for a language, does` nvim-lint` use all of them? or the first one it finds on the system or?

Tastyep:
> if there are multiple linter defined for a language, does` nvim-lint` use all of them? or the first one it finds on the system or?

All of them

abzcoding:
> > if there are multiple linter defined for a language, does` nvim-lint` use all of them? or the first one it finds on the system or?
> 
> All of them

would be nice if we can disable the linting or make it on demand like how the formatting works with 
`O.format_on_save` and `<leader>lf`

abzcoding (approved):
LGTM :rocket:

ChristianChiarulli:
Will you be adding anymore or should we merge?

Tastyep:
This should be ready, I can always make a new PR to add support for other linters.
My last concern is that for python it reports A LOT since there are many configured linters.

abzcoding:
> This should be ready, I can always make a new PR to add support for other linters.
> My last concern is that for python it reports A LOT since there are many configured linters.

for python can we keep only one of them and add a sample to `lv-config.example.lua` and how to change it or add more?

ChristianChiarulli:
if the python linters aren't installed then they shouldn't be active tho right? Like if you only have flake8 then that's all that will show up.

Tastyep:
> if the python linters aren't installed then they shouldn't be active tho right? Like if you only have flake8 then that's all that will show up.

That's true, we should probably aim for maximum support and let virtual environments do this job.
I'll still add an example in the lv-config.example files

ChristianChiarulli:
I think this should be ready, I'm going to leave a not that we can get async linting by running the linter on the `TextChanged` or 
`InsertLeave` events.

PR #984, opened 2021-07-16
[Feature] Allow users to inspect all settings inside LV

### LINKED ISSUE #944: [Feature]: Autogenerate lv-config.example.lua
### Feature motivation

As used plugins are added/removed/refactored, it is hard to keep existing `lv-config.example.lua` up to date.

Having a tool to generate up to date example of such file would help.

Create a neovim command or CI/CD task, which takes existing content of lunarvim directory, picks all to `O.something.option.parameter` stuff togather with well maintained documentation strings being around and create sample `lv-config.example.lua` file.

There are many options to generate the file (using treeparser one config file by one, extract the stuff, or using some external lua parser and templates).

Great solution would be e.g. neovim command `:LunarVimConfigExample` which would generate the sample config and read it into new buffer.

An alternative is manually updating the sample config.

### Additional context

_No response_

abzcoding:
# Description

This PR will let users inspect all LV settings.

Fixes #944 

## How Has This Been Tested?

run the following and the inspect `~/.config/nvim/lv-settings.lua`:
```bash
cd ~/.config/nvim
nvim --headless +'lua require("lv-utils").generate_settings()' +qa && sort -o lv-settings.lua{,}
```

### DIFF
--- a/lua/lv-utils/init.lua
+++ b/lua/lv-utils/init.lua
@@ -1,5 +1,61 @@
 local lv_utils = {}
 
+-- recursive Print (structure, limit, indent)
+local function rPrint(s, l, i)
+  l = l or 100 -- default item limit
+  i = i or "." -- indent string
+  if l < 1 then
+    print "ERROR: Item limit reached."
+    return l - 1
+  end
+  local ts = type(s)
+  if ts ~= "table" then
+    if ts == nil then
+      io.write("-- O", i:sub(2), " = nil\n")
+      return l - 1
+    elseif ts == "string" then
+      -- escape sequences
+      s = string.format("%q", s)
+    end
+    i = i:gsub("%.%[", "%[")
+    if type(s) == "function" then
+      -- don't print functions
+      io.write("-- O", i:sub(2), " = function ()\n")
+    else
+      io.write("O", i:sub(2), " = ", tostring(s), "\n")
+    end
+    return l - 1
+  end
+
+  if ts == nil then
+    ts = ""
+  end
+  for k, v in pairs(s) do
+    -- replace non alpha keys wih ["key"]
+    if tostring(k):match "[^%a_]" then
+      k = '["' .. tostring(k) .. '"]'
+    end
+    l = rPrint(v, l, i .. "." .. tostring(k))
+    if l < 0 then
+      break
+    end
+  end
+  return l
+end
+
+function lv_utils.generate_settings()
+  -- Opens a file in append mode
+  local file = io.open("lv-settings.lua", "w")
+
+  -- sets the default output file as test.lua
+  io.output(file)
+
+  rPrint(O, 10000, ".")
+
+  -- closes the open file
+  io.close(file)
+end
+
 function lv_utils.reload_lv_config()
   vim.cmd "source ~/.config/nvim/lv-config.lua"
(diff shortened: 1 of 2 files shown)


abzcoding:
this is the result of running the command:
just rename it from `.lua.txt` to `.lua` ( github won't allow lua :( )
[lv-settings.lua.txt](https://github.com/ChristianChiarulli/LunarVim/files/6829045/lv-settings.lua.txt)

lua/lv-utils/init.lua
```diff
@@ -1,5 +1,61 @@
 local lv_utils = {}
 
+-- recursive Print (structure, limit, indent)
+local function rPrint(s, l, i)
```
Tastyep, line 4: Even though you added a note above, I still think one char variable name is bad because it makes the rest of the code harder to read

lua/lv-utils/init.lua
```diff
@@ -1,5 +1,61 @@
...
+  l = l or 100 -- default item limit
+  i = i or "." -- indent string
+  if l < 1 then
+    print "ERROR: Item limit reached."
+    return l - 1
+  end
+  local ts = type(s)
+  if ts ~= "table" then
+    if ts == nil then
```
Tastyep, line 13: `ts == nil` if `s == nil`, in that case you can move this condition (and use s instead of ts) above  `local ts = type(s)`.

lua/lv-utils/init.lua
```diff
@@ -1,5 +1,61 @@
...
+    if type(s) == "function" then
+      -- don't print functions
+      io.write("-- O", i:sub(2), " = function ()\n")
+    else
+      io.write("O", i:sub(2), " = ", tostring(s), "\n")
+    end
+    return l - 1
+  end
+
+  if ts == nil then
```
Tastyep, line 30: This condition is always false

lua/lv-utils/init.lua
```diff
@@ -1,5 +1,61 @@
...
+    ts = ""
+  end
+  for k, v in pairs(s) do
```
Tastyep, line 33: You could probably put this in a `if ts == 'table'`condition on top and return from there (the motivation is that it contains the recursion) and as a consequence you would be able to unwrap the other type checks from the `if ts ~= "table"` condition.

lua/lv-utils/init.lua
```diff
@@ -1,5 +1,61 @@
 local lv_utils = {}
 
+-- recursive Print (structure, limit, indent)
+local function rPrint(s, l, i)
+  l = l or 100 -- default item limit
+  i = i or "." -- indent string
+  if l < 1 then
```
Tastyep, line 7: Can you explain what the limit is for ?

Tastyep (changes requested):

lua/lv-utils/init.lua
abzcoding: agreed, gonna fix it asap :running_man:

lua/lv-utils/init.lua
abzcoding: just to make sure if my code has issues it won't loop forever, but might remove it after refactoring

Tastyep (approved):

kylo252 (approved):

PR #986, opened 2021-07-16
add scala and sbt support

abzcoding:
# Description

Add scala, sbt LSP and formatting support thanks to @kpritam

## How Has This Been Tested?

add the following to your `lv-config.lua`
```lua
O.lang.scala.metals.active = true
```


abzcoding:
@Pocco81, @kpritam can you guys test if this work properly :pleading_face:

dimas-cyriaco:
Tried this and got this error on the `O.lang.scala.metals.active = true` line:

```
Error detected while processing BufWritePost Autocommands for "lv-config.lua":
E5108: Error executing lua /Users/dimascyriaco/.config/nvim/lua/lv-utils/init.lua:4: Vim(source):E5113: Error while calling lua chunk: /Users/dimascyriaco/.config/nvim/lv-config.lua:47: attempt to index field 'scala' (a nil value)
```

Had to add a field in the `lua/default-config.lua`:

```
lang = {
	scala = {
		metals = {},
	},
}
```

After that I got this error:

```
Error executing vim.schedule lua callback: /Users/dimascyriaco/.config/nvim/lua/lang/scala.lua:54: attempt to call field 'common_on_attach' (a nil value)
```

lua/lang/scala.lua
```diff
@@ -0,0 +1,82 @@
...
+end
+
+M.lint = function()
+  -- TODO: implement linters (if applicable)
+  return "No linters configured!"
+end
+
+M.lsp = function()
+  local function scala_on_attach(client, bufnr)
+    require("lsp").common_on_attach(client, bufnr)
```
ChristianChiarulli, line 54: I think this line is throwing an error

ChristianChiarulli (changes requested):

lua/lang/scala.lua
kpritam: I dont see metals plugin in plugins.lua

```lua
use {
    'scalameta/nvim-metals',
    disable = not O.lang.scala.metals.active,
}

```

abzcoding: 🤦 forgot to remove that line

abzcoding: > I dont see metals plugin in plugins.lua
> 
> ```lua
> use {
>     'scalameta/nvim-metals',
>     disable = not O.lang.scala.metals.active,
> }
> ```

🏃‍♂️ adding it rn

abzcoding:
> Tried this and got this error on the `O.lang.scala.metals.active = true` line:
> 
> ```
> Error detected while processing BufWritePost Autocommands for "lv-config.lua":
> E5108: Error executing lua /Users/dimascyriaco/.config/nvim/lua/lv-utils/init.lua:4: Vim(source):E5113: Error while calling lua chunk: /Users/dimascyriaco/.config/nvim/lv-config.lua:47: attempt to index field 'scala' (a nil value)
> ```
> 
> Had to add a field in the `lua/default-config.lua`:
> 
> ```lua
> lang = {
>   scala = {
>     metals = {},
>   },
> }
> ```
> 
> After that I got this error:
> 
> ```
> Error executing vim.schedule lua callback: /Users/dimascyriaco/.config/nvim/lua/lang/scala.lua:54: attempt to call field 'common_on_attach' (a nil value)
> ```
> 
> on `require("lsp").common_on_attach(client, bufnr)`.

sorry for the inconvenience 😅 
can you pull the latest changes and test it again 🥺

lua/lang/scala.lua
```diff
@@ -0,0 +1,80 @@
...
+M.config = function()
+  O.lang.scala = {
+    metals = {
+      active = false,
+      server_version = "0.10.2+33-c6441eb4-SNAPSHOT",
+      excluded_packages = {
+        "akka.actor.typed.javadsl",
+        "com.github.swagger.akka.javadsl",
+        "akka.stream.javadsl",
+      },
```
kpritam, line 12: I think, default list can be empty and user can exclude packages as per their needs.

lua/lang/scala.lua
```diff
@@ -0,0 +1,80 @@
+local M = {}
+
+M.config = function()
+  O.lang.scala = {
+    metals = {
+      active = false,
+      server_version = "0.10.2+33-c6441eb4-SNAPSHOT",
```
kpritam, line 7: Latest stable version of metals is `0.10.5`

lua/lang/scala.lua
```diff
@@ -0,0 +1,80 @@
...
+      excluded_packages = {
+        "akka.actor.typed.javadsl",
+        "com.github.swagger.akka.javadsl",
+        "akka.stream.javadsl",
+      },
+      show_implicit_arguments = true,
```
kpritam, line 13: False is a good default for implicit and inferred types.

kpritam:
Getting nil index error, scala config needs to be sourced in `lua/default-config.lua` file.

`require("lang.scala").config()`

dimas-cyriaco:
Now I'm getting this error:

```
lua/plugins.lua:262: attempt to index field 'scala' (a nil value)
```

abzcoding:
> Getting nil index error, scala config needs to be sourced in `lua/default-config.lua` file.
> 
> `require("lang.scala").config()`

yep add it , i think i've been drunk while coding this , my bad guys 😢

dimas-cyriaco:
It's working now. 👍

abzcoding:
@kpritam  @dimas-cyriaco 
i'm sorry for the 💩 code 
thank you for helping me with this, you guys are 💯

dimas-cyriaco:
The `nvim-metals` docs [recommends](https://github.com/scalameta/nvim-metals#prerequisites) to set `shortness` without `F`. Wouldn't it be better to set this by default?

abzcoding:
> The `nvim-metals` docs [recommends](https://github.com/scalameta/nvim-metals#prerequisites) to set `shortness` without `F`. Wouldn't it be better to set this by default?

i've updated the code, can you confirm that it works?

kpritam:
Metals document also recommends not to use `nvim-metals` with `nvim-lspconfig`, it works independent of lspconfig.

https://github.com/scalameta/nvim-metals#installation

Here is the recommended configuration from the author of nvim-metals.
https://github.com/scalameta/nvim-metals/discussions/39

cc @ckipp01

dimas-cyriaco:
I don't know if this is an issue with my config or with `nvim-metals` itself, but there are some options that does not work. Like pressing `gd` to go to definition. If I press `g` the `vim.lsp.buf.definition()` option does not appear in the whichkey menu. But if I run`:lua vim.lsp.buf.definition()` it works fine.

dimas-cyriaco:
@abzcoding the `shortmess` worked fine. 👍

abzcoding:
> Metals document also recommends not to use `nvim-metals` with `nvim-lspconfig`, it works independent of lspconfig.
> 
> https://github.com/scalameta/nvim-metals#installation
> 
> Here is the recommended configuration from the author of nvim-metals.
> [scalameta/nvim-metals#39](https://github.com/scalameta/nvim-metals/discussions/39)
> 
> cc @ckipp01

i've updated the code to not use lspconfig for `scala` can you confirm if it works or not?

kpritam:
@abzcoding I tried basic functionalities like `gd`,`gr`, `K`, `worksheets` etc. and can confirm that they are working as expected. 👍

abzcoding:
> I don't know if this is an issue with my config or with `nvim-metals` itself, but there are some options that does not work. Like pressing `gd` to go to definition. If I press `g` the `vim.lsp.buf.definition()` option does not appear in the whichkey menu. But if I run`:lua vim.lsp.buf.definition()` it works fine.

can you test it again, it should be fixed now

dimas-cyriaco:
@abzcoding it's not working for me. Uninstalled LunarVim completely and reinstalled. Added `O.lang.scala.metals.active = true` to `lv-config.lua` and the shortcuts are not working.

abzcoding:
> @abzcoding it's not working for me. Uninstalled LunarVim completely and reinstalled. Added `O.lang.scala.metals.active = true` to `lv-config.lua` and the shortcuts are not working.

ok let me test it somewhere locally just to make sure

ckipp01:
> The nvim-metals docs recommends to set shortness without F. Wouldn't it be better to set this by default?

This is only listed in the docs based on how `nvim-metals` starts, with an filetype autocmd. I have no idea how `LunarVim` works, but if it does rely on a filetype autocmd, then removing `F` would be necessary in order for the message to be shown which prompts the user for various things.

> i've updated the code to not use lspconfig for scala can you confirm if it works or not?

Again, not really familiar with `LunarVim` at all, but the reason I have this in here is since `nvim-metals` works different then `nvim-lspconfg` in that manages the entire lifecycle of the session by itself: the initialize, attaching, etc. If you enable Metals with `nvim-lspconfig` and use `nvim-metals` at the same time you'll end up with two different metals instances attatches screwing all sorts of things up.

abzcoding:
@dimas-cyriaco I've identified the root cause,
basically, we define those keybindings inside `lua/lsp/init.lua` and since we didn't call the `common_on_change` we need to at least initialize it to set those keybindings.

if should be fixed now as I've tested locally.

@ChristianChiarulli i think it is ready to be merged

lua/lang/scala.lua
```diff
@@ -0,0 +1,80 @@
...
+
+M.config = function()
+  O.lang.scala = {
+    metals = {
+      active = false,
+      server_version = "0.10.5",
+      excluded_packages = {},
+      show_implicit_arguments = false,
+      show_inferred_type = true,
+      status_bar_provider = false,
```
ckipp01, line 11: I'd probably recommend removing all of these here. Most of them are the defaults like `exluded_packages`,  `show_implicit_argumens`, and `status_bar_provider`. It's a bit odd that the only one you're really setting is `show_inferred_type`. I'd recommend just removing all of these and then letting the user set them.

lua/lang/scala.lua
```diff
@@ -0,0 +1,80 @@
...
+    },
+    formatter = {
+      exe = "scalafmt",
```
ckipp01, line 14: So a bit unsure on what this is actually doing, but if it's doing what I think it is and actually executing `scalafmt` directly I'd recommend not doing this. Metals has formatting build it using Scalafmt, so all the user needs to do is send in a `vim.lsp.buf.formating()` request while Metals is attached and it will format your file.

lua/lang/scala.lua
```diff
@@ -0,0 +1,80 @@
...
+    metals_config.handlers["textDocument/publishDiagnostics"] =
+      vim.lsp.with(vim.lsp.diagnostic.on_publish_diagnostics, {
+        virtual_text = {
+          prefix = "",
+        },
+      })
+    metals_config.settings = {
+      showImplicitArguments = O.lang.scala.metals.show_implicit_arguments,
+      showInferredType = O.lang.scala.metals.show_inferred_type,
+      excludedPackages = O.lang.scala.metals.excluded_packages,
```
ckipp01, line 67: Again I probably wouldn't se these at all and instead let the user set them. Since there are a _lot_ more options than this.

abzcoding: yes these `O` thingies are user configurable

abzcoding:
> > The nvim-metals docs recommends to set shortness without F. Wouldn't it be better to set this by default?
> 
> This is only listed in the docs based on how `nvim-metals` starts, with an filetype autocmd. I have no idea how `LunarVim` works, but if it does rely on a filetype autocmd, then removing `F` would be necessary in order for the message to be shown which prompts the user for various things.
> 
> > i've updated the code to not use lspconfig for scala can you confirm if it works or not?
> 
> Again, not really familiar with `LunarVim` at all, but the reason I have this in here is since `nvim-metals` works different then `nvim-lspconfg` in that manages the entire lifecycle of the session by itself: the initialize, attaching, etc. If you enable Metals with `nvim-lspconfig` and use `nvim-metals` at the same time you'll end up with two different metals instances attatches screwing all sorts of things up.

thank you for the awesome `nvim-metals` project
as per @kpritam recommendations, I've enabled `nvim-metals` without `lspconfig` and have set the shortmess according to your recommendations, 
and when I tested it locally, it did show the `:MetalsInstall` and after that everything did work.
from jumping around for formatting and more

lua/lang/scala.lua
abzcoding: this is user-configurable as well
user can enable auto formatting by settings this in their config
```lua
O.format_on_save = true
```
also user can manually format the code using:
```
<leader>lf
```

ckipp01: But is this executing `scalafmt` directly? It will be slower this way then just relying on the built-in way to do this via lsp.

lua/lang/scala.lua
ckipp01: So is this the only way to set settings? If so, there is a lot missing.

lua/lang/scala.lua
abzcoding: we need to have these somewhere so when the user is trying to change them in their config
they see sth like this and also the default value

metals

lua/lang/scala.lua
abzcoding: well we can always improve it later on :sweat_smile:

lua/lang/scala.lua
kpritam: Yes, I agree with @ckipp01 here. `metals` does formatting automatically using scalafmt. We do not need to call it explicitly here.

lua/lang/scala.lua
abzcoding: @ckipp01 this is a better picture 😅 
Screen Shot 2021-07-16 at 11 33 20 PM

lua/lang/scala.lua
abzcoding: let me remove it then :running_man:

abzcoding: @ChristianChiarulli what do you think about this?
I can override the `<leader>lf` or `autoformat augroup` for scala so  instead of running `scalafmt` it just runs
`vim.lsp.buf.formatting()` but sure if it is a good idea or not:

lua/lang/scala.lua
abzcoding: @ckipp01 also they can do sth like
```lua
O.lang.scala.metals = {
-- override all things,  even the one we didn't provide
}
```

dimas-cyriaco:
@abzcoding its working 🤘

lua/lang/scala.lua
abzcoding: as described in #996 , we want to have the ability to use builtin formatter for other langs as well.
so we need a bit of refactoring. but for the moment we gonna let this stay as is

@kpritam @ckipp01 thank you for bringing up this issue

w-r-z-k:
The workstation feature doesn't seem to work.  When you edit a file such as *.worksheet.sc it is not active.  Is there an option that needs to be set?

PR #987, opened 2021-07-16
Explicit stdin for formatters

abzcoding:
# Description

I tried to be smart :facepalm:  and used some weird logic in the formatter `stdin`
there is absolutely no reason for that :sweat_smile:

### DIFF
--- a/lua/lang/clang.lua
+++ b/lua/lang/clang.lua
@@ -13,6 +13,7 @@ M.config = function()
     formatter = {
       exe = "clang-format",
       args = {},
+      stdin = true,
     },
     debug = {
       adapter = {
@@ -29,7 +30,7 @@ M.format = function()
       return {
         exe = O.lang.clang.formatter.exe,
         args = O.lang.clang.formatter.args,
-        stdin = not (O.lang.clang.formatter.stdin ~= nil),
+        stdin = O.lang.clang.formatter.stdin,
         cwd = vim.fn.expand "%:h:p",
       }
     end,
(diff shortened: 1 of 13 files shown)


Tastyep (approved):

maxxnino:
Should we use stdin true or false. Which one is better

abzcoding:
> Should we use stdin true or false. Which one is better. If true is better, can you edit zig to stdin=true

the only difference between them is this:

some formatters accept input from stdin:
```
cat my_shitty_code.sth | awesome_formatter -format_my_code (hence stdin = true)
```

some does not:
```
another_formatter -format -file my_shitty_code.sth
```

maxxnino:
> > Should we use stdin true or false. Which one is better. If true is better, can you edit zig to stdin=true
> 
> the only difference between them is this:
> 
> some formatters accept input from stdin (hence `stdin = true`):
> 
> ```shell
> cat my_shitty_code.sth | awesome_formatter -format_my_code
> ```
> 
> some does not (hence `stdin = false`):
> 
> ```shell
> another_formatter -format -file my_shitty_code.sth
> ```

If the formater can handle both. Which one is better.

abzcoding:
> > > Should we use stdin true or false. Which one is better. If true is better, can you edit zig to stdin=true
> > 
> > 
> > the only difference between them is this:
> > some formatters accept input from stdin (hence `stdin = true`):
> > ```shell
> > cat my_shitty_code.sth | awesome_formatter -format_my_code
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > some does not (hence `stdin = false`):
> > ```shell
> > another_formatter -format -file my_shitty_code.sth
> > ```
> 
> If the formater can handle both. Which one is better.

the default is ` stdin = true `

PR #992, opened 2021-07-16
Overrideable configs for core plugins

rebuilt:
# Description
Allows users to override default configs for core plugins with their own settings.  This addresses users concerns that they can't modify some settings.  This also means that going forward, if some new obscure setting gets released for a plugin, the user can try it out immediately without waiting for an entry to get added to a table.  

I made the following plugins override-able: autopairs, bufferline, compe, dap, dashboard, galaxyline, nvimtree, telescope, terminal, treesitter, zen

I skipped gitsigns, because I couldn't make it work with this method.

I also skipped formatter because there's nothing user configurable there
And which-key because I think this will have a different solution.  Can't we just expose the default whichkey mappings from the O object and let the user rewrite the default bindings if they wish?  The user_which_key table can stay in case they want the default bindings AND their own as well.  

Fixes #(issue)
https://github.com/ChristianChiarulli/LunarVim/issues/960

## How Has This Been Tested?
```lua
O.plugin.autopairs.config = function()
  print "loading user config for autopairs"
  -- VALID_CONFIG_HERE
end

O.plugin.bufferline.config = function()
  print "Loaded user config for bufferline"
  -- VALID_CONFIG_HERE
end

O.completion.config = function()
  print "Loaded user config for compe"
  -- VALID_CONFIG_HERE
end

O.plugin.dap.active = true
O.plugin.dap.config = function()
  print "Loaded user config for dap"
  -- VALID_CONFIG_HERE
end

O.plugin.dashboard.active = true
O.plugin.dashboard.config = function()
  print "Loaded user config for dashboard"
  -- VALID_CONFIG_HERE
end

O.plugin.galaxyline.config = function()
  print "Loaded user config for galaxyline"
  -- VALID_CONFIG_HERE
end

O.plugin.nvimtree.config = function()
  print "Loaded user config for nvimtree"
  -- VALID_CONFIG_HERE
end

O.plugin.telescope.config = function()
  print "Loaded user config for telescope"
  -- VALID_CONFIG_HERE
end

O.plugin.terminal.active = true
O.plugin.terminal.config = function()
  print "Loaded user config for terminal"
  -- VALID_CONFIG_HERE
end

O.treesitter.config = function()
  print "Loaded user config for treesitter"
  -- VALID_CONFIG_HERE
end

O.plugin.zen.active = true
O.plugin.zen.config = function()
  print "Loaded user config for zen"
  -- VALID_CONFIG_HERE
end
```

I added valid configs to make sure each plugin respected modifications

### DIFF
--- a/lua/plugins.lua
+++ b/lua/plugins.lua
@@ -59,9 +59,7 @@ return require("packer").startup(function(use)
   -- Autopairs
   use {
     "windwp/nvim-autopairs",
-    config = function()
-      require("core.autopairs").setup()
-    end,
+    config = require("core.autopairs").setup(),
   }
 
   -- Snippets
--- a/lua/plugins.lua
+++ b/lua/plugins.lua
@@ -60,11 +60,7 @@ return require("packer").startup(function(use)
   use {
     "windwp/nvim-autopairs",
     config = function()
-      if O.plugin.autopairs.config then
-        O.plugin.autopairs.config()
-      else
-        require("core.autopairs").setup()
-      end
+      require("core.autopairs").setup()
     end,
   }
 
--- a/lua/plugins.lua
+++ b/lua/plugins.lua
@@ -151,7 +151,7 @@ return require("packer").startup(function(use)
   use {
     "romgrk/barbar.nvim",
     config = function()
-      require "core.bufferline"
+      require("core.bufferline").setup()
     end,
     event = "BufWinEnter",
   }
--- a/lua/plugins.lua
+++ b/lua/plugins.lua
@@ -44,7 +44,9 @@ return require("packer").startup(function(use)
   -- Telescope
   use {
     "nvim-telescope/telescope.nvim",
-    config = require("core.telescope").setup(),
+    config = function()
+      require("core.telescope").setup()
+    end,
   }
 
   -- Autocomplete
--- a/lua/plugins.lua
+++ b/lua/plugins.lua
@@ -142,7 +142,7 @@ return require("packer").startup(function(use)
   use {
     "glepnir/galaxyline.nvim",
     config = function()
-      require "core.galaxyline"
+      require("core.galaxyline").setup()
     end,
     event = "BufWinEnter",
     disable = not O.plugin.galaxyline.active,
--- a/lua/plugins.lua
+++ b/lua/plugins.lua
@@ -59,7 +59,9 @@ return require("packer").startup(function(use)
   -- Autopairs
   use {
     "windwp/nvim-autopairs",
-    config = require("core.autopairs").setup(),
+    config = function()
+      require("core.autopairs").setup()
+    end,
   }
 
   -- Snippets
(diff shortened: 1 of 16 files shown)


lua/plugins.lua
```diff
@@ -44,7 +44,7 @@ return require("packer").startup(function(use)
   -- Telescope
   use {
     "nvim-telescope/telescope.nvim",
-    config = [[require('core.telescope').setup()]],
+    config = require("core.telescope").setup(),
```
Tastyep, line 47: Why calling directly setup instead of assigning it as callback?

rebuilt: Thanks,  I meant to change it so it would match the style of the others but I forgot to surround it in a function.

rebuilt: Ok.  I wrapped it.

Tastyep (approved):

kylo252 (commented):
I think it's looking pretty good. Nice work!

One observation though, I'm not still not sure what the best approach is for allowing partial overriding. So what is the use-case, or how often would a user want to completely overwrite the LV config? I had the same dilemma with the key-maps but it was gonna get pretty hacky. 

My suggestion or vision is to implement `vim.tbl_extend` or `vim.tbl_deep_extend` for these options. I think it should work pretty well since we're dealing with distinct keys and it's also the method that plugins use anyway.

rebuilt:
> So what is the use-case, or how often would a user want to completely overwrite the LV config?

We had a user who wanted to get rid of all the devicons used in all the configs so he wouldn't need to install a nerd font  https://github.com/ChristianChiarulli/LunarVim/pull/941
That pull request added many many keys that won't be used by most people.

Rants like this one prove we have power users who want full control over anything that bothers them... while still being able to pull the latest updates.
https://github.com/ChristianChiarulli/LunarVim/issues/960

Some users may have their own configurations for things like galaxyline that they spent a lot of time on.  This provides a way for them to just copy and paste their old configs into LunarVim

It's near impossible to support all options for all plugins since they're constantly changing.  This provides an easy way to work around that.

Some plugins (not all)  include a setup function where you pass a table of settings.  For those plugins, I think your solution(tbl_extend or tbl_deep_extend) for extending the functionality.  

But I haven't tested this approach.  I'm not sure how difficult it would be to make it work.  What made it hacky?

kylo252:
> Some plugins (not all) include a setup function where you pass a table of settings. For those plugins, I think your solution(tbl_extend or tbl_deep_extend) might work for extending the functionality.

Yeah my idea is to basically unify the interfaces for modifying the options. So right now, what takes priority? is it: 
1- `O.plugin.dashboard.config.search_handler`
2- `O.plugin.dashboard.search_handler`

I'm also worried that it's not clear enough that using `.config` will actually completely override the defaults. But yeah, it's a tough one. 

> What made it hacky?

It got confusing because I'd need to start parsing strings (like `<C-l>`) to see if they exist.  Otherwise, because of the autocmd then that file might get reloaded a few times, then having `nvim_del_key` for example will start throwing errors, because the key got unmapped the first time.

rebuilt:
Redundant after 1042

PR #999, opened 2021-07-17
Add feature Add openSUSE support in install.sh

richardschembri:
# Description

Fixes #(issue)

## How Has This Been Tested?

- Run command `:mycommand`
- Check logs
- ...

### DIFF
--- a/utils/installer/install.sh
+++ b/utils/installer/install.sh
@@ -31,6 +31,10 @@ installnodefedora() {
     sudo dnf install -y npm
 }
 
+installnodeopensuse() {
+	sudo zypper in nodejs-common
+}
+
 installnodegentoo() {
     echo "Printing current node status..."
     emerge -pqv net-libs/nodejs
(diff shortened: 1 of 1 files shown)


utils/installer/install.sh
```diff
@@ -163,6 +173,14 @@ installonfedora() {
     pip3 install wheel ueberzug
 }
 
+installonopensuse() {
+	sudo zypper in ripgrep fzf ranger
+	sudo zypper in python-devel python3-devel
+    pip3 install ueberzug
+    pip3 install neovim-remote
+    npm install -g tree-sitter-cli
```
kylo252, line 181: 1- This is not checking if `pip3` or `npm` are available beforehand. And it doesn't help that there isn't error handling.
2- I think it should be `pip3 install --user ...`

kylo252 (changes requested):

utils/installer/install.sh
richardschembri: Thank you for the feedback!

1. This function is a modified version of `installonubuntu` which does not have checks for `pip3` and no error handling. I think `installonubuntu` assumes to be run after the `npm` and `pip3` checks that are at the bottom of the script.
2. Same as point one, command copied from `installonubuntu`. Tested on local machine and worked.

Shall I still make the changes?

kylo252: You're right, I misread the order of the functions. 

Weirdly enough that `asktoinstallpip` is actually commented-out for some reason.  I generally think this entire script needs an overhaul.

I'd still change the pip command to install to user level. I'm also not sure that `neovim-remote` is needed, it's usually enough with `pynvim`.

richardschembri: I think so too, seems very WIP. But I think the changes should be done in a separate issue/pull request since this one specifically targets openSUSE. The proposed changes would also need to be applied on the functions relating to other OSs.

PR #1002, opened 2021-07-17
[Feature] Add swift support

abzcoding:
# Description

Adds swift support using [sourcekit-lsp](https://github.com/apple/sourcekit-lsp) 
credit goes to @DanielOberg


kylo252:
It says it's using `clangd` for objc and it doesn't support formatting either. So maybe keep that integration intact and just use the new settings for swift(? )

abzcoding:
> It says it's using `clangd` for objc and it doesn't support formatting either. So maybe keep that integration intact and just use the new settings for swift(? )

agreed and update :)

kylo252 (approved):

PR #1004, opened 2021-07-17
Check that plugin is install prior to calling config

### LINKED ISSUE #1003: [Bug]: Installation error after nvimtree refactoring (1f5fd93f80) on rolling branch
### Problem description

With the factoring of nvimtree in config() and setup() functions I'm experience an error installing due to a null value in default-config.lua

line 162 in `default-config.lua` reads: 
```lua
require("core.nvimtree").config()
```
Before running `PackagerInstall` the statement 
```lua
require("core.nvimtree")
```
 will return null and cause the error when trying to call `config()`
```lua
-- lua/core/nvimtree.lua
local status_ok, nvim_tree_config = pcall(require, "nvim-tree.config")
if not status_ok then
  return
end
```
Note also that the `setup()` function depends requires that` O.plugin.nvimtree` is defined (line 55), and this variable is defined in the `config function()`.

The fix that worked for me was to define a function that performs a call to the plugin, and check the status prior to call the configure().
```lua
-- lua/default-config.lua

local function configure_plugin(plugin_name)
  status_ok, plugin = pcall("require", plugin_name)
  if status_ok then
    plugin.config()
  end
end
configure_plugin("core.nvimtree")
```

I also had to add a check in `lua/core/nvimtree.lua:54` (setup function) as the setup is run at `PackerInstall`.
```lua
-- lua/core/nvimtree.lua:54

if not O.plugin.nvimtree  then
    return
end
```

 

### LunarVim version

rolling ddb627d004d1f6c3bbd593a381ed93bbe54a24fd

### Neovim version (>= 0.5)

v0.6.0-dev+8-g75f758b99

### Relevant log output

_No response_

### Screenshots

_No response_

### I have

- [X] Read the readme
- [X] Read the wiki

PellePedro:
# Description

The nvim-tree causes fresh installation to fail

Fixes #1003

## How Has This Been Tested?

Install a fresh neovim

### DIFF
--- a/lua/default-config.lua
+++ b/lua/default-config.lua
@@ -149,7 +149,7 @@ O = {
   },
 }
 
-lvutils = require("lv-utils")
+lvutils = require "lv-utils"
 
 require "core.status_colors"
 require("core.gitsigns").config()
--- a/lua/default-config.lua
+++ b/lua/default-config.lua
@@ -166,7 +166,7 @@ require("core.zen").config()
 require("core.telescope").config()
 require("core.treesitter").config()
 require("core.which-key").config()
-configure_plugin("core.nvimtree")
+configure_plugin "core.nvimtree"
 
 require("lang.clang").config()
 require("lang.cmake").config()
--- a/lua/default-config.lua
+++ b/lua/default-config.lua
@@ -149,12 +149,7 @@ O = {
   },
 }
 
-local function configure_plugin(plugin_name)
-  status_ok, plugin = pcall("require", plugin_name)
-  if status_ok then
-    plugin.config()
-  end
-end
+lvutils = require("lv-utils")
 
 require "core.status_colors"
 require("core.gitsigns").config()
@@ -166,7 +161,7 @@ require("core.zen").config()
 require("core.telescope").config()
 require("core.treesitter").config()
 require("core.which-key").config()
-configure_plugin "core.nvimtree"
+lvutils.configure_plugin "core.nvimtree"
 
 require("lang.clang").config()
 require("lang.cmake").config()
--- a/lua/default-config.lua
+++ b/lua/default-config.lua
@@ -149,6 +149,13 @@ O = {
   },
 }
 
+local function configure_plugin(plugin_name)
+  status_ok, plugin = pcall("require", plugin_name)
+  if status_ok then
+    plugin.config()
+  end
+end
+
 require "core.status_colors"
 require("core.gitsigns").config()
 require("core.compe").config()
@@ -159,7 +166,7 @@ require("core.zen").config()
 require("core.telescope").config()
 require("core.treesitter").config()
 require("core.which-key").config()
-require("core.nvimtree").config()
+configure_plugin("core.nvimtree")
 
 require("lang.clang").config()
 require("lang.cmake").config()
--- a/lua/core/nvimtree.lua
+++ b/lua/core/nvimtree.lua
@@ -52,7 +52,7 @@ end
 M.setup = function()
   local g = vim.g
 
-  if not O.plugin.nvimtree  then
+  if not O.plugin.nvimtree then
     return
   end
 
--- a/lua/core/nvimtree.lua
+++ b/lua/core/nvimtree.lua
@@ -52,6 +52,10 @@ end
 M.setup = function()
   local g = vim.g
 
+  if not O.plugin.nvimtree  then
+    return
+  end
+
   for opt, val in pairs(O.plugin.nvimtree) do
     g["nvim_tree_" .. opt] = val
   end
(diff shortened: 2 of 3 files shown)


lua/default-config.lua
```diff
@@ -149,6 +149,13 @@ O = {
   },
 }
 
+local function configure_plugin(plugin_name)
+  status_ok, plugin = pcall("require", plugin_name)
+  if status_ok then
+    plugin.config()
+  end
+end
+
```
kylo252, line 158: Could you move this function to `lv-utils`?

lua/default-config.lua
```diff
@@ -159,7 +166,7 @@ require("core.zen").config()
 require("core.telescope").config()
 require("core.treesitter").config()
 require("core.which-key").config()
-require("core.nvimtree").config()
+configure_plugin "core.nvimtree"
```
kylo252, line 169: I would just keep the function call in `nvimtree.config()`. I don't see why we can't keep there tbh.

lua/default-config.lua
PellePedro: Sure

lua/default-config.lua
PellePedro: So I understand you correct, the statement require("core.nvimtree") returns null before the nvimtree plugin is downloaded in :PackerInstall. Hence, we cannot call config() on a null object. As I understand we need to check if the plugin exist prior to calling config (if the nvimtree.config() is called from default-config.lua), the alternative would be to call config() from  nvimtree.lua at line 5, after the call check.

It would be great if you can elaborate a bit more about this comment.

kylo252: My idea is to change `configure_plugin` to be `is_plugin_loaded` and then call it from inside the usual `M.config()` so it would maybe look something like this: 

```lua
M.config = function()
  if lv-utils.is_plugin_loaded('nvimtree') == false then
    return
  end
  O.plugin.nvimtree = {
   side = "left",
   ---
   --
   }
end    
```

I think we probably also need to move the first `pcall` into the setup function, which I think is probably the real culprit here.
https://github.com/ChristianChiarulli/LunarVim/blob/11ac140eb57c70dd30b5ec34d31da1679c4151ce/lua/core/nvimtree.lua#L2

PellePedro: Cool, thanks, I'll give that a try.

PellePedro: I tried to update nvimtree.lua according to the comment above, but if I leave the 
require("core.nvimtree").config() in default-config.lua then I stilll get the Error at the initial PackagerInstall (before the plugins are loaded).

The config in nvimtree is used for defining the configurable parameters, so I also successfully tried to move calling config to the plugin as.
  -- NvimTree
  use {
    "kyazdani42/nvim-tree.lua",
    -- event = "BufWinOpen",
    -- cmd = "NvimTreeToggle",
    commit = "fd7f60e242205ea9efc9649101c81a07d5f458bb",
    config = function()
      require("core.nvimtree").config()
      require("core.nvimtree").setup()
    end,
  }

Therefore the two options that works for me are.

#1. Move the nvimtree  require("core.nvimtree").config() from default-config.lua / default-config and call the require("core.nvimtree").config()  from the plugin.lua

#2. Keep the config() call in default-config.lua / default-config and call lv_utils.configure_plugin()
to use pcall before config() is called.


Comments welcome.

ChristianChiarulli:
does this fix the issue on install as is?

rebuilt:
I've tested this pr as-is.  It fixes the problem.  It also changes the default behavior of nvimtree so it does not toggle open automatically on start

@kylo252  is correct.

The pcalls in core.nvimtree.lua are set as locals outside the functions for toggle_tree and setup.  When you call config and setup those locals will not get loaded, hence the nil errors.

You can leave default-config.lua as-is and just move this pcall into the setup function

```
local status_ok, nvim_tree_config = pcall(require, "nvim-tree.config")
if not status_ok then
	return
end

```

And this pcall into the toggle_tree function

```
local view_status_ok, view = pcall(require, "nvim-tree.view")
if not view_status_ok then
	return
end

```



I like @kylo252 's function definition since it shortens the pcalls and the method name makes the intent clearer
```
if lv-utils.is_plugin_loaded('nvimtree') == false then
    return
  end
```

rebuilt:
User's will also have to -rf the plugin folder after any fix to this problem

kylo252:
I think this might be superseded by #1011 
Since the we have a pattern that is problematic.

abzcoding:
Fixed by #1012

abzcoding:
@PellePedro, thank you for putting so much time and effort into this.
I just wanted to apologize for forcing my own fix and not using this. It was getting a bit out of hand.

Issue #996, opened 2021-07-16
[Feature]: User should be able to choose formatting provider


abzcoding:
### Feature motivation

some plugins support `vim.lsp.buf.formatting()` and they run faster than calling a cmd using `formatter.nvim`

an example would be `scala` using [nvim-metals](https://github.com/scalameta/nvim-metals) as talked about [here](https://github.com/ChristianChiarulli/LunarVim/pull/986#discussion_r671463631)

### Describe the expected solution

letting users override the autoformat capability to run `vim.lsp.formatting()` instead of `formatter.nvim` for specified languages

### Describe the alternatives you've considered

**1. overriding the keymap for it inside the `lua/lang/sth.lua`**
```lua
M.lsp = function()
-- ...
     O.user_which_key = {
       l = {
         f = { "<cmd>lua vim.lsp.buf.formatting()<cr>", "Format" },
       },
     }
-- ...
end
```
**2. add a new keymap and disable format_on_save and lint_on_save:**
```lua
M.config = function()
O.lang.sth ={
  formatters = {},
  linters = {},
}
end

M.lsp = function()
-- ...
     O.user_which_key = {
       l = {
         F = { "<cmd>lua vim.lsp.buf.formatting()<cr>", "Format with native lsp" },
       },
     }
-- ...
end
```

**3. and override/removing the `autoformat augroup`**

but it seems more like a hack than a solution

### Additional context

#986
#888
#1005

sethigeet:
I guess we could add a function on to the global object (`_G`) and then call it in all the places from where formatting is done (autocmds, keybindings, etc.) and that function could check the formatter selected for the current `filetype` in the global options object (`O`) every time we format a file.

By adding the function to the global object(`_G`), it will be loaded at the start and hence will be very fast

Tastyep:
First, about the which key mapping, I think that we should expose only one mapping for formatting, more mappings are just confusing for the user.
Also, the augroup for autoformatting should stay, we don't want to remove the functionality, we just want it to behave differently based on the user's settings.
As you know, both  of them (whichkey and augroup) call `FormatWrite`, the straightforward solution for adding a customization point would  be to use one function we would declare somewhere (maybe formatter.lua, maybe somewhere else) which would dispatch to the preferred way of formatting. This means this function will either call `vim.lsp.buf.formatting()` or `FormatWrite`.

[...]

PR #1005, opened 2021-07-17
Fix Clang integration

kylo252:
# Description

Fix Clang integration

- Rename `O.lang.clang` to `O.lang.cpp`
- Use built-in support of `clangd` for `clang-tidy` and `clang-format`
- User can use `<leader>lF` which is mapped to `vim.lsp.buf.formatting`
to perform formatting using the `clangd`.
- Allow the user to override all the options for `clangd`
- Allow the user to specify an external linter and formatter By explcitly
setting `O.lang.cpp.linter` and `O.lang.cpp.formatter` respectively.

Fixes #(issue)

Clang is not a language 😄 

## How Has This Been Tested?

Open any *.c or *.cc file and test formatting with `<leader>lF` and try to trigger clang-tidy, if it's not already complaining about something.

### DIFF
--- a/lua/core/which-key.lua
+++ b/lua/core/which-key.lua
@@ -118,6 +118,7 @@ M.config = function()
           "Workspace Diagnostics",
         },
         f = { "<cmd>silent FormatWrite<cr>", "Format" },
+        F = { "<cmd>lua vim.lsp.buf.formatting()<CR>", "Format (native-lsp)" },
         i = { "<cmd>LspInfo<cr>", "Info" },
         j = {
           "<cmd>lua vim.lsp.diagnostic.goto_next({popup_opts = {border = O.lsp.popup_border}})<cr>",
--- a/lua/lang/cpp.lua
+++ b/lua/lang/cpp.lua
@@ -1,24 +1,22 @@
 local M = {}
 
 M.config = function()
-  O.lang.clang = {
+  O.lang.cpp = {
     diagnostics = {
       virtual_text = { spacing = 0, prefix = "" },
       signs = true,
       underline = true,
     },
-    cross_file_rename = true,
-    header_insertion = "never",
     filetypes = { "c", "cpp", "objc" },
-    formatter = {
-      exe = "clang-format",
-      args = {},
-      stdin = true,
+    clangd_flags = {
+      "--background-index",
+      "--header-insertion=never",
+      "--cross-file-rename",
+      "--clang-tidy",
+      "--clang-tidy-checks=-*,llvm-*,clang-analyzer-*",
     },
-    linters = {
-      "cppcheck",
-      "clangtidy",
-    },
+    formatter = {},
+    linters = {},
     debug = {
       adapter = {
         command = "/usr/bin/lldb-vscode",
@@ -29,12 +27,15 @@ M.config = function()
 end
 
 M.format = function()
+  if O.lang.cpp.format == nil then
+    return
+  end
   local shared_config = {
     function()
       return {
-        exe = O.lang.clang.formatter.exe,
-        args = O.lang.clang.formatter.args,
-        stdin = O.lang.clang.formatter.stdin,
+        exe = O.lang.cpp.formatter.exe,
+        args = O.lang.cpp.formatter.args,
+        stdin = O.lang.cpp.formatter.stdin,
         cwd = vim.fn.expand "%:h:p",
       }
     end,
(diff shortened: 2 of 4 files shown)


lua/lang/cpp.lua
```diff
@@ -1,24 +1,22 @@
...
-      "clangtidy",
+    clangd_flags = {
+      "--background-index",
+      "--header-insertion=never",
+      "--cross-file-rename",
+      "--clang-tidy",
+      "--clang-tidy-checks=-*,llvm-*,clang-analyzer-*,modernize-*",
     },
+    formatter = {},
+    linters = {},
```
abzcoding, line 19: these lines pretty much make `format_on_save` and `lint_on_save` useless, do we want that?
or do we want to leave it as is and fix #996 ?

ChristianChiarulli:
Having multiple ways to format can be confusing, should <leader>f just be format? 

We need to really think about how the end user will see this not understanding what lsp formatting is vs "offline" formatting.

kylo252:
> Having multiple ways to format can be confusing, should f just be format?
> 
> We need to really think about how the end user will see this not understanding what lsp formatting is vs "offline" formatting.

I don't think anyone would want to use "offline" formatting for C/C++ but not want a proper "intellisense" support.

 I just believe that this is a different use-case than other languages (looking at interpreted ones for example). As anyone who actually bothers to format C/C++ would almost definitely would want to work on the code extensively and therefore require `clangd`. So if you do actually have `clangd` as a prerequisite for a good editing experience, then why attempt to add more tools to the mixture? 
 
 Let `clangd` do its thing and manage `clang-tidy` asynchronously and also potentially get the benefit of having the LS do proper formatting for you.

jameswalmsley (approved):
I'd agree with @kylo252 on this.. for cpp you should really have the lot installed clangd and clang-format.

I suppose this comes back to detecting if the language server is installed.. if not then fallback to clang-format directly?

lua/core/which-key.lua
```diff
@@ -118,6 +118,7 @@ M.config = function()
           "Workspace Diagnostics",
         },
         f = { "<cmd>silent FormatWrite<cr>", "Format" },
+        F = { "<cmd>lua vim.lsp.buf.formatting()<CR>", "Format (native-lsp)" },
```
Tastyep, line 121: We don't need two bindings for formatting, instead 'f' should call a function which would dispatch to the preferred option

lua/lang/cpp.lua
Tastyep: IMO we should let the user have the choice.
Also, clangd doesn't run  cppcheck.

lua/lang/cpp.lua
```diff
@@ -1,24 +1,22 @@
...
-    },
-    linters = {
-      "cppcheck",
-      "clangtidy",
+    clangd_flags = {
+      "--background-index",
+      "--header-insertion=never",
+      "--cross-file-rename",
+      "--clang-tidy",
+      "--clang-tidy-checks=-*,llvm-*,clang-analyzer-*,modernize-*",
```
Tastyep, line 16: Why providing a bunch of enabled checks ?

Tastyep (changes requested):

lua/core/which-key.lua
kylo252: I agree, but that would require some extra steps. What if I want to use `shfmt` but still let `clangd` deal with the formatting for me. We need some way to test if the client has formatting ability before having a fallback on an external tool.

lua/lang/cpp.lua
kylo252: The whole point is that the user can override `O.lang.cpp.linters` if they want. I just made them empty by default.

It's some weird behaviour by `clangd` where it doesn't know how to run `clang-tidy` without some pre-defined checks. I'm only using the basic ones. 

I forgot to remove "modernize" which is useful for testing but may overwhelm the user.

lua/core/which-key.lua
Tastyep: Ok, maybe this is out of the scope of this PR.
I'm not sure if I understand your example with `shfmt` and `clangd`, but the idea would indeed be that instead of calling `Format` we call an internal function that makes the dispatching (probably depending on a config value).

lua/lang/cpp.lua
Tastyep: Ok :+1:  I wasn't aware of that since I use clang-format and clang-tidy separately.

kylo252:
Fixed by #1078.

PR #1008, opened 2021-07-17
[Feature] Add support for svelte

abzcoding:
# Description

Added support for `svelte` thanks to @hulufei

## How Has This Been Tested?

- `:LspInstall svelte`


lua/lang/svelte.lua
```diff
@@ -0,0 +1,35 @@
...
+
+M.lsp = function()
+  if require("lv-utils").check_lsp_client_active "svelte" then
+    return
+  end
+
+  require("lspconfig").svelte.setup {
+    cmd = { DATA_PATH .. "/lspinstall/svelte/node_modules/.bin/svelteserver", "--stdio" },
+    filetypes = { "svelte" },
+    root_dir = require("lspconfig.util").root_pattern("package.json", ".git"),
```
Tastyep, line 25: I think we should avoid overwriting the default values, lspconfig should be well enough configured, and also it creates bugs on our side because lspconfig get updates, but we keep outdated configs.

Tastyep (changes requested):

lua/lang/svelte.lua
abzcoding: aren't they the exact same?
I think if we want to do that, we need to do it for all `core/lang` files and avoid overwriting default values?
so do I create a new PR and change all of them, or is there something specifically wrong with this config?

Tastyep: They are the same, but if lspconfig updates their defaults, we don't, you get that.
This needs to be done for all language servers, indeed, so I'll approve this in the meantime.
I started doing it today, but there is something I don't know how to do yet.
They apply the given user configuration on top of their default config, which is good, but this also means that it overwrites both the "cmd" entry and the options part. Also, I didn't manage to access their default config.

Tastyep (approved):

lua/lang/svelte.lua
abzcoding: I completely agree with you, even the maintainer of `nvim-lspconfig` was talking about this in discord

so i think we can ask him `mjlbach` in discord about that

PR #1011, opened 2021-07-18
[Refactor] Clean-up redundant module-load checks

### LINKED ISSUE #1003: [Bug]: Installation error after nvimtree refactoring (1f5fd93f80) on rolling branch

kylo252:
# Description

It's unncessary to check if a plugin is loaded during `config` part in Packer, since it "Specifies code to run after this plugin is loaded."

See: `:h packer.use()`

Also remove the redundant `augroup` logic for linter and formatter since each file is only sourced ones.

Fixes #1003

## How Has This Been Tested?

Run on a clean install

```shell
rm -rf ~/.local/share/nvim ~/.config/nvim/plugin/*
nvim --headless +'autocmd User PackerComplete sleep 100m | qall' +PackerInstall
nvim --headless +'autocmd User PackerComplete sleep 100m | qall' +PackerSync
nvim +PackerCompile
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,8 +4,7 @@ repos:
       - id: shfmt
         name: shfmt
         minimum_pre_commit_version: 2.4.0
-        language: golang
-        additional_dependencies: [mvdan.cc/sh/v3/cmd/shfmt@v3.2.2]
+        language: system
         entry: shfmt
         args: [-i=2, -ci, -w]
         types: [shell]
@@ -21,7 +20,7 @@ repos:
         language: rust
         entry: stylua
         types: [lua]
-        args: [.]
+        args: ['-']
       - id: luacheck
         name: luacheck
         language: system
--- a/lua/core/terminal.lua
+++ b/lua/core/terminal.lua
@@ -1,5 +1,4 @@
 local M = {}
-local Log = require "core.log"
 local utils = require "utils"
 
 M.config = function()
@@ -46,20 +45,11 @@ M.config = function()
 end
 
 M.setup = function()
-  local status_ok, terminal = pcall(require, "toggleterm")
-  if not status_ok then
-    Log:get_default().error "Failed to load toggleterm"
-    print(terminal)
-    return
-  end
+  local terminal = require "toggleterm"
   for _, exec in pairs(lvim.builtin.terminal.execs) do
     require("core.terminal").add_exec(exec[1], exec[2], exec[3])
   end
   terminal.setup(lvim.builtin.terminal)
-end
-
-local function is_installed(exe)
-  return vim.fn.executable(exe) == 1
 end
 
 M.add_exec = function(exec, keymap, name)
@@ -85,8 +75,9 @@ end
 
 M._exec_toggle = function(exec)
   local binary = M._split(exec)[1]
-  if is_installed(binary) ~= true then
-    print("Please install executable " .. binary .. ". Check documentation for more information")
+  if vim.fn.executable(binary) ~= 1 then
+    local Log = require "core.log"
+    Log:get_default().error("Unable to run executable " .. binary .. ". Please make sure it is installed properly.")
     return
   end
   local Terminal = require("toggleterm.terminal").Terminal
--- a/lua/core/which-key.lua
+++ b/lua/core/which-key.lua
@@ -1,5 +1,4 @@
 local M = {}
-local Log = require "core.log"
 M.config = function()
   lvim.builtin.which_key = {
     active = false,
@@ -230,14 +229,7 @@ M.config = function()
 end
 
 M.setup = function()
-  -- if not package.loaded['which-key'] then
-  --  return
-  -- end
-  local status_ok, which_key = pcall(require, "which-key")
-  if not status_ok then
-    Log:get_default "Failed to load whichkey"
-    return
-  end
+  local which_key = require "which-key"
 
   which_key.setup(lvim.builtin.which_key.setup)
 
@@ -247,10 +239,8 @@ M.setup = function()
   local mappings = lvim.builtin.which_key.mappings
   local vmappings = lvim.builtin.which_key.vmappings
 
-  local wk = require "which-key"
-
-  wk.register(mappings, opts)
-  wk.register(vmappings, vopts)
+  which_key.register(mappings, opts)
+  which_key.register(vmappings, vopts)
 end
 
 return M
--- a/lua/plugins.lua
+++ b/lua/plugins.lua
@@ -131,15 +131,9 @@ return {
     "terrortylor/nvim-comment",
     event = "BufRead",
     config = function()
-      local status_ok, nvim_comment = pcall(require, "nvim_comment")
-      if not status_ok then
-        local Log = require "core.log"
-        Log:get_default().error "Failed to load nvim-comment"
-        return
-      end
-      nvim_comment.setup()
+      require("nvim_comment").setup()
       if lvim.builtin.comment.on_config_done then
-        lvim.builtin.comment.on_config_done(nvim_comment)
+        lvim.builtin.comment.on_config_done(require "nvim_comment")
       end
     end,
   },
--- a/lua/core/compe.lua
+++ b/lua/core/compe.lua
@@ -1,5 +1,4 @@
 local M = {}
-local Log = require "core.log"
 M.config = function()
   lvim.builtin.compe = {
     enabled = true,
@@ -61,11 +60,7 @@ end
 M.setup = function()
   vim.g.vsnip_snippet_dir = lvim.vsnip_dir
 
-  local status_ok, compe = pcall(require, "compe")
-  if not status_ok then
-    Log:get_default().error "Failed to load compe"
-    return
-  end
+  local compe = require "compe"
 
   compe.setup(lvim.builtin.compe)
 
--- a/lua/core/dap.lua
+++ b/lua/core/dap.lua
@@ -1,5 +1,4 @@
 local M = {}
-local Log = require "core.log"
 M.config = function()
   lvim.builtin.dap = {
     active = false,
@@ -13,11 +12,7 @@ M.config = function()
 end
 
 M.setup = function()
-  local status_ok, dap = pcall(require, "dap")
-  if not status_ok then
-    Log:get_default().error "Failed to load dap"
-    return
-  end
+  local dap = require "dap"
 
   vim.fn.sign_define("DapBreakpoint", lvim.builtin.dap.breakpoint)
   dap.defaults.fallback.terminal_win_cmd = "50vsplit new"
--- a/lua/core/gitsigns.lua
+++ b/lua/core/gitsigns.lua
@@ -1,5 +1,4 @@
 local M = {}
-local Log = require "core.log"
 M.config = function()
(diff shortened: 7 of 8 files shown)


rebuilt:
I tested.  It works.  

Should we be using pcalls or require statements inside configuration files for core plugins?  

I think the logic behind using pcalls was to prevent the user from seeing a bunch of errors on first start.  That was when we loaded the entire configuration file whole unlike now where we load configurations with the 'config' key.  

Now that we've split up the configuration file into 'config' and 'setup' functions.  And those functions get called just when they're needed, I  think we should prefer require statements over pcalls.

kylo252:
> Should we be using pcalls or require statements inside configuration files for core plugins?

After this, there's only once instance where `pcall` is used inside a `.config()` function and it's for telescope. I'm not actually sure if it's safe to remove or not, especially if we revert to lazy-loading it.

Otherwise I think as long as it's inside a function that's called by packer then it should be sufficient to use `require(..)`

ChristianChiarulli:
I need a few people to sign off on this actually working for new installs, is there any reason not to use the pcalls for added safety?

lua/plugins.lua
```diff
@@ -82,15 +82,15 @@ return require("packer").startup(function(use)
   use {
     "mhartington/formatter.nvim",
     config = function()
-      require "core.formatter"
+      require("core.formatter").generic_config()
```
Tastyep, line 85: Please use `setup` instead of `generic_config` like for other plugins

Tastyep (changes requested):
I'll test a fresh install and come back to this.

Thank you for these changes! I'll need them before begin able to move on [PR938](https://github.com/ChristianChiarulli/LunarVim/pull/938)

Tastyep:
It works after a fresh install :+1:

lua/plugins.lua
kylo252: This one is tricky, because we're not actually setting it up just yet here. That is done per language. Then I also considered `setup_generic()` but it's not that much better either. 
Alternatively, I was considering brining back `autocmds.lua` or maybe it was never there (?)
What do you think?

Tastyep: This is still a behavior that we set up only once the plugin is loaded, I would keep the same naming as IMO we don't need to differentiate them from other plugins.

Tastyep: Also, the config callback is just a hook provided by packer so that additional behavior can be defined, which is exactly what we do by defining augroups

lua/default-config.lua
```diff
@@ -153,6 +153,7 @@ O = {
 }
 
 require "core.status_colors"
+require("core.which-key").config() -- We need to to get which-key as early as possible
```
Tastyep, line 156: Good catch :clap:

Tastyep (approved):

lua/plugins.lua
```diff
@@ -57,7 +57,7 @@ return {
   {
```
Tastyep, line 57: You forgot to update
- autopairs
- galaxyline

Tastyep (changes requested):

Tastyep:
Probably related: https://github.com/wbthomason/packer.nvim/pull/508

Tastyep:
Did you  abandon this ?

kylo252:
> Did you abandon this ?

I wasn't sure how to proceed with this, I'll take a look at the current state of the repo and see how this can help.

lua/core/terminal.lua
```diff
@@ -85,7 +75,7 @@ end
 
 M._exec_toggle = function(exec)
   local binary = M._split(exec)[1]
-  if is_installed(binary) ~= true then
+  if vim.fn.executable(binary) ~= true then
```
Tastyep, line 78: `vim.fn.executable` returns 1 on success not `true`
``` lua
print(1 == true) -> false
```

lua/core/autopairs.lua
```diff
@@ -1,54 +1,63 @@
...
-  return
+local M = {}
+
+function M.config()
+  lvim.builtin.autopairs = {
+    active = true,
+    ---@usage  map <CR> on insert mode
+    map_cr = true,
+    ---@usage auto insert after select function or method item,
+    map_complete = vim.bo.filetype ~= "tex",
```
Tastyep, line 9: Maybe add a note above this as it only works if the first file you open is a tex file, not when switching from Markdown to tex for example.

lua/plugins.lua
```diff
@@ -68,11 +68,12 @@ return {
     -- event = "InsertEnter",
     after = "nvim-compe",
     config = function()
-      require "core.autopairs"
+      require("core.autopairs").setup()
       if lvim.builtin.autopairs.on_config_done then
         lvim.builtin.autopairs.on_config_done(require "nvim-autopairs")
       end
     end,
+    disable = not lvim.builtin.autopairs.active,
```
Tastyep, line 76: While at it, would you extend this to other builtins ?

Tastyep (changes requested):

lua/core/terminal.lua
kylo252: oops.. thanks copy-paste!

lua/core/autopairs.lua
kylo252: Yeah I'm not sure how any of this works. Feel free to write the note you'd like to see here, and I'll add it verbatim.

lua/plugins.lua
kylo252: Tracked in #1318

lua/core/autopairs.lua
Tastyep: Smth like
`-- NOTE: This should be wrapped into a function so that it is re-evaluated when opening new files`

lua/core/autopairs.lua
```diff
@@ -1,54 +1,63 @@
...
-local npairs = require "nvim-autopairs"
-local Rule = require "nvim-autopairs.rule"
 
--- skip it, if you use another global object
-_G.MUtils = {}
+M.setup = function()
+  -- skip it, if you use another global object
+  _G.MUtils = {}
+  local npairs = require "nvim-autopairs"
+  local Rule = require "nvim-autopairs.rule"
```
Tastyep, line 24: lowercase `rule`

lua/plugins.lua
```diff
@@ -131,15 +132,9 @@ return {
...
-      if not status_ok then
-        local Log = require "core.log"
-        Log:get_default().error "Failed to load nvim-comment"
-        return
-      end
-      nvim_comment.setup()
+      require("nvim_comment").setup()
       if lvim.builtin.comment.on_config_done then
-        lvim.builtin.comment.on_config_done(nvim_comment)
+        lvim.builtin.comment.on_config_done(require, "nvim_comment")
```
Tastyep, line 137: extra comma after require

lua/core/which-key.lua
```diff
@@ -218,14 +217,7 @@ M.config = function()
...
 M.setup = function()
-  -- if not package.loaded['which-key'] then
-  --  return
-  -- end
-  local status_ok, which_key = pcall(require, "which-key")
-  if not status_ok then
-    Log:get_default "Failed to load whichkey"
-    return
-  end
+  local which_key = require "which-key"
```
Tastyep, line 220: Can you also remove the duplicate line 230 ?

Tastyep (changes requested):
If I uninstall a plugin (toggleterm), close and reopen lvim, I get this message

image

Maybe this should be an issue on packer's side

kylo252:
> If I uninstall a plugin (toggleterm), close and reopen lvim, I get this message
> Maybe this should be an issue on packer's side

It works for me but I had to force it a few time. I think it might be a Packer problem indeed. 
I really would like a way to delete `packer_compiled.lua` before we even source anything.

This is what I see in the log:

```log
[WARN  Sat 14 Aug 2021 02:02:04 PM CEST 3965981047435] .../site/pack/packer/start/packer.nvim/lua/packer/clean.lua:79: Cleaning cancelled!
[WARN  Sat 14 Aug 2021 02:03:43 PM CEST 4065424164345] .../site/pack/packer/start/packer.nvim/lua/packer/clean.lua:79: Cleaning cancelled!
[ERROR Sat 14 Aug 2021 02:03:43 PM CEST 4065879981962] .../site/pack/packer/start/packer.nvim/lua/packer/async.lua:20: Error in coroutine: ...narvim/site/pack/packer/start/packer.nvim/lua/packer.lua:692: attempt to index a nil value
[ERROR Sat 14 Aug 2021 02:19:36 PM CEST 5018613856386] .../site/pack/packer/start/packer.nvim/lua/packer/async.lua:20: Error in coroutine: ...ite/pack/packer/start/packer.nvim/lua/packer/compile.lua:598: attempt to index a nil value
[ERROR Sat 14 Aug 2021 02:19:50 PM CEST 5032306546542] .../site/pack/packer/start/packer.nvim/lua/packer/async.lua:20: Error in coroutine: ...ite/pack/packer/start/packer.nvim/lua/packer/compile.lua:598: attempt to index a nil value
[WARN  Sun 15 Aug 2021 02:54:46 PM CEST 2133135548228] .../site/pack/packer/start/packer.nvim/lua/packer/clean.lua:79: Cleaning cancelled!

```

lua/core/autopairs.lua
kylo252: This is now in #1318 since it's different scope.

Tastyep (approved):

.pre-commit-config.yaml
```diff
@@ -21,7 +20,7 @@ repos:
         language: rust
         entry: stylua
         types: [lua]
-        args: [.]
+        args: ['-']
```
abzcoding, line 23: shouldn't this be `.` instead of `-`?

abzcoding (approved):

PR #1012, opened 2021-07-18
fixing broken nvimtree

abzcoding:
# Description

Fixes nvimtree failing during clean install


geclos:
Still experiencing installation issue after this:  `Not an editor command: PackerInstall` using the one-line install from README.

PR #1014, opened 2021-07-18
remove ranger and ueberzug

luxus:
# Description
removed ranger and ueberzug from the install.sh as it not used anymore by lunarvim

### DIFF
--- a/utils/installer/install.sh
+++ b/utils/installer/install.sh
@@ -134,38 +134,30 @@ asktoinstallpip() {
 }
 
 installonmac() {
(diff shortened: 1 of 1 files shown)


abzcoding:
> # Description
> 
> removed ranger and ueberzug from the install.sh as it not used anymore by lunarvim

can you test this on a fresh system to make sure if it works :pleading_face:

luxus:
tested on Mac and ubuntu.. the changes are very save... I just removed both packages from the installer part on every install function

PR #1016, opened 2021-07-18
[Feature] Add Vue formatter (prettier) setup

PasiBergman:
# Description

1. Add lang config for vue filetype
2. Add formatter setup for vue filetype
2. Use nvim-lspconfig fuels default root_dir value
3. Setup ts-fmt-lint with vue filetype
4. Add `lv_utils.gsub_args` as general substitution function for args

## How Has This Been Tested?

With vue filetype (*.vue)
- vuels activates
- efm activate
- `:Format` and `<leader>lf` formats file content
- linting (eslint or eslint_d) is working through efm
- fixing (e.g. `:lua vim.lsp.buf.formatting()`) is available if eslint_d is used.
-


williamgoulois:
Do we need to separate vue from other javascript like files for formatting ?
We could just add the filetype in `lua/lsp/tsserver-ls.lua` ?

ChristianChiarulli:
@williamgoulois vue uses a separate langserver afaik called vetur

PasiBergman:
Yes, Vue uses Vetur (vuels) for language server, but for formatter....we could consider using the tsserver's **formatter** settings.

PR #1035, opened 2021-07-19
[Feature]: Add sorbet lsp for ruby

will:
# Description

The [sorbet lsp server is included](https://github.com/neovim/nvim-lspconfig/blob/master/CONFIG.md#sorbet) in nvim-lspconfig. This adds support for this in addition to the solargraph support that was already there.

## How Has This Been Tested?

I tested it on my project that uses sorbet, and it worked with no other setup:

image

### DIFF
--- a/lua/lang/ruby.lua
+++ b/lua/lang/ruby.lua
@@ -41,24 +41,26 @@ M.lint = function()
 end
 
 M.lsp = function()
-  if require("lv-utils").check_lsp_client_active "solargraph" then
-    return
+  if not require("lv-utils").check_lsp_client_active "sorbet" then
+    require("lspconfig").sorbet.setup {}
   end
 
-  -- If you are using rvm, make sure to change below configuration
-  require("lspconfig").solargraph.setup {
-    cmd = { DATA_PATH .. "/lspinstall/ruby/solargraph/solargraph", "stdio" },
-    on_attach = require("lsp").common_on_attach,
-    handlers = {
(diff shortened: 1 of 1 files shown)


lua/lang/ruby.lua
```diff
@@ -41,24 +41,26 @@ M.lint = function()
 end
 
 M.lsp = function()
-  if require("lv-utils").check_lsp_client_active "solargraph" then
```
abzcoding, line 44: this line was to avoid spawning a new `solargraph` lsp for each nvim instance, so ...

abzcoding:
can you fix the formatting plz :
```bash
stylua --config-path ./utils/.stylua.toml *.lua lua
```

will:
> can you fix the formatting plz :
> 
> ```shell
> stylua --config-path ./utils/.stylua.toml *.lua lua
> ```

Oh yeah, thanks! Just pushed

Tastyep (approved):
Looks legit, I don;t have ruby files so I can;t test it but the code makes sens :+1: 

Also, it's unfortunate, but LspInstall doesn't support sorbet.

will:
> Also, it's unfortunate, but LspInstall doesn't support sorbet.

Yeah, but I think it might be because [sorbet](https://sorbet.org) is not a general purpose ruby LSP. You need to make pretty big changes to your project (add types everywhere, and pragma comments to the start of every file, etc) to support it. And also they push a new version like every day, so if you have two projects using it, you might want to have different versions per project, not one global one for your user.

PR #1036, opened 2021-07-19
Support prettierd

maxime50:
This supports prettierd. If the prettierd package (@fsouza/prettierd) is installed locally, it will be prioritized. Otherwise, it will load the global prettierd.

related issue: https://github.com/ChristianChiarulli/LunarVim/issues/418

### DIFF
--- a/lua/lsp/tsserver-ls.lua
+++ b/lua/lsp/tsserver-ls.lua
@@ -1,15 +1,28 @@
 vim.cmd "let proj = FindRootDirectory()"
 local root_dir = vim.api.nvim_get_var "proj"
 
--- use the global prettier if you didn't find the local one
-local prettier_instance = root_dir .. "/node_modules/.bin/prettier"
-if vim.fn.executable(prettier_instance) ~= 1 then
-  prettier_instance = O.lang.tsserver.formatter.exe
+local formatter_exe = O.lang.tsserver.formatter.exe
+
+local get_prettier_module_name = function()
+  -- get local module name
+	if formatter_exe == "prettierd" then
+		return "@fsouza/prettierd"
+	end
+	return formatter_exe
+end
+
+local get_prettier_instance = function()
+	-- prioritize local instance over global
+  local local_instance = root_dir .. "/node_modules/.bin/" .. get_prettier_module_name()
+	if vim.fn.executable(local_instance) == 1 then
+		return local_instance
+	end
+	return formatter_exe
 end
 
 O.formatters.filetype["javascriptreact"] = {
   function()
-    local args = { "--stdin-filepath", vim.fn.fnameescape(vim.api.nvim_buf_get_name(0)) }
+    local args = { vim.fn.fnameescape(vim.api.nvim_buf_get_name(0)) }
     local extend_args = O.lang.tsserver.formatter.args
 
     if extend_args then
@@ -19,7 +32,7 @@ O.formatters.filetype["javascriptreact"] = {
     end
 
     return {
-      exe = prettier_instance,
+      exe = get_prettier_instance(),
       args = args,
       stdin = true,
     }
--- a/lua/lsp/tsserver-ls.lua
+++ b/lua/lsp/tsserver-ls.lua
@@ -3,17 +3,9 @@ local root_dir = vim.api.nvim_get_var "proj"
 
 local formatter_exe = O.lang.tsserver.formatter.exe
 
-local get_prettier_module_name = function()
-  -- get local module name
-  if formatter_exe == "prettierd" then
-    return "@fsouza/prettierd"
-  end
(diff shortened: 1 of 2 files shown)


maxime50:
How can I make this formatting check happy? I just ran stylua with the default configs

abzcoding:
> How can I make this formatting check happy? I just ran stylua with the default configs
this I think:
```bash
stylua --config-path ./utils/.stylua.toml *.lua
```

lua/lsp/tsserver-ls.lua
```diff
@@ -1,15 +1,28 @@
...
+  local local_instance = root_dir .. "/node_modules/.bin/" .. get_prettier_module_name()
+  if vim.fn.executable(local_instance) == 1 then
+    return local_instance
+  end
+  return formatter_exe
 end
 
 O.formatters.filetype["javascriptreact"] = {
   function()
-    local args = { "--stdin-filepath", vim.fn.fnameescape(vim.api.nvim_buf_get_name(0)) }
```
abzcoding, line 12: are you sure `prettier` doesn't need this argument?
```
--stdin-filepath <path>  Path to the file to pretend that stdin comes from.
```

maxime50: Good question. I had to remove `--stdin-filepath` for `prettierd` (otherwise it didn't work)
I also tried with normal prettier to make sure it still work and everything seems ok.
Should we put this argument condtionally if using `prettier`? idk

abzcoding: I don't know either 🤦 , somebody else can help probably 🥺

williamgoulois: According to the documentation https://github.com/fsouza/prettierd#editor-integration we only need to give `vim.api.nvim_buf_get_name(0)` to args !

```lua
-- prettierd
       function()
          return {
            exe = "prettierd",
            args = {vim.api.nvim_buf_get_name(0)},
            stdin = true
          }
        end
```

lua/lsp/tsserver-ls.lua
```diff
@@ -1,15 +1,28 @@
...
-  prettier_instance = O.lang.tsserver.formatter.exe
+local formatter_exe = O.lang.tsserver.formatter.exe
+
+local get_prettier_module_name = function()
+  -- get local module name
+  if formatter_exe == "prettierd" then
+    return "@fsouza/prettierd"
+  end
+  return formatter_exe
+end
```
williamgoulois, line 12: You don't need this ! The local module name is still
`node_modules/.bin/prettierd` and not `node_modules/.bin/@fsouza/prettierd` !

lua/lsp/tsserver-ls.lua
```diff
@@ -1,15 +1,28 @@
...
+
+local get_prettier_instance = function()
+  -- prioritize local instance over global
+  local local_instance = root_dir .. "/node_modules/.bin/" .. get_prettier_module_name()
```
williamgoulois, line 16:
```suggestion
  local local_instance = root_dir .. "/node_modules/.bin/" .. formatter_exe
```

williamgoulois (changes requested):
I tested with local installation of prettierd and this does not work. The local module_name is still `node_modules/.bin/prettierd`

lua/lsp/tsserver-ls.lua
maxime50: Alright thank you for this! 😄

lua/lsp/tsserver-ls.lua
maxime50: Fixed! 🐼

williamgoulois (approved):

lua/lsp/tsserver-ls.lua
maxime50: Thanks for the information - does `prettier` still need the argument tho? If yes, we should put the argument conditonally then. Everything works as expected on `prettier` and `prettierd` without the argument on my side, but I don't want to break anything 🙃

abzcoding: @williamgoulois can you test this with the `prettier` as well, just wanna make sure that it doesn't need the `--stdin-filepath`, bear in mind that instead of directly opening the file, you need to open it through telescope file picker or sth to be sure

williamgoulois: Just tested with prettier local and global this does work !

williamgoulois:
I just saw that we have the same code for java and css we could refactor it in prettier.lua and call it in the ftplugins files

abzcoding (approved):
@ChristianChiarulli LGTM 🚀

ChristianChiarulli:
prettier is now supported through null-ls

PR #1042, opened 2021-07-20
[Feature]: Add callbacks for plugins with a config entry for packer

devtoi:
# Description

This adds a feature for callbacks that are run when a plugin's config is run. This will allow users to change settings for plugins after the fact instead of fully relying on lunarvim. One example is being able to run whick_keys register function to hook up less standard keybindings.

I am terrible with naming, so please make sure they are good :)

Example syntax:

```lua
O.plugin.which_key.on_config_done = function(module)
  module.register({
    s = {
      name = "Search",
      t = {
        ":lua search_selected()<CR>",
        "Search Selected Text",
      },
    },
  }, {
    mode = "v",
    prefix = "<leader>",
  })
end
```

## How Has This Been Tested?

To run the callbacks insert this into `lv-config.lua`:

```lua
O.plugin.lspinstall.on_config_done = function(module)
  print("lspinstall: " .. vim.inspect(module ~= nil))
end

O.plugin.telescope.on_config_done = function(module)
  print("telescope: " .. vim.inspect(module ~= nil))
end

O.plugin.compe.on_config_done = function(module)
  print("compe: " .. vim.inspect(module ~= nil))
end

O.plugin.autopairs.on_config_done = function(module)
  print("autopairs: " .. vim.inspect(module ~= nil))
end

O.plugin.treesitter.on_config_done = function(module)
  print("treesitter: " .. vim.inspect(module ~= nil))
end

O.plugin.formatter.on_config_done = function(module)
  print("formatter: " .. vim.inspect(module ~= nil))
end

O.plugin.lint.on_config_done = function(module)
  print("lint: " .. vim.inspect(module ~= nil))
end

O.plugin.nvimtree.on_config_done = function(module)
  print("nvimtree: " .. vim.inspect(module ~= nil))
end

O.plugin.gitsigns.on_config_done = function(module)
  print("gitsigns: " .. vim.inspect(module ~= nil))
end

O.plugin.which_key.on_config_done = function(module)
  print("which_key: " .. vim.inspect(module ~= nil))
end

O.plugin.comment.on_config_done = function(module)
  print("comment: " .. vim.inspect(module ~= nil))
end

O.plugin.rooter.on_config_done = function(module)
  print("rooter: " .. vim.inspect(module ~= nil))
end

O.plugin.galaxyline.on_config_done = function(module)
  print("galaxyline: " .. vim.inspect(module ~= nil))
end

O.plugin.bufferline.on_config_done = function(module)
  print("bufferline: " .. vim.inspect(module ~= nil))
end

O.plugin.dap.on_config_done = function(module)
  print("dap: " .. vim.inspect(module ~= nil))
end

O.plugin.dashboard.on_config_done = function(module)
  print("dashboard: " .. vim.inspect(module ~= nil))
end

O.plugin.terminal.on_config_done = function(module)
  print("terminal: " .. vim.inspect(module ~= nil))
end

O.plugin.zen.on_config_done = function(module)
  print("zen: " .. vim.inspect(module ~= nil))
end
```


Tastyep:
Thank you for your contribution, this is indeed something missing and will allow the user to implement their own logic.
Now, I'm not a big fan of how this is done :dagger:

There is a function in `plugins.lua` while I'd like to keep it as a table of plugins.
Also, this function could be simplified into
``` lua
if O.plugin[plugin_variable_name].on_config_done then
    O.plugin[plugin_variable_name].on_config_done(plugin)
end
```
IMO, this code should simply be at the end of the `setup()` function of each plugin, this way the plugin is already loaded and required, so you can pass it to `on_config_done`
Additionally, this removes the need of mapping the plugin name.

Finally, as a general rule, don't suffix your variables with `_variable`

Tastyep (changes requested):
Thank you for your contribution, this is indeed something missing and will allow the user to implement their own logic.
Now, I'm not a big fan of how this is done :dagger:

There is a function in `plugins.lua` while I'd like to keep it as a table of plugins.
Also, this function could be simplified into
``` lua
if O.plugin[plugin_variable_name].on_config_done then
    O.plugin[plugin_variable_name].on_config_done(plugin)
end
```
IMO, this code should simply be at the end of the `setup()` function of each plugin, this way the plugin is already loaded and required, so you can pass it to `on_config_done`
Additionally, this removes the need of mapping the plugin name.

Finally, as a general rule, don't suffix your variables with `_variable`

devtoi:
Thanks for the feedback! I marked this PR as WIP (coming from gitlab where that is the standard xD) as I assumed people would have some good feedback on it. Then I noticed github does things differently and I can't seem to convert this into a draft. Ohh well.

I was thinking about having it in the setup() for each plugin. But the problem is that not every plugin has a set up and having it in the config entry directly seemed to be the most consistent and clear.

Good idea with the shortening of the if. It is legacy because I didn't start out having the change of the plugin table. With the smaller if I could just refactor it to not use a function and have those 3 lines instead.

devtoi:
I now refactored it to not use a function but just those 3 if lines instead. Should be more concise and clear now.

rebuilt:
Pulling to test this pr.

rebuilt (approved):
Thank you very much for this.  I tested your example and then I also made endwise rules for myself in autopairs.  Magic.  This can be merged.

Tastyep (approved):

rebuilt:
You can clear all the default whichkey bindings by calling reset()

rebuilt:
Endwise rule

```
O.plugin.autopairs.on_config_done = function(module)
  local endwise = require('nvim-autopairs.ts-rule').endwise
  module.add_rules(
  {
endwise('def', 'end',nil, nil)
    })
end
```

PR #1043, opened 2021-07-20
[Feature] Allow for custom path to a language server binary

ichigo-gyuunyuu:
# Description

Users will now be able to use a different language server binary. The path can be set in `lv-confg.lua`, e.g. 
`O.lang.lua.lsp_path = "path/to/different/binary"`

Some languages like dart (already has `sdk_path`) and java (it's a special child :) ) have been left as is.

## How Has This Been Tested?

- Confirm that the langauge server attaches by checking `:LspInfo`
- Add a custom path in `lv-config.lua` and confirm that the change is reflected


Tastyep (approved):
Legit to me, we could maybe have done `lsp.path` instead of `lsp_path`, but since we don't have a use for other attributes, it's fine :+1:

ichigo-gyuunyuu:
> Legit to me, we could maybe have done `lsp.path` instead of `lsp_path`, but since we don't have a use for other attributes, it's fine 👍

Yeah, that is more future-proof. Time for more refactoring if the lsp section gets more attributes 😃

abzcoding:
> > Legit to me, we could maybe have done `lsp.path` instead of `lsp_path`, but since we don't have a use for other attributes, it's fine 👍
> 
> Yeah, that is more future-proof. Time for more refactoring if the lsp section gets more attributes 😃

I also agree with tasty, and would appreciate it if you use `lsp.path` instead of `lsp_path` 🥺

init.lua
```diff
@@ -1,3 +1,4 @@
+print "hi"
```
abzcoding, line 1: can you remove this line plz

ichigo-gyuunyuu: I told myself to remove that before comitting 😞

abzcoding (approved):
tyvm

Tastyep (approved):

PR #1048, opened 2021-07-20
WIP: Minor comments changes, and proposed refactor/cleanup to lang/rust.lua

Uzaaft:
# Description

* Fixed an issue with the lv-config.example.lua rust_tools example.
* Moved the @usage comment for rustfmt down to linter, and created a separate comment for the formatter.
* Proposed changes:
* @abzcoding I am tagging you here since you seem to be the person who has written most of the rust lang config.  I am proposing to remove the values where we are using the default values. This would remove what I consider "unnessecary" code(unless there is a reason behind it that I am not thinking about), and clean up the rust file somewhat. 
* Add propper clippy(or another linter) support for rust files. 

Fixes #(issue)

## How Has This Been Tested?

TBD

### DIFF
--- a/lua/lang/rust.lua
+++ b/lua/lang/rust.lua
@@ -7,12 +7,13 @@ M.config = function()
       parameter_hints_prefix = "<-",
       other_hints_prefix = "=>", -- prefix for all the other hints (type, chaining)
     },
-    -- @usage can be clippy
+    -- @usage can be rustfmt
     formatter = {
       exe = "rustfmt",
       args = { "--emit=stdout", "--edition=2018" },
       stdin = true,
     },
+    -- @usage can be clippy
     linter = "",
     diagnostics = {
       virtual_text = { spacing = 0, prefix = "" },
--- a/lua/lang/rust.lua
+++ b/lua/lang/rust.lua
@@ -42,7 +42,7 @@ end
 
 M.lint = function()
   -- TODO: implement linters (if applicable)
-  require("lint).linters_by_fit =  {
+  require("lint").linters_by_fit =  {
     rust = O.lang.rust.linters,
   }
   -- return "No linters configured!"
--- a/lua/lang/rust.lua
+++ b/lua/lang/rust.lua
@@ -42,7 +42,10 @@ end
 
 M.lint = function()
   -- TODO: implement linters (if applicable)
-  return "No linters configured!"
+  require("lint).linters_by_fit =  {
+    rust = O.lang.rust.linters,
+  }
+  -- return "No linters configured!"
 end
 
 M.lsp = function()
--- a/utils/installer/lv-config.example.lua
+++ b/utils/installer/lv-config.example.lua
@@ -68,7 +68,7 @@ O.lang.python.analysis.use_library_code_types = true
 O.lang.tsserver.linter = nil
 
 -- rust
--- O.lang.rust.rust_tools = true
+-- O.lang.rust.rust_tools.active = true
 -- O.lang.rust.formatter = {
 --   exe = "rustfmt",
 --   args = {"--emit=stdout", "--edition=2018"},


lua/lang/rust.lua
```diff
@@ -41,7 +42,10 @@ end
 
 M.lint = function()
   -- TODO: implement linters (if applicable)
-  return "No linters configured!"
+  require("lint).linters_by_fit =  {
```
ichigo-gyuunyuu, line 45: closing quote missing, `require("lint")` maybe?

Uzaaft: Ah I was looking after that missing quote! Thanks xD

Uzaaft:
Note to self: Consider contributing to the nvim-lint repo.

lua/lang/rust.lua
```diff
@@ -7,12 +7,13 @@ M.config = function()
...
     },
-    -- @usage can be clippy
+    -- @usage can be rustfmt
     formatter = {
       exe = "rustfmt",
       args = { "--emit=stdout", "--edition=2018" },
       stdin = true,
     },
+    -- @usage can be clippy
     linter = "",
```
abzcoding, line 17: can you change this to `linters = {}`

abzcoding:
> Note to self: Consider contributing to the nvim-lint repo.

that would be wonderful 😍

lua/lang/rust.lua
Uzaaft: Already done locally.

Uzaaft:
Just to get an explicit answer. Is it alright if I go ahead and remove all the default values we are declaring? @abzcoding

abzcoding:
> Just to get an explicit answer. Is it alright if I go ahead and remove all the default values we are declaring? @abzcoding

wait what , which ones?

Uzaaft:
If you look at https://github.com/simrat39/rust-tools.nvim you will see that the values assigned at row 64-82 are the standard values. To clean stuff up I think we should remove these.

abzcoding:
> If you look at https://github.com/simrat39/rust-tools.nvim you will see that the values assigned at row 64-82 are the standard values. To clean stuff up I think we should remove these.

you are right sir, lets do that 👍

Uzaaft:
Since I am already on it, I might as well try and implement a linter for javascript, typescript, both react and none-react.
Is there a way we could use one file for all four languages?

abzcoding:
> Since I am already on it, I might as well try and implement a linter for javascript, typescript, both react and none-react.
> Is there a way we could use one file for all four languages?

if you want to do that, please do it in a separate PR

abzcoding:
`WIP` only works for GitLab, it is useless in GitHub 😞

abzcoding:
@Uzaaft you might wanna pause dev on this PR, a lot of stuff is about to change
check `new_contract` branch for more info

Uzaaft:
Ah snap. I`ll pause it.

abzcoding:
rust_tools are gone 😭 
rust support is provided via `null-ls`

Uzaaft:
> rust_tools are gone 
> rust support is provided via `null-ls`

Ah, that's a shame. Made LunarVim less useful out of the box when it comes to rust development.

abzcoding:
> > rust_tools are gone
> > rust support is provided via `null-ls`
> 
> Ah, that's a shame. Made LunarVim less useful out of the box when it comes to rust development.

yeah hopefully after we've created video series for each lang, we can provide a sample `lv-config.lua` tuned for each lang so that users can use those as a reference

Uzaaft:
I would very much like to help out with creating a sample lv-config for rust, since I pretty much live in rust code xD

abzcoding:
> I would very much like to help out with creating a sample lv-config for rust, since I pretty much live in rust code xD

can you add a `rust.md` [here](https://github.com/LunarVim/lunarvim.org/tree/master/docs/languages)

Uzaaft:
> > I would very much like to help out with creating a sample lv-config for rust, since I pretty much live in rust code xD
> 
> can you add a `rust.md` [here](https://github.com/LunarVim/lunarvim.org/tree/master/docs/languages)

On it captain

PR #1052, opened 2021-07-20
[Refactor] Installer v2 with support for backup

kylo252:
# Description

- Provide an install script that is more coherent and doesn't require root privileges. Instead it will prompt the user to install any missing requirements with a hint about the possible command.
- Add built-in support for backing up any existing lunarvim folders.
- Add support for cloning from a fork `LV_REMOTE=..`
- Add support for overriding all the destination folders `$LUNARVIM_RUNTIME_DIR`, `$LUNARVIM_CONFIG_DIR`, `$INSTALL_PREFIX`
- Add support for updating automatically if it was already installed and there are no conflicts (--ff-only)

## How Has This Been Tested?

```shell
wget https://raw.githubusercontent.com/kylo252/LunarVim/installer_v2/utils/installer/install.sh
LVBRANCH=installer_v2 LV_REMOTE="kylo252/lunarvim.git" INSTALL_PREFIX=/tmp/t1 LUNARVIM_CONFIG_DIR=/tmp/t1 LUNARVIM_RUNTIME_DIR=/tmp/t2 bash ./install.sh
```

### DIFF
--- a/utils/installer/install.sh
+++ b/utils/installer/install.sh
@@ -64,17 +64,22 @@ EOF
 
   echo "Detecting platform for managing any additional neovim dependencies"
   detect_platform
-  check_system_deps
 
-  __add_separator "80"
+  # skip this in a Github workflow
+  if [ -z "$GITHUB_ACTIONS" ]; then
+    check_system_deps
 
-  echo "Would you like to check noevim's nodejs dependencies?"
-  read -p "[y]es or [n]o (default: no) : " -r answer
-  [ "$answer" != "${answer#[Yy]}" ] && install_npm_deps
+    __add_separator "80"
 
-  echo "Would you like to check noevim's python dependencies?"
-  read -p "[y]es or [n]o (default: no) : " -r answer
-  [ "$answer" != "${answer#[Yy]}" ] && install_pip_deps
+    echo "Would you like to check noevim's nodejs dependencies?"
+    read -p "[y]es or [n]o (default: no) : " -r answer
+    [ "$answer" != "${answer#[Yy]}" ] && install_npm_deps
+
+    echo "Would you like to check noevim's python dependencies?"
+    read -p "[y]es or [n]o (default: no) : " -r answer
+    [ "$answer" != "${answer#[Yy]}" ] && install_pip_deps
+
+  fi
 
   __add_separator "80"
 
@@ -150,8 +155,7 @@ function check_system_deps() {
   for dep in "${!__system_deps[@]}"; do
     if ! command -v "${__system_deps[$dep]}" &>/dev/null; then
       print_missing_dep_msg "$dep"
-      # do not abort in a Github workflow
-      [ -z "$GITHUB_ACTIONS" ] && exit 1
+      exit 1
     fi
   done
 }
--- a/utils/installer/install.sh
+++ b/utils/installer/install.sh
@@ -49,15 +49,15 @@ function main() {
   # Welcome
   __add_separator "80"
 
-  cat <<"EOF"
-88\                                                   88\               
-88 |                                                  \__|              
-88 |88\   88\ 888888$\   888888\   888888\ 88\    88\ 88\ 888888\8888\  
-88 |88 |  88 |88  __88\  \____88\ 88  __88\\88\  88  |88 |88  _88  _88\ 
-88 |88 |  88 |88 |  88 | 888888$ |88 |  \__|\88\88  / 88 |88 / 88 / 88 |
-88 |88 |  88 |88 |  88 |88  __88 |88 |       \88$  /  88 |88 | 88 | 88 |
-88 |\888888  |88 |  88 |\888888$ |88 |        \$  /   88 |88 | 88 | 88 |
-\__| \______/ \__|  \__| \_______|\__|         \_/    \__|\__| \__| \__|
+  cat <<'EOF'
+      88\                                                   88\               
+      88 |                                                  \__|              
+      88 |88\   88\ 888888$\   888888\   888888\ 88\    88\ 88\ 888888\8888\  
+      88 |88 |  88 |88  __88\  \____88\ 88  __88\\88\  88  |88 |88  _88  _88\ 
+      88 |88 |  88 |88 |  88 | 888888$ |88 |  \__|\88\88  / 88 |88 / 88 / 88 |
+      88 |88 |  88 |88 |  88 |88  __88 |88 |       \88$  /  88 |88 | 88 | 88 |
+      88 |\888888  |88 |  88 |\888888$ |88 |        \$  /   88 |88 | 88 | 88 |
+      \__| \______/ \__|  \__| \_______|\__|         \_/    \__|\__| \__| \__|
 EOF
 
   __add_separator "80"
@@ -107,7 +107,7 @@ EOF
     echo "Updating LunarVim"
     update_lvim
   else
-    clone_lvim "$@"
+    clone_lvim
     setup_lvim
   fi
 
@@ -150,7 +150,8 @@ function check_system_deps() {
   for dep in "${!__system_deps[@]}"; do
     if ! command -v "${__system_deps[$dep]}" &>/dev/null; then
       print_missing_dep_msg "$dep"
-      exit 1
+      # do not abort in a Github workflow
+      [ -z "$GITHUB_ACTIONS" ] && exit 1
     fi
   done
 }
@@ -203,17 +204,10 @@ function install_packer() {
 
 function clone_lvim() {
   echo "Cloning LunarVim configuration"
-  case "$@" in
-    *--testing*)
-      cp -r "$(pwd)" "$LUNARVIM_RUNTIME_DIR/lvim"
-      ;;
-    *)
-      local CLONE_CMD="git clone --progress --branch $LVBRANCH \
-        --depth 1 https://github.com/$LV_REMOTE $LUNARVIM_RUNTIME_DIR/lvim"
-      printf "Running: %s" "$CLONE_CMD"
-      eval "$CLONE_CMD"
-      ;;
-  esac
+  local CLONE_CMD="git clone --progress --branch $LVBRANCH \
+    --depth 1 https://github.com/$LV_REMOTE $LUNARVIM_RUNTIME_DIR/lvim"
+  printf "Running: %s" "$CLONE_CMD"
+  eval "$CLONE_CMD"
 }
 
 function lvim() {
@@ -224,14 +218,25 @@ function lvim() {
   fi
 }
 
+function setup_shim() {
+  if [ ! -d "$INSTALL_PREFIX/bin" ]; then
+    mkdir -p "$INSTALL_PREFIX/bin"
+  fi
+  cat >"$INSTALL_PREFIX/bin/lvim" <<EOF
+#!/usr/bin/env bash
+
+declare -r LUNARVIM_RUNTIME_DIR="$LUNARVIM_RUNTIME_DIR"
+declare -r LUNARVIM_CONFIG_DIR="$LUNARVIM_CONFIG_DIR"
+
+exec nvim -u "$LUNARVIM_RUNTIME_DIR/lvim/init.lua" --cmd "set runtimepath+=$LUNARVIM_RUNTIME_DIR/lvim" "\$@"
+EOF
+}
+
 function setup_lvim() {
 
   echo "Installing LunarVim shim"
 
-  if [ ! -d "$INSTALL_PREFIX/bin" ]; then
-    mkdir -p "$INSTALL_PREFIX/bin"
-  fi
-  cp "$LUNARVIM_RUNTIME_DIR/lvim/utils/bin/lvim" "$INSTALL_PREFIX/bin"
+  setup_shim
   chmod u+x "$INSTALL_PREFIX/bin/lvim"
 
   echo "Preparing Packer setup"
--- a/utils/installer/install.sh
+++ b/utils/installer/install.sh
@@ -269,7 +269,7 @@ function setup_lvim() {
 
   echo "Thank you for installing LunarVim!!"
   echo "You can start it by running: $INSTALL_PREFIX/bin/lvim"
-  echo "Do not forget to use a font with ligatures support [https://github.com/ryanoasis/nerd-fonts]"
+  echo "Do not forget to use a font with glyphs (icons) support [https://github.com/ryanoasis/nerd-fonts]"
 }
 
 function update_lvim() {
--- a/utils/installer/install.sh
+++ b/utils/installer/install.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -euo pipefail
+set -eo pipefail
 
 #Set branch to master unless specified by the user
 declare -r LVBRANCH="${LVBRANCH:-master}"
--- a/utils/installer/install.sh
+++ b/utils/installer/install.sh
@@ -107,9 +107,6 @@ EOF
 
   __add_separator "80"
 
-  echo "Thank you for installing LunarVim!!"
-  echo "I recommend you also install and activate a font \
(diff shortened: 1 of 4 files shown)


utils/installer/install.sh
```diff
@@ -1,225 +1,191 @@
...
+  echo "Would you like to check noevim's python dependencies?"
+  read -p "[y]es or [n]no (default: no) : " -r answer
+  [ "$answer" != "${answer#[Yy]}" ] && install_pip_deps
+  case "$@" in
+    *--overwrite*)
+      echo "!!Warning!! -> Removing all lunarvim related config \
+        because of the --overwrite flag"
+      read -p "Would you like to continue? [y]es or [n]o : " -r answer
+      [ "$answer" == "${answer#[Yy]}" ] && exit 1
+      for dir in "${__lvim_dirs[@]}"; do
```
abzcoding, line 60: we might need to remove `NEOVIM_CACHE_DIR` as well (`~/.cache/nvim`)

utils/installer/install.sh
```diff
@@ -1,225 +1,191 @@
...
+      local CLONE_CMD="git clone https://github.com/$LV_REPO --branch $LVBRANCH --depth 1 \
+        $LUNARVIM_RUNTIME_DIR/lvim"
+      echo "$CLONE_CMD"
+      eval "$CLONE_CMD"
+      ;;
+  esac
+
+  if [ -f "$LUNARVIM_RUNTIME_DIR/lvim/utils/bin/lvim" ]; then
+    echo "Install LunarVim shim"
+    cp "$LUNARVIM_RUNTIME_DIR/lvim/utils/bin/lvim" "$HOME/.local/bin"
```
devtoi, line 172: Can we expect that the user had this path in the PATH?

utils/installer/install.sh
```diff
@@ -1,225 +1,191 @@
...
+  if [ -e "$LUNARVIM_RUNTIME_DIR/site/pack/packer/start/packer.nvim" ]; then
+    echo "packer already installed"
+  else
+    install_packer
+  fi
+
+  if [ -e "$LUNARVIM_RUNTIME_DIR/lvim/init.lua" ]; then
+    echo "LunarVim already installed"
+    echo "checking for the latest updates instead"
+    local CMD="git -C $LUNARVIM_RUNTIME_DIR/lvim pull origin $LVBRANCH"
```
devtoi, line 75: Should this case perhaps complain if the existing remote doesn't match the expected?

abzcoding:
also we need to check if `git` is available or not
(adding `git` and `neovim` as dependencies )

utils/installer/install.sh
```diff
@@ -1,225 +1,191 @@
...
+function clone_lvim() {
+  echo "Backing up old LunarVim configuration"
+  backup_old_config
+  echo "Cloning LunarVim configuration"
+  case "$@" in
+    *--testing*)
+      cp -r "$(pwd)" "$LUNARVIM_RUNTIME_DIR/lvim"
+      ;;
+    *)
+      local CLONE_CMD="git clone https://github.com/$LV_REPO --branch $LVBRANCH --depth 1 \
```
abzcoding, line 163: apparently `--depth 1` caused some issues.
cc: @rebuilt

utils/installer/install.sh
```diff
@@ -1,225 +1,191 @@
...
+
+function install_packer() {
+  git clone --depth 1 https://github.com/wbthomason/packer.nvim \
+    "$LUNARVIM_RUNTIME_DIR/site/pack/packer/start/packer.nvim"
+}
+
+function clone_lvim() {
+  echo "Backing up old LunarVim configuration"
+  backup_old_config
+  echo "Cloning LunarVim configuration"
```
abzcoding, line 157: on fresh install do we need to check for the existence of `LUNARVIM_RUNTIME_DIR` ?

.github/workflows/format.yaml
```diff
@@ -39,5 +39,5 @@ jobs:
 
       - name: Check formatting
         run: |
-          shfmt -l -d .
+          shfmt -i 2 -ci -l -d .
```
abzcoding, line 42: cc: @ChristianChiarulli

abzcoding: @kylo252 if we are changing this, we might wanna change the default formatting on `core/lang/sh.lua` as well
https://github.com/ChristianChiarulli/LunarVim/blob/0064b446a02be13dba0f16f084e073b347716710/lua/lang/sh.lua#L13-L17

utils/installer/install.sh
kylo252: That's actually already handled by the backup function. I took special note to keep that logic.

utils/installer/install.sh
kylo252: So I'm relying on the fact that people should have this, since it's used by `pip` for example. 

I'm also considering removing this all together and use a shell function and append it to the user bashrc/zshrc. But then we'd need to ask the user every time.

utils/installer/install.sh
kylo252: You mean if it's not called "origin" right? damn that got quite a bit more complicated just now. I should create a function for it which will actually prompt the user before doing anything to the repo.

.github/workflows/format.yaml
kylo252: I'm gonna move that shfmt style commit to a separate PR since it's affecting a few files already.

utils/installer/install.sh
kylo252: Yeah but if it's holding the sessions_dir then that will affect people's neovim's setup. So I'm not sure how to gracefully do this.

utils/installer/install.sh
devtoi: Yeah, or if the supplied LV_REPO isn't the same as origin

utils/installer/install.sh
kylo252: > also we need to check if `git` is available or not
> (adding `git` and `neovim` as dependencies )

I had a system_dep check that I can bring back for this.

utils/installer/install.sh
abzcoding: this is the `--overwrite` flag so the user has asked for it :smiling_imp:

utils/installer/install.sh
devtoi: I think the current one in this PR is the lesser evil. Changing user configs are gonna be iffy at best, imo.

.github/workflows/format.yaml
abzcoding: sounds good to me :+1:

utils/installer/install.sh
kylo252: Yeah this is way too complicated. I'm gonna leave this for a separate PR with regards on how to handle updating lunarvim. 
There are simply way too many possible cases so it's not trivial. I guess we could open a poll and see what people think the workflow should look like.

sperezconesa:
Some feedback installing in Ubuntu:
+ The script asks to `sudo apt install -y  nvim` it should be `neovim` I think.
+ Somewhere it is written "noevim".
+ I get this when installing:
image

kylo252:
> Some feedback installing in Ubuntu:
> 
> * The script asks to `sudo apt install -y  nvim` it should be `neovim` I think.
> * Somewhere it is written "noevim".
> * I get this when installing:

I rebased the script and added update functionality (when safe), can you try again please?

sperezconesa:
Installation works now! But if I try to start lvim in:
```bash
.local/bin/lvim
./test_lvim/bin/lvim
```
I just get regular plain neovim. I guess environment variables are missing.

kylo252:
> Installation works now! But if I try to start lvim in:
> ```bash
> .local/bin/lvim
> ./test_lvim/bin/lvim
> ```
> I just get regular plain neovim. I guess environment variables are missing.

I forgot to make the shim to be dynamic as well. Any ideas how to handle that, other than writing a new file?

abzcoding:
@kylo252 
can't we install all the dependencies if the `--testing` flag is given, or fix the installation github workflow

Tastyep (changes requested):
Looks neat :+1: 

My only complaint is that we now require fd-find, I don't have it, am I missing out on something ?

kylo252:
> Looks neat 👍
> 
> My only complaint is that we now require fd-find, I don't have it, am I missing out on something ?

I'm surprised it's not the default one in Telescope. You should definitely install it if you don't have it already. I'm also considering dropping it along with ripgrep from this table.

utils/installer/install.sh
```diff
@@ -1,272 +1,271 @@
...
 
-  if [ -e "$HOME/.local/share/lunarvim/lvim/init.lua" ]; then
-    echo 'config.lua already present'
-  else
-    cp "$HOME/.local/share/lunarvim/lvim/utils/installer/config.example.lua" "$HOME/.config/lvim/config.lua"
+function setup_shim() {
+  if [ ! -d "$INSTALL_PREFIX/bin" ]; then
+    mkdir -p "$INSTALL_PREFIX/bin"
   fi
+  cat >"$INSTALL_PREFIX/bin/lvim" <<EOF
```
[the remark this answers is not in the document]
kylo252: I would generally agree, but try to write that `sed` command in a way that doesn't expand variables. I hate it when I have a hard-coded expansion of `$HOME` in my scripts, and I would like to keep those variables. That's why I opted for this approach which is way easier to read/maintain, but it's also very possible that there's a nice way to do it with `sed`.

Some notes to keep in mind about making sure it's cross-platform:
- `sed -E` is different between Linux and Darwin
- `awk` isn't guaranteed to be available and/or work reliably on BSD

time to write it in `perl`? 😄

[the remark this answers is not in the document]
Tastyep: I don't understand what you mean about the sed command expanding variables, that's the whole point.
Please explain to me how having two versions of the same file, `bin/lvim` and this one hidden in the setup script is more maintainable.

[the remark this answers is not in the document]
kylo252: We don't need to have `util/bin/lvim` as a static file anymore. And there's nothing to maintain that isn't handled by the installer.

utils/installer/install.sh
```diff
@@ -1,272 +1,271 @@
...
+    git -C "$LUNARVIM_RUNTIME_DIR/lvim" pull --ff-only --progress ||
+      echo "Unable to guarantee data integrity while updating. Please do that manually instead." && exit 1
   fi
-fi
+  echo "Your LunarVim installation is up to date!"
+}
+
+function __add_separator() {
+  local DIV_WIDTH="$1"
+  printf "%${DIV_WIDTH}s\n" ' ' | tr ' ' -
```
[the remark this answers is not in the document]
kylo252: It's so that I don't have to escape the quote marks since the variable name would become `DIV_WIDTHs`, notice the `s`.

utils/installer/install.sh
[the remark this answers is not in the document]
Tastyep: What am i missing here ?
https://github.com/LunarVim/LunarVim/pull/1052/files#diff-fe4492969d1077edb685fdb1cc8cf4e6a7d93dcf6eb20d6b4e8fdd908077a5fa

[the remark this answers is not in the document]
kylo252: I meant I'm gonna delete it in the next patch 😄

utils/installer/install.sh
```diff
@@ -1,272 +1,280 @@
...
-}
+#!/usr/bin/env bash
+set -eo pipefail
 
-installnodeubuntu() {
-  sudo apt install nodejs
-  sudo apt install npm
-}
+#Set branch to master unless specified by the user
+declare -r LVBRANCH="${LVBRANCH:-master}"
```
Tastyep, line 5: `declare -r LVBRANCH=${LVBRANCH:-"master"}`

utils/installer/install.sh
```diff
@@ -1,272 +1,280 @@
...
-  apt install ripgrep fzf
-  pip install neovim-remote
-  npm install -g tree-sitter-cli
+# we try to install the missing one with cargo even though it's unlikely to be found
+function install_rust_deps() {
+  if ! command -v fd &>/dev/null; then
+    __attempt_to_install_with_cargo "fd-find"
+  fi
+  if ! command -v rg &>/dev/null; then
+    __attempt_to_install_with_cargo "ripgrpe"
```
Tastyep, line 194: ripgrep

utils/installer/install.sh
```diff
@@ -1,272 +1,280 @@
...
-fi
-
-if [ "$(uname)" != "Darwin" ]; then
-  if [ -e "$HOME/.local/share/applications/lvim.desktop" ]; then
-    echo 'Desktop file already available'
-  else
-    mkdir -p "$HOME/.local/share/applications"
-    cp "$HOME/.local/share/lunarvim/lvim/utils/desktop/lvim.desktop" "$HOME/.local/share/applications/lvim.desktop"
+function setup_lvim() {
+
```
Tastyep, line 244: blank line

Tastyep (changes requested):

utils/installer/install.sh
kylo252: That's not necessary. I needed to change it to rolling anyway.

Tastyep: I'd like that you quote the inner most variable instead of outer so that is also matches other lines

utils/installer/install.sh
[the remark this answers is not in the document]
Tastyep: Sorry to come back on the topic, but I still think this should not be generated from the script.
These lines are unnecessarily hidden in the setup script, what didn't you like about the solution I offered ?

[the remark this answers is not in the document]
kylo252: It's does not qualify as "hidden" because: 
1- It's not doing anything out of the ordinary, as `nvim -u` is a very common flag.
2- the binary is entirely dependent on the install script, and there is not a single useful thing that it does which is not controlled by the installer. 

This pattern is not as uncommon as you think it is, especially for a trivial shim. You can also see similar patterns in homebrew or nix for example. 

In a later version, the user will be prompted to add `lvim` as a shell alias/function, that will also help in maintaining autocompletion. or allow running as a default neovim config by adjusting some variables and adding a symlink to `vim.fn.stdpath "config"`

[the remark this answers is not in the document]
Tastyep: I've given my arguments, if you don't want to do it, so be it

Tastyep (approved):

abzcoding:
this doesn't work in mac os 😢 
Screen Shot 2021-08-22 at 8 49 07 PM

here is the difference between linux/osx
Screen Shot 2021-08-22 at 8 52 53 PM

kylo252:
> this doesn't work in mac os 😢 
> Screen Shot 2021-08-22 at 8 49 07 PM
> 
> here is the difference between linux/osx
> Screen Shot 2021-08-22 at 8 52 53 PM
> 

That's what I tried to avoid originally.
@Tastyep, any ideas/workarounds?

utils/installer/install.sh
```diff
@@ -1,272 +1,287 @@
...
-  brew install yarn
-}
+#!/usr/bin/env bash
+set -eo pipefail
 
-installnodeubuntu() {
-  sudo apt install nodejs
-  sudo apt install npm
-}
+#Set branch to master unless specified by the user
```
abzcoding, line 4: can we update this comment as well :)

abzcoding:
btw can we have a warning or sth that mac users need to have `$HOME/.local/` in their path or that the `lvim` binary is located at `~/.local/lvim` and let users know to move it somewhere else if they want for example
```sh
sudo mv ~/.local/lvim /usr/local/bin/lvim
```

utils/installer/install.sh
kylo252: oops :)

It cannot be safely used for master until https://github.com/LunarVim/LunarVim/pull/1354 is there.

abzcoding (approved):
LGTM 🚀

PR #1058, opened 2021-07-21
[Feature] Support multiple formatters

Tastyep:
# Description

This provides support for running multiple formatters.
I also define formatter profiles, so that the interface for configuring formatters and linters is the same.
Finally, I made sure that formatter parameters are interpreted at formatting time by wrapping them into lambdas.

Fixes #[933](https://github.com/ChristianChiarulli/LunarVim/issues/933)

## How Has This Been Tested?

Using `:Format` you can see the formatter's logs.
I tested with some languages (not all of them), more testing is required before merging this.
With python files, we can see that both black and isort run.
I checked that the arguments defined within a lambda are indeed evaluated at formatting time (not when configuring lvim or its formatters)

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ O.treesitter.highlight.enabled = true
 
 -- lua
 O.lang.lua.autoformat = false
-O.lang.lua.formatter = 'lua-format'
+O.lang.lua.formatters = { 'stylua' }
 
 -- javascript
 O.lang.tsserver.formatter = 'prettier'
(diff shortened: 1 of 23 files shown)


kylo252:
I thought we also needed a plan to handle #996 first, which is related to figuring out if the language server can handle formatting. Maybe we could add the LS as one of the profiles.  And we should probably explore the idea of using the global object `_G` for this, as suggested by @sethigeet

Also related: #1005

williamgoulois:
#1036 Related too

Tastyep:
> I thought we also needed a plan to handle #996 first, which is related to figuring out if the language server can handle formatting.
You are right, this should probably be taken care of first, I'll give it some thought tomorrow and reply and the issue to offer my ideas.

@williamgoulois Agreed too, only this work was already quite consequent, and I didn't want to do everything in the same PR.

williamgoulois:
@Tastyep I agree we should merge 996 and 1036 first then this PR and then we could refactor prettier (I can do it if needed)

williamgoulois:
> @williamgoulois Agreed too, only this work was already quite consequent, and I didn't want to do everything in the same PR.
Also related mfussenegger/nvim-lint#48

We already have efm for linting (used with eslint and eslint_d and it is also used in zsh.lua) why do you want to use nvim-lint instead ?

Tastyep:
> > @williamgoulois Agreed too, only this work was already quite consequent, and I didn't want to do everything in the same PR.
> > Also related [mfussenegger/nvim-lint#48](https://github.com/mfussenegger/nvim-lint/pull/48)
> 
> We already have efm for linting (used with eslint and eslint_d and it is also used in zsh.lua) why do you want to use nvim-lint instead ?

efm might go someday, also it never runs eslint on my setup

ChristianChiarulli:
should be able to get all of our formatters from here: https://github.com/jose-elias-alvarez/null-ls.nvim/pull/53

PR #1059, opened 2021-07-21
Fix slow CI by installing a neovim binary directly

kylo252:
# Description

Fix the slow CI workflow by using a neovim binary directly from the github release page instead of installing all the build dependencies. 

## How Has This Been Tested?

We'll test it directly and then install any missing deps individually.

### DIFF
--- a/utils/installer/install-neovim-from-release
+++ b/utils/installer/install-neovim-from-release
@@ -2,7 +2,7 @@
 
 set -eu pipefall
 
-declare -r INSTALL_PREFIX="${INSTALL_PREFIX:-"$HOME/.local"}"
+declare -r LV_INSTALL_PREFIX="${INSTALL_PREFIX:-"$HOME/.local"}"
 declare -r RELEASE_VER="${RELEASE_VER:-latest}" # can be set to nightly
 
 declare ARCHIVE_NAME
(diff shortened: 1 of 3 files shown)


williamgoulois:
Using a bash script to install neovim from release is quite overkill you can install nightly version just with this :
`brew tap jason0x43/homebrew-neovim-nightly && brew install --cask neovim-nightly`

utils/installer/install-neovim-from-release
```diff
@@ -0,0 +1,83 @@
...
+    exit 1
+  fi
+  echo "Download complete!"
+}
+
+function verify_neovim() {
+  echo "Verifying the installation.."
+  DOWNLOADED_SHA="$(sha256sum "$DOWNLOAD_DIR/$ARCHIVE_NAME.tar.gz" | awk '{print $1}')"
+
+  if [ "$RELEASE_SHA" != "$DOWNLOADED_SHA" ]; then
```
williamgoulois, line 57: You can use `sha256sum --check $RELEASE_SHA $DOWNLOADED_SHA` to verify

utils/installer/install-neovim-from-release
```diff
@@ -0,0 +1,83 @@
...
+
+function install_neovim() {
+
+  echo "Installing Neovim.."
+  pushd "$DOWNLOAD_DIR"
+  tar -xzf "$DOWNLOAD_DIR/$ARCHIVE_NAME.tar.gz"
+  popd
+  # https://dev.to/ackshaey/macos-vs-linux-the-cp-command-will-trip-you-up-2p00
+  cp -r "$DOWNLOAD_DIR/$RELEASE_NAME/." "$INSTALL_PREFIX"
+  echo "Installtion complete!"
```
williamgoulois, line 73:
```suggestion
  echo "Installation complete!"
```

utils/installer/install-neovim-from-release
kylo252: Thanks for the tip! I still want to print the  `$DOWNLOADED_SHA` in case of an error, so I'll keep this way for now.

PR #1072, opened 2021-07-23
we don't need to require 50 times

abzcoding:
# Description
what the title says :)


rebuilt:
Another option would be to move the on_attach assignment to the configuration for lspinstall.  

```
			local lspinstall = require("lspinstall")
			local common_on_attach = require("lsp").common_on_attach
			lspinstall.setup()
			local servers = lspinstall.installed_servers()
			for _, server in pairs(servers) do
				require("lspconfig")[server].setup({
					on_attach = common_on_attach,
				})
			end

```

abzcoding:
> Another option would be to move the on_attach assignment to the configuration for lspinstall.
> 
> ```lua
> local lspinstall = require("lspinstall")
> local common_on_attach = require("lsp").common_on_attach
> lspinstall.setup()
> local servers = lspinstall.installed_servers()
> for _, server in pairs(servers) do
> 	require("lspconfig")[server].setup({
> 		on_attach = common_on_attach,
>         })
> end
> ```
if we do it this way,
can users change the `on_attach` for a single language if they wanted to?

rebuilt:
> if we do it this way,
> can users change the `on_attach` for a single language if they wanted to?

Yes because the common_on_attach method doesn't change.  

```
function lsp_config.common_on_attach(client, bufnr)
  if O.lsp.on_attach_callback then
    O.lsp.on_attach_callback(client, bufnr)
  end
end

```

But rereading the codebase tsserver-ls.lua would need to change because it uses it's own method ts_server_on_attach  and mixes in capabilities settings.  Typescript still uses common_on_attach, it just also adds capabilities in the same method.  Every other language keeps these two things separate.

rebuilt:
I'll also say that using require many times is ugly and verbose but it shouldn't have performance costs since require makes sure the file is only loaded into memory once.

abzcoding:
> I'll also say that using require many times is ugly and verbose but it shouldn't have performance costs since require makes sure the file is only loaded into memory once.

i actually didn't know that 😅 but it looks really ugly

rebuilt:
From the docs

```
Roughly, require does the same job as dofile, but with two important differences. First, require searches for the file in a path; second, require controls whether a file has already been run to avoid duplicating the work. Because of these features, require is the preferred function in Lua for loading libraries. 
```

abzcoding:
> From the docs
> 
> ```
> Roughly, require does the same job as dofile, but with two important differences. First, require searches for the file in a path; second, require controls whether a file has already been run to avoid duplicating the work. Because of these features, require is the preferred function in Lua for loading libraries. 
> ```

does it cache the functions as well?
like for capabilites?

rebuilt:
I'm not sure any performance gains would be measurable.  The first thing the lua reference says about performance optimization is "don't".   I don't know enough lua to tell you about the memory management.

abzcoding:
> I don't know enough lua to tell you about the memory management. I'm not sure any performance gains would be measurable. The first thing the lua reference says about performance optimization is "don't".

ok fair enough

Tastyep (approved):

PR #1077, opened 2021-07-24
Fix rust_tools in example lv-config

midnightexigent:


ChristianChiarulli:
rust tools has been removed, I will show how to install it on the website in the future

PR #1078, opened 2021-07-24
Enable querying lang-server formatting capabilities

kylo252:
# Description

Enable checking the language server capabilities to determine if it can handle formatting natively.

## How Has This Been Tested?

Add   { "abzcoding/null-ls.nvim", branch = "feature/extra-formatters" },
Added a shortcut `<leader>lc` that quickly displays the provided capabilities by the server.
Check that null-ls doesn't start a formatter if it's already supported natively, when `clang` is used for example.


rebuilt:
new_contract just got merged into rolling.

abzcoding:
can you fix these plz
```
lua/default-config.lua:299:24: accessing undefined variable common_capabilities
lua/default-config.lua:780:24: accessing undefined variable common_capabilities
```

kylo252:
> can you fix these plz
> 
> ```
> lua/default-config.lua:299:24: accessing undefined variable common_capabilities
> lua/default-config.lua:780:24: accessing undefined variable common_capabilities
> ```

It was probably due to the merge. I may have to revert it and do it manually.

rebuilt:
Thanks for this pr.  I looks promising.  

Here are some errors I got when testing

When a formatter is not installed it throws errors.  

Complaining about missing luacheck.  Error goes away after installing luacheck.  
```
Error executing vim.schedule lua callback: /home/nelson/.local/share/lunarvim/lvim/lua/lsp/null-ls.lua:46: Unable to find the path for: [luacheck]
```

Javascript and typescript.  I assume I'm missing some formatters but all I see is a memory address for a table.  I tried to inspect the contents of the table but I think it's an empty table.  Can you confirm?

```
Error executing vim.schedule lua callback: /home/nelson/.local/share/lunarvim/lvim/lua/lsp/null-ls.lua:29: unable to find the path for: table: 0x7f36247a2c30
```

kylo252:
> When a formatter is not installed it throws errors.
> 
> Complaining about missing luacheck. Error goes away after installing luacheck.

That's according to plan! The error is not actually fatal. I wish I could a find a way to make `vim.notify` work but it seems anything other than `error()` is just disappearing somewhere..


> 
> Javascript and typescript. I assume I'm missing some formatters but all I see is a memory address for a table. I tried to inspect the contents of the table but I think it's an empty table. Can you confirm?
> 
> ```
> Error executing vim.schedule lua callback: /home/nelson/.local/share/lunarvim/lvim/lua/lsp/null-ls.lua:29: unable to find the path for: table: 0x7f36247a2c30
> ```

I was using the wrong variable in the print function by mistake. It should be fixed now.

rebuilt:
Ok.  That makes it clearer.  I'm missing eslint.

kylo252:
> Ok. That makes it clearer. I'm missing eslint.

Yeah, no more blind-guessing about why null-ls isn't attaching 😅

rebuilt:
I suggest using notify
```
    vim.notify(string.format("Unable to find the path for: [%s]", provider), 1)

```

kylo252:
> I suggest using notify
> 
> ```
>     vim.notify(string.format("Unable to find the path for: [%s]", provider), 1)
> ```

Yeah It's actually working now, I have no idea why it was refusing to work before. I changed it, and also removed the redundant print anyway.

rebuilt (approved):
This makes it clear to the user why null-ls is failing.

kylo252:
I forgot to set command path correctly for local nodejs modules but then I got hit by a weird bug where I can't actually use `null_ls.builtins[ft][exe]._opts.command` for lua files. It seems to be only available for certain languages, so I had to do redundant check instead of just passing the path either way 😕

ChristianChiarulli:
I'll have to take a look at this on stream

rebuilt:
This is how I test prs:

```
git fetch origin pull/1078/head:ls-caps
git checkout ls-caps
```

PR #1090, opened 2021-07-25
Fix ignored user autocommands

kylo252:
# Description

- Rename `lvim.autocommands` -> `lvim.autocommands.custom_group`. THis makkes it clearer that we're expecting a table and not a single entry. And it also allows us to expose `lvim.autocommands` to the user to allow for modification.
- Add `core.autocmds` to allow for evaluating the augroups _after_ `lvim.autocommands.custom_groups` has been updated.

## Fixes

Fix ignored user autocommands

## How Has This Been Tested?

Uncomment the example lines at the `lv-config[-examle].lua` to see the effect, or copy these lines manually

```lua
lvim.autocommands.custom_groups = {
  { "BufWinEnter", "*.lua", "setlocal ts=8 sw=8" },
}

```

### DIFF
--- a/lua/core/autocmds.lua
+++ b/lua/core/autocmds.lua
@@ -1,0 +1,112 @@
+local autocommands = {}
+
+
+lvim.autocommands = {
+  _general_settings = {
+    {
+      "TextYankPost",
+      "*",
+      "lua require('vim.highlight').on_yank({higroup = 'Search', timeout = 200})",
+    },
+    {
+      "BufWinEnter",
+      "*",
+      "setlocal formatoptions-=c formatoptions-=r formatoptions-=o",
+    },
+    {
+      "BufWinEnter",
+      "dashboard",
+      "setlocal cursorline signcolumn=yes cursorcolumn number",
+    },
+    {
+      "BufRead",
+      "*",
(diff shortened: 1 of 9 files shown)


lua/core/autocmds.lua
```diff
@@ -0,0 +1,112 @@
+local autocommands = {}
```
abzcoding, line 1: what is this variable used for?

lua/core/autocmds.lua
```diff
@@ -0,0 +1,112 @@
...
+      vim.cmd(command)
+    end
+
+    vim.cmd "augroup END"
+  end
+end
+
+
+
+return autocommands
```
abzcoding, line 112: and why are we returning a `{}`

lua/core/autocmds.lua
kylo252: It's used to return the `define_augroups` function that we need to call after `lvim.autocommands` has been updated by the user. For now it's called from `init.lua`

abzcoding (approved):

lua/core/autocmds.lua
kylo252: It's just called `autocommands` instead of the usual `M={}` that we use.

PR #1093, opened 2021-07-25
[Bugfix] Show correct indentation settings in galaxyline

etienne-k:
Display "Tab size: " instead of "Spaces: " if indenting with tabs.

# Description

In the galaxyline, LunarVim displays a text like "Spaces: _n_", where _n_ is the amount of spaces the user indents with, extracted from `shiftwidth`.
However, if a user indents with tabs, LunarVim still displays "Spaces: _n_", which can be very misleading.

This fix additionally checks the `expandtab` option and changes sets the text to "Tab size: _n_" if `expandtab` is `false`.

Just like the check on `shiftwidth`, I check the option for the current buffer.

Fixes #(issue)

## How Has This Been Tested?

Alternatingly set different values for the following options and confirm the expected behavior:
`set tabstop=4`
`set shiftwidth=4`
`set expandtab` / `set noexpandtab`

### DIFF
--- a/lua/core/galaxyline.lua
+++ b/lua/core/galaxyline.lua
@@ -266,7 +266,11 @@ table.insert(gls.right, {
 table.insert(gls.right, {
   Tabstop = {
     provider = function()
-      return "Spaces: " .. vim.api.nvim_buf_get_option(0, "shiftwidth") .. " "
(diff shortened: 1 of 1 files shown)


rebuilt (approved):
Tested.  It works.

PR #1094, opened 2021-07-25
Some user mappings not working due to timeoutlen=100

qvieth:
Changing timeoutlen from 100 to 300 fixes the issue

### DIFF
--- a/lua/settings.lua
+++ b/lua/settings.lua
@@ -27,7 +27,7 @@ local default_options = {
   splitright = true, -- force all vertical splits to go to the right of current window
   swapfile = false, -- creates a swapfile
   termguicolors = true, -- set term gui colors (most terminals support this)
-  timeoutlen = 100, -- time to wait for a mapped sequence to complete (in milliseconds)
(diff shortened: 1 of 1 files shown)


rebuilt (approved):
Yeah, timeout length is also tied to the speed whichkey opens.  There is a tug of war with this setting.  I think 300 is a more sane default.   Whichkey still feels quick enough.   But 300 it's still too quick for more complex mappings, e.g. vim-surround.   There are 16 closed issues related to timeoutlen and 4 pull requests surrounding it.  One was my own where I increased it to 200.  I'll approve and Chris can decide.  There are enough issues surrounding timeoutlen that I think it warrants another look.

rebuilt:
For anyone else reading this, you can still set it in your own config with
```
vim.cmd("set timeoutlen=500")

```

qvieth:
Yes!! I think sacrificing a little speed of whichkey for user mappings compatibility would worth it (also, I think some users would refer a little delay in whichkey's show up time). This would reduce the number of users who have to figure out why are some mappings not being recognized. (I was about opening an issue why my keys aren't working but make a lucky guess about timeoutlen, and it works)(this was quite annoying)

PR #1097, opened 2021-07-25
splits up the settings file.  settings.lua gets required before lv-co…

rebuilt:
Splits up the settings file.  settings.lua remains required before lv-config.   toggleable-settings depends on lv-config settings and must be required after.

Re-enables these configuration options:
lvim.line_wrap_cursor_movement
lvim.transparent_window

Fixes #(issue)
https://github.com/ChristianChiarulli/LunarVim/issues/1088

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -12,12 +12,13 @@ vim.cmd [[
 ]]
 -- vim.opt.rtp:append() instead of vim.cmd ?
 require "default-config"
-require "settings"
+require("settings").options()
 local status_ok, error = pcall(vim.cmd, "luafile ~/.config/lvim/lv-config.lua")
 if not status_ok then
   print "something is wrong with your lv-config"
   print(error)
 end
+require("settings").commands()
 require("core.autocmds").define_augroups(lvim.autocommands)
 
 require "keymappings"
--- a/init.lua
+++ b/init.lua
@@ -12,13 +12,13 @@ vim.cmd [[
 ]]
 -- vim.opt.rtp:append() instead of vim.cmd ?
 require "default-config"
-require("settings").options()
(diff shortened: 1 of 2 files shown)


ChristianChiarulli:
we could split settings into functions and call at different times?

ichigo-gyuunyuu:
The pattern followed so far is first load `lv-config`, then in all other places we read the options in `lv-config` and proceed accordingly. If we wanted to follow that pattern, we need to expose an lvim setting to override `vim.opt` stuff.

rebuilt:
I'll rewite it as functions so they can be called when needed.

rebuilt:
> The pattern followed so far is first load `lv-config`, then in all other places we read the options in `lv-config` and proceed accordingly. If we wanted to follow that pattern, we need to expose an lvim setting to override `vim.opt` stuff.

That pattern is what we used to do but it got confusing for people.  It was sort of arbitrary what things couldn't be changed with a lunarvim global.  The choice was either to duplicate all settings or none.

rebuilt:
Settings file now has this format

```
local M = {}

M.options = function()
  -- vim.opt related things
end

M.commands = function()
 -- vim.cmd related things
end

return M
```

Tastyep (approved):
Looks good :+1: 

I just have a minor style suggestion about the method's name, IMO names like `options()` or `commands()` should be used for functions returning something, here I would use a verb.
For example, `load_options()` and `load_commands()`, what do you think ?

rebuilt:
>  For example, `load_options()` and `load_commands()`, what do you think ?

Yeah.  That works.  I changed the names of the functions.

abzcoding (approved):

PR #1111, opened 2021-07-26
BufWritePre enables format on save for lv-config.lua

rebuilt:
# Description
Currently, lv-config.lua does not get autoformatted. Changes the autocommand from `BufWritePost` to `BufWritePre`

## How Has This Been Tested?
Open lv-config.lua
set `lvim.format_on_save = true`
Save lv-config.lua.  Autoformat should be enabled

### DIFF
--- a/lua/lv-utils/init.lua
+++ b/lua/lv-utils/init.lua
@@ -71,7 +71,7 @@ local toggle_autoformat = function()
   end
 
   if not lvim.format_on_save then
-    vim.cmd [[if exists('#autoformat#BufWritePost')
+    vim.cmd [[if exists('#autoformat#BufWritePre')
   :autocmd! autoformat
   endif]]
   end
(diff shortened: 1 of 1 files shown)


Tastyep (approved):

abzcoding:
doing it this way was causing neovim to ask you to select a formatter on each save
is that fixed?

rebuilt:
> doing it this way was causing neovim to ask you to select a formatter on each save
> is that fixed?

That was apparently a problem before switching to BufWritePre.  Hatsu was working on that problem.  It must have gotten resolved because I no longer experience it.

abzcoding (approved):

abzcoding (changes requested):
line 74 needs to change as well
from this
```lua
    vim.cmd [[if exists('#autoformat#BufWritePost')
```
to this:
```lua
    vim.cmd [[if exists('#autoformat#BufWritePre')
```

rebuilt:
I tested python, c, c++, ruby, javascript, and typescript.  Autoformatting continues to work.  No errors are thrown

abzcoding (approved):
LGTM :rocket:

PR #1117, opened 2021-07-26
[Bugfix] Add LSP fields for `tex` files

ichigo-gyuunyuu:
# Description

lvim would complain on opening `.tex` files as there were no `lang` fields for them. This adds the required formatter, linter and lsp path for tex files

## How Has This Been Tested?

Verify that `texlab` attaches on opening a `.tex` file and that formatting and linting works.


ichigo-gyuunyuu:
small github thing, not sure why the commits when i fetch upstream on my fork are added as well. is there anyway to avoid that?

PR #1128, opened 2021-07-27
[Bugfix] Wrap lsp default_keybindings in function instead of load on 1st require

devtoi:
# Description

I noticed an issue where you could no longer set `lvim.lsp.default_keybinds = false`. This is because the lsp keybindings were set up on the first require of that module. Which happens before reading the user lv-config.lua. I simply wrapped it in a function instead and called it after keymappings.

## How Has This Been Tested?
- Set `lvim.lsp.default_keybinds = false` in `~/.config/lvim/lv-config.lua`.
- Open a file where an LSP would run.
- Press tab in normal mode on something you expect to have signature help.
- Nothing happens because the binding no longer exist.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -22,6 +22,7 @@ require("settings").load_commands()
 require("core.autocmds").define_augroups(lvim.autocommands)
 
 require "keymappings"
+require("lsp").setup_default_bindings()
 
 local plugins = require "plugins"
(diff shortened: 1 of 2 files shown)


rebuilt (approved):

rebuilt:
Tested.  Fix confirmed.

PR #1130, opened 2021-07-27
Conditionally enable TAB to trigger completion based on filetype. Disable completion for simple text files

rebuilt:
# Description
Currently, hitting TAB triggers autocomplete in all files.  In simple text files hitting TAB should insert a tab character.  
This pr allows a user to specify filetypes in which they don't want Compe mappings.
By default this pr excludes markdown and txt files.  
The list of filetypes can be specified with

``` 
lvim.builtin.compe.exclude_filetypes = { "md", "markdown", "mdown", "mkd", "mkdn", "mdwn", "text", "txt" },
```

Fixes #(issue)
https://github.com/ChristianChiarulli/LunarVim/issues/1121

## How Has This Been Tested?
Editing markdown files and making sure a tab character is inserted. 
Editing other files to make sure compe behavior remains the same

### DIFF
--- a/lua/core/autocmds.lua
+++ b/lua/core/autocmds.lua
@@ -54,6 +54,10 @@ lvim.autocommands = {
   _markdown = {
     { "FileType", "markdown", "setlocal wrap" },
     { "FileType", "markdown", "setlocal spell" },
+    { "FileType", "markdown", "map <Tab> <Tab>" },
+    { "FileType", "markdown", "map <S-Tab> <S-Tab>" },
+    { "FileType", "markdown", "imap <Tab> <Tab>" },
+    { "FileType", "markdown", "imap <S-Tab> <S-Tab>" },
   },
   _buffer_bindings = {
     { "FileType", "floaterm", "nnoremap <silent> <buffer> q :q<CR>" },
--- a/lua/core/autocmds.lua
+++ b/lua/core/autocmds.lua
@@ -54,12 +54,9 @@ lvim.autocommands = {
   _markdown = {
     { "FileType", "markdown", "setlocal wrap" },
     { "FileType", "markdown", "setlocal spell" },
-    { "FileType", "markdown", "imap <Tab> <Tab>" },
-    { "FileType", "markdown", "imap <S-Tab> <S-Tab>" },
-    { "FileType", "md", "setlocal wrap" },
-    { "FileType", "md", "setlocal spell" },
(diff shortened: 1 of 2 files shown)


rebuilt:
Control+space will still trigger completion.   This is useful for autocompleting emoji in text files.  (type dash `-` and hit control+space to cycle through emoji)

abzcoding:
well Chris just removed that `tab` mapping altogether  😅

rebuilt:
This pr applies to the `TAB` mapping for insert mode, not normal mode.

abzcoding:
> This pr applies to the `TAB` mapping for insert mode, not normal mode.

yep noticed my mistake 🤦 and deleted the comment 😅

etienne-k:
Many thanks for the PR and sorry for replying just now.

I fear this change comes at a cost:

- In non-source code text files, autocomplete still makes some sense (considering the completed words were typed at least once already). In certain languages, certain words are long to type, so the autocomplete is always a welcome feature.
  - This is why other modern editors also provide autocomplete in simple text files.
  - I haven't had the time to test yet, but looking at the change, it seems like we _completely disabled autocomplete for simple text files_.

- The behavior described in the linked issue mentioned a use case for Markdown only, but I realize this does not just affect simple text files.
  - Example: I need the same tab-inserting behavior when writing comments in source files, where I sometimes write them in Markdown format.
  - Again, I couldn't test yet, but it seems like the described _issue is present in source files as well_.

What do you think? Do you share my point of view? Should I open a new issue for this?

rebuilt:
Compe-enabled autocomplete was not removed entirely, just the `TAB` keymap.  To trigger autocomplete press `Control`+`Space`.  

Add this to your lv-config.lua to opt out entirely.  
```
lvim.builtin.compe.exclude_filetypes = {  }
```

etienne-k:
You're right about the text files, autocomplete still works there - triggered automatically or with ctrl+space.

But the issue still exists with source files -> It's impossible to enter a tab/spaces after a dash character:
test

rebuilt:
That's on purpose.  In source files you still trigger autocomplete with the `TAB` key.   If you want to insert a tab character in those files, add it to the exclude list.

```
lvim.builtin.compe.exclude_filetypes = { "c", "cpp","md", "markdown", "mdown", "mkd", "mkdn", "mdwn", "text", "txt" },
```

PR #1132, opened 2021-07-27
feat: Add pre-commit hook for linting and formatting

kylo252:
# Description

Add pre-commit hook for linting and formatting that can be version controlled to make contributors' lives a bit easier :)
Powered by: https://pre-commit.com/

## How Has This Been Tested?

```shell
pip install --user pre-commit
cd /path/to/repo
pre-commit install
```
it should not let you commit something with a linter/formatter violation.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -22,17 +22,13 @@ jobs:
       
       - name: Run luacheck
         run: luacheck *.lua lua/
-    
-  shell-linter:
-    name: "Linting with shellcheck"
-    runs-on: ubuntu-20.04
+
+  shellcheck:
+    name: Shellcheck
+    runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      
-      - name: Use shellcheck
-        run: sudo apt-get install shellcheck
-      
-      - name: Run shellcheck
-        run:  |
-          pwd
-          shellcheck $(find . -name "*.sh")
+    - uses: actions/checkout@v2
+    - name: Run ShellCheck
+      uses: ludeeus/action-shellcheck@master
+      with:
+       scandir: './utils'
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,29 @@
+repos:
+  - repo: local
+    hooks:
+      - id: shfmt
+        name: shfmt
+        minimum_pre_commit_version: 2.4.0
+        language: golang
+        additional_dependencies: [mvdan.cc/sh/v3/cmd/shfmt@v3.2.2]
+        entry: shfmt
+        args: [-i=2, -ci, -w]
+        types: [shell]
+      - id: shellcheck
+        name: shellcheck
+        language: system
+        types: [shell]
+        entry: bash
+        args: [-c, "shfmt -f $(git rev-parse --show-toplevel) | xargs shellcheck"]
+      - id: stylua
+        name: StyLua
+        language: rust
+        entry: stylua
+        types: [lua]
+        args : [.]
+      - id: luacheck
+        name: luacheck
+        language: system
+        entry: luacheck
+        types: [lua]
+        args : [.]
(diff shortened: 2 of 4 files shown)


utils/bin/jdtls
```diff
@@ -41,20 +41,20 @@ fi
...
 GRADLE_HOME=$HOME/gradle "$JAVACMD" \
-	-Declipse.application=org.eclipse.jdt.ls.core.id1 \
-	-Dosgi.bundles.defaultStartLevel=4 \
-	-Declipse.product=org.eclipse.jdt.ls.core.product \
-	-Dlog.protocol=true \
-	-Dlog.level=ALL \
-	-javaagent:$HOME/.local/share/nvim/lspinstall/java/lombok.jar \
-	-Xms1g \
-	-Xmx2G \
-	-jar $(echo "$JAR") \
```
Tastyep, line 52: @ChristianChiarulli Tagging you because I know I'm guilty of having done the same change in the past, and it seems that it breaks things for you.
Does it need to expand into multiple strings ?

ChristianChiarulli: I'm not exactly sure what caused the break, but until we can confirm this doesn't break things in the future I think we should exclude this file somehow.

.github/workflows/lint.yaml
```diff
@@ -29,10 +29,17 @@ jobs:
...
+      - name: install dependencies
+        run: |
+          sudo apt-get install shellcheck
+          GO111MODULE=on go get mvdan.cc/sh/v3/cmd/shfmt
       
       - name: Run shellcheck
         run:  |
           pwd
-          shellcheck $(find . -name "*.sh")
+          shfmt -f "$(git rev-parse --show-toplevel)" | xargs shellcheck
```
Tastyep, line 45: You will probably have to revert it to `$(find . -name "*.sh")` since shfmt checks the shebang and includes `jdtls`

kylo252: I don't believe that's a good solution. I can disable the rule for shellcheck, but the whole point of it is to complain about potential issues. I can at least replace `which` with `command -v`.

Tastyep: I'm not the one you need to convince ;)

utils/bin/jdtls
kylo252: is this okay now? I've added a directive ` # shellcheck disable=SC2116`. And fixed the other ones, since they were pretty straight-forward.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,29 @@
+repos:
+  - repo: local
+    hooks:
+      - id: shfmt
+        name: shfmt
+        minimum_pre_commit_version: 2.4.0
+        language: golang
```
abzcoding, line 7: `golang` ?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,29 @@
...
+        additional_dependencies: [mvdan.cc/sh/v3/cmd/shfmt@v3.2.2]
+        entry: shfmt
+        args: [-i=2, -ci, -w]
+        types: [shell]
+      - id: shellcheck
```
Tastyep, line 12: Maybe some newlines between each id to mark a clear separation :) ?

utils/bin/jdtls
```diff
@@ -46,10 +48,10 @@ GRADLE_HOME=$HOME/gradle "$JAVACMD" \
   -Declipse.product=org.eclipse.jdt.ls.core.product \
   -Dlog.protocol=true \
   -Dlog.level=ALL \
-  -javaagent:$HOME/.local/share/nvim/lspinstall/java/lombok.jar \
+  -javaagent:"$HOME/.local/share/nvim/lspinstall/java/lombok.jar" \
   -Xms1g \
   -Xmx2G \
-  -jar $(echo "$JAR") \
+  -jar "$(echo "$JAR")" \
```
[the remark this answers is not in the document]
kylo252: > Does putting it on top disables it globally ?

Yes it disables for the entire file, so that should be fine for now.

.pre-commit-config.yaml
kylo252: It's now been formatted according to `prettier` :)
We can iterate on this file in the feature once everybody has used it a bit more.

PR #1133, opened 2021-07-27
include `package-lock.json` & `package.json` in root patterns

rafamadriz:
* Just adds package.json and package-lock.json for the root patterns of
  vim-rooter

* The other ones are the default patterns, see `:help g:rooter_patterns`

### DIFF
--- a/lua/plugins.lua
+++ b/lua/plugins.lua
@@ -140,6 +140,16 @@ return {
     "airblade/vim-rooter",
     config = function()
       vim.g.rooter_silent_chdir = 1
+      vim.g.rooter_patterns = {
+        ".git",
+        "_darcs",
+        ".hg",
+        ".bzr",
+        ".svn",
+        "Makefile",
+        "package.json",
+        "package-lock.json",
+      }
       if lvim.builtin.rooter.on_config_done then
         lvim.builtin.rooter.on_config_done()
       end


ChristianChiarulli:
is there any way to append this rather than override?

rafamadriz:
> is there any way to append this rather than override?

By looking at the vim-rooter README and `help` it doesn't seem like that's how it works, that's why I added all the defaults ones just in case.

rafamadriz:
I just did a test and just adding package.json and package-lock.json doesn't append, but overrides the whole thing and leaves  package.json and package-lock.json as the only root patterns. The keep the default patterns they need to be added too.

ChristianChiarulli:
do you think it would be possible to upstream this?

rafamadriz:
> do you think it would be possible to upstream this?

Not sure if I understand, do you mean to send a PR to vim-rooter to accept appending to the `let g:rooter_patterns` variable ?

ChristianChiarulli:
either that or add package.json to the default roots in vimrooter

rafamadriz:
> add package.json to the default roots in vimrooter

Oh I see that makes more sense, yeah sure I can send a PR. For now I think is save to merge this for node projects that don't have a `git` folder, until the PR to vim-rooter is merged, if it is ever accepted.

PR #1141, opened 2021-07-28
null-ls formatter name is different from its command

abzcoding:
# Description

1. null ls formatter name is different from the actual command
for example `clang_format` is the `null-ls` formatter but `clang-format` is the actual command so we need to check for the existence of the command, not the formatter

2. show formatter inside `galaxyline`, otherwise, we wouldn't know if `null-ls` is activated or not 😅 

3. a couple of langs are supported natively by `null-ls` so I've added them

## How has this been tested?

1. Save this as `Program.cs`
```cs
using System;
using System.Collections.Generic;
using System.Linq;
using System.ServiceProcess;
using System.Text;
using System.Threading.Tasks;

namespace OWINTest.Service
{
    static class Program
    {
        /// <summary>
        /// The main entry point for the application.
        /// </summary>
        static void Main()
        {
            ServiceBase[] ServicesToRun;
            ServicesToRun = new ServiceBase[]
            {
                new APIServiceTest()
            };
            ServiceBase.Run(ServicesToRun);
        }
    }
}
```

2. install `clang-format`
3. remove `omnisharp` lsp if you have it installed
4. before my PR if you open and try to format it, nothing would change
5. but after using my PR it would format it

the reason is that it looks for `clang_format` but it should look for `clang-format`

### DIFF
--- a/lua/default-config.lua
+++ b/lua/default-config.lua
@@ -119,7 +119,7 @@ lvim.lang = {
   },
   cs = {
     formatter = {
-      exe = "",
+      exe = "clang_format",
       args = {},
     },
     linters = {},
@@ -139,7 +139,7 @@ lvim.lang = {
   },
   cmake = {
     formatter = {
-      exe = "clang_format",
+      exe = "cmake_format",
       args = {},
     },
     linters = {},
@@ -194,8 +194,8 @@ lvim.lang = {
   },
   dart = {
     formatter = {
-      exe = "dart",
-      args = { "format" },
+      exe = "dart_format",
+      args = {},
       stdin = true,
     },
     linters = {},
@@ -233,7 +233,7 @@ lvim.lang = {
   elixir = {
     formatter = {
       exe = "mix",
-      args = { "format" },
+      args = {},
       stdin = true,
     },
     linters = {},
@@ -250,7 +250,7 @@ lvim.lang = {
   },
   elm = {
     formatter = {
-      exe = "",
+      exe = "elm_format",
       args = {},
       stdin = true,
     },
@@ -273,7 +273,7 @@ lvim.lang = {
   },
   erlang = {
     formatter = {
-      exe = "",
+      exe = "erlfmt",
       args = {},
     },
     linters = {},
@@ -291,7 +291,7 @@ lvim.lang = {
   emmet = { active = false },
   fish = {
     formatter = {
-      exe = "",
+      exe = "fish_indent",
       args = {},
     },
     linters = {},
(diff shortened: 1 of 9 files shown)


abzcoding:
also if lsp supports formatting, we should not use null-ls, not sure if this is the case rn, or de need to do sth about it

rebuilt:
Yeah, it's very unclear to me what takes precedence.

lua/lsp/null-ls.lua
```diff
@@ -43,7 +43,7 @@ local function setup_ls(exe, type)
       end
     end
   else
-    if vim.fn.executable(exe) == 1 then
+    if null_ls.builtins[type][exe] ~= nil and vim.fn.executable(null_ls.builtins[type][exe]._opts.command) then
```
Tastyep, line 46: `~= nil` isn't necessary

Tastyep (approved):
Did you add all the formatters you defined in you PR for null-ls ?

abzcoding:
> Did you add all the formatters you defined in you PR for null-ls ?

actually no 😅 i should probably do that

lua/default-config.lua
```diff
@@ -117,9 +143,24 @@ lvim.lang = {
       },
     },
   },
+  crystal = {
```
Tastyep, line 146: Will this work without the corresponding ftplugin file ?

abzcoding: we have the ftplugin file though?

Tastyep: I'm blind, sorry :laughing:

rebuilt:
It's not formatting for me.  Clang-format is installed.  Omnisharp is not.   Writing the file has no effect.  Null-ls doesn't seem to be attached to the buffer.  

Passing the format command to the command line works `:!clang-format -i %`

clangformat

I'm on the latest push 
```
bc26f83 - (HEAD -> null-ls) nil is unnecessary (3 hours ago) <abzcoding>
```

rebuilt:
I installed this:   https://www.npmjs.com/package/clang-format

Is that the right clang-format?

Tastyep:
> I installed this: https://www.npmjs.com/package/clang-format
> 
> Is that the right clang-format?

clang-format is usually packaged with llvm, but if you can run `clang-format` in your shell or vim this should work.
What's weird is that you have to lsp active

rebuilt:
>  What's weird is that you have no active lsp

Yeah.  This is true.  `.cs` and  `.jl` don't register any formatters for me. It's true on rolling and in this pr as well.  Other languages report formatters as expected: ruby, javascript, typescript, c, cpp, latex, lua

rebuilt:
I'll try installing llvm

rebuilt:
Installing llvm worked. 


clang_format

PR #1146, opened 2021-07-28
[Refactor] Cleanup lsp init file

Tastyep:
# Description

Split `lsp/init.lua` into configurations (`lvim` variable and keymappings), setup (`.config()` and `.setup()`), and services (`lsp/service.lua`).
Also removed ~80 LOC by reworking some code.

## How Has This Been Tested?

By starting lvim and checking the status bar to see that formatters and linters were still there.


ChristianChiarulli:
I just noticed this, this file definitely needed to be cleaned up.

Tastyep:
Well, appparently that means that I need to repeat the work ...

PR #1147, opened 2021-07-28
feat: add diagnostics source name

kylo252:
# Description

- Add diagnostics' source name and to help distinguish where it's coming from.
- Add diagonstics' code to make it easier for the user to turn it off if the provider supports that.

## Fixes

`setup_handlers()` from lsp wasn't getting called also.

## How Has This Been Tested?

image

### DIFF
--- a/lua/default-config.lua
+++ b/lua/default-config.lua
@@ -41,6 +41,7 @@ lvim = {
       },
       signs = true,
       underline = true,
+      severity_sort = true,
     },
     override = {},
     document_highlight = true,
--- a/lua/lsp/handlers.lua
+++ b/lua/lsp/handlers.lua
@@ -9,6 +9,40 @@ function M.setup()
     underline = lvim.lsp.document_highlight,
   })
 
+  vim.lsp.handlers["textDocument/publishDiagnostics"] = function(_, _, params, client_id, _)
+    local config = { -- your config
+      virtual_text = lvim.lsp.diagnostics.virtual_text,
+      signs = lvim.lsp.diagnostics.signs,
+      underline = lvim.lsp.diagnostics.underline,
+      update_in_insert = lvim.lsp.diagnostics.update_in_insert,
+      severity_sort = lvim.lsp.diagnostics.severity_sort,
+    }
+    local uri = params.uri
+    local bufnr = vim.uri_to_bufnr(uri)
+
+    if not bufnr then
+      return
+    end
+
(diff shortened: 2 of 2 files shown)


lua/lsp/init.lua
```diff
@@ -208,6 +231,7 @@ function lsp_config.common_on_attach(client, bufnr)
     lvim.lsp.on_attach_callback(client, bufnr)
   end
   lsp_highlight_document(client)
+  lsp_config.setup_handlers()
```
Tastyep, line 234: `setup_handlers` is already called from init.lua

lua/lsp/init.lua
```diff
@@ -65,11 +65,34 @@ end
...
+    local bufnr = vim.uri_to_bufnr(uri)
+
+    if not bufnr then
+      return
+    end
+
+    local diagnostics = params.diagnostics
+
+    for i, v in ipairs(diagnostics) do
+      diagnostics[i].message = string.format("%s: %s", v.source, v.message)
```
Tastyep, line 85: Have you considered adding the code too ? Maybe smth like `"%s [%s]: %s"`
See [here](https://github.com/jose-elias-alvarez/null-ls.nvim/pull/57/files#diff-c2c076648e1b1e04ff69ac1030444a8ffbfda480d12fe8771b719a77cba92d7fR142) for examples of code values.

Tastyep (changes requested):
Looks nice :+1:

ChristianChiarulli:
there is a dedicated file for handlers now, could you use that, this looks like a really useful feature.

kylo252:
> there is a dedicated file for handlers now, could you use that, this looks like a really useful feature.

I'll fix this once #1078 is merged.

ChristianChiarulli:
May need a slight refactor now, but looks good.

lua/lsp/init.lua
kylo252: > Have you considered adding the code too ? Maybe smth like `"%s [%s]: %s"`
> See [here](https://github.com/jose-elias-alvarez/null-ls.nvim/pull/57/files#diff-c2c076648e1b1e04ff69ac1030444a8ffbfda480d12fe8771b719a77cba92d7fR142) for examples of code values.

I'll try to see what I can come up with.
Thanks for all the amazing work for null-ls btw :)

Tastyep (approved):
Looks great :+1:

PR #1152, opened 2021-07-29
[Feature]: Add path on status line galaxyline

deryrahman:
# Description

Just small changes, adding the path status :)

Before:
Screen Shot 2021-07-29 at 11 24 30

After:
Screen Shot 2021-07-29 at 11 25 22

## How Has This Been Tested?

- Open lvim for specific file path:
    - `/example.txt` will be printed as `/example.txt`
    - `/tmp/example.txt` will be printed as `/tmp/example.txt`
    - `/tmp/files/path/example.txt` will be printed as `../path/example.txt`

### DIFF
--- a/lua/core/galaxyline.lua
+++ b/lua/core/galaxyline.lua
@@ -155,6 +155,37 @@ table.insert(gls.left, {
   },
 })
 
+table.insert(gls.left, {
+  FilePath = {
+    provider = function()
+      -- reduce longer path
+      -- eg. path/foo/bar/file.html -> ../bar/file.html
+      local path_div = function(path)
+        if path == nil then
+          return path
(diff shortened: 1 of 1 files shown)


Tastyep:
Just being curious, in what situation do you need this ?
Personally, when I need to see what file I'm on, I'm checking the tab name or hitting <space>e to open the tree.

deryrahman:
@Tastyep oh never mind, I've just seen the tab is also showing the path folder name when there're 2 files with the same name under different folders. My hypothesis was wrong, I thought the tab name would have the same name. 😛 
The intention is just to tell the user which file we're currently open.

Close this PR since it doesn't solve any issue.

PR #1153, opened 2021-07-29
[Feature]: Add support for saving a file without formatting

sethigeet:
# Description

Now you can save a file without formatting even if format on save is enabled
You no longer have to go and disable the format on save option if you want to save a file without formatting just once!

Also, if you have format on save disabled, now you can save a file and also format it at the same time with the new command!

# Changes

I recommend that you go through the changes once but here's a summary:
 - Added `SaveWithoutFormatting`, `SaveWithFormatting`, and `SaveFile` commands
 - Added `SaveWithoutFormattingAndQuit`, `SaveWithFormattingAndQuit`, and `SaveFileAndQuit` commands
 - Changed the built in vim command `w`, and `wq` to use these new commands

# What do the new commands do?

 - `SaveWithoutFormatting`: Save the file without formatting
 - `SaveWithFormatting`: Save the file and format it
 - `SaveFile`: Save the file but respect the `format_on_save` option from the user's config file
 - The other three commands do the same actions but also quit after that!

## How Has This Been Tested?

Just try to do save a file without formatting using vim's built in commands or the new ones that I have created or just use the provided keybindings(defined in `which-key`)

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -30,8 +30,6 @@ plugin_loader:load { plugins, lvim.plugins }
 vim.g.colors_name = lvim.colorscheme -- Colorscheme must get called after plugins are loaded or it will break new installs.
 vim.cmd("colorscheme " .. lvim.colorscheme)
 
-local utils = require "utils"
-utils.toggle_autoformat()
 local commands = require "core.commands"
 commands.load(commands.defaults)
 
--- a/lua/core/commands.lua
+++ b/lua/core/commands.lua
@@ -11,18 +11,19 @@ M.defaults = {
   endfunction
   ]],
 
+  -- NOTE: Make sure that `<q-args>` is quoted with single quotes only or there will be errors!
   -- Save files
   -- Save file without formatting the file
-  [[command! -bang -bar SaveWithoutFormatting lua require 'utils'.save_file(false, "<bang>" == "!")]],
+  [[command! -nargs=* -bang -bar SaveWithoutFormatting lua require 'utils'.save_file(false, "<bang>" == "!", nil, <q-args>)]],
   -- Save file and also format the file
-  [[command! -bang -bar SaveWithFormatting  lua require 'utils'.save_file(true, "<bang>" == "!")]],
+  [[command! -nargs=* -bang -bar SaveWithFormatting  lua require 'utils'.save_file(true, "<bang>" == "!", nil, <q-args>)]],
   -- Save file but respect auto format settings
-  [[command! -bang -bar SaveFile lua require 'utils'.save_file(nil, "<bang>" == "!")]],
+  [[command! -nargs=* -bang -bar SaveFile lua require 'utils'.save_file(nil, "<bang>" == "!", nil, <q-args>)]],
 
   -- Like above command but also quit immediately
-  [[command! -bang -bar SaveWithoutFormattingAndQuit lua require 'utils'.save_file(false, "<bang>" == "!", true)]],
-  [[command! -bang -bar SaveWithFormattingAndQuit  lua require 'utils'.save_file(true, "<bang>" == "!", true)]],
-  [[command! -bang -bar SaveFileAndQuit lua require 'utils'.save_file(nil, "<bang>" == "!", true)]],
+  [[command! -nargs=* -bang -bar SaveWithoutFormattingAndQuit lua require 'utils'.save_file(false, "<bang>" == "!", true, <q-args>)]],
+  [[command! -nargs=* -bang -bar SaveWithFormattingAndQuit  lua require 'utils'.save_file(true, "<bang>" == "!", true, <q-args>)]],
+  [[command! -nargs=* -bang -bar SaveFileAndQuit lua require 'utils'.save_file(nil, "<bang>" == "!", true, <q-args>)]],
 
   -- Map vim commands to our new commands
   [[cnoreabbrev w SaveFile]],
--- a/lua/core/commands.lua
+++ b/lua/core/commands.lua
@@ -10,6 +10,25 @@ M.defaults = {
     endif
   endfunction
   ]],
+
+  -- Save files
+  -- Save file without formatting the file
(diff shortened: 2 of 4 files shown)


init.lua
```diff
@@ -31,7 +31,6 @@ vim.g.colors_name = lvim.colorscheme -- Colorscheme must get called after plugin
 vim.cmd("colorscheme " .. lvim.colorscheme)
 
 local utils = require "utils"
```
abzcoding, line 33: `unused variable utils`

sethigeet: Done!

lua/core/commands.lua
```diff
@@ -10,6 +10,25 @@ M.defaults = {
...
+  -- Save file but respect auto format settings
+  [[command! -bang -bar SaveFile lua require 'utils'.save_file(nil, "<bang>" == "!")]],
+
+  -- Like above command but also quit immediately
+  [[command! -bang -bar SaveWithoutFormattingAndQuit lua require 'utils'.save_file(false, "<bang>" == "!", true)]],
+  [[command! -bang -bar SaveWithFormattingAndQuit  lua require 'utils'.save_file(true, "<bang>" == "!", true)]],
+  [[command! -bang -bar SaveFileAndQuit lua require 'utils'.save_file(nil, "<bang>" == "!", true)]],
+
+  -- Map vim commands to our new commands
+  [[cnoreabbrev w SaveFile]],
```
abzcoding, line 28: can  we not change the default neovim behavior, and only use
`<leader>w` or `<leader>q` or at least make this configurable?

abzcoding:
@ChristianChiarulli which one do you prefer?
#1153 or #898

lua/core/commands.lua
sethigeet: A lot of people still use `:w` to save the file! If we would not do this remap, then they would not be able to format the file on save using `:w` but will have to type `:SaveFile`

But I guess that's their problem for not using `<leader>w`

But I can add and option for the user so that they can decide whether they want to the remap to be there or not! Should I?

abzcoding: let chris and others comment on these, don't change anything yet :)

rebuilt:
The happy path works
```
    Added SaveWithoutFormatting, SaveWithFormatting, and SaveFile commands
    Added SaveWithoutFormattingAndQuit, SaveWithFormattingAndQuit, and SaveFileAndQuit commands
    Changed the built in vim command w, and wq to use these new commands
```

But I'm sorry to report this breaks renaming files and also breaks opening a new file and saving.

rebuilt:
newbuffer

rebuilt:
save_as

rebuilt:
It's also a bit jarring that hitting space after 'w' changes it to 'SaveFile`.  I've been using vim long enough that it feels like something is happening that I didn't intend.

sethigeet:
Thanks for pointing that out @rebuilt! I ad totally forgotten that you could pass in filenames to the write command. That has now been fixed!

About the other thing, I don't think that that can be fixed as that is how abbreviations work in (Neo)Vim. If you have a better idea to map the built in commands to the new ones, please tell me. But I do not think that that should be much of an issue because:
 - the user will only see that behavior when the press a space and want to put in a filename. When they just use `:w` and press enter they would not see that expansion taking place!
 - the user will still have to only type the old commands `:w`, `:wq`, etc and they automatically expand, so everything is happening automatically and the user does not have to do anything!
I think we can make such a minor change (which will be noticed very rarely) in order to add a very, very important functionality, save without formatting.

rebuilt:
Vim provides a different option, :update.  Abbreviated as :up

Can we override :up instead of :w?   It's a more obscure command.

Also, remember that you're overriding :w not just when the user types it in but also for any remappings to :w they've done

For instance this remap saves a file each time you exit insert mode

```
vim.api.nvim_set_keymap("i", "<Esc>", "<Esc>:w<CR>", { noremap = true, silent = true })

```

You would also need to reorder the following so :up doesn't autoclose 

```
  [[cnoreabbrev up SaveFileAndQuit]],
  [[cnoreabbrev up! SaveFileAndQuit!]],
  [[cnoreabbrev up SaveFile]],
  [[cnoreabbrev up! SaveFile!]],

```

rebuilt:
```
							*:up* *:update*
:[range]up[date][!] [++opt] [>>] [file]
			Like ":write", but only write when the buffer has been
			modified.

```

sethigeet:
I don't think that that is a possible solution as then the people who use `:w` to save the file would not be able to use the new functionality and would wonder why the file is not being formatted or vice versa

sethigeet:
The remaps made by the user would work if they do not set `noremap` to `true`. We can mention this in the docs!

```
// Would not work as `noremap = true`
vim.api.nvim_set_keymap("i", "<Esc>", "<Esc>:w<CR>", { noremap = true, silent = true })

// Would work as expected as `noremap = nil/false`
vim.api.nvim_set_keymap("i", "<Esc>", "<Esc>:w<CR>", { silent = true })
```

rebuilt (approved):
Ok.  I can get used to `w` getting changed to `SaveFile`. Users might be surprised by the alias quirk but it shouldn't break anything and the name of the command is clear. 

The change adds enough value to be included in LunarVim.

rebuilt:
We will probably have to document this behavior somewhere.  Maybe we should have a FAQ page on the documentation site?

sethigeet:
I think that we will have to add it to an FAQ page on the website eventually!
But I think we should add it to the wiki as soon as this PR merges and also mention it in the readme for a short period of time till everyone knows about it.

rebuilt:
After giving this more thought, I grow less and less comfortable hijacking the `:w` command.  It's possible there are no problems, but it's also possible there are edge cases where this breaks things for users.

kylo252:
Seeing as this outdated now, please feel free to open a new one based on #1937

PR #1156, opened 2021-07-29
[Feature] Expose lsp config

Tastyep:
# Description

@ChristianChiarulli

[PR-1146](https://github.com/ChristianChiarulli/LunarVim/pull/1146)

I removed utils functions because:
- `has_value` reimplements vim.tbl_contains
- `is_string` & `is_table` don't encapsulate any logic and make code more indirect.
https://stackoverflow.com/questions/47309651/should-you-use-a-function-for-only-one-line-of-code


Tastyep:
Rebased, there are quite a few things that need to be done:

- We should follow some style guidelines, I'll open an issue about that: https://github.com/luarocks/lua-style-guide
- When I saw that we had a logging method I was happy, then very sad when I got to see the implementation
- The code of null-ls.lua is unreadable, it looks more like a playground than production code, this is also very sad.

abzcoding:
> Rebased, there are quite a few things that need to be done:
> 
>     * We should follow some style guidelines, I'll open an issue about that: https://github.com/luarocks/lua-style-guide
> 
>     * When I saw that we had a logging method I was happy, then very sad when I got to see the implementation
> 
>     * The code of null-ls.lua is unreadable, it looks more like a playground than production code, this is also very sad.

i do agree with all of those
1. can you update [this](https://www.lunarvim.org/dev/) and add the resources/references so we all can learn from them.
2. yeah we need a better logging solution for debugging and stuff.
3. but you are fixing it 😈

abzcoding:
can you fix the conflicts plz, i just merged your other PR about keybindings 😢

PR #1161, opened 2021-07-29
[Bugfix] Disable dashboard when telescope is inactive

chaeing:
# Description

Disable Dashboard when Telescope is inactive since it depends on it. 

## How Has This Been Tested?

lv-config.lua
```
lvim.builtin.dashboard.active = true
lvim.builtin.telescope.active = false
```

### DIFF
--- a/lua/plugins.lua
+++ b/lua/plugins.lua
@@ -1,3 +1,8 @@
+if lvim.builtin.dashboard.active and not lvim.builtin.telescope.active then
+  lvim.builtin.dashboard.active = false
+  print "Telescope plugin must be active in order to use Dashboard"
+end
+
 return {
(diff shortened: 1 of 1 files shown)


abzcoding:
atm `lvim.builtin.telescope.active` doesn't do anything
you can check this in `lua/plugins.lua` line 23 or sth ( or using `:PackerStatus`)
the telescope plugins load either way

chaeing:
> atm `lvim.builtin.telescope.active` doesn't do anything (meaning you cannot disable telescope rn)
> you can check this in `lua/plugins.lua` line 23 or sth ( or using `:PackerStatus`)
> the telescope plugin load either way

That is right.
Closing it.

PR #1163, opened 2021-07-29
Refactor/remove logic from plugins

Tastyep:
# Description

This removes logic from plugins.lua (as it should be).
What stays in plugins.lua is all the hardcoded configuration liek the package name, events lazy loading the plugin, ect ...
The logic that's configurable by the user like the disable flag or the new callback provided once the plugin is loaded would be automatically handled by the plugin-loader.
Doing this we make sure that the plugin configuration is taken into account and minimize the risk of forgetting things while refactoring / implementing new features.
One [example](https://github.com/ChristianChiarulli/LunarVim/blob/0.5.1/lua/core/which-key.lua#L4) among probably many more.

This PR will stay a draft as long as [packer relies on strings.dump](https://github.com/wbthomason/packer.nvim/discussions/513)
I had this on a separate branch, but I think it's better to share it with everyone.

### DIFF
--- a/lua/core/autopairs.lua
+++ b/lua/core/autopairs.lua
@@ -1,51 +1,52 @@
--- if not package.loaded['nvim-autopairs'] then
---   return
--- end
-local status_ok, _ = pcall(require, "nvim-autopairs")
-if not status_ok then
-  return
-end
-local npairs = require "nvim-autopairs"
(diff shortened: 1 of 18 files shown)


kylo252:
Is this still useful @Tastyep?

PR #1164, opened 2021-07-29
[Feature]: Add gdscript LSP support

### LINKED ISSUE #358: Add LSP and Syntax Highlighting for GDScript
Currently there is no syntax highlighting or LSP for GDScript.
[this may be helpful](https://github.com/neovim/nvim-lspconfig/blob/master/CONFIG.md#gdscript)

Menchen:
# Description

Close #358 by adding LSP support for gdscript, **syntax highlighting is already supported by tree-sitter.**

LSP config reference: https://github.com/neovim/nvim-lspconfig/blob/master/CONFIG.md#gdscript

One small problem about using `nc` as executable is that **some** version of `nc` doesn't work.
In my testing the GNU `nc` installed with homebrew failed to run, but BSD `nc` bundled with macOs work just fine, also `ncat` from `nmap` work fine too.

Furthermore, windows user *may* want to use their own version of `nc` binary replacement.

To use custom `nc`, small hack in `lv-config.lua`:
```lua
lvim.lang.gdscript.lsp.setup.cmd[1] = "ncat"
```
or just fix path..

Fixes
#358

## How Has This Been Tested?

With Godot Engine opened, LspInfo show that it's attached and code completion is working.

### DIFF
--- a/lua/default-config.lua
+++ b/lua/default-config.lua
@@ -1000,6 +1000,22 @@ lvim.lang = {
       },
     },
   },
+  gdscript = {
+    formatter = {},
+    linters = {},
+    lsp = {
+      provider = "gdscript",
+      setup = {
+        cmd = {
+          "nc",
+          "localhost",
+          "6008",
+        },
+        on_attach = common_on_attach,
+        capabilities = common_capabilities,
+      },
+    },
+  },
 }
 
 require("core.which-key").config()
--- a/ftplugin/gdscript.lua
+++ b/ftplugin/gdscript.lua
@@ -1,0 +1,1 @@
+require("lsp").setup "gdscript"


lua/default-config.lua
```diff
@@ -1000,6 +1000,26 @@ lvim.lang = {
       },
     },
   },
+  gdscript = {
+    formatter = {
```
Tastyep, line 1004: Can this just be `formatter = {}` ?

lua/default-config.lua
```diff
@@ -1000,6 +1000,26 @@ lvim.lang = {
...
+      exe = "",
+      args = {},
+      stdin = false,
+    },
+    linters = {},
+    lsp = {
+      provider = "gdscript",
+      setup = {
+        cmd = {
```
Tastyep, line 1013: It would be better if you don't specify this as it's already defaulted by [lspconfig](https://github.com/neovim/nvim-lspconfig/blob/master/CONFIG.md#gdscript)

Tastyep (changes requested):

lua/default-config.lua
Menchen: I would like to let the user override the binary for `nc` using `lvim.lang.gdscript.lsp.setup.cmd[1] = "ncat"` as I stated above, also the user may want to use another port or IP instead of the default `6008` and `localhost` E.g when using Lunarvim inside of WSL2, the LSP server(Godot Engine in this case) is not `localhost`.

lua/default-config.lua
Menchen: Fixed.

lua/default-config.lua
Tastyep: You are right, it's more of a problem on our side since we force the values to be duplicated

Tastyep (approved):

PR #1170, opened 2021-07-30
Add custom root_dir to lv-config

### LINKED ISSUE #1169: [Feature]: Expose the ability to set root markers for the LSP
### Feature motivation

I work in a monorepo for my day job that uses Bazel files to delineate packages and services.  I had worked around this in previous versions of LunarVim by modifying the lspconfigs to include a bespoke list of `root_files`, however, the latest move to lvim has changed a lot.  Nlsp-settings allegedly allows config files to be written in lua rather than json but I have yet to get that feature to produce usable results.

A simpler solution would be to allow users to set a list of root markers for all lsps to allow the file system root to be found more easily.

A single variable exposed to lv-config that will contain a list of root markers.

Alternatives are directly modifying the lvim config files themselves.

### Additional context

_No response_

vlaetansky:
# Description

Fixes #1169

## How Has This Been Tested?

- uncomment both lines in config
- create a dir and execute git init
- open a .py file


kylo252:
Already handled now with LSP override: https://www.lunarvim.org/languages/#overriding-the-default-configuration

PR #1171, opened 2021-07-30
[Refactor] Make autopairs configuration more consistent

devtoi:
# Description

Autopairs configuration were run when module was loaded with require. This was inconsistent with other plugins. As well as not being able to disable the builtin plugin as with other ones. This makes it more consistent with other plugins and you can disable it with the expected variables.

## How Has This Been Tested?
Tested with `lvim.builtin.autopairs.active = false` and `lvim.builtin.autopairs.active = true` in my `~/.config/lvim/lv-config.lua` with a `:PackerSync` in between

### DIFF
--- a/lua/core/autopairs.lua
+++ b/lua/core/autopairs.lua
@@ -1,7 +1,7 @@
 local M = {}
 
 M.config = function()
-  lvim.builtin.dap = {
+  lvim.builtin.autopairs = {
     active = true,
   }
 end
--- a/lua/core/autopairs.lua
+++ b/lua/core/autopairs.lua
@@ -15,15 +15,10 @@ M.setup = function()
 
   vim.g.completion_confirm_key = ""
   MUtils.completion_confirm = function()
-    if vim.fn.pumvisible() ~= 0 then
-      if vim.fn.complete_info()["selected"] ~= -1 then
-        return vim.fn["compe#confirm"](npairs.esc "<cr>")
-      else
-        return npairs.esc "<cr>"
-      end
-    else
-      return npairs.autopairs_cr()
+    if vim.fn.complete_info()["selected"] ~= -1 then
+      return vim.fn["compe#confirm"](npairs.esc "<cr>")
     end
+    return npairs.esc "<cr>"
   end
 
   if package.loaded["compe"] then
--- a/lua/core/autopairs.lua
+++ b/lua/core/autopairs.lua
@@ -15,10 +15,13 @@ M.setup = function()
 
   vim.g.completion_confirm_key = ""
   MUtils.completion_confirm = function()
-    if vim.fn.complete_info()["selected"] ~= -1 then
-      return vim.fn["compe#confirm"](npairs.esc "<cr>")
+    if vim.fn.pumvisible() ~= 0 then
+      if vim.fn.complete_info()["selected"] ~= -1 then
+        return vim.fn["compe#confirm"](npairs.esc "<cr>")
+      end
+      return npairs.esc "<cr>"
     end
-    return npairs.esc "<cr>"
+    return npairs.autopairs_cr()
   end
 
   if package.loaded["compe"] then
--- a/lua/core/autopairs.lua
+++ b/lua/core/autopairs.lua
@@ -7,10 +7,6 @@ M.config = function()
 end
 
 M.setup = function()
-  local status_ok, _ = pcall(require, "nvim-autopairs")
-  if not status_ok then
-    return
-  end
   local npairs = require "nvim-autopairs"
   local Rule = require "nvim-autopairs.rule"
 
@@ -55,6 +51,8 @@ M.setup = function()
     Rule("%", "%", "lua"):with_pair(ts_conds.is_ts_node { "string", "comment" }),
     Rule("$", "$", "lua"):with_pair(ts_conds.is_not_ts_node { "function" }),
   }
+
+  return npairs
 end
 
 return M
--- a/lua/core/autopairs.lua
+++ b/lua/core/autopairs.lua
@@ -1,51 +1,60 @@
--- if not package.loaded['nvim-autopairs'] then
---   return
--- end
-local status_ok, _ = pcall(require, "nvim-autopairs")
-if not status_ok then
-  return
-end
-local npairs = require "nvim-autopairs"
-local Rule = require "nvim-autopairs.rule"
+local M = {}
 
--- skip it, if you use another global object
-_G.MUtils = {}
-
-vim.g.completion_confirm_key = ""
-MUtils.completion_confirm = function()
-  if vim.fn.pumvisible() ~= 0 then
-    if vim.fn.complete_info()["selected"] ~= -1 then
-      return vim.fn["compe#confirm"](npairs.esc "<cr>")
-    else
-      return npairs.esc "<cr>"
-    end
-  else
-    return npairs.autopairs_cr()
-  end
-end
-
-if package.loaded["compe"] then
-  require("nvim-autopairs.completion.compe").setup {
-    map_cr = true, --  map <CR> on insert mode
-    map_complete = true, -- it will auto insert `(` after select function or method item
+M.config = function()
+  lvim.builtin.dap = {
+    active = true,
   }
 end
 
-npairs.setup {
-  check_ts = true,
-  ts_config = {
-    lua = { "string" }, -- it will not add pair on that treesitter node
-    javascript = { "template_string" },
-    java = false, -- don't check treesitter on java
-  },
-}
+M.setup = function()
+  local status_ok, _ = pcall(require, "nvim-autopairs")
+  if not status_ok then
+    return
+  end
+  local npairs = require "nvim-autopairs"
+  local Rule = require "nvim-autopairs.rule"
 
-require("nvim-treesitter.configs").setup { autopairs = { enable = true } }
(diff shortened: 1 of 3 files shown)


lua/core/autopairs.lua
```diff
@@ -1,51 +1,60 @@
...
+    active = true,
+  }
 end
-local npairs = require "nvim-autopairs"
-local Rule = require "nvim-autopairs.rule"
 
--- skip it, if you use another global object
-_G.MUtils = {}
+M.setup = function()
+  local status_ok, _ = pcall(require, "nvim-autopairs")
```
Tastyep, line 10: Can you try removing the pcall, uninstalling nvim-autopairs and reinstalling it again (comment the entry in `plugins.lua`)
If that works without issues, then remove these lines

lua/core/autopairs.lua
```diff
@@ -1,51 +1,60 @@
...
-      return vim.fn["compe#confirm"](npairs.esc "<cr>")
+  -- skip it, if you use another global object
+  _G.MUtils = {}
+
+  vim.g.completion_confirm_key = ""
+  MUtils.completion_confirm = function()
+    if vim.fn.pumvisible() ~= 0 then
+      if vim.fn.complete_info()["selected"] ~= -1 then
+        return vim.fn["compe#confirm"](npairs.esc "<cr>")
+      else
```
Tastyep, line 25: I know it's not your code, but while at it, do you mind removing the else blocks after `return`
``` lua
      if vim.fn.complete_info()["selected"] ~= -1 then
        return vim.fn["compe#confirm"](npairs.esc "<cr>")
      end
      return npairs.esc "<cr>"
```

The same applies for L:29

Tastyep (changes requested):
Thanks a lot for making this change :+1: 
While this replicates work done by [PR-1011](https://github.com/ChristianChiarulli/LunarVim/pull/1011).

I hope that we finally get to merge these PRs

lua/core/autopairs.lua
Tastyep, line 10: You can remove this up to L13. Since this is only called once the plugin is loaded, it is always available

rebuilt:
Thank you for your work on this.  The file does need to be refactored.  We need to take a few more steps on this pull to avoid breaking the callbacks.   

broken autopairs

rebuilt:
Can you please return `npairs` from the setup function
```
  return npairs

```
And then use it in the plugins file
```
    config = function()
      local autopairs = require("core.autopairs").setup()
      if lvim.builtin.autopairs.on_config_done then
        lvim.builtin.autopairs.on_config_done(autopairs)
      end
    end,

```

Then test to make sure it works for you as well

rebuilt:
For reference, this is the relevant autopairs rule

``` lua
lvim.builtin.autopairs.on_config_done = function(module)
	local endwise = require("nvim-autopairs.ts-rule").endwise
	module.remove_rule("`")
	local Rule = require("nvim-autopairs.rule")
	module.add_rule(Rule("`", "`", { "lua", "ruby", "javascript" }))
	module.add_rule(Rule("```", "```", "markdown"))
end

```

lua/core/autopairs.lua
devtoi: I am not sure why, but done.

devtoi:
> For reference, this is the relevant autopairs rule
> 
> ```lua
> lvim.builtin.autopairs.on_config_done = function(module)
> 	local endwise = require("nvim-autopairs.ts-rule").endwise
> 	module.remove_rule("`")
> 	local Rule = require("nvim-autopairs.rule")
> 	module.add_rule(Rule("`", "`", { "lua", "ruby", "javascript" }))
> 	module.add_rule(Rule("```", "```", "markdown"))
> end
> ```

Using this with the new (just updated) code seems to work as expected.

rebuilt (approved):
Retested with updates.  Callbacks work.

lua/core/autopairs.lua
```diff
@@ -1,51 +1,53 @@
...
+M.setup = function()
+  local npairs = require "nvim-autopairs"
+  local Rule = require "nvim-autopairs.rule"
+
+  -- skip it, if you use another global object
+  _G.MUtils = {}
 
-vim.g.completion_confirm_key = ""
-MUtils.completion_confirm = function()
-  if vim.fn.pumvisible() ~= 0 then
```
Tastyep, line 16: Wasn't this necessary ?

devtoi: I think I misunderstood your previous comment. I think it should be correct now.

lua/default-config.lua
```diff
@@ -1002,6 +1002,7 @@ lvim.lang = {
   },
 }
 
+require("core.autopairs").config()
```
Tastyep, line 1005: This will be fixed by your next PR

Tastyep (approved):
Looking good :+1:

ChristianChiarulli:
subtle weird things happen when playing around with autopairs/compe/treesitter, I will need to test this pretty extendibly before merging.

Tastyep:
> subtle weird things happen when playing around with autopairs/compe/treesitter, I will need to test this pretty extendibly before merging.

Well, this just wraps the code into a function and calls it, it doesn't change anything right ?

rebuilt:
@ChristianChiarulli @devtoi @Tastyep @abzcoding @Tastyep 

I think we were having trouble with autopairs for a while because packer loads things in the order they appear.  We load treesitter after compe and autopairs.  But because treesitter is autostarted and we manually require the dependencies in each of the core.configs, it doesn't cause problems now.  Also, any changes to the plugins file invalidates packer_compiled and it's easy to forget to remove it or recompile.  

The next bit, I'm still on shaky ground on, and I'd like correction where I'm wrong:

For performance reasons, `require` makes sure a module doesn't get loaded twice.  This also means that if plugins don't manually check for state changes, the object will be immutable on the next require.  This is why callbacks are necessary for some plugins and not others.  Some plugins manually check for updates to state and other don't.   I was wrong in a previous comment about getting two objects back.  If you `require` a plugin twice, it will print the same memory address for both tables.  So it's the same object getting returned.  The difference is the first time you require a plugin you can change state but the second time, it's immutable unless it manually checks for updates.  

To update state, some plugins `unload` a package manually so that it will get reloaded into memory on the next require.  

For example, this code from telescope
```
-- Hot reload stuff
package.loaded['telescope'] = nil
package.loaded['telescope.init'] = nil
package.loaded['telescope.picker'] = nil
package.loaded['telescope.finder'] = nil

```

And some other plugins
```
rg plenary.reload
opt/which-key.nvim/lua/which-key/init.lua
105:  require("plenary.reload").reload_module("which-key")

start/popup.nvim/lua/tests/basic_popup_spec.lua
1:require('plenary.reload').reload_module('popup')

start/telescope.nvim/scratch/simplest_test.lua
1:require('plenary.reload').reload_module('telescope')

```

Tastyep:
> Also, any changes to the plugins file invalidates packer_compiled and it's easy to forget to remove it or recompile.

About loading order, packer provides a `after` field, [example](https://github.com/ChristianChiarulli/LunarVim/blob/rolling/lua/plugins.lua#L71), would this solve this issue ?

Normally, [this](https://github.com/ChristianChiarulli/LunarVim/blob/rolling/lua/core/autocmds.lua#L67) takes car of it, but I agree, many times I had to remove the file.

> This also means that if plugins don't manually check for state changes, the object will be immutable on the next require.

I'm not sure if I understand, first what state are you speaking about ? The one passed to `module.config(...)` ?
The immutability thing seems very weird to me, do you have a link to some official doc explaining this ?

rebuilt:
> > Also, any changes to the plugins file invalidates packer_compiled and it's easy to forget to remove it or recompile.
> 
> About loading order, packer provides a `after` field, [example](https://github.com/ChristianChiarulli/LunarVim/blob/rolling/lua/plugins.lua#L71), would this solve this issue ?
> 
> Normally, [this](https://github.com/ChristianChiarulli/LunarVim/blob/rolling/lua/core/autocmds.lua#L67) takes car of it, but I agree, many times I had to remove the file.
> 
> > This also means that if plugins don't manually check for state changes, the object will be immutable on the next require.
> 
> I'm not sure if I understand, first what state are you speaking about ? The one passed to `module.config(...)` ?
> The immutability thing seems very weird to me, do you have a link to some official doc explaining this ?

From the nanotee lua guide.   I think this applies to situations like trying to remove a section from galaxyline.  

```
require():
    is a built-in Lua function. It allows you to take advantage of Lua's module system
    searches for modules in lua/ folders in your 'runtimepath'
    keeps track of what modules have been loaded and prevents a script from being parsed and executed a second time. If you change the file containing the code for a module and try to require() it a second time while Neovim is running, the module will not actually update
```

Tastyep:
It says the file won't be reinterpreted because it is cached, but still, the objects are loaded in memory and are not constant, so you can update them, just you can't update constants by modifying the file. I think there was a misunderstanding because you started talking about immutability, so please tell me if I understood you correctly.

rebuilt:
> It says the file won't be reinterpreted because it is cached, but still, the objects are loaded in memory and are not constant, so you can update them, just you can't update constants by modifying the file. I think there was a misunderstanding because you started talking about immutability, so please tell me if I understood you correctly.

You are right.  I was just misunderstanding the documentation.

Tastyep:
Fixed by #1318

PR #1174, opened 2021-07-30
Fix:  Enables galaxyline callbacks to work properly

rebuilt:
Galaxyline callbacks don't work properly.  The example on the wiki stopped working.  
https://github.com/ChristianChiarulli/LunarVim/wiki/Extending-Configurations-for-Core-Plugins#modifying-galaxyline

Galaxyline now has a setup function that passes back a reference to the galaxyline object.  This reference then gets passed on to the callback function.  

Fixes #(issue)
https://github.com/ChristianChiarulli/LunarVim/issues/1144


rebuilt:
This is how it was tested

galaxyline

rebuilt:
This other pull applies the corrections
https://github.com/ChristianChiarulli/LunarVim/pull/1175

PR #1177, opened 2021-07-30
[Feature]: Add capability to override keymapping + refactor keymappings

### LINKED ISSUE #1140: [Feature]: Allow overwriting keymaps rather than replacing them
### Feature motivation

If you add any keybind to `lvim.keys` for a given mode all of the default keybinds for that mode are removed. I feel that it'd be more intuitive for custom keybinds to extend or overwrite the defaults rather than replace them entirely.

Extending the default keybinds.

Copying all of the default keybinds from `keymappings.lua` into `lv-config.lua`.

### Additional context

_No response_

deryrahman:
# Description

I'm adding the capability to override the default keymap without replacing the whole of them. The current behaviour is replacing the default keymap entirely (see #1140). Also I'm doing some refactoring on keymappings so that the code is more readable.

The way we can add / override the keymap is just by adding something like this on `lv-config`

```lua
lvim.keys.normal_mode['po'] = '<ESC>' -- use default opt based on mode
lvim.keys.normal_mode['po'] = {'<ESC>', { silent = true, noremap = true } } -- if you want to use custom opt
```

Fixes #1140 

## How Has This Been Tested?

- Add custom keymapping on lvim config:
```lua
lvim.keys.insert_mode["po"] = 'echo "Hello"<CR>'
```
- Check the keymapping is working
- Check default keymapping is working (not replaced)


deryrahman:
Some plugins keymap will be on the folder `keymappings/preset/core`. So far some builtin plugin's keymaps that I've already moved are bufferline and compe.

Just wondering, is it good if we move all builtin plugin's keymap outside the `core` folder? Planning to move all keymaps from that folder to `keymappings/preset/core` so other dev will know, where to see compiled keymaps in one place. Make it easier, since the existing keymaps is a bit scattered

deryrahman:
Close, no need refactor. Pull out the custom opt capability to different PR https://github.com/ChristianChiarulli/LunarVim/pull/1215

PR #1183, opened 2021-07-30
[Bugfix] Use correct exe name for clang-format

tafryn:
# Description

Fix typo that broke `clang-format` calls.

## How Has This Been Tested?

* Called `vim.lsp.buf.formatting()` on CPP and C files with the fix.


abzcoding:
this has been fixed in upstream

PR #1186, opened 2021-07-31
feat: get null-ls registered providers by filetype

kylo252:
# Description

Adds a utility function to get the registered `null-ls` providers by the file type. 
This helps clean up the logic in `galaxyline` quite a bit, and it's useful for debugging purposes.

## How Has This Been Tested?

`null-ls` providers are still showing up correctly in `galaxyline`


kylo252:
This is appending non-unique values, I'll fix that ASAP.

abzcoding:
@kylo252 can you fix the conflict so i don't mess up again

PR #1190, opened 2021-07-31
Wraps the logic in parenthesis so it works correctly

rebuilt:
With the previous commit, the following way to set keybindings broke
```lua
	lvim.leader = "\\"
```

This pr add parenthesis to fix the logic
```
vim.g.mapleader = (lvim.leader == "space" and " ") or lvim.leader

```

### DIFF
--- a/lua/keymappings.lua
+++ b/lua/keymappings.lua
@@ -99,7 +99,7 @@ if vim.fn.has "mac" == 1 then
   keymaps.normal_mode[8][1] = "<A-Right>"
 end
 
-vim.g.mapleader = lvim.leader == "space" and " " or lvim.leader
+vim.g.mapleader = (lvim.leader == "space" and " ") or lvim.leader
 
 -- navigate tab completion with <c-j> and <c-k>
 -- runs conditionally


Tastyep:
I don't understand why is it necessary to wrap between parentheses

If I run
``` lua
local k = "\\"
local v = k == "space" and " " or k

print(v)
```
https://www.jdoodle.com/execute-lua-online/

It prints what's expected, can you explain the rational ?

rebuilt:
You're right.  It does work as expected.  Looking at it again, the mistake was on my end.  Abzcoding asked anyone to test the latest merges.

I did but found I got this error on startup. I'd been using the latest rolling up to that point.
```
/home/nelson/.local/share/lunarvim/lvim/init.lua:
E5113: Error while calling lua chunk: ...e/nelson/.local/share/lunarvim/lvim/lua/utils/key
map.lua:17: bad argument #1 to 'ipairs' (table expected, got string)
```

A blank config worked fine.  But my config threw that error.  
I commented out all my keybindings to find the one that was causing an issue. 
This is the line in my config that caused an issue.  I removed it and neovim started up fine.
``` lua
lvim.keys.leader_key = "space"
```

This is wrong usage but I didn't realize that at the time.  I read through the lunarvim keybinding code to see where the leader was used.  I added parenthesis, then went to my personal keybindings.  But because I had just read the logic, I set the keybinding correctly with 

``` lua
lvim.leader = "space"
```
It worked and I incorrectly attributed the fix to adding parenthesis and not to incorrect usage.

Tastyep:
Good to know where it comes from, that's not a big deal anyway :+1:

PR #1197, opened 2021-08-01
[Feature] - Add terminal filetype cmd

barlingo:
# Description

This PR enables an interface to add terminal commands to be launch based on filetype. 
Original idea by @abzcoding 

## How Has This Been Tested?

Tested with python execute so far.

## Usage:
-- Add commands on your config.lua file by filling the following table:

`lvim.builtin.terminal.ft_cmds = {{ filetype, keymap, cmd, name}}`

For example:

`lvim.builtin.terminal.ft_cmds = {
	{ "python", "xr", "python " .. vim.fn.expand("%") .. "; read", "Run" },
}`

The keymap does not include the <leader> key since it is attached automatically.


abzcoding:
@barlingo can you pull the latest changes, we had some issues with formatting that should be fixed now

barlingo:
There might be a better way of doing this tbh. I just thought of not having any dependencies on terminal being active.

If the terminal plugin does not load on start the autocommands don't get executed which means that they are not valid.

If the autocommands live in other place we will need to ensure that terminal is active.

Code can also be refactored to create an autocommand function instead of copying over the one from lua/core/autocmds.lua

abzcoding:
> There might be a better way of doing this tbh. I just thought of not having any dependencies on terminal being active.
> 
> If the terminal plugin does not load on start the autocommands don't get executed which means that they are not valid.
> 
> If the autocommands live in other place we will need to ensure that terminal is active.
> 
> Code can also be refactored to create an autocommand function instead of copying over the one from lua/core/autocmds.lua

First off, thank you for your contribution 
but I'm kind of against adding these kinds of extra stuff to LV 😅 
i don't see why we should consider this a core feature, if users want this, they can add it themselves tbh

barlingo:
I agree it only adds a bit of syntactic help on creating the user command.

For instance, instead of doing this:
`{ "Filetype", "c,cpp", "nnoremap <leader>r <cmd>lua require('core.terminal')._exec_toggle('make run;read')<CR>" },`

You'll just need to do
`{ "c,cpp", "r", "make run; read", "Run" }`

Since commands based on file type to launch a process are fairly common in any vimrc I thought this could simplify it.

There might be a way to do this in a cleaner way.

rebuilt:
Maybe we should come back to this after 1.0 is released?   Adding feature is taking away time from fixing bugs.  The more features we add, the more features we have to support.  And our team is not growing as fast as the features.

PR #1199, opened 2021-08-01
[Refactor] Introduce service and config modules

Tastyep:
# Description

Encapsulate configuration logic in the config module.
Provide an interface for loading a user's configuration on top of LVIM's default settings.
Introduce core services, for people familiar with DDD, this would correspond to domain services. I moved utils/keymap.lua in there.
Providing core services will allow to unbloat the `utils` module which should not contain the application's logic.

Ideally, I would have moved all builtins from `core/` into `core/builtin/`, plugin-loader would be in `core/service/`, and plugins would be part of `core/builtin/`. However, this would create too many conflicts with other PRs.

Additionally, I removed some code introduced by [this PR](https://github.com/ChristianChiarulli/LunarVim/pull/1193/files#diff-0fc5c43f6f1d7605fd7d16c088ba6b83f096eee8e2049517407bbaa27aa3a4f7R14) as I think that the error message is explicit (see below).

image

## How Has This Been Tested?

Open lv-config.lua, change the value of format_on_save to false, it should not format, change again, the config should be formatted.


lua/config/init.lua
```diff
@@ -0,0 +1,35 @@
+local M = {
+  USER_CONF_PATH = "~/.config/lvim/config.lua",
```
rebuilt, line 2: We were still supporting lv-config.lua.  This makes it so that filename is no longer supported.  I think we should keep supporting it for a bit longer.

rebuilt (changes requested):

Tastyep:
Since I'll be back on the 12th, this PR will be outdated, I'll redo the work if necessary

PR #1203, opened 2021-08-01
[Feature] Add exec to lvim binary

regzon:
# Description

Currently, the nvim process is launched as a child of the shell process. In my case, this made the tmux think that the running process is a shell script rather than the nvim. This led to the broken tmux-nvim integration, wrong tmux window title and etc. Also, I believe that this update has many other user cases.

## Solution

`exec` shell command replaces the shell process with the given command.

## How Has This Been Tested?

I've updated the shell script locally and tested that the LunarVim itself is working fine and that the tmux determines the process correctly.

### DIFF
--- a/utils/bin/lvim
+++ b/utils/bin/lvim
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-nvim -u ~/.local/share/lunarvim/lvim/init.lua --cmd "set runtimepath+=~/.local/share/lunarvim/lvim" "$@"
+exec nvim -u ~/.local/share/lunarvim/lvim/init.lua --cmd "set runtimepath+=~/.local/share/lunarvim/lvim" "$@"


rebuilt:
Does this work in a remote tmux session?

regzon:
> Does this work in a remote tmux session?

You mean connect to the remote host, start tmux there and launch a LunarVim instance inside tmux?

rebuilt:
Yes.  I don't expect it to break anything.  But can we sanity check this use case?

regzon:
> Yes. I don't expect it to break anything. But can we sanity check this use case?

Yeah, I've checked. Everything works fine.

regzon:
I can't test this on MacOS (cause I don't have one). I haven't heard of any issue with `exec` on Macintosh platforms, but it'd be nice to check.

rebuilt:
`exec` is a posix command so it should be available in both bash and zsh; the shells macos uses.  So it should work.  Or am I wrong about this?

abzcoding:
> `exec` is a posix command so it should be available in both bash and zsh; the shells macos uses. So it should work. Or am I wrong about this?

can confirm that it works in mac as well

abzcoding (approved):
seems to be working on linux/mac

rebuilt (approved):

PR #1204, opened 2021-08-01
[Feature] multiple lang server providers for single language

### LINKED ISSUE #1107: [Feature]: Lsp provider should be a table or a string
### Feature motivation

some langs need multiple lang servers

so we will loop through the table and set them or if it's just a string set the one

none

### Additional context

_No response_

dklymenk:
# Description

The changes in this PR allow users to supply multiple language servers to `lvim.lang.<lang>.lsp`.
The table will be looped over and every language server will be initialized.

Fixes #1107 

## How Has This Been Tested?

1. Starting lvim with the default config causes no errors.
2. Adding something something like this correctly initializes 2 language servers:
```lua
lvim.lang.php.lsp = {
  intelephense = {
    provider = "intelephense", -- we shouldn't need this line anymore because we can use the key above for provider name
    setup = {
      cmd = {
        DATA_PATH .. "/lspinstall/php/node_modules/.bin/intelephense",
        "--stdio",
      },
      filetypes = { "php", "phtml" },
      settings = {
        intelephense = {
          environment = {
            phpVersion = "7.4",
          },
        },
      },
    },
  },
  phpactor = {
    provider = "phpactor",  -- we shouldn't need this line anymore because we can use the key above for provider name
    setup = {
      handlers = {
        -- rely on intelephense for hover
        ["textDocument/hover"] = function() end,
      },
    },
  },
}
```

If you are going to test this PR with my php example above, you'll need to have [phpactor](https://phpactor.readthedocs.io/en/master/usage/standalone.html) and `intelephense` installed.
For `intelephense`, `:LspInstall php` should do the trick, but for `phpactor` you will need to follow the instructions [here](https://phpactor.readthedocs.io/en/master/usage/standalone.html).

Thanks.

### DIFF
--- a/lua/lsp/init.lua
+++ b/lua/lsp/init.lua
@@ -71,14 +71,14 @@ function M.setup(lang)
   local lsp = lvim.lang[lang].lsp
 
   -- normalize the lsp config to be a table of providers
-  if lsp['provider'] ~= nil then
-    lsp = {[lsp.provider] = lsp}
+  if lsp["provider"] ~= nil then
+    lsp = { [lsp.provider] = lsp }
   end
 
   -- initialize language server for each provider
   for provider, single_lang_server in pairs(lsp) do
     if not require("utils").check_lsp_client_active(provider) then
-        require("lspconfig")[provider].setup(single_lang_server.setup)
+      require("lspconfig")[provider].setup(single_lang_server.setup)
     end
   end
 end
--- a/lua/lsp/init.lua
+++ b/lua/lsp/init.lua
@@ -69,12 +69,18 @@ end
 
 function M.setup(lang)
   local lsp = lvim.lang[lang].lsp
-  if require("utils").check_lsp_client_active(lsp.provider) then
-    return
+
+  -- normalize the lsp config to be a table of providers
+  if lsp['provider'] ~= nil then
+    lsp = {[lsp.provider] = lsp}
   end
 
-  local lspconfig = require "lspconfig"
-  lspconfig[lsp.provider].setup(lsp.setup)
+  -- initialize language server for each provider
+  for provider, single_lang_server in pairs(lsp) do
+    if not require("utils").check_lsp_client_active(provider) then
+        require("lspconfig")[provider].setup(single_lang_server.setup)
+    end
+  end
 end
 
 return M


lua/lsp/init.lua
```diff
@@ -69,12 +69,18 @@ end
 
 function M.setup(lang)
   local lsp = lvim.lang[lang].lsp
-  if require("utils").check_lsp_client_active(lsp.provider) then
-    return
+
+  -- normalize the lsp config to be a table of providers
+  if lsp["provider"] ~= nil then
```
Tastyep, line 74: If you want to add support to multiple providers, please change the config structure to show that it supports this.

Tastyep (changes requested):

lua/lsp/init.lua
dklymenk: My idea was to make this feature optional and not do any changes to `lua/default-config.lua`, as it might break the setup for some users on `rolling`.

I can add my php example to `utils/installer/config.example.lua` if that's what you mean.

Tastyep: No I mean that we shouldn't have some kind of adapter, we should adapt the structure of the configuration to our needs, and it should reflect on Lunarvim's capabilities.

Tastyep: So probably it should be `providers = {...}`

dklymenk: > No I mean that we shouldn't have some kind of adapter, we should adapt the structure of the configuration to our needs, and it should reflect on Lunarvim's capabilities.

Ok, I agree

> So probably it should be `providers = {...}`

What exactly do you think should be inside the `{...}`? Just the names of providers, or setup functions too?

dklymenk: I propose a structure like this:
```
lvim.lang.php.lsp = {
  intelephense = {
    setup = { },
    },
  },
  phpactor = {
    setup = {
    },
  },
}
```

Basically the one I had in my example, but we remove the redundant provider key.

I can refactor the default config to be structured like this, if you're sure that's ok.

dklymenk: Hello, I am still interested in contributing this as well as refactor the layout of default configs for lsp, I just want to agree on some specific structure, so I don't have to redo it.

Tastyep: The structure should be

``` lua
lvim.lang.php.lsp = {
  providers = {
    {
      name="provider_name",
      setup = {}
    }
  }
}
```

But I don't think configuring multiple LSPs just work out of the box, for instance how does it work if both provide formatting capabilities ?

dklymenk: Ok, I agree with your suggested structure.

> But I don't think configuring multiple LSPs just work out of the box, for instance how does it work if both provide formatting capabilities ?

You can see in my example config in the first post I am passing custom handlers to make sure there is only one hover box when I press `Shift+k`. In my experience some of the stuff you disable via `capabilities` and some via `handlers`, but I think it should be up to user to decide which language server they want to use for each functionality. It is up to us, however, to make sure that there is a way to disable the functionalities and it's well documented. I will update sample config with an example.

dklymenk: @Tastyep Hello, I have updated the default config with new `lsp.providers` structure, removed normalization code and made it somewhat more streamlined overall.

Also updated the example configs with infos on how to add 2 providers and added a note about disabling capabilities via handlers. The `["textDocument/hover"] = function() end,` trick works for pretty much anything I have tested:
* textDocument/formatting
* textDocument/hover
* textDocument/publishDiagnostics
* textDocument/definition
* textDocument/declaration

One thing I didn't find a way to remove duplication for is `textDocument/completion`. The handler trick does nothing for it, and neither does setting capability to `nil` or false. Same for overwriting a callback. I am under impression that this should be resolved via `compe`, but I'm not sure there is a way to disable LSP source for just one particular provider.

Really hope this gets merged before any other changes to default configs are made, resolving configs there won't be pleasant.

utils/installer/config.example-no-ts.lua
```diff
@@ -76,6 +76,37 @@ lvim.builtin.treesitter.highlight.enabled = true
 --     args = {}
 --   }
 -- }
+-- you can define multiple providers for one language
```
Tastyep, line 79: `you can define multiple language servers`

lua/lsp/init.lua
```diff
@@ -128,22 +128,19 @@ function M.common_on_attach(client, bufnr)
...
     end
   end
 
-  if lsp.provider ~= nil and lsp.provider ~= "" then
-    local lspconfig = require "lspconfig"
-    lspconfig[lsp.provider].setup(lsp.setup)
+  -- initialize language server for each provider
+  local providers = lvim.lang[lang].lsp.providers
+  for _, provider in pairs(providers) do
+    if provider.name ~= nil and provider.name ~= "" and not require("lsp.utils").is_client_active(provider.name) then
```
Tastyep, line 141: I know you just adapted the existing code, but our code didn't make sens
can you do
``` lua
local lsp_utils = require("lsp.utils") -- Put above providers
if not lsp_utils.is_client_active(provider.name) then
   ....
end
```

Can you then do a pass on default-config to change:

``` lua
      providers = {
        {
          name = "",
          setup = {},
        },
      },
```

into

``` lua
 providers = {},
```

Tastyep (changes requested):
I'll ask on the dev channel to postpone merges touching `default-config`

utils/installer/config.example-no-ts.lua
dklymenk: changed the wording here and in another example file

lua/lsp/init.lua
dklymenk: changed

Tastyep: Please resolve the conversation :)

lua/lsp/init.lua
```diff
@@ -128,22 +128,20 @@ function M.common_on_attach(client, bufnr)
...
       return
     end
   end
 
-  if lsp.provider ~= nil and lsp.provider ~= "" then
-    local lspconfig = require "lspconfig"
-    lspconfig[lsp.provider].setup(lsp.setup)
+  -- initialize language server for each provider
+  local lsp_utils = require "lsp.utils"
+  local providers = lvim.lang[lang].lsp.providers
```
abzcoding, line 140: what happens if `lvim.lang[lang]` is nil?
for example for a lang that we didn't define in `default-config.lua`

dklymenk: Managed to test this by deleting an entry from `default-config.lua`. It does indeed throw an error.
The one-line change I pushed resolves the issue by adding a fallback `{}` if that value is empty.

lua/lsp/init.lua
```diff
@@ -128,22 +128,20 @@ function M.common_on_attach(client, bufnr)
...
       return
     end
   end
 
-  if lsp.provider ~= nil and lsp.provider ~= "" then
-    local lspconfig = require "lspconfig"
-    lspconfig[lsp.provider].setup(lsp.setup)
+  -- initialize language server for each provider
+  local lsp_utils = require "lsp.utils"
+  local providers = (lvim.lang[lang] and lvim.lang[lang].lsp.providers or {})
```
Tastyep, line 140: That's how you should do it

``` lua
function M.setup(lang)
  if not lvim.lang[lang] or not lvim.lang[lang].lsp then
    return
  end
```

Tastyep: Take a look at this
https://github.com/luarocks/lua-style-guide

abzcoding:
this happens for me, but not sure yet what is causing this 😢 
Screen Shot 2021-08-16 at 4 05 39 PM

abzcoding:
sidenote: 
this is a breaking change for people that installed their own langservers
or tried to configure `json_schemas` using this syntax
Screen Shot 2021-08-16 at 4 12 16 PM

Tastyep (approved):

abzcoding:
after a vote from core devs, we've decided to shelf this for now
Screen_Shot_2021-08-19_at_9 58 02_PM

dklymenk:
I'm sorry if I am missing some vital context here, but if you don't have time to work on this feature, I can work on it myself.

If you don't like the fact it changes the config structure, we can use my original proposal that makes it so old single provider configs still work.

If you don't like the actual suggested code (i'm pointing out the "yet another thing that can fail" comment on a screenshot), I can implement any number of additional checks you want. I personally don't really see why that simple loop is more prone to errors than the current version of the code in `rolling`.

My point is, I would really like to have this feature in core lvim, so I don't have to resolve conflicts every time there are updates lsp/init.lua, and I am willing to do everything I can to make it merged before 1.0.

P.S. One big reason why you might want to merge this before 1.0 is because if what @Tastyep said is still your priority:

> No I mean that we shouldn't have some kind of adapter, we should adapt the structure of the configuration to our needs, and it should reflect on Lunarvim's capabilities.

Implementing a breaking change to config structure after 1.0 release would make for a pretty bad user experience.

abzcoding:
> I'm sorry if I am missing some vital context here, but if you don't have time to work on this feature, I can work on it myself.
> 
> If you don't like the fact it changes the config structure, we can use my original proposal that makes it so old single provider configs still work.
> 
> If you don't like the actual suggested code (i'm pointing out the "yet another thing that can fail" comment on a screenshot), I can implement any number of additional checks you want. I personally don't really see why that simple loop is more prone to errors than the current version of the code in `rolling`.
> 
> My point is, I would really like to have this feature in core lvim, so I don't have to resolve conflicts every time there are updates lsp/init.lua, and I am willing to do everything I can to make it merged before 1.0.
> 
> P.S. One big reason why you might want to merge this before 1.0 is because if what @Tastyep said is still your priority:
> 
> > No I mean that we shouldn't have some kind of adapter, we should adapt the structure of the configuration to our needs, and it should reflect on Lunarvim's capabilities.
> 
> Implementing a breaking change to config structure after 1.0 release would make for a pretty bad user experience.

yeah sorry I should've made it more clear, basically, we had an extended argument about pros/cons and decided that this would cause some cons that are not worth the `x% of users, x<5`  get out of it
so we shelved it for now, but we might come back to it before `1.0`, bear in mind we are still at `0.5.2` 😅 

here are some of those arguments
image

Screen Shot 2021-08-19 at 10 27 29 PM

kylo252:
> My point is, I would really like to have this feature in core lvim, so I don't have to resolve conflicts every time there are updates lsp/init.lua, and I am willing to do everything I can to make it merged before 1.0.

It's only appropriate that I chime-in on this. I'm really sorry if you felt ignored, but I can say at least that the discussion about this has been going on for a while.

Can you tell me what problems you faced while starting the server on your own? It's a way bigger problem if servers can't be started independently of core.

dklymenk:
> Can you tell me what problems you faced while starting the server on your own? It's a way bigger problem if servers can't be started independently of core.

:facepalm: WOW. I have only just realized I can do this directly in  `config.lua` and it won't break anything
```lua
require("lspconfig").phpactor.setup {
  handlers = {
    ["textDocument/hover"] = function() end,
  },
}
```

Thanks for your comment. I didn't have much experience with nvim lua or lua at all when I first installed lunarvim and thought that I need to change the core files to facilitate my need for those 2 language servers. And I guess after I got tunnel-visioned into it, there was no going back.

kylo252:
> Thanks for your comment. I didn't have much experience with nvim lua or lua at all when I first installed lunarvim and thought that I need to change the core files to facilitate my need for those 2 language servers. And I guess after I got tunnel-visioned into it, there was no going back.

No worries, I'm just glad it worked out 🥳 

BTW, according to the maintainers of lspconfig, this operation is very cheap [^1]

> "requires evaluating less lua at startup" -> the total time for evaluating a setup call across all language servers in the repo was around 1 ms (combined) last time I checked

https://github.com/neovim/nvim-lspconfig/issues/970

dklymenk:
@abzcoding Thanks for the screenshots, most of those points don't really make sense to me or are factually wrong, but it doesn't matter anyway since I am content with just adding 5 lines of code to my `config.lua`.

If/when this feature is back on your radar and you need someone to work on it, ping me. I'll happily run the config structure change macro and adapt the changes to `lua/lsp/init.lua` regardless of how it will look like in the future.

Btw, I think. you should update the original issue #1107, so ppl don't have to read this PR to get insights.

paulofaria:
@abzcoding just out of curiosity, how did you reach the <5% number?

abzcoding:
> @abzcoding just out of curiosity, how did you reach the <5% number?

Screen Shot 2021-09-16 at 1 45 05 AM
my name is not `Hatsu` 😅

abzcoding:
> @abzcoding just out of curiosity, how did you reach the <5% number? In other words, where did this number come from? Does Lunarvim has some kind of analytics going on? Was there a poll?

no that was just a symbolic number to signify that not many people would need it, and to clarify we don't have any sort of analytics as you can see in the code

paulofaria:
Oh, I was just curious if there is some analytics or a recurrent poll or something. Having this kind of information as close to reality as possible would really help with decision making. Any analytics would have to be opt-in, of course.

paulofaria:
Maybe polls would be less intrusive, but maybe less precise. I personally don't like analytics, but for an open source project like this, I might consider.

abzcoding:
> Oh, I was just curious if there is some analytics or a recurrent poll or something. Having this kind of information as close to reality as possible would really help with decision making. Any analytics would have to be opt-in, of course.

oh i doubt analytics would be appreciated in the community, but we can run a poll or sth in discord, or create an issue(discussion) in github and ask the users 🤷

paulofaria:
I'll create an issue later on. Would be nice to know how are people using lvim! Thanks for your hard work, by the way. @abzcoding

abzcoding:
> I'll create an issue later on. Would be nice to know how are people using lvim! Thanks for your hard work, by the way. @abzcoding

tyvm sir 👍

PR #1205, opened 2021-08-01
[Feature] Make keymaps of bufferline and compe configurable

Tastyep:
# Description

Make compe and bufferline keymaps overridable.


lua/core/bufferline.lua
```diff
@@ -1,2 +1,16 @@
-vim.api.nvim_set_keymap("n", "<S-l>", ":BufferNext<CR>", { noremap = true, silent = true })
-vim.api.nvim_set_keymap("n", "<S-h>", ":BufferPrevious<CR>", { noremap = true, silent = true })
+lvim.builtin.bufferline = {
+  keymap = {
+    values = {
+      insert_mode = {
+        { "<S-l>", ":BufferNext<CR>" },
```
abzcoding, line 5: these should be normal mode instead of insert mode i think

PR #1207, opened 2021-08-01
feat: Add an async logger using plenary

kylo252:
# Description

- Add an async logger using Plenary's log API.
- Replaces all calls to `lvim_log` that were used for handling `null-ls` with the calls new `utils.log` object.
- Added two new key bindings `<leader>Ll`  to open the logfile, and `<leader>Li` or view it interactively using [lnav](https://github.com/tstack/lnav)

TODOs
- [ ] use a standardized log entry format
- [ ] fix highlighting in `lnav`
- [x] add a log entry for all essential logic
     - [x] LSP
     - [x] plugins??
- [x] interactive logs should open in a split instead of floating mode
- [x] add a fallback for viewing the logs (probably `less +F`

## How Has This Been Tested?

- add `lvim.debug = true` to your `config.lua`
- Use any of the aforementioned shortcuts or directly open `~/.cache/nvim/lunarvim.log`


rebuilt (changes requested):
It would be nice to explain this in the example configuration as well.

```lua
-- To turn on debugging add these settings and install lnav
-- https://github.com/tstack/lnav
-- Press  <leader>Ll to open the logfile, and <leader>Li or view it interactively using lnav
-- lvim.debug = true
-- lvim.builtin.terminal.active = true
```

The logging works for null-ls but I don't see any logs for lsp problems.  Is the scope of this pr to include only formatter errors?  

I would suggest adding these keybindings also
```
      L = {
        name = "lunarvim",
        f = { "<cmd>edit ~/.cache/nvim/lunarvim.log<cr>", "Open the Formatting logfile" },
        n = { "<cmd>edit ~/.cache/nvim/log<cr>", "Open the Neovim logfile" },
        l = { "<cmd>edit ~/.cache/nvim/lsp.log<cr>", "Open the LSP logfile" },
        p = { "<cmd>edit ~/.cache/nvim/packer.nvim.log<cr>", "Open the Packer logfile" },
      },

```

rebuilt:
Keybindings sorted alphabetically

```
      L = {
        name = "lunarvim",
        f = { "<cmd>edit ~/.cache/nvim/lunarvim.log<cr>", "Open the Formatting logfile" },
        l = { "<cmd>edit ~/.cache/nvim/lsp.log<cr>", "Open the LSP logfile" },
        n = { "<cmd>edit ~/.cache/nvim/log<cr>", "Open the Neovim logfile" },
        p = { "<cmd>edit ~/.cache/nvim/packer.nvim.log<cr>", "Open the Packer logfile" },
      },

```

rebuilt:
And the execs 

```
{ "lnav ~/.cache/nvim/lunarvim.log", "LF", "Interactive formatter log [lnav]" },
{ "lnav ~/.cache/nvim/lsp.log", "LL", "Interactive lsp log [lnav]" },
{ "lnav ~/.cache/nvim/log", "LN", "Interactive neovim log [lnav]" },
{ "lnav ~/.cache/nvim/packer.nvim.log", "LP", "Interactive packer log [lnav]" },

```

kylo252:
Any comments about `lnav`, alternatives or maybe provide a simple `less +F` instead/fallback?

rebuilt:
I've never used lnav or any tool like it.  Honestly I'm a little lost in it.  

I think a fallback is a great idea.  You could fall back by grepping for errors in the log file, filtering the last 10 errors and sending that to `less`.   
```
lnav ~/.cache/nvim/lsp.log || grep ERROR ~/.cache/nvim/lsp.log | tail  | less
```

rebuilt:
I'm not sure what changed but I think you have an error in lsp/init.lua

I'm getting this error now
```
Error detected while processing /home/nelson/.local/share/lunarvim/lvim/ftplugin/lua.lua:
E5113: Error while calling lua chunk: /home/nelson/.local/share/lunarvim/lvim/lua/lsp/init.lua:98: attempt to index globa
l 'u' (a nil value)
Error detected while processing FileType Autocommands for "*":
Error while processing file: /home/nelson/.local/share/lunarvim/lvim/ftplugin/lua.lua
/home/nelson/.local/share/lunarvim/lvim/lua/lsp/init.lua:98: attempt to index global 'u' (a nil value)
```

kylo252:
> I've never used lnav or any tool like it. Honestly I'm a little lost in it.

It's just a tool to provide an easy way to view logs with some syntax highlighting and other nifty features like severity flags, error rate, search, etc.

> I think a fallback is a great idea. You could fall back by grepping for errors in the log file, filtering the last 10 errors and sending that to `less`.

We don't need `lnav`  for that, we can simply use `less` like this but we lose the colored output

`less +F ~/.cache/nvim/lunarvim.log` 

or we can use `bat` for the syntax highlighting, but lose the "live" view since it doesn't support the follow, i,e: `+F` option.

`bat -l log ~/.cache/nvim/lunarvim.log`

rebuilt:
When I opened lnav I didn't get any syntax highlighting.  I'll try on another machine when your code is updated.

kylo252:
> I'm not sure what changed but I think you have an error in lsp/init.lua

it should be sorted now. it was a merge error 😅

kylo252:
> When I opened lnav I didn't get any syntax highlighting. I'll try on another machine when your code is updated.

yeah it's not working correctly yet because the format from plenary seems to be non-standard. take a look here: <https://docs.lnav.org/en/latest/formats.html#log-formats>

also
> New log formats can be defined by placing JSON configuration files in subdirectories of the ~/.lnav/formats/ directory.

Try it with some system logfile, e,g: `/var/log/pacman.log`, to see how it usually looks 😃

rebuilt:
Let me know when I should test this again.

kylo252:
This is probably blocked by #1256 .

rebuilt (dismissed):
A lot of the log statements here are aimed at null-ls and formatting.  But if you're trying to figure out what's going wrong with formatting you also need to print something to the log when calling utils.toggle_autoformat() so the user knows that autoformatting is off/on.  

I think it might also be useful to log when a callback is being called.  Not as a debug log but maybe an info log.  I'm not sure.  If they're not using callbacks, nothing would get listed but if they, are, it would be nice to know the callback is being registered properly.  I'm not sure about this one, so I'll leave it up to you.

kylo252:
@rebuilt, is this good enough now? I'm thinking it might be better to adjust the log format upstream in Plenary instead of here.

rebuilt:
> @rebuilt, is this good enough now? I'm thinking it might be better to adjust the log format upstream in Plenary instead of here.

I think it's great.  There's just one question I had.  I added a log statement in utils.reload_lv_config()
`    Log:get_default().info "Reloaded configuration"` 

But that statement never gets logged.  I tried adding print statements there but that didn't work either.  I'm not sure if my log statement is incorrectly declared or if something is blocking that autocommand from running correctly.

rebuilt (approved):
Looks great!  I think it's ready to go.

rebuilt:
All the log statements work.  The only case where a log statement does not get printed is the "Reloaded configuration" message. And that's only if the config file is being symlinked.  I don't think we need to account for this edge case. I wanted to note it here for completeness.

abzcoding (approved):
LGTM 🚀

PR #1210, opened 2021-08-02
[Refactor] Adopt which key mapping style

Tastyep:
# Description

Convert our keymappings to match which-key style.
This will allow us to migrate all our keymappings to which key once we will be able to do it.
It is also more explicit for the user in case he wants to replace a specific keymap.

! It will break user-defined keymappings in config.lua as the format has changed.

## How Has This Been Tested?

`:map`, `:nmap` ...

### DIFF
--- a/lua/keymappings.lua
+++ b/lua/keymappings.lua
@@ -9,72 +9,72 @@ local opts = {
 local keymaps = {
   insert_mode = {
     -- I hate escape
-    { "jk", "<ESC>" },
-    { "kj", "<ESC>" },
-    { "jj", "<ESC>" },
+    ["jk"] = { "<ESC>" },
+    ["kj"] = { "<ESC>" },
+    ["jj"] = { "<ESC>" },
     -- Move current line / block with Alt-j/k ala vscode.
-    { "<A-j>", "<Esc>:m .+1<CR>==gi" },
-    { "<A-k>", "<Esc>:m .-2<CR>==gi" },
+    ["<A-j>"] = { "<Esc>:m .+1<CR>==gi" },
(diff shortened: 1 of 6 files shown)


lua/keymappings.lua
```diff
@@ -9,94 +9,94 @@ local opts = {
...
-    { "<C-Down>", ":resize +2<CR>" },
-    { "<C-Left>", ":vertical resize -2<CR>" },
-    { "<C-Right>", ":vertical resize +2<CR>" },
+    ["<C-Up>"] = { ":resize -2<CR>" },
+    ["<C-Down>"] = { ":resize +2<CR>" },
+    ["<C-Left>"] = { ":vertical resize -2<CR>" },
+    ["<C-Right>"] = { ":vertical resize +2<CR>" },
 
     -- Tab switch buffer
     -- { "<TAB>", ":bnext<CR>" },
```
rebuilt, line 39: Looks like there are some leftover comments here as well.  Do you want to remove them or update them to the new syntax?   I'm for removing them.

rebuilt (changes requested):
That syntax is cleaner and easier to read.  I like it.   Did you manually change all these or did you use a macro?

rebuilt:
I tested all the keymaps.  Everything continues to work fine.

ChristianChiarulli:
I like this

lua/keymappings.lua
Tastyep: Too late, but to answer you I did it manually

rebuilt: This is the perfect situation for macros.

gnmearacaun: @Tasty, you're oldschool!!

Issue #490, opened 2021-06-14
are all features documented somewhere?


AggamR:
are all the features documented somewhere? I wanna learn how to use all the features of the plugins but I don't know where to start.

rebuilt:
Start on the project page for each plugin.  That's where they'll be documented best.

AggamR:
> Start on the project page for each plugin. That's where they'll be documented best.

and where can i see the list of plugins? or their keybinds?

rebuilt:
https://github.com/ChristianChiarulli/LunarVim#important-configuration-files

rebuilt:
Read the sections for Leader and Whichkey, other key bindings, and important configuration files

AggamR:
> Read the sections for Leader and Whichkey, other key bindings, and important configuration files

Hi, sorry, I can't find the keybind for IntelliSense-like features like seeing definitions and going to them and stuuf like that

rebuilt:
They are defined in the whichkey configuration file

```lua
   l = {
      
      
                name = "+LSP",
      
      
                a = {"&lt;cmd&gt;Lspsaga code_action&lt;cr&gt;", "Code Action"},
      
      
                A = {"&lt;cmd&gt;Lspsaga range_code_action&lt;cr&gt;", "Selected Action"},
      
      
                d = {"&lt;cmd&gt;Telescope lsp_document_diagnostics&lt;cr&gt;", "Document Diagnostics"},
      
      
                D = {"&lt;cmd&gt;Telescope lsp_workspace_diagnostics&lt;cr&gt;", "Workspace Diagnostics"},
      
      
                f = {"&lt;cmd&gt;LspFormatting&lt;cr&gt;", "Format"},
      
      
                i = {"&lt;cmd&gt;LspInfo&lt;cr&gt;", "Info"},
      
      
                l = {"&lt;cmd&gt;Lspsaga lsp_finder&lt;cr&gt;", "LSP Finder"},
      
      
                L = {"&lt;cmd&gt;Lspsaga show_line_diagnostics&lt;cr&gt;", "Line Diagnostics"},
      
      
                p = {"&lt;cmd&gt;Lspsaga preview_definition&lt;cr&gt;", "Preview Definition"},
      
      
                q = {"&lt;cmd&gt;Telescope quickfix&lt;cr&gt;", "Quickfix"},
      
      
                r = {"&lt;cmd&gt;Lspsaga rename&lt;cr&gt;", "Rename"},
      
      
[...]

PR #1213, opened 2021-08-02
[Refactor] Allow editing default keymaps

### LINKED ISSUE #1194: [Feature]: Add keybindings defined in defualt plugin configs to which-key (if possible?)
### Feature motivation

When pressing "g" and waiting for which-key to show keybindings for "g", the description for "gi" is that of the default. But we override "gi" with "vim.lsp.buf.implementation()" in the LSP config. See attached image below. 

I realize that this is a very minor issue and it's possible that this is an issue that should be brought up with with which-key and not LunarVim? However, I'm not sure thought I'd try here to see if anyone has an idea of how to work around this.

Screenshot 2021-07-31 at 18 59 44

Also, as a side-note, it would be great if we could override the LSP keybindings. For example, I'd prefer to stick with the default "gi" and map "vim.lsp.buf.implementation()" to "gI" instead. I guess this is related to https://github.com/ChristianChiarulli/LunarVim/issues/966 though.

_No response_

_No response_

### Additional context

_No response_

kylo252:
# Description

A refactor of the core logic for keybindings

- Allow editing any default mapping _individually_ with autocomplete support in `~/.config/lvim/config.lua`
- Add a new key binding to view all the defaults `<leader>Lk`

Fixes #1194 and partially #490

## How Has This Been Tested?

See the examples in the README

### DIFF
--- a/lua/keymappings.lua
+++ b/lua/keymappings.lua
@@ -1,109 +1,157 @@
-local opts = {
-  insert_mode = { noremap = true, silent = true },
-  normal_mode = { noremap = true, silent = true },
-  visual_mode = { noremap = true, silent = true },
-  visual_block_mode = { noremap = true, silent = true },
-  term_mode = { silent = true },
+local M = {}
+
+local generic_opts_any = { noremap = true, silent = true }
+
+local mode_adapters = {
+  insert_mode = "i",
+  normal_mode = "n",
+  term_mode = "t",
+  visual_mode = "v",
+  visual_block_mode = "x",
 }
 
-local keymaps = {
-  insert_mode = {
-    -- I hate escape
-    ["jk"] = { "<ESC>" },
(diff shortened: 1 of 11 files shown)


lua/keymappings.lua
```diff
@@ -1,109 +1,157 @@
-local opts = {
```
Tastyep, line 1: We go back to mixing logic and data/configuration :(

abzcoding: this doesn't notify anyone btw 😅 
you might want to tag someone if you want them to notice

kylo252: I thought it was just redundant for default keymaps (we want them silent and non-recursive). I can restore these opts if you think it's better though. It's not a big deal for me.

I also purposefully didn't remove `load()` and `load_mode()`, so we can always re-use them again if you'd like.

PR #1218, opened 2021-08-03
feat: add lvim.lsp.smart_cwd

kylo252:
# Description

- Enable querying the language-server for the `root_dir`
- Use `root_dir` to set the current working-directory (CWD) and enable it by default
- Make vim-rooter configurable and add an option to disable it

Inspired by "ahmedkhalf/lsp-rooter.nvim"

## How Has This Been Tested?

1- Disable vim-rooter

```lua
lvim.builtin.rooter.active = false
```

2- Open any file with LSP support and see the "smart_cwd" in action :)

### DIFF
--- a/lua/lsp/null-ls.lua
+++ b/lua/lsp/null-ls.lua
@@ -24,10 +24,14 @@ function M.get_registered_providers_by_filetype(ft)
 end
 
 local function validate_nodejs_provider(requests, provider)
-  vim.cmd "let root_dir = FindRootDirectory()"
-  local root_dir = vim.api.nvim_get_var "root_dir"
+  local ts_client = require("utils").get_active_client_by_ft "typescript"
+  if ts_client == nil then
+    u.lvim_log "Unable to determine root directory since tsserver didn't start correctly"
+    return
+  end
+  local root_dir = ts_client.config.root_dir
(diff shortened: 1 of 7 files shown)


ChristianChiarulli:
This will break formatting with `node_modules`

kylo252:
> This will break formatting with `node_modules`

Do you mean if we disable vim-rooter? we can re-use the same `client.config.root_dir` query for that as well.

ChristianChiarulli:
right, I'm saying as is

kylo252:
> right, I'm saying as is

can you try it now?

lua/lsp/null-ls.lua
```diff
@@ -24,10 +24,14 @@ function M.get_registered_providers_by_filetype(ft)
 end
 
 local function validate_nodejs_provider(requests, provider)
-  vim.cmd "let root_dir = FindRootDirectory()"
-  local root_dir = vim.api.nvim_get_var "root_dir"
+  local ts_client = require("utils").get_active_client_by_ft "typescript"
```
abzcoding, line 27: I'm using prettier for markdown files, and this breaks that 
unknown

kylo252: resolved in 6c6fb67a8835cdcf911ef48ec2c19d553fe7d1a1

PR #1227, opened 2021-08-04
Allow calling `add_lsp_keybindings` from anywhere

### LINKED ISSUE #1219: [Bug] No lsp key bindings when language server is set up by plugin
I'm not sure if this is a bug or a feature but since this commit https://github.com/ChristianChiarulli/LunarVim/commit/77e283bd9c33166937756250918b12e349caf050 I lost my lsp key bindings for the rust_analyzer. I'm using the [rust-tools](https://github.com/simrat39/rust-tools.nvim) plugin which sets up the `rust_analyzer` on it's own. For it to work properly I have to include `lvim.lsp.override = { "rust" }` in my `config.lua` to avoid the plugin to conflict with the `lspconfig` plugin. Since the commit mentioned above the override takes away the keybindings (e.g. "gd", "gD", "K" etc.) which was not the case before. If this is a feature, what would be the best way to work around this issue?

kylo252:
# Description

Allow calling `add_lsp_keybindings` from anywhere.

Fixes #1219

## How Has This Been Tested?

Press `g` and wait for which-key prompt to confirm they're still working.

### DIFF
--- a/lua/keymappings.lua
+++ b/lua/keymappings.lua
@@ -151,6 +151,33 @@ function M.config()
   end
 end
 
+--- Add keybindings for LSP helper function (goto definition, etc)
+---@param bufnr will be used if specified, otherwise they will be global
+function M.add_lsp_keybindings(bufnr)
+  local keys = {
+    ["K"] = { "<cmd>lua vim.lsp.buf.hover()<CR>", "Show hover" },
+    ["gd"] = { "<cmd>lua vim.lsp.buf.definition()<CR>", "Goto Definition" },
+    ["gD"] = { "<cmd>lua vim.lsp.buf.declaration()<CR>", "Goto declaration" },
+    ["gr"] = { "<cmd>lua vim.lsp.buf.references()<CR>", "Goto references" },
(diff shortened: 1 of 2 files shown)


kylo252:
I'm not sure if we actually need this or not. We can judge based on the linked issue how smooth the current workflow is. 
We also have the possibility to move this to `core.which-key` but I figured it's better to have it in a more visible place.

abzcoding:
> I'm not sure if we actually need this or not. We can judge based on the linked issue how smooth the current workflow is.
> We also have the possibility to move this to `core.which-key` but I figured it's better to have it in a more visible place.

we don't need this, as i mentioned in the actual issue
```lua
function M.common_on_attach(client, bufnr)
  if lvim.lsp.on_attach_callback then
    lvim.lsp.on_attach_callback(client, bufnr)
  end
  add_lsp_buffer_keybindings(bufnr)  -- <- we have this already
  if lvim.lsp.smart_cwd then
    vim.api.nvim_set_current_dir(client.config.root_dir)
    require("core.nvimtree").change_tree_dir(client.config.root_dir)
  end
  require("lsp.null-ls").setup(vim.bo.filetype)
end
```

abzcoding:
that issue was related to `rust_tools` :)

PR #1241, opened 2021-08-05
Add LunarVim info panel (Experimental)

### LINKED ISSUE #1187: [Feature]: Create `:LvimLspInfo`
# Description

It would be nice to create a custom `LspInfo` like `LspInfoExtended` or `LvimLspInfo` or something. It would show what we should have setup in case there's a problem. This can also show and some other internal info about the config parameters that were sent to the language server.

After merging #1086 we would be to retrieves a list of the currently used null-ls providers by filetype.

## Some Ideas of what such a function could include

@rebuilt 
> - Which formatter/linter is actually being used
> - Formatter/linter options for that specified filetype
> - Attached lsp servers.
> - The output of :messages.
> - The output of grep ERROR ~/.cache/nvim/lsp.log  and also maybe the other log files (dap.log, packer.nvim.log)  
> - The output of :checkhealth 
> - lspconfig The output of :checkhealth treesitter
> - The important bits from nvim -v
> - Missing default plugins.

kylo252:
# Description

Add an info panel with some useful status reports for the current lunarvim configuration. 
It operates per files, and tries to give suggestions about setting up linters and formatters. 

Todos:
- [x] show missing providers
- [ ] accessibility concerns: colors, line wrapping, etc.
- [ ] use icons?
- [ ] add more/different info?

### Lua
image

### C (clangd supports formatting)
c

### typescript (forced formatting off)
ts

Fixes #1187 

## How Has This Been Tested?

```lua
-- Add this to your `config.lua` 
lvim.builtin.which_key.mappings["Li"] = {
	"<cmd>lua package.loaded['core.info']=nil; require('core.info').toggle_display(vim.bo.filetype)<cr>",
	"Toggle LunarVim Info",
}
```

### DIFF
--- a/lua/core/info.lua
+++ b/lua/core/info.lua
@@ -1,13 +1,8 @@
 local M = {}
 local lspui = require "lspconfig/_lspui"
-local job = require "plenary.job"
 local u = require "utils"
+local null_ls_handler = require "lsp.null-ls"
 local indent = "  "
-local sep = {
-  "",
-  "-------------------------------------------------------------------",
-  "",
-}
 
 M.banner = {
   "",
@@ -31,7 +26,7 @@ local function get_formatter_suggestion_msg(ft)
   return {
     "-------------------------------------------------------------------",
     "",
-    "HINT!",
+    "  HINT ",
     "",
     indent .. "* List of supported formatters: " .. str_list(supported_formatters),
     "",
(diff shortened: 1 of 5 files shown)


rebuilt:
Huh.  After yesterday's patch to lsp.init I expected lsp formatting to be on if no formatters were explicitly set  but it's still off, which this info panel makes clear.  What am I missing?  Why is lsp formatting off?

kylo252:
> Huh. After yesterday's patch to lsp.init I expected lsp formatting to be on if no formatters were explicitly set but it's still off, which this info panel makes clear. What am I missing? Why is lsp formatting off?

This info is about LSP client's capabilities, so it would show if they can actually do native formatting.

or did I get your question wrong?

rebuilt:
I think it's ready to merge as-is.  What's left is visual cleanup and that can be a separate pr.

PR #1244, opened 2021-08-05
[Feature] Add PowerShell language server support

PasiBergman:
# Description

Adds support for PowerShell language server (powershell_es) configuration

## How Has This Been Tested?

1. Installed Microsoft PowerShell: https://github.com/PowerShell/PowerShell/releases
2. Installed PowerShellEditorServices: https://github.com/PowerShell/PowerShellEditorServices/releases
3. Added configuration e.g. `lvim.lang.ps1.lsp.setup.bundle_path = "/path/to/PowerShellEditorServices"`
4. Opened a PowerShell `.ps1` file and saw powershell_es language server activated.

### DIFF
--- a/lua/default-config.lua
+++ b/lua/default-config.lua
@@ -1234,7 +1234,7 @@ lvim.lang = {
     },
   },
   gdscript = {
-    formatter = {},
+    formatters = {},
     linters = {},
     lsp = {
       provider = "gdscript",
@@ -1251,7 +1251,7 @@ lvim.lang = {
     },
   },
   ps1 = {
-    formatter = {},
+    formatters = {},
     linters = {},
     lsp = {
       provider = "powershell_es",
--- a/lua/default-config.lua
+++ b/lua/default-config.lua
@@ -1250,6 +1250,19 @@ lvim.lang = {
       },
     },
   },
+  ps1 = {
+    formatter = {},
+    linters = {},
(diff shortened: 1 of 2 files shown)


lua/default-config.lua
```diff
@@ -1250,6 +1250,19 @@ lvim.lang = {
       },
     },
   },
+  ps1 = {
+    formatter = {},
```
abzcoding, line 1254: it should be `formatters` i think

PasiBergman: Good point. I copy pasted the configuration above which (also) has the error. I will fix both.

abzcoding (approved):

kylo252:
Nice work!
1- I guess it's not supported by LspInstall? If that's the case, maybe add the instructions somewhere, probably in `ps1.lua` as a reference at least until we figure out a better way.
2- Does this support v7 only? If so, maybe we should add a warning or a disclaimer about that.

abzcoding:
document it [here](https://www.lunarvim.org/languages/) if you want

PasiBergman:
@kylo252 
Good idea, I'll look into it. I think only cross-platform PowerShell (basically v7+) is supported as the LS startup requires `pwsh`.

@abzcoding 
Thx. I'll document PowerShell and C# (csharp) as they are a bit niche.

PR #1248, opened 2021-08-06
bufferline broke because the added active toggle will always be false…

rebuilt:
…. Set it in default_config to fix behavior

# Description
New installs will not have barbar active.  An active check was added to plugins but it will always be false.  Call barbar.config from default_config to load the default value: true

Fixes #(issue)
Barbar not installed on fresh install or removed after a :PackerSync


rebuilt:
User reported error

kahgeh — Today at 1:45 AM
seems like the latest rolling install have broken barbar or it could be just some plugins may have broken it?

PR #1262, opened 2021-08-08
[Bugfix] Use vim.lsp.buf.formatting instead of vim.lsp.buf.formatting_sync

Tastyep:
I don't know why yet, but it lets us use multiple formatters.

# Description

Self-explanatory title. While reviewing [sequential formatting](https://github.com/jose-elias-alvarez/null-ls.nvim/pull/93) from null-ls, I observed that using `vim.lsp.buf.formatting_sync` did only run `black`, (if `isort` runs, the buffer is not updated).
I don't understand yet the rationale, so let's discuss the change before merging it. Maybe @jose-elias-alvarez could give us a hand.

## How Has This Been Tested?

``` lua
lvim.lang.python.formatters = {
  {
    exe = "black",
  },
  {
    exe = "isort",
  },
}

```

### DIFF
--- a/lua/utils/init.lua
+++ b/lua/utils/init.lua
@@ -64,7 +64,7 @@ function utils.toggle_autoformat()
         {
           "BufWritePre",
           "*",
-          ":silent lua vim.lsp.buf.formatting_sync()",
+          ":silent lua vim.lsp.buf.formatting()",
         },
       },
     }


jose-elias-alvarez:
As far as I can tell, the issue is that `buf.formatting_sync` uses `client.request_sync`, which uses a [sync handler](https://github.com/neovim/neovim/blob/a2909aa35faf941a30b1496e74c278accc3fb15b/runtime/lua/vim/lsp.lua#L919) that will wait for and return the first result it gets. `buf.formatting`, on the other hand, uses `client.request`, which in this case will [use the default formatting handler](https://github.com/neovim/neovim/blob/a2909aa35faf941a30b1496e74c278accc3fb15b/runtime/lua/vim/lsp.lua#L890) (which we can call as many times as we want). 

For LunarVim, I think changing the autocommand is a simple solution to produce the expected behavior. Fixing this on the null-ls side is a little trickier (overriding `client.request_sync`?) and I'll most likely recommend switching to `buf.formatting` there, too.

abzcoding:
@rebuilt didn't we face some issues with `BufWritePre` and `formatting` ?

kylo252:
What about manually triggering a format? are there problems and/or benefits to each with multiple formatters attach?

It's a niche use-case imo.

rebuilt:
> @rebuilt didn't we face some issues with `BufWritePre` and `formatting` ?

Yes.  We were trying to do formatting after the write: `BufWritePost`.  Changing the order to first run `formatting`, and then `write` worked correctly:  `BufWritePre`

So I tested the method above and it worked correctly for null-ls formatting but not for formatting performed by the LSP.  This change means that when using LSP for formatting, the formatting appears to take place but doesn't actually get written to disk.  This is confusing because if you write the file, you see it autoformat.  Then if you try to quit neovim says you have unwritten changes.  

I tested on typescript.  Python LSP formatting doesn't work for me.  I'm not sure why.

Tastyep:
This is [fixed](https://github.com/jose-elias-alvarez/null-ls.nvim/pull/93#issuecomment-894972832) :tada:

PR #1267, opened 2021-08-08
[Feature]: Add snippets navigation with c-j / c-k

leadrien:
# Description

Allow to cycle through snippets placeholders with `<C-j>` and `<C-k>`, the same way `<Tab>` / `<S-Tab>` works.
Navigation in completion menu still works the same (down and up selection) with those keymaps.

## How Has This Been Tested?

With python and c snippets.


ChristianChiarulli:
wouldn't ctrl-j and ctrl-k behave like tab normally with this change?

leadrien:
It will.

I kept the down/up menu selection with c-j/c-k because it was already in keymappings.lua
It's redundant with c-n/c-p, but it's consistent with lvim Telescope mapping which also uses c-j/c-k for selection.

This PR only aims to bring coc and other c-j/c-k snippets tools habits.

ChristianChiarulli:
I think I would need a separate function for c-j, c-k or find a way to parameterize the current one.

PR #1269, opened 2021-08-09
[Feature] Add LvimInfo command

PasiBergman:
# Description

Add command `:LvimInfo` to toggle/show LunarVim info panel (in addition to `<leader>Li`). 

## How Has This Been Tested?

- Open a file(type) for which you want to check LSP, formatter and/or linter info.
- Run command `:LvimInfo`

### DIFF
--- a/lua/core/commands.lua
+++ b/lua/core/commands.lua
@@ -10,6 +10,8 @@ M.defaults = {
     endif
   endfunction
   ]],
+  -- :LvimInfo
+  [[command! LvimInfo lua require('core.info').toggle_popup(vim.bo.filetype)]],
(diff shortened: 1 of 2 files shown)


leadrien:
Cool!
I guess you could use it in whichkey  `Li`

kylo252:
Nice! yeah I've had that on the back-burner for a while. 

Could you move it to [lua/core/commands.lua](https://github.com/LunarVim/LunarVim/blob/rolling/lua/core/commands.lua) instead?

PasiBergman:
Sure. lua/core/commands.lua looks like the right place.

Issue #1274, opened 2021-08-10
[Feature]: Can we talk about the change from ["<C-j>"] = actions.move_selection_next, to ["<C-j>"] = actions.cycle_history_next,


afridley:
### Feature motivation

I would like to change 
         ``` ["<C-j>"] = actions.cycle_history_next,```
          ```["<C-k>"] = actions.cycle_history_prev,```
back to
          ```["<C-j>"] = actions.move_selection_next,```
         ``` ["<C-k>"] = actions.move_selection_previous,```
and have
          ```["<C-n>"] = actions.cycle_history_next,```
          ```["<C-p>"] = actions.cycle_history_prev,```

To me being able to hold control and move up and down telescopes search is more useful then cycling the history.
It also matches my expectation with being able to hold ctrl j,k to move up and down lsp menus.

I would also be fine if there was just a config option that would switch it.

But maybe I'm missing something as to why it was changed

abzcoding:
you can change them to whatever you want
```lua
lvim.keys.normal_mode["<C-j>"] = "whatever"
lvim.keys.normal_mode["<C-n>"] = "whatever"
-- ...
```

afridley:
@abzcoding This is great as far as a config solution.

I still wonder if the default shouldn't be the other way around though to match the existing navigation for qflist and lsp menus. Since in both those you can use some form of j and k

afridley:
@abzcoding @ChristianChiarulli Also it looks like they are still j and k in normal mode just not insert mode so we should probably make those consistent.

      mappings = {
        i = {
          ["<C-n>"] = actions.move_selection_next,
          ["<C-p>"] = actions.move_selection_previous,
          ["<C-c>"] = actions.close,
          ["<C-j>"] = actions.cycle_history_next,
          ["<C-k>"] = actions.cycle_history_prev,
          ["<C-q>"] = actions.smart_send_to_qflist + actions.open_qflist,
          ["<CR>"] = actions.select_default + actions.center,
          -- To disable a keymap, put [map] = false
          -- So, to not map "<C-n>", just put
          -- ["<c-t>"] = trouble.open_with_trouble,
          -- ["<c-x>"] = false,
          -- ["<esc>"] = actions.close,
          -- Otherwise, just set the mapping to the function that you want it to be.
          -- ["<C-i>"] = actions.select_horizontal,
          -- Add up multiple actions
          -- You can perform as many actions in a row as you like
[...]

Issue #1137, opened 2021-07-27
Telescope lvim keybinding for Next/Previous item in insert mode are wierd


vlcinsky:
### Problem description

Open any Telescope search, e.g. to search a file `<space>f`

See few items and try to select 3rd file from the bottom.

I am used in lists of items to pick from to use <Ctrl-P> and <Ctrl-N> for Previous/Next item movement in the list which is visible.

Current keybinding in LunarVim are surprising as they do not change position in visible list but swap search text to next/previous item from search history.

### LunarVim version

rolling branch commit 66bf89c

### Neovim version (>= 0.5)

v0.5.0

### I have

- [X] Read the readme
- [X] Read the wiki
- [X] Searched for similar issues

ChristianChiarulli:
idk if it's a bug but it's something you could definitely change, what keybinds do you think should cycle history?

vlcinsky:
I would swam keymapping for next/previous history item with next/previous found item.
So for history it would be Ctrl-j and Ctrl-k so at https://github.com/ChristianChiarulli/LunarVim/blob/rolling/lua/core/telescope.lua#L46 it would change from original:
```lua
      mappings = {
        i = {
          ["<C-n>"] = actions.cycle_history_next,
          ["<C-p>"] = actions.cycle_history_prev,
          ["<C-c>"] = actions.close,
          ["<C-j>"] = actions.move_selection_next,
          ["<C-k>"] = actions.move_selection_previous,
          ["<C-q>"] = actions.smart_send_to_qflist + actions.open_qflist,
          ["<CR>"] = actions.select_default + actions.center,
```
to 
```lua
      mappings = {
        i = {
          ["<C-n>"] = actions.move_selection_next,
          ["<C-p>"] = actions.move_selection_previous,
          ["<C-c>"] = actions.close,
          ["<C-j>"] = actions.cycle_history_next,
          ["<C-k>"] = actions.cycle_history_prev,
          ["<C-q>"] = actions.smart_send_to_qflist + actions.open_qflist,
          ["<CR>"] = actions.select_default + actions.center,
```

PS: "Bug" is result of the issue template which seemed to be the most relevant.

vlcinsky:
[Default telescope.nvim mapping](https://github.com/nvim-telescope/telescope.nvim/blob/79644ab67731c7ba956c354bf0545282f34e10cc/lua/telescope/mappings.lua#L12) is also using
```lua
[...]

PR #1276, opened 2021-08-10
Please... reverse history and cycle next telescope bindings

afridley:
# Description

Reverses the bindings for cycle history and cycle next in telescope so that ctrl j and k go back to navigating through telescope lists while ctrl n and p will cycle the history of searches.

Fixes #(issue)
#1274 
## How Has This Been Tested?

- Open a project
- select find files
- use ctrl j and k to move up and down the list.
- use ctrl n and p to cycle history


ChristianChiarulli:
I liked this behavior too, could we just cycle history with tab and s-tab?

abzcoding:
as mentioned in #1137 and also [the Default telescope.nvim mapping](https://github.com/nvim-telescope/telescope.nvim/blob/79644ab67731c7ba956c354bf0545282f34e10cc/lua/telescope/mappings.lua#L12)

i think it should stay the way it is

abzcoding:
the-office-dont-you-dare

afridley:
@abzcoding This is great as far as a config solution.

I still wonder if the default shouldn't be the other way around though to match the existing navigation for qflist and lsp menus. Since in both those you can use some form of j and k.

though n and p are used in lsp menus as well but not on quickfix lists. Maybe quickfix navigation should also get n and p alongside j and k for consistency across the app.

afridley:
@abzcoding @ChristianChiarulli I think we need something like
```
lvim.builtin.which_key.mappings["t"]
```
except for telescope
```
lvim.builtin.telescope.mappings["i"]
lvim.builtin.telescope.mappings["n"]
```

I created a feature request for this here
https://github.com/LunarVim/LunarVim/issues/1279

afridley:
This works big thanks to @abzcoding 

```lua
lvim.builtin.telescope.on_config_done = function()
  local actions = require "telescope.actions"
  lvim.builtin.telescope.defaults.mappings.i["<C-j>"] = actions.move_selection_next
  lvim.builtin.telescope.defaults.mappings.i["<C-j>"] = actions.move_selection_next
  lvim.builtin.telescope.defaults.mappings.i["<C-k>"] = actions.move_selection_previous
  lvim.builtin.telescope.defaults.mappings.i["<C-n>"] = actions.cycle_history_next
  lvim.builtin.telescope.defaults.mappings.i["<C-p>"] = actions.cycle_history_prev
end
```

afridley:
@ChristianChiarulli @abzcoding This should be good to go. I couldn't do anything about the quickfix list as it just like navigating normally and I wasn't about to rebind n, p ctrl-n or ctrl-p to be up and down when editing text.

So I rebound n and p to move selection in telescopes normal mode as well as insert. Then added a comment in both configs that will overwrite both telescopes insert and normal mode to use j and k

PR #1277, opened 2021-08-10
[Refactor/Bugfix] Improve null ls handler

### LINKED ISSUE #1295: [Bug]: LvimInfo not listing eslint_d as supported linter for JavaScript family.
### Problem description

LvimInfo not listing eslint_d as supported linter for JavaScript family. Reason is, that there is no specific config for eslint_d linter, but elint config is used with cmd = eslint_d. LvimInfo should add eslint_d as supported linter when eslint is detected.

### LunarVim version

0.5.1

### Neovim version (>= 0.5)

0.5

### Relevant log output

_No response_

### Screenshots

_No response_

### I have

- [X] Read the readme
- [X] Read the wiki

### LINKED ISSUE #1299: [Bug]: LunarVim Info provides misleading instructions about configuring formatters
### Problem description

Steps to reproduce:
1. open e.g. python file thing.py
2. use `<space>Li` to show Lunarvim info page
3. it shows text as follows
```
  * List of supported formatters: [  ]
  * Configured formatter needs to be installed and executable.
  * Enable installed formatter(s) with following config in ~/.config/lvim/config.lua

    lvim.lang.dashboard.formatting = { { exe = '' } }
```
However, consulting https://github.com/LunarVim/LunarVim/blob/a3344203818658441d176fe279e1978505940432/lua/default-config.lua#L879-L886 the parameters for configuring should read:

```
  * List of supported formatters: [  ]
  * Configured formatters needs to be installed and executable.
  * Enable installed formatter(s) with following config in ~/.config/lvim/config.lua

    lvim.lang.dashboard.formatters = { { exe = '' } }
```

### LunarVim version

0.5.1 + (4fd72b1be2c1220a519d5b8d58219939a11df0c9)

### Neovim version (>= 0.5)

v0.5.0

### Relevant log output

```shell
logs are not relevant here, it is typo
```

### Screenshots

_No response_

### I have

- [X] Read the readme
- [X] Read the wiki

### LINKED ISSUE #1307: [Bug]: null-ls error on .lua files
### Problem description

I have an error when trying to open `.lua` file.
I don't set any formatters for `.lua` in my config.

### LunarVim version

rolling 03b7da

### Neovim version (>= 0.5)

0.6

### Relevant log output

_No response_

### Screenshots

image

_______
image

### I have

- [X] Read the readme
- [X] Read the wiki

Tastyep:
# Description

Improve the coding style for `null-ls.lua` (mainly).
Please tell me If you would like to change things.
Reviewing the changes might be difficult (side-by-side instead of split view might help greatly)

Fixes #1307
Fixes #1295 
Fixes #1299 

## How Has This Been Tested?

My config and languages I tested:

``` lua
lvim.lang.lua.formatters = {
  {
    exe = "stylua",
  },
}
lvim.lang.lua.linters = {
  {
    exe = "luacheck",
  },
  -- {
  --   exe = "selene",
  -- },
  {
    exe = "toti",
  },
}
lvim.lang.javascript.linters = {
  {
    exe = "eslint",
  },
}
lvim.lang.python.formatters = {
  {
    exe = "black",
  },
  {
    exe = "isort",
  },
}
lvim.lang.sh.formatters = {
  {
    exe = "shfmt",
    args = { "-i", "2", "-ci" },
  },
}
```

Locally installed providers are found.
`Using linter: /home/luc/Projects/Pi-OpenCast/webapp/node_modules/.bin/eslint`


lua/core/info.lua
```diff
@@ -134,36 +123,42 @@ function M.toggle_popup(ft)
...
 
+  local null_ls_providers = null_ls_handler.registered_providers_name(ft)
   local null_ls_info = {
     "Configured providers: " .. table.concat(null_ls_providers, "  , ") .. "  ",
     "",
   }
   vim.list_extend(buf_lines, null_ls_info)
 
-  local missing_formatters_status
+  local missing_formatters = null_ls_handler.langs[ft].errors.formatters
```
abzcoding, line 146: `null_ls_handler.langs[ft]` can never be `nil`, right?

kylo252 (commented):
It needed a refactor but I still think it can be better. I still don't like the fact that the setup function is doing a little bit of everything, with no clear and specific role. The formatter/linter validation needs to be extracted from this IMO.

lua/lsp/null-ls.lua
```diff
@@ -1,141 +1,136 @@
...
-        Log:get_default().info("Using linter provider", builtin_diagnoser.name)
+    local builtin_linter = adapt_linter(linter)
+    if not builtin_linter then
+      logger.error("Not a valid linter:", linter.exe)
+    elseif not vim.tbl_contains(M.langs[filetype].registered.linters, builtin_linter) then
+      local linter_cmd = find_command(builtin_linter._opts.command)
+      if not linter_cmd then
+        logger.warn("Not found:", builtin_linter._opts.command)
+        table.insert(M.langs[filetype].errors.linters, linter.exe)
       else
```
kylo252, line 124: no need for the else part if you're doing `if not ..`

lua/lsp/null-ls.lua
```diff
@@ -1,141 +1,136 @@
...
+      registered = {
+        formatters = {},
+        linters = {},
+      },
+      errors = {
+        formatters = {},
+        linters = {},
+      },
+    }
+  end
```
kylo252, line 97: I think we should go a for a proper metatable because of this much attached info, at least they can provide some functions.
I had that idea originally but ended up scrapping it.

lua/lsp/null-ls.lua
```diff
@@ -1,141 +1,136 @@
...
-    return
-  end
-  -- NOTE: we can't use provider.name because eslint_d uses eslint name
-  if vim.tbl_contains(nodejs_local_providers, provider._opts.command) then
-    return validate_nodejs_provider(provider)
-  end
-  if vim.fn.executable(provider._opts.command) ~= 1 then
-    Log:get_default().warn("Unable to find the path for", vim.inspect(provider))
-    return
+function M.registered_providers_name(ft)
```
kylo252, line 73: I think it might be a better with a more descriptive name, no pun intended :)
Is it returning a list/table/string?

lua/lsp/null-ls.lua
```diff
@@ -1,141 +1,136 @@
...
-    root_dir = vim.api.nvim_get_var "root_dir"
-  else
-    --- use LSP to set root_dir
-    local ts_client = require("utils").get_active_client_by_ft "typescript"
-    if ts_client == nil then
-      Log:get_default().error "Unable to determine root directory since tsserver didn't start correctly"
-      return
-    end
-    root_dir = ts_client.config.root_dir
+local function adapt_linter(linter)
```
kylo252, line 64: this looks like adapt "eslint" rather than adapt linter

lua/lsp/null-ls.lua
```diff
@@ -1,141 +1,136 @@
...
-    if vim.tbl_contains(provider.filetypes, ft) then
-      local provider_name = provider.name
+  return ts_client.config.root_dir
+end
 
-      table.insert(matches, provider_name)
+local function find_command(command)
+  if local_installations[command] then
+    local root_dir = find_root_dir()
+    local local_command = local_installations[command](root_dir, command)
```
kylo252, line 52: I feel like this is abstraction in the wrong direction. 
It's kinda confusing that `local_installations` is a list of callbacks instead of the actual path. From a glance you wouldn't guess that it would be. 
Secondly, why is it called `local_installation` when in fact it's relative to the `root_dir`?  where one  would expect it to be any arbitrary path that's set somehow. Instead there's the hidden dependency that the first arg needs to actually be fed as start point so that it can find it.

I'm not sure how this is better than using the clearly defined `validate_nodejs_providers`. Another side effect, is that we would lose the ability to log those failed register attempts more granularly.

lua/lsp/null-ls.lua
```diff
@@ -1,141 +1,136 @@
...
-
-  if vim.fn.executable(local_nodejs_command) == 1 then
-    command_path = local_nodejs_command
-  elseif vim.fn.executable(provider._opts.command) == 1 then
-    Log:get_default().debug("checking in global path instead for node module", provider._opts.command)
-    command_path = provider._opts.command
-  else
-    Log:get_default().debug("Unable to find node module", provider._opts.command)
-  end
-  return command_path
```
kylo252, line 74: These are the granular logs I was talking about. You would exactly know what it was searching for (local vs global) and why it couldn't find it.

lua/lsp/null-ls.lua
```diff
@@ -1,141 +1,136 @@
...
+      return local_command
     end
   end
 
-  return matches
-end
-
-local function register_failed_request(ft, provider, operation)
-  if not lvim.lang[ft][operation]._failed_requests then
-    lvim.lang[ft][operation]._failed_requests = {}
```
kylo252, line 42: Aren't you registering the failed request twice without this check?

lua/core/info.lua
Tastyep: I assume that [this](https://github.com/LunarVim/LunarVim/pull/1277/files#diff-61e47b385a127e61dcc5c97b396900ce66b93bf502422b58dcf30f5ba0566a09R86) should be called first

lua/lsp/null-ls.lua
```diff
@@ -1,141 +1,136 @@
...
+    elseif not vim.tbl_contains(M.langs[filetype].registered.formatters, builtin_formatter) then
+      local formatter_cmd = find_command(builtin_formatter._opts.command)
+      if not formatter_cmd then
+        logger.warn("Not found:", builtin_formatter._opts.command)
+        table.insert(M.langs[filetype].errors.formatters, formatter.exe)
       else
-        -- mark it here to avoid re-doing the lookup again
-        register_failed_request(filetype, formatter.exe, "formatters")
+        logger.info("Using formatter:", formatter_cmd)
+        table.insert(M.langs[filetype].registered.formatters, builtin_formatter.with { command = formatter_cmd })
```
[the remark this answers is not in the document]
Tastyep: Do you want to make the `langs` attribute private ?

lua/lsp/null-ls.lua
Tastyep: Yes, however, I don't see how does this check it either ;)

lua/lsp/null-ls.lua
Tastyep: I'll rework this, but I'd like to not make it specific to eslint in case there are more linters to support

lua/lsp/null-ls.lua
Tastyep: IMO, it was quite explicit by reading the logs whether the local or global installation is picked.

If it provides a local path, then the local binary is used:
image

Insist and I'll add those logs back :)

lua/lsp/null-ls.lua
Tastyep: If the linter is not found, then add it as failed, else add it as registered.

Tastyep:
Thank you for the reviews :heart: this is still a WIP, but I feel like we are getting there

lua/lsp/null-ls.lua
kylo252: > Yes, however, I don't see how does this check it either ;)

As long as we don't have duplicate entries I'm happy! I originally extracted `register_failed_request()` to reduce the number of nested if statements.

lua/lsp/null-ls.lua
kylo252: Oh I didn't know it was printing very often. We can just raise the log-level entry to debug in that case :)

lua/lsp/null-ls.lua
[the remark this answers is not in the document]
kylo252: As I said in another comment. they might be better served as a "class" and then we only expose certain get functions.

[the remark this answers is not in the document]
Tastyep: I kept the table approach (which is IMO the neatest way of implementing classes in Lua) and made the attribute a local variable.
The local variable is now only accessible via the object's (table) getters

Tastyep:
> can we merge this ASAP? It is so much better than what we currently have

Few things still need to be addressed:
- eslint_d special case is not handled yet (not that important)
- reloading on config updates is not 100% functional

I'll try to do this in the plane on my way back today ^^

lua/lsp/null-ls/init.lua
```diff
@@ -0,0 +1,62 @@
...
+    table.insert(names, name)
+  end
+
+  return names
+end
+
+function M.list_provider_names(filetype, options)
+  local names = {}
+
+  for _, providers in pairs(providers_by_ft[filetype]) do
```
abzcoding, line 20: can we check for `providers_by_ft[filetype]`not being nil 🥺

abzcoding:
Screen Shot 2021-08-11 at 8 58 22 PM


if i put this in my `ftplugin/python.lua`
```lua

lvim.lang.python.linters = {
  {
    -- can be pylint or flake8
    exe = "",
    args = {},
  },
}

```
we might not care about it though 😅

Tastyep:
> Screen Shot 2021-08-11 at 8 58 22 PM
> 
> if i put this in my `ftplugin/python.lua`
> 
> ```lua
> lvim.lang.python.linters = {
>   {
>     -- can be pylint or flake8
>     exe = "",
>     args = {},
>   },
> }
> ```
> 
> we might not care about it though sweat_smile

Well, that's because `""` is not a valid linter after all :sweat_smile:, to indicate that there is no linter it makes more sens to have

``` lua
lvim.lang.python.linters = {}
```

I already refactored `default-config.lua` to remove these invalid definitions

lua/lsp/null-ls/init.lua
Tastyep: Don't we have the guarantee that it can't be `nil` because the setup function would be called before ?

lua/lsp/utils.lua
```diff
@@ -0,0 +1,23 @@
...
+      return true
+    end
+  end
+  return false
+end
+
+function M.get_active_client_by_ft(filetype)
+  local clients = vim.lsp.get_active_clients()
+  for _, client in pairs(clients) do
+    if client.name == lvim.lang[filetype].lsp.provider then
```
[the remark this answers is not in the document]
Tastyep: Yep, we should care, I didn't pay attention to it while moving it.
For my defense, it already crashes :) :
https://github.com/LunarVim/LunarVim/blob/9a68500333660af76a2b7e75c3d7dcd38f8481b2/lua/utils/init.lua#L115

lua/lsp/null-ls/services.lua
```diff
@@ -0,0 +1,55 @@
+local M = {}
+
+local logger = require("core.log"):get_default()
```
kylo252, line 3: This is still not safe :)

lua/lsp/null-ls/services.lua
```diff
@@ -0,0 +1,55 @@
...
+    local local_command = local_providers[command].find(command)
+    if local_command and vim.fn.executable(local_command) == 1 then
+      return local_command
+    end
+  end
+
+  if vim.fn.executable(command) == 1 then
+    return command
+  end
+  return nil
```
kylo252, line 52: Can we add some kind of log entry here? It would go a long way. You can leave at "DEBUG" level if you'd like.

lua/lsp/utils.lua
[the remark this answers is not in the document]
kylo252: Yeah there's also another corner-case with `markdown` files needing to support `prettier`. So we need to make sure to not de-reference `lvim.lang[filetype].lsp.provider` in those cases :\

lua/utils/init.lua
```diff
@@ -99,55 +99,10 @@ function utils.reload_lv_config()
...
-  local clients = vim.lsp.get_active_clients()
-  for _, client in pairs(clients) do
-    if client.name == lvim.lang[filetype].lsp.provider then
-      return client
-    end
-  end
-  return nil
-end
+  local null_ls = require "lsp.null-ls"
+  null_ls.setup(vim.bo.filetype, { force_reload = true })
```
kylo252, line 103: Is this necessary? I feel like it might slow down the reload unintentionally, or is it about hot-reloading the configs?

lua/lsp/null-ls/init.lua
```diff
@@ -0,0 +1,78 @@
...
+  end
+  return list_provider_names(providers_by_ft[filetype].formatters, { filter = "unsupported" })
+end
+
+function M.list_unsupported_linter_names(filetype)
+  if not providers_by_ft[filetype] then
+    return {}
+  end
+  return list_provider_names(providers_by_ft[filetype].linters, { filter = "unsupported" })
+end
```
kylo252, line 57: nit: move these to their respective module and also provide an installer function to replace this 
```lua
  providers_by_ft[filetype] = {}
  providers_by_ft[filetype].formatters = null_formatters.list_configured(lvim.lang[filetype].formatters)
  providers_by_ft[filetype].linters = null_linters.list_configured(lvim.lang[filetype].linters)
  null_ls.register { sources = providers_by_ft[filetype].formatters.supported }
  null_ls.register { sources = providers_by_ft[filetype].linters.supported }
```
It will allow for changing the implementation details, if necessary, without modifying the interface.

kylo252 (approved):
Great work as always, much appreciated!

lua/utils/init.lua
Tastyep: Yep, this is done to hot reload providers, it works visually (since galaxyline shows them), but I'm not sure null-ls handles it (yet ?)

lua/lsp/null-ls/services.lua
Tastyep: My intent by removing logs from this function was to log only from the caller because it could provide additional context and also because the function can be reused in other configurations where we wouldn't want to log.
[This](https://github.com/LunarVim/LunarVim/pull/1277/files#diff-156f26605a183454e3e66901ae23c560faa0c1fd41377a40cd1ee0b022f4441cR30) is where I log this scenario.

lua/lsp/null-ls/services.lua
Tastyep: This one might be safe because it relies on null-ls being loaded (with plenary as dep), but that's not really the issue here.
IMO, the issue is that we return a nil logger when `plenary` is missing, this forces us to have `Log:get_default()` everywhere we want to log, instead of instantiating a module logger, this I really don't like.
Also, because of that, we still mix `print` calls where we know that the dependency could not yet be available.
A quick solution to solve this issue could be to provide a basic logger which would implement the same interface as `planary`'s one and dispatch the calls to `print`, this way we at least keep the same logging abstraction while providing basic capabilities.
Do you mind leaving this as is, expect `plenary` to be installed, and leave it to another PR instead ?

kylo252: Yeah I mentioned this same exact issue yesterday. We need to figure out a way to either have `plenary` as a direct dependency, (e,g. clone it while installing, or as sub-module,.. etc) or have own logger module, that could optionally be 1:1 feature-wise with `plenary.log`.

I'll create a ticket for this, so that we can see the pros and cons of each approach.

abzcoding:
@Tastyep 
can you change this line inside lua/core/info
```lua
    indent .. "  lvim.lang." .. tostring(ft) .. [[.formatting = { { exe = ']] .. table.concat(
```
to 
```lua
    indent .. "  lvim.lang." .. tostring(ft) .. [[.formatters = { { exe = ']] .. table.concat(
```

abzcoding (approved):

Tastyep:
So, I ran `git reset --hard upstream/rolling` and `git push -f`, while thinking I was on my master branch ...
Does anyone have a local copy of my work or do I need to recode everything ?

Tastyep:
Thank you, @abzcoding, for helping out, I'm saved (this time)

rebuilt:
Our docs tell users to set up formatters in this style:
```
lvim.lang.lua.formatters = {
	{
		exe = "stylua",
		args = {},
	},
}

```

But this refactor requires that users remove the empty `args` declaration.  

Otherwise they'll get the following errors

```
Error executing vim.schedule lua callback: ...e/pack/packer/
start/null-ls.nvim/lua/null-ls/helpers.lua:167: error in gen
erator output: error: The following required arguments were
not provided:
    <files>...
```

Is it possible to handle the empty args gracefully or do all users have to fix their configs?

Tastyep:
> Our docs tell users to set up formatters in this style:
> 
> ```
> lvim.lang.lua.formatters = {
> 	{
> 		exe = "stylua",
> 		args = {},
> 	},
> }
> ```
> 
> But this refactor requires that users remove the empty `args` declaration.
> 
> Otherwise they'll get the following errors
> 
> ```
> Error executing vim.schedule lua callback: ...e/pack/packer/
> start/null-ls.nvim/lua/null-ls/helpers.lua:167: error in gen
> erator output: error: The following required arguments were
> not provided:
>     <files>...
> ```
> 
> Is it possible to handle the empty args gracefully or do all users have to fix their configs?

TBH, I was hoping that nobody would notice :smile:, there is a problem when "extending" arguments.
[Null-ls](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/helpers.lua#L281) does a `vim_tbl_deep_extend`, but this doesn't provide the right behavior for extending builtins arguments.
example:

``` lua
local opts = {
  args = {
    -- Disable printing version information to stderr
    "-q",
    -- Disable executing \input statements
    "-I0",
    -- Format output
    "-f%l:%c:%d:%k:%m\n",
  },
}

local override = {
  args = { "-v" },
}

print("OPTS:", vim.inspect(vim.tbl_deep_extend("force", opts, override)))
```

Outputs:

``` bash
OPTS: {
  args = { "-v" }
}
```

So when providing an empty table `args = {}`, which is not the desired behavior just for the reason that the output format specification is lost, meaning that the builtin's output couldn't be parsed anymore.

A workaround would probably be to use `vim.list_extend` on the private attributes, I'll also open an issue on `null-ls` repo.

Tastyep:
It's not possible ATM to extend a builtin's args, but it might be soon

https://github.com/jose-elias-alvarez/null-ls.nvim/issues/105

Tastyep:
IMO this is ready to be merged.
Yes, there is the issue @rebuilt mentioned about empty args, but it's something that will be fixed on null-ls' side.

rebuilt (approved):
The empty argument problem isn't specific to this pull request.  It seems like they don't work in rolling either.  So, regardless, users will need to remove empty args from their configs until the issue is resolved upstream.

abzcoding (approved):
just tested this again locally, seems to be working fine
@ChristianChiarulli just waiting for you to take a look 😅 
so we can merge it

Tastyep:
Removed `eslint_d` adapter as it's now provided by null-ls
https://github.com/jose-elias-alvarez/null-ls.nvim/commit/4eb37a9a6d094edd2974be4ce195082e5dd894a4

ChristianChiarulli:
just tested it, seems like it works to me, so you guys can merge when ready @abzcoding @Tastyep @kylo252

PR #1281, opened 2021-08-10
[Refactor] LunarVim Info UI

PasiBergman:
# Description

Make LunarVim Info UI more compact.

- Run command `:LvimInfo` on LS activated filetype buffer

### DIFF
--- a/lua/core/info.lua
+++ b/lua/core/info.lua
@@ -4,15 +4,20 @@ local null_ls_handler = require "lsp.null-ls"
 local indent = "  "
 
 M.banner = {
+  indent .. "Press q to close",
   "",
-  "",
-  "⠀⣿⡟⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀   ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀  ⠀⠀     ⠀⠀⠀   ⠀⠀ ⣺⡿⠀⠀⠀⠀⠀⠀⠀ ⠀⠀⠀",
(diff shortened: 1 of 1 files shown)


lua/core/info.lua
```diff
@@ -4,15 +4,20 @@ local null_ls_handler = require "lsp.null-ls"
 local indent = "  "
 
 M.banner = {
+  indent .. "Press q to close",
```
kylo252, line 7: My only real complaint is this. I think it looks off. I tried giving it a big offset but I'm not sure it's that much better anyway
```lua
  local function get_adaptive_dismiss_hint(window_width)
    local msg = " "
    local hint = "Press q to dismiss"
    local padding = 4
    for i = 1, (window_width - (string.len(hint) + padding)) do
      msg = msg .. " "
    end
    return { "", msg .. hint, "" }
  end

  vim.list_extend(buf_lines, get_adaptive_dismiss_hint(math.floor(vim.o.columns * 0.7)))
```

Maybe we should see how Telescope embeds it in the frame.

lua/core/info.lua
```diff
@@ -126,42 +133,47 @@ function M.toggle_popup(ft)
...
     "",
   }
   vim.list_extend(buf_lines, header)
 
   local lsp_info = {
-    "Associated language-server: " .. client.name,
-    indent .. "* Active: " .. tostring(is_client_active) .. ", id: " .. tostring(client.id),
-    indent .. "* Formatting support: " .. tostring(client.resolved_capabilities.document_formatting),
-    indent .. "* Capabilities list: " .. table.concat(vim.list_slice(client_enabled_caps, 1, num_caps / 2), ", "),
+    indent .. "Language Server",
```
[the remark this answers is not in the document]
PasiBergman: Renamed line to suggested one - line is not too long.

lua/core/info.lua
PasiBergman: Removed "Press q to close"

ChristianChiarulli:
@kylo252 are we happy with this yet?

kylo252 (approved):

PR #1284, opened 2021-08-10
fix: do not overwrite plugins bindings in reload

kylo252:
# Description

We should source the keybindings before loading the plugins, so that we don't overwrite any new ones.

## How Has This Been Tested?

Some plugins like "aserowy/tmux.nvim" uses C-hjkl for movement, and these should take effect now.


abzcoding (approved):
seems fine

PR #1287, opened 2021-08-10
Seach hidden files via telescope (excluding git)

gikari:
# Description

Telescope functions like `find_files` did not include hidden
files in the results. These include like files from GNU stow repos
(`.config`, `.zshrc` etc) and files from various programming projects
like `.clang-format` and `.dockerignore`. Not including these in search
results was kinda strange.

Fixes #(issue)

## How Has This Been Tested?

- Run command `:mycommand`
- Check logs
- ...


tami5:
One of the reasons we haven't made as default is that hidden files are considered secondary when working with files and only applicable to system level configurations.

Though this gave me an idea, maybe a shortcut in the picker to toggle hidden files 🤔

tami5:
One of the reasons we haven't made as default in telescope is that hidden files are considered secondary when working with files and only applicable to system level configurations.

Though this gave me an idea, maybe a shortcut in the picker to toggle hidden files 🤔

gikari:
> One of the reasons we haven't made as default in telescope is that hidden files are considered secondary when working with files and only applicable to system level configurations.
> 
> Though this gave me an idea, maybe a shortcut in the picker to toggle hidden files 🤔

Ideally I would rather have a **single** configuration toggle, like with nvim_tree. Is it ok to go this way? Toggle is also nice, but pressing the one extra button when you're working on various programming projects with with hidden files would be annoyning and unnessary.

kylo252:
I'm not sure this is the best way to do it. I was going to suggest just using a different argument to `fd` but then I noticed the find command isn't even specified..

gikari:
>  So maybe we should actually force these as defaults.

Do you mean, that we should ask Telescope maintainers to make their built-in pickers search for hidden files by default?

kylo252:
> Do you mean, that we should ask Telescope maintainers to make their built-in pickers search for hidden files by default?

Yeah both `fd` and `rg` can do that while still respecting gitignore.

gikari:
It seems like Telescope devs won't set hidden as a default, at least according to the https://github.com/nvim-telescope/telescope.nvim/issues/525

yassinebridi:
This is what I did, searching for files and text using [ripgrep](https://github.com/BurntSushi/ripgrep)

```lua
-- Finding files, this will search all files including the hidden ones, but the folders specified below.

lvim.builtin.which_key.mappings["f"] = { "<cmd>:lua require'telescope.builtin'.find_files({ find_command = {'rg', '--files', '--hidden', '-g', '!.git', '-g', '!node_modules', '-g', '!.next' }})<CR>", "Find Files" }

-- Finding text, this will search all text in all the files in the current workspace including hidden files, but the folders and files specified below.

lvim.builtin.which_key.mappings["rg"] = { "<cmd>:lua require'telescope.builtin'.live_grep({ vimgrep_arguments = { 'rg', '--hidden','--color=never', '--no-heading', '--with-filename', '--line-number', '--column', '--smart-case', '-u', '-g', '!yarn.lock', '-g', '!.git', '-g', '!node_modules', '-g', '!.next' } })<CR>", "Find Text" }
```

lua/core/telescope.lua
```diff
@@ -32,7 +32,9 @@ function M.config()
         vertical = { mirror = false },
       },
       file_sorter = require("telescope.sorters").get_fzy_sorter,
-      file_ignore_patterns = {},
+      file_ignore_patterns = {
+        ".git/",
```
felippepuhle, line 36: can we use some custom patterns here?

gikari: Do you mean the config option should be exposed, so the user could set the ignore patterns in the config.lua?

felippepuhle: yeah that would be awesome

kylo252:
Sorry for the late response, I got busy with all the other refactors that I didn't get a chance to give a proper feedback for a while now.

If you take a look at the implementation of [file_ignore_pattern](https://github.com/nvim-telescope/telescope.nvim/blob/ba41f0eeb13796c26f6c5f3e8c44b2397a466e39/lua/telescope/pickers.lua#L1009), you'll see that it doesn't exactly do what you'd expect. Instead we want to set a _better_ default command for `find_files`, which is what I wanted to add in https://github.com/nvim-telescope/telescope.nvim/pull/1304, but I got a response from connie that we can already achieve by setting these options:
https://github.com/LunarVim/LunarVim/blob/a3b03e304a3e38d4c84e6e2b1d11e6c8b2e5352c/lua/core/telescope.lua#L49-L52

I would really appreciate some feedback there to see what else we should tweak in [feat: better telescope integration (WIP) #1702](https://github.com/LunarVim/LunarVim/pull/1702).

---

Alternatively, there's a native way to do this: 
- fd: https://github.com/kylo252/dotfiles/blob/950f353d46a52ed34eea8c006127022e247da782/.config/fd/ignore#L1
- ripgrep: https://github.com/kylo252/dotfiles/blob/950f353d46a52ed34eea8c006127022e247da782/.config/ripgrep/ripgreprc#L13-L16

PR #1289, opened 2021-08-11
Fix lunarvim info nil issue

abzcoding:
# Description

fix nill issues in LunarVim info when there are no active LS or the lang is not in `default_config.lua`

### DIFF
--- a/lua/core/info.lua
+++ b/lua/core/info.lua
@@ -130,13 +130,26 @@ end
 
 function M.toggle_popup(ft)
   local client = u.get_active_client_by_ft(ft)
-  local is_client_active = not client.is_stopped()
-  local client_enabled_caps = require("lsp").get_ls_capabilities(client.id)
+  local is_client_active = false
+  local client_enabled_caps = {}
+  local client_name = ""
+  local client_id = 0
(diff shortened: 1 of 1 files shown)


abzcoding:
image

lua/core/info.lua
```diff
@@ -130,13 +130,26 @@ end
...
+  local missing_linters = {}
+  local missing_formatters = {}
   local num_caps = vim.tbl_count(client_enabled_caps)
   local null_ls_providers = null_ls_handler.get_registered_providers_by_filetype(ft)
-
-  local missing_linters = lvim.lang[ft].linters._failed_requests or {}
-  local missing_formatters = lvim.lang[ft].formatters._failed_requests or {}
+  if client ~= nil then
+    is_client_active = not client.is_stopped()
+    client_enabled_caps = require("lsp").get_ls_capabilities(client.id)
```
PasiBergman, line 144: Should the `num_caps` be calculated again after getting them?

abzcoding: yes they should you are right

PasiBergman (approved):
Looks fine to me.

PR #1294, opened 2021-08-11
feat: make autopairs configurable

kylo252:
# Description

Let autopairs be configurable through `lvim.builtin.autopairs`

## How Has This Been Tested?

Check that autopairs still works.


lua/core/autopairs.lua
```diff
@@ -1,54 +1,63 @@
...
+  local ts_conds = require "nvim-autopairs.ts-conds"
 
-local ts_conds = require "nvim-autopairs.ts-conds"
+  -- TODO: can these rules be safely added from "config.lua" ?
+  -- press % => %% is only inside comment or string
+  npairs.add_rules {
+    Rule("%", "%", "lua"):with_pair(ts_conds.is_ts_node { "string", "comment" }),
+    Rule("$", "$", "lua"):with_pair(ts_conds.is_not_ts_node { "function" }),
+  }
+end
```
Tastyep, line 61: Return `npairs` so that it can be passed to on_config_done

lua/plugins.lua
```diff
@@ -68,11 +68,12 @@ return {
     -- event = "InsertEnter",
     after = "nvim-compe",
     config = function()
-      require "core.autopairs"
+      require("core.autopairs").setup()
```
Tastyep, line 71:
``` lua
local autopairs = require("core.autopairs").setup()
...
lvim.builtin.autopairs.on_config_done(autopairs)
```

lua/core/autopairs.lua
```diff
@@ -1,54 +1,63 @@
...
-  return
+local M = {}
+
+function M.config()
+  lvim.builtin.autopairs = {
+    active = true,
+    ---@usage  map <CR> on insert mode
+    map_cr = true,
+    ---@usage auto insert after select function or method item,
+    map_complete = vim.bo.filetype ~= "tex",
```
Tastyep, line 9: Probably out of scope for this PR, but this only works if the first file I open is a `tex` file.
Should we make this a function and call `require("nvim-autopairs.completion.compe").setup` when the file is changed ? IDK, in the meantime maybe add a note.

Tastyep (commented):
I tested it, and it stopped working for me

``` lua
local t = {
```

PR #1303, opened 2021-08-12
[Feature] Add a new option to show the mode name in galaxyline

mhcerri:
# Description

Add the new config variable "lvim.builtin.galaxyline.show_mode" to
control if the mode name should be shown in the galaxyline. The option
defaults to false to preserve the current behavior.

## How Has This Been Tested?

Tested with `lvim.builtin.galaxyline.show_mode = false`, `lvim.builtin.galaxyline.show_mode = true` and without `lvim.builtin.galaxyline.show_mode` in the `config.lua`


mhcerri:
Fixed luacheck warning related non local function.

PR #1310, opened 2021-08-13
feat: use telescope to quickly open lunarvim files

kylo252:
# Description

Add two simple functions that allow opening and grepping LunarVim Files from anywhere.

## How Has This Been Tested?

Use `<leader>L` to see the new options.


abzcoding (approved):
seems to be working 🎉

PR #1315, opened 2021-08-14
[Refactor]: Remove vim-rooter and smart-cwd; then use project.nvim

### LINKED ISSUE #1304: [Mega Refactor]: Remove vim-rooter and smart-cwd; then use project.nvim
### Feature motivation

I am always frustrated when vim-rooter triggers on useless filetypes like nvim-tree or help.
I am always frustrated when vim-rooter globs my entire directory for all matches, instead of stopping at the first match.
I want a 100% Lua solution. I want telescope integration.

Introducing project.nvim. It provides:

- Libuv based file indexing. (It was originally multithreaded, but after @mjlbach gave me some tips, I decided to stop the program complexity and make it synchronous because it uses only 0.12ms per call on my system.)
- Has both an LSP and a glob engine. So the patterns from vim-rooter will work as expected in project.nvim.
- Only triggers on real files. Not dashboard. Not function signature popups. Not help...
- View project history using Telescope :nerd_face:
  - and as a plus, the history get's read asynchronously at startup so it will not slow down startup time.

Will help fix: #910

So I talked to @kylo252 over on discord and he said "my plan was to only use vim-rooter as a backup option in this context, but I think we decided to leave it as default for compatibility reason". And by context, he means the default on_attach handler for lunarvim's language server configs.

Unfortunately, this is impossible under the current way neovim handles LSP. Furthermore, it will lead to bugs if someone is working on many projects in the same neovim session. Allow me to explain:

The current way things work in neovim is by calling a function on_attach that is a callback to when a certain lsp server attaches itself to a buffer. The problems:
1. this callback is only called once.

This is problematic when having two LSP-based projects on the same neovim session. Guess what: currently, lunarvim will only change the directory to one of those projects and will ignore the latter.

2. this callback can get called anytime
3. an lsp is virtually always not attached to the buffer `OnBufEnter`

Those two are also big problems because it means we can't just have a backup as @kylo252 says, without sacrificing some performance by having both a buf_enter autocmd and an attach callback.

### Additional context

This is how telescope integration looks like:

https://user-images.githubusercontent.com/36672196/129220312-88cd8ae1-0f78-4d26-81e1-9eb23b86624a.mp4

Finally, I need to have the "ok" from major contributors of this configuration, as I don't want to waste time making a PR, only for it to be rejected.

@ChristianChiarulli @abzcoding @PasiBergman @kylo252 ...

ahmedkhalf:
# Description

Remove vim-rooter and smart-cwd; then use project.nvim.
This is the first step towards project management.

Fixes #1304

## How Has This Been Tested?

TBA


lua/core/project.lua
```diff
@@ -0,0 +1,49 @@
+local M = {}
+function M.config()
+  lvim.builtin.project = {
```
Tastyep, line 3: Instead of duplicating your default config here, can you instead extend `lvim.builtin.project` ?
Otherwise, if you add entries to your defaults, this is not updated.
This should probably be done for all other plugins ... (not in the scope of this PR)

lua/core/project.lua
```diff
@@ -0,0 +1,49 @@
...
+    -- detection_methods
+    patterns = { ".git", "_darcs", ".hg", ".bzr", ".svn", "Makefile", "package.json" },
+
+    -- When set to false, you will get a message when project.nvim changes your
+    -- directory.
+    silent_chdir = true,
+  }
+end
+function M.setup()
+  require("project_nvim").setup {
```
Tastyep, line 28: Also, instead of having to match manually all keys, it would be better if you put the plugin's setup config in a separate table, which you would be able to pass directly here instead.

lua/core/project.lua
ahmedkhalf: I can't do this unless I am in M.setup() this way I can import the defaults:
```lua
lvim.builtin.project = require("project_nvim.config").defaults
```
But then it would break the autocompletion.

Tastyep: Arf, I hate it that we duplicate this, but again it's not specific to this so let's postpone.

ahmedkhalf: okay :+1:

lua/core/project.lua
ahmedkhalf: Done

abzcoding:
Also we might need someone to test this on frontend stuff `(eslint/prettier/...)`

@ChristianChiarulli maybe you can help us here 🥺

abzcoding:
project

ahmedkhalf:
packersync then packercompile @abzcoding

abzcoding:
> packersync then packercompile @abzcoding

did that already, didn't help

abzcoding:
should be merged after #1318

lua/core/project.lua
```diff
@@ -0,0 +1,56 @@
...
+    pcall(vim.cmd, [[packadd telescope.nvim]])
+  end
+
+  if package.loaded["telescope"] then
+    lvim.builtin.dashboard.custom_section["b"] = {
+      description = { "  Recent Projects    " },
+      command = "Telescope projects",
+    }
+    require("telescope").load_extension "projects"
+  end
```
kylo252, line 53: I feel like this is done a bit backwards, shouldn't the call that is wrapped and not the definition? 

How about something like this?

```lua
lvim.builtin.dashboard.custom_section["b"] = {
  description = { "  Recent Projects    " },
  command = "<cmd>lua require('telescope').load_extension('projects'); require('telescope').extensions.projects.projects()<CR>"
}
pcall(require, ("telescope").load_extension("projects"))
```

ahmedkhalf:
@kylo252 Please test the new changes and give me ur feedback :)

rebuilt (commented):
So far everything works as expected.  Switching projects changes the root directory.   Opening a new project adds it to the list of recent projects.  Editing a file with `:e ~/directory/filename.txt` also changes the working directory.  The edit action does not add an entry in telescope projects.   I think this is the default behavior of that plugin.  `Projects` correctly changes the root directory without dashboard being installed.  `Projects` can be used on its own and triggered with `Telescope projects`.

ChristianChiarulli:
Does everyone feel good about this?

ChristianChiarulli:
Everything seems to work for me so far.

lua/core/project.lua
```diff
@@ -0,0 +1,48 @@
...
+
+    -- Methods of detecting the root directory. **"lsp"** uses the native neovim
+    -- lsp, while **"pattern"** uses vim-rooter like glob pattern matching. Here
+    -- order matters: if one is not detected, the other is used as fallback. You
+    -- can also delete or rearangne the detection methods.
+    detection_methods = { "lsp", "pattern" },
+
+    -- All the patterns used to detect root dir, when **"pattern"** is in
+    -- detection_methods
+    patterns = { ".git", "_darcs", ".hg", ".bzr", ".svn", "Makefile", "package.json" },
```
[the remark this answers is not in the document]
ahmedkhalf: Honestly I just copied the patterns from the vim rooter guys, so they must have a good reason to add them, also yes, glob like patterns will work just fine.

PR #1317, opened 2021-08-14
[Bugfix] Support extending null builtin args

Tastyep:
# Description

Support [has been added](https://github.com/jose-elias-alvarez/null-ls.nvim/pull/106) to null-ls for extending builtins args :tada: 

## How Has This Been Tested?

``` lua
lvim.lang.sh.formatters = {
  {
    exe = "shfmt",
    args = { "-i", "2", "-ci" },
  },
}
```

### DIFF
--- a/lua/lsp/null-ls/formatters.lua
+++ b/lua/lsp/null-ls/formatters.lua
@@ -59,7 +59,7 @@ function M.list_configured(formatter_configs)
         errors[fmt_config.exe] = {} -- Add data here when necessary
       else
         logger.info("Using formatter:", formatter_cmd)
-        formatters[fmt_config.exe] = formatter.with { command = formatter_cmd, args = fmt_config.args }
+        formatters[fmt_config.exe] = formatter.with { command = formatter_cmd, extra_args = fmt_config.args }
       end
     end
   end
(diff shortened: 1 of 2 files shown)


abzcoding:
do these args extend the default args in `null-ls` and if so, how can one go about replacing them instead of extending them?

Tastyep:
> do these args extend the default args in `null-ls` and if so, how can one go about replacing them instead of extending them?

They do extend instead of overwrite, there is a discussion about it [here](https://github.com/jose-elias-alvarez/null-ls.nvim/issues/105)

TLDR: null-ls only provides arguments that are useful for itself [example](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/diagnostics.lua#L251), these should be left unchanged.
I don;t see a situation where you would want to overwrite them, but you would do it like this: `builtin.with({args = your_args})`

abzcoding (approved):
seems to be working fine, but we should warn users about updating their null-ls or sth 😅

Tastyep:
> seems to be working fine, but we should warn users about updating their null-ls or sth sweat_smile

Why ? I don't think it would break anything.

abzcoding:
> > seems to be working fine, but we should warn users about updating their null-ls or sth sweat_smile
> 
> Why ? I don't think it would break anything.

some might have tried to override the `null-ls` config for example for `isort` instead of `--profile black` they had sth else, but now I don't know what happens since we are extending instead of overriding, but i think we should merge this anyways

Tastyep:
This is an issue on null-ls side, I'll review the builtins tomorrow and PR removal of unnecessary parameters.

PR #1320, opened 2021-08-14
[Feature] Encapsulate interface logic

Tastyep:
# Description

Encapsulate popup creation.
Provide a function for aligning text relatively to its container.
Provide a function for pretty formatting tables.
Use `strings.format` instead of the `..` operator

Make it easier for implementing new popups. 

I add to change the banner because it used non-ASCII characters which caused invalid text alignment.
I used [this website](http://www.patorjk.com/software/taag/#p=testall&f=Graffiti&t=LunarVim), pick another one if you prefer.

## How Has This Been Tested?

`<space>Li`

Small:
image

Medium:
image

Big:
image

No LSP
image


lua/interface/text.lua
```diff
@@ -0,0 +1,27 @@
+local M = {}
+
+--- Center align lines relatively to its container
```
Tastyep, line 3: to its container  -> to the parent container

kylo252 (approved):
LGTM!

lua/core/info.lua
```diff
@@ -136,67 +57,89 @@ function M.toggle_popup(ft)
...
+    fmt("* Associated server:    %s", client_name),
+    fmt("* Active:               %s (id: %d)", tostring(is_client_active), client_id),
+    fmt("* Supports formatting:  %s", tostring(document_formatting)),
   }
-  vim.list_extend(buf_lines, lsp_info)
+  local caps_text = "* Capabilities list:    "
+  local caps_text_len = caps_text:len()
+  local enabled_caps = text.format_table(client_enabled_caps, 3, " | ")
+  enabled_caps = text.shift_left(enabled_caps, caps_text_len)
+  enabled_caps[1] = fmt("%s%s", caps_text, enabled_caps[1]:sub(caps_text_len + 1))
```
abzcoding, line 84: if you mistakenly open it in terminal 😅 
Screen Shot 2021-08-16 at 3 39 35 PM

abzcoding: also if i open a `test.fish` file and run the same command 😢

lua/interface/popup.lua
```diff
@@ -0,0 +1,58 @@
...
+
+  local default_layout = {
+    relative = "editor",
+    height = math.ceil(vim.o.lines * 0.9),
+    width = math.floor(vim.o.columns * 0.8),
+    style = "minimal",
+    border = "rounded",
+  }
+  default_layout.col = (vim.o.columns - default_layout.width) / 2
+  default_layout.row = (vim.o.lines - default_layout.height) / 2
```
PasiBergman, line 24: I.e. if height calculation is change, then L24 should be something like
```lua
default_layout.row = (vim.o.lines - default_layout.height) / 2 - vim.o.cmdheight
```

lua/core/info.lua
```diff
@@ -1,128 +1,49 @@
...
+local fmt = string.format
+
 local function str_list(list)
-  return "[ " .. table.concat(list, ", ") .. " ]"
+  return fmt("[ %s ]", table.concat(list, ", "))
 end
 
 local function get_formatter_suggestion_msg(ft)
   local null_formatters = require "lsp.null-ls.formatters"
   local supported_formatters = null_formatters.list_available(ft)
```
PasiBergman, line 20: Maybe we could add condition after L20
```lua
if #supported_formatters == 0 then
  return { }
end
```

lua/core/info.lua
```diff
@@ -1,128 +1,49 @@
...
+    "* Configured formatter needs to be installed and executable.",
+    fmt("* Enable installed formatter(s) with following config in %s", USER_CONFIG_PATH),
     "",
+    fmt("  lvim.lang.%s.formatters = { { exe = '%s' } }", ft, table.concat(supported_formatters, "│")),
   }
 end
 
 local function get_linter_suggestion_msg(ft)
   local null_linters = require "lsp.null-ls.linters"
   local supported_linters = null_linters.list_available(ft)
```
PasiBergman, line 34: Maybe we could add condition after L34
```lua
if #supported_linters == 0 then
  return { }
end
```

lua/core/info.lua
PasiBergman, line 84: `enabled_caps` could be empty => cannot index [1]

PasiBergman (changes requested):

lua/core/info.lua
abzcoding: https://github.com/LunarVim/LunarVim/pull/1320#discussion_r689443364

PasiBergman (approved):

PR #1323, opened 2021-08-15
[Feature] Expose null ls setup to user config

dklymenk:
# Description

This feature allows user to pass a custom table to null_ls setup function.

## How Has This Been Tested?

I have tested this with `root_dir` overwrite.
Example config: 
```
lvim.lsp.null_ls.setup.root_dir = require("lspconfig").util.root_pattern("Makefile", ".git", "node_modules")
```

With this I can see null_ls root set to the `app` folder when opening a file in a project with structure like this
```
repo/
├── .git/
└── app/
    ├── node_modules/
    ├── App.js
    ├── package.json
    └── yarn.lock
```

## Why is this useful?

Eslint and babel sometimes have weird interactions, where if you're running eslint or eslint_d from repo root and not `app` folder (see project structure above), you might run into errors where eslint can't find your babel config because it either looks for it from cwd and upwards or only cwd.

Making sure `null_ls` root is set correctly is the easiest workaround for my case (if you don't want to add random empty makefiles around the projects you work on).

### DIFF
--- a/utils/installer/config.example-no-ts.lua
+++ b/utils/installer/config.example-no-ts.lua
@@ -73,7 +73,7 @@ lvim.builtin.treesitter.highlight.enabled = true
 --   elseif vim.bo.filetype == "php" then
 --     return require("lspconfig/util").root_pattern("Makefile", ".git", "composer.json")(fname) or vim.fn.getcwd()
 --   else
---     return require("lspconfig/util").root_pattern("Makefile", ".git")(fname)
+--     return require("lspconfig/util").root_pattern("Makefile", ".git")(fname) or require("lspconfig/util").path.dirname(fname)
 --   end
 -- end
 
--- a/utils/installer/config.example-no-ts.lua
+++ b/utils/installer/config.example-no-ts.lua
@@ -63,18 +63,18 @@ lvim.builtin.treesitter.highlight.enabled = true
 -- end
 -- you can overwrite the null_ls setup table (useful for setting the root_dir function)
 -- lvim.lsp.null_ls.setup = {
---   root_dir = require("lspconfig").util.root_pattern("Makefile", ".git", "node_modules")
+--   root_dir = require("lspconfig").util.root_pattern("Makefile", ".git", "node_modules"),
 -- }
 -- or if you need something more advanced
 -- lvim.lsp.null_ls.setup.root_dir = function(fname)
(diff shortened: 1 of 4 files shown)


Desideriadez (approved):

utils/installer/config.example-no-ts.lua
```diff
@@ -61,6 +61,21 @@ lvim.builtin.treesitter.highlight.enabled = true
...
+--   root_dir = require("lspconfig").util.root_pattern("Makefile", ".git", "node_modules"),
+-- }
+-- or if you need something more advanced
+-- lvim.lsp.null_ls.setup.root_dir = function(fname)
+--   if vim.bo.filetype == "javascript" then
+--     return require("lspconfig/util").root_pattern("Makefile", ".git", "node_modules")(fname)
+--       or require("lspconfig/util").path.dirname(fname)
+--   elseif vim.bo.filetype == "php" then
+--     return require("lspconfig/util").root_pattern("Makefile", ".git", "composer.json")(fname) or vim.fn.getcwd()
+--   else
```
Tastyep, line 75: Why did you drop the [fallback](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/9ffbb7db173b9c3441efa3cd6378a094460d0b7b/lua/null-ls/lspconfig.lua#L17)

Also, since else after return is not useful, this can be
`return util.root_pattern("Makefile", ".git")(fname) or util.path.dirname(fname)`

Tastyep (changes requested):
Just a small detail, other than that it looks good :+1:

utils/installer/config.example-no-ts.lua
dklymenk: That code does not represent my personal config. I didn't put much thought into it. I just wanted to add as much random stuff there as possible to show what you can do with it:
* different setup depending on filetype
* ability to use both `require("lspconfig/util").root_pattern` and basic stuff like `vim.fn.getcwd()`

I can replace that example code with anything you want, if you can come up with example that both makes sense and demonstrates how it might be useful.

Tastyep: I think the example is great, just the else part should be what null-ls defines as default
`return util.root_pattern("Makefile", ".git")(fname) or util.path.dirname(fname)`

PR #1329, opened 2021-08-15
[Feature] switch galaxyline to lualine

chaeing:
# Description

Switching default statusline from galaxyline to lualine.

**Fully configurable** ([Lualine README.md](https://github.com/hoob3rt/lualine.nvim/blob/master/README.md))
Users will have everything that lualine provides.
config.lua
```lvim.builtin.lualine.ANY.OPTIONS.BY.LUALINE = VALUE```

**Hot reload**

**Provide predefined styles** 
This is not an option from lualine, but from Lunarvim.
Valid options are "lvim" or "default" or "none"
```lvim.builtin.lualine.style = "lvim"```
"lvim" style is set by default
If invalid options are set, it will fallback to "lvim"

**Theme fallback**
It will look for ```lvim.builtin.lualine.options.theme``` then, ```lvim.colorscheme```
and set it as lualine theme.
If none of them are defined or invalid theme is set, 
it will default to "auto" which goes well with current colorscheme.

## How Has This Been Tested?
config.lua
```
lvim.builtin.lualine.active = true
local custom_gruvbox = require "lualine.themes.gruvbox"
custom_gruvbox.normal.c.bg = "#ffffff"
lvim.builtin.lualine.options.theme = custom_gruvbox
lvim.builtin.lualine.style = "default"
lvim.builtin.lualine.sections.lualine_c = { "branch" }
lvim.builtin.lualine.sections.lualine_a = { "filetype" }

```

### CODE REVIEW

File: lua/utils/init.lua [hunk 1]
```diff
@@ -129,6 +131,23 @@ function utils.is_file(filename)
   return stat and stat.type == "file" or false
 end
 
+function utils.clone_table(t) -- deep-copy a table
```
Tastyep, line 134: Doesn't `vim.deepcopy` answer your need ?

lua/plugins.lua
```diff
@@ -156,15 +156,12 @@ return {
...
     config = function()
-      require "core.galaxyline"
-      if lvim.builtin.galaxyline.on_config_done then
-        lvim.builtin.galaxyline.on_config_done(require "galaxyline")
-      end
+      require("core.lualine").setup()
     end,
-    event = "BufWinEnter",
-    disable = not lvim.builtin.galaxyline.active,
+    disable = not lvim.builtin.lualine.active,
```
abzcoding, line 164: we can make it load on `BufWinEnter` so people can show off their faster `--staruptime` 🤣

lua/utils/init.lua
```diff
@@ -99,9 +100,10 @@ function utils.reload_lv_config()
   vim.cmd ":PackerCompile"
   vim.cmd ":PackerInstall"
   -- vim.cmd ":PackerClean"
+  vim.g.colors_name = lvim.colorscheme
+  vim.cmd("colorscheme " .. lvim.colorscheme)
```
abzcoding, line 104: do we need both of these, i think we had some issue, and added the extra one, but never checked again if we actually need both or not

abzcoding: it might be better to deal with this in another PR

abzcoding:
this weird issue happens for me, basically it throws an error no matter what i put in `lvim.colorscheme` but everything seems to be working afterwards 😢 
Screen Shot 2021-08-16 at 3 50 59 PM

abzcoding:
this happens on a fresh install as well
Screen Shot 2021-08-16 at 3 55 31 PM

lua/utils/init.lua
chaeing: It does! Thanks for the review as always.

chaeing:
> this weird issue happens for me, basically it throws an error no matter what i put in `lvim.colorscheme` but everything seems to be working afterwards 
> Screen Shot 2021-08-16 at 3 50 59 PM

It means that there is no "pablo" lualine theme found, therefore setting it to "auto".
If you want to get rid of the message set the the theme to "auto" explicitly like,
```
lvim.builtin.lualine.options.theme = "auto"
```

lua/plugins.lua
chaeing: I want it as well.
However hot reload won't work this way atm.

Should we give up hot reload for lazyload?

chaeing:
> this happens on a fresh install as well
> Screen Shot 2021-08-16 at 3 55 31 PM

Resolved, thank you :)

lua/core/lualine/styles.lua
```diff
@@ -0,0 +1,221 @@
...
+        function(msg)
+          msg = msg or "LSP Inactive"
+          local buf_clients = vim.lsp.buf_get_clients()
+          if next(buf_clients) == nil then
+            return msg
+          end
+          local buf_ft = vim.bo.filetype
+          local buf_client_names = {}
+          -- local null_ls_providers = require("lsp.null-ls").get_registered_providers_by_filetype(buf_ft)
+          local null_ls_providers = require("lsp.utils").get_active_client_by_ft(buf_ft)
```
Tastyep, line 125: If what you want to show are the active language servers, then rename the variable accordingly :)

lua/plugins.lua
abzcoding: not at all, we will deal with the lazyload later, hot reload is more important rn

abzcoding:
> > this weird issue happens for me, basically it throws an error no matter what i put in `lvim.colorscheme` but everything seems to be working afterwards
> > Screen Shot 2021-08-16 at 3 50 59 PM
> 
> It means that there is no "pablo" lualine theme found, therefore setting it to "auto".
> If you want to get rid of the message set the the theme to "auto" explicitly like,
> 
> ```
> lvim.builtin.lualine.options.theme = "auto"
> ```

but it fails the fresh installations as well
context https://github.com/LunarVim/LunarVim/pull/1329#issuecomment-899434833

lua/core/lualine/conditions.lua
```diff
@@ -0,0 +1,15 @@
...
+  buffer_not_empty = function()
+    return vim.fn.empty(vim.fn.expand "%:t") ~= 1
+  end,
+  hide_in_width = function()
+    return vim.fn.winwidth(0) > 80
+  end,
+  check_git_workspace = function()
+    local filepath = vim.fn.expand "%:p:h"
+    local gitdir = vim.fn.finddir(".git", filepath .. ";")
+    return gitdir and #gitdir > 0 and #gitdir < #filepath and vim.fn.winwidth(0) > 80
```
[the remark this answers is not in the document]
chaeing: tbh, I even thought ```return #gitdir > 0``` would be enough.
I just copied the example code from lualine official repo to be safe.
Thought there might be some cases that I am not aware of.

Should we still change anyways?

lua/core/lualine/init.lua
```diff
@@ -0,0 +1,50 @@
+local M = {}
+M.config = function()
+  lvim.builtin.lualine = {
+    active = true,
```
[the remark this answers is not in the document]
chaeing: Not against doing it now
but how about handling this on future pr for consistency?

lua/core/lualine/init.lua
```diff
@@ -0,0 +1,50 @@
...
+    style = "lvim",
```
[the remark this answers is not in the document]
chaeing: Yes just like other plugins config.
Are you trying to trim the size of the table concerning startup time?

lua/core/lualine/conditions.lua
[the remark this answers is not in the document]
Tastyep: Keep it there then :)

lua/core/lualine/init.lua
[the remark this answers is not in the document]
Tastyep: As long as it doesn't conflict, I'm fine with modifying it later :+1:

lua/core/lualine/init.lua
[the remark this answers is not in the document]
Tastyep: Let's leave it like that for now, I'll do a pass later

lua/core/lualine/styles.lua
```diff
@@ -0,0 +1,226 @@
...
+  tabline = {},
+  extensions = { "nvim-tree" },
+}
+
+function M.update()
+  local config = lvim.builtin.lualine
+  local style = config.style
+
+  local style_keys = vim.tbl_keys(styles)
+  if not vim.tbl_contains(style_keys, style) then
```
Tastyep, line 186:
``` lua
  if not vim.tbl_contains(style_keys, style) then
    local Log = require "core.log"
    local logger = Log:get_default()
    logger.error(
      "Invalid lualine style",
      string.format('"%s"', style),
      "options are: ",
      string.format('"%s"', table.concat(style_keys, '", "'))
    )
    logger.info '"lvim" style is applied.'
    style = "lvim"
  end

```

Tastyep (changes requested):

lua/core/lualine/styles.lua
chaeing: Perfect! Thank you :)

lua/core/lualine/utils.lua
```diff
@@ -0,0 +1,13 @@
+local M = {}
+
+function M.validate_theme()
+  local theme = lvim.builtin.lualine.options.theme
+
+  local lualine_loader = require "lualine.utils.loader"
+  local ok, validated_theme = pcall(lualine_loader.load_theme, theme)
+  if not ok and type(validated_theme) ~= "table" then
```
Tastyep, line 8: What is the second check for ? if ok is false, then validated_theme is always nil, right ?

lua/core/lualine/styles.lua
```diff
@@ -103,6 +104,24 @@ styles.lvim = {
       },
     },
     lualine_x = {
+      {
+        -- Python Env
+        function()
+          if vim.bo.filetype == "python" then
+            local venv = os.getenv "CONDA_DEFAULT_ENV"
+            if venv ~= nil then
```
Tastyep: Simpler syntax: `if venv then`

lua/core/lualine/styles.lua
```diff
@@ -103,6 +104,24 @@ styles.lvim = {
...
+              return "  (" .. require("core.lualine.utils").env_cleanup(venv) .. ")"
```
Tastyep, line 111: Can you define `utils` above ?

``` lua
local utils = require("core.lualine.utils")
.. 
  return string.format("  (%s)", utils.env_cleanup(venv))
```

lua/core/lualine/styles.lua
```diff
@@ -150,13 +150,24 @@ styles.lvim = {
...
           local active_client = require("lsp.utils").get_active_client_by_ft(buf_ft)
           for _, client in pairs(buf_clients) do
             if client.name ~= "null-ls" then
               table.insert(buf_client_names, client.name)
             end
           end
           vim.list_extend(buf_client_names, active_client or {})
+
+          -- add formatter
+          local active_formatters = require("lsp.null-ls.formatters").list_supported_names(buf_ft)
```
Tastyep, line 164: https://github.com/luarocks/lua-style-guide#modules

Tastyep (changes requested):

lua/core/lualine/utils.lua
chaeing: That is what I expected but no.
It is a function from lualine and it returns string even if ok is false

lua/core/lualine/styles.lua
chaeing: Thanks always.
I do not mean to make you over again with a same thing :(
But could you check it again and tell me more info pls?
It is under ```components.lua``` now

lua/core/lualine/utils.lua
Tastyep: Ah yep, I think it returns an error as second argument.
So knowing that, we can remove the second part of the condition :)

lua/core/lualine/components.lua
```diff
@@ -118,21 +118,20 @@ return {
...
+      local chars = { "__", "▁▁", "▂▂", "▃▃", "▄▄", "▅▅", "▆▆", "▇▇", "██" }
+      local scroll_index = 1
       if current_line == 1 then
-        index = 1
+        scroll_index = 1
       elseif current_line == total_lines then
-        index = #chars
+        scroll_index = #chars
       else
         local line_no_fraction = vim.fn.floor(current_line) / vim.fn.floor(total_lines)
```
Tastyep, line 128: I have some doubts about this block of code.
My first concern is that the two conditions above should not be necessary.
Then `local line_no_fraction = vim.fn.floor(current_line) / vim.fn.floor(total_lines)` sound weird to me, does `vim.fn.floor` works the same way as [math.floor](https://devdocs.io/lua~5.4/index#pdf-math.floor) ? also, this function is being passed line numbers, aren't they already integers ? Finally, line_no_fraction will be a fraction if we end up dividing two integers, am I missing smth ?

lua/core/lualine/utils.lua
chaeing: Lol you are right.
Thank you :)

lua/core/lualine/components.lua
chaeing: You are not missing anything :)

lua/core/lualine/components.lua
```diff
@@ -119,19 +119,13 @@ return {
...
-        scroll_index = 1
-      elseif current_line == total_lines then
-        scroll_index = #chars
-      else
-        local line_no_fraction = vim.fn.floor(current_line) / vim.fn.floor(total_lines)
-        scroll_index = vim.fn.float2nr(line_no_fraction * #chars)
-        if scroll_index == 0 then
-          scroll_index = 1
-        end
+      local index
```
Tastyep, line 122: This should work :)
``` lua
local line_ratio = current_line / total_lines
local index = math.ceil(line_ratio * #chars)
return chars[index]
```

Tastyep (changes requested):

lua/core/lualine/components.lua
```diff
@@ -0,0 +1,134 @@
...
+      for _, client in pairs(buf_clients) do
+        if client.name ~= "null-ls" then
+          table.insert(buf_client_names, client.name)
+        end
+      end
+      vim.list_extend(buf_client_names, active_client or {})
+
+      -- add formatter
+      local formatters = require "lsp.null-ls.formatters"
+      local supported_formatters = formatters.list_supported_names(buf_ft)
```
Tastyep, line 85: `list_supported_names` always returns a valid list

Tastyep (changes requested):
Re reviewed everything, it looks great :+1:  way better than what we had with galaxyline.

lua/core/lualine/components.lua
chaeing: Thank you @Tastyep
Your kind reviews made me a better dev :)

Tastyep (approved):
image

abzcoding (approved):

lua/core/lualine/components.lua
```diff
@@ -0,0 +1,130 @@
...
+    end,
+    left_padding = 0,
+    right_padding = 0,
+    condition = conditions.hide_in_width,
+  },
+  branch = {
+    "branch",
+    icon = " ",
+    condition = function()
+      return conditions.hide_in_width() and conditions.check_git_workspace()
```
shadmansaleh, line 17: lualines `branch` component will check whether buffer is in a git workspace.
No need to check twice by placing it in `condition` .

Also since you guys have gitsigns you can just reuse the branch & diff info from gitsigns
instead of leaving lualine to figure them out .
https://github.com/shadmansaleh/lualine.nvim/wiki/Component-snippets#using-external-source-for-branch

abzcoding: cc: @chaeing

chaeing: Thank you for this useful information :)

Issue #1271, opened 2021-08-09
[Feature]: Windows support


Dulanjala007:
### Feature motivation

I love Lunarvim's experience in my linux machines so much, and so used to all the keybindings and ways of going around the files and projects. It now feels really uncomfortable to work now on windows. so I made Lunarvim work on windows by butchering around the code here and there. (not pretty and can't update to latest versions.) but if we could have something we can use directly it would really amazing...!

### Describe the expected solution

few problematic things standout:

1. dealings with neovim's default path structure, and where we install Lunarvim and packer and config on windows (easily reproducible) 
2. all the path configurations in Lunarvim to be windows path compatible (and less use of expansion for performance? )
3. all the things related to external process execution and related things (shell, env) to be windows compatible.

these are few things i had to change to make it work on windows.
would Love a proper implementation of these...

marwan38:
Me too, there's a PR up for it already:
https://github.com/LunarVim/LunarVim/pull/1261

Xyhlon:
did you consider using wsl2 and using it in there thats what i am doing

PR #1261, opened 2021-08-08
feat: add installation script for Windows

MarcSchaetz:
# Description
This pull request adds a `install.ps1` for an easy installation on Windows, an `uninstall.ps1` for removing the lunarvim folder and adds a usage description in `README.md`.

This is basically a conversion from bash to powershell. To work on windows however two lua scripts are changed in the installation process.
- ~/.local/share/lunarvim/lvim/lua/default-config.lua
- ~/.local/share/lunarvim/lvim/init.lua

On both these files the string `os.getenv "HOME"` is replaced with `os.getenv "HOMEDRIVE" .. os.getenv "HOMEPATH"` because on windows a `HOME` environment variable does not exist.

Due to the lack of onboard package management of windows `winget` is used for installing dependencies. If `winget` is not installed the script stops but tells the user the reason.

Fixes #804 

## How Has This Been Tested?
- Powershell 5.1
- Windows 10 Pro (20H2) x64
- nvim 5.0
- Python 3.9.6 with pip 21.2.2
- winget 1.0.11692

The script was tested against both the master and rolling branches.

PR #1333, opened 2021-08-16
[Feature] switched os.getenv('HOME') to a basepath variable for better cross platform support

ahsan-a:
# Description

The current method of using `os.env("home")` for a basepath doesn't work on windows. The install script in #1261 works around this by creating a backup file of `init.lua` and `default-config.lua` and replacing all occurrences with a windows alternative. Using this instead sets a `BASE_PATH` variable to `os.getenv("HOME")`, or if it doesn't exist (meaning the OS being used is windows) sets it to `os.getenv "HOMEDRIVE" .. os.getenv "HOMEPATH"`.

Improves situation with #1271 

## How Has This Been Tested?
- Replace files with modified files
- See if lvim opens

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -22,7 +22,8 @@ local function file_exists(name)
   end
 end
 
-local lvim_path = os.getenv "HOME" .. "/.config/lvim/"
+BASE_PATH = os.getenv "HOME" or os.getenv "HOMEDRIVE" .. os.getenv "HOMEPATH"
+local lvim_path = BASE_PATH .. "/.config/lvim/"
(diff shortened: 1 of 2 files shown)


kylo252:
This is a good step forward, but it's missing quite a few things sadly. See the discussion in https://github.com/LunarVim/LunarVim/pull/1261#issuecomment-899584099 and take a look at the patch I had in there.

vlcinsky (changes requested):
You may use `vim.loop.os_homedir()` to get home directory in cross platform compatible way.

kylo252:
can you please rebase on #1354 and follow the new pattern?

PR #1335, opened 2021-08-16
Check if emmet is active.  Enable emmet completion if emmet language server is active

rebuilt:
Fixes #(issue)
Emmet completion does not work correctly.  An expression like `nav>ul>li` should produce
```html
<nav>
    <ul>
        <li></li>
    </ul>
</nav>
```

Instead, it just inserts a tab character.  

This pr adds a check to see if emmet_ls is active.  Enables emmet completion if emmet language server is active.   
```lua
    elseif is_emmet_active() then
      return vim.fn["compe#complete"]()
```
## How Has This Been Tested?
emmet

To set up emmet-ls  add this to your plugins.  

```lua
lvim.plugins ={
		{
			"aca/emmet-ls",
			config = function()
				local lspconfig = require("lspconfig")
				local configs = require("lspconfig/configs")

				local capabilities = vim.lsp.protocol.make_client_capabilities()
				capabilities.textDocument.completion.completionItem.snippetSupport = true
				capabilities.textDocument.completion.completionItem.resolveSupport = {
					properties = {
						"documentation",
						"detail",
						"additionalTextEdits",
					},
				}

				if not lspconfig.emmet_ls then
					configs.emmet_ls = {
						default_config = {
							cmd = { "emmet-ls", "--stdio" },
							filetypes = {
								"html",
								"css",
								"javascript",
								"typescript",
								"eruby",
								"typescriptreact",
								"javascriptreact",
								"svelte",
								"vue",
							},
							root_dir = function(fname)
								return vim.loop.cwd()
							end,
							settings = {},
						},
					}
				end
				lspconfig.emmet_ls.setup({ capabilities = capabilities })
			end,
		},
}
```

### DIFF
--- a/lua/core/compe.lua
+++ b/lua/core/compe.lua
@@ -83,9 +83,12 @@ M.setup = function()
   end
 
   local is_emmet_active = function()
-    local utils = require "lsp.utils"
-    if utils.is_client_active "emmet_ls" then
-      return true
+    local clients = vim.lsp.buf_get_clients()
+
+    for _, client in pairs(clients) do
+      if client.name == "emmet_ls" then
+        return true
+      end
     end
     return false
   end
--- a/lua/core/compe.lua
+++ b/lua/core/compe.lua
@@ -98,10 +98,10 @@ M.setup = function()
       return t "<C-n>"
     elseif vim.fn.call("vsnip#jumpable", { 1 }) == 1 then
       return t "<Plug>(vsnip-jump-next)"
-    elseif is_emmet_active() then
-      return vim.fn["compe#complete"]()
     elseif check_back_space() then
       return t "<Tab>"
+    elseif is_emmet_active() then
+      return vim.fn["compe#complete"]()
     else
       return t "<Tab>"
     end
--- a/lua/core/compe.lua
+++ b/lua/core/compe.lua
@@ -82,6 +82,14 @@ M.setup = function()
     end
   end
 
+  local is_emmet_active = function()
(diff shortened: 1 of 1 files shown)


lua/core/compe.lua
```diff
@@ -92,8 +100,9 @@ M.setup = function()
       return t "<Plug>(vsnip-jump-next)"
     elseif check_back_space() then
       return t "<Tab>"
+    elseif is_emmet_active() then
```
PasiBergman, line 103: As discussed, please add `not check_back_space() and` condition before is_emmet_active() to allow use of <tab> at the beginning of the line before non-space chars have been inserted.

lua/core/compe.lua
```diff
@@ -82,6 +82,14 @@ M.setup = function()
     end
   end
 
+  local is_emmet_active = function()
+    local utils = require "lsp.utils"
+    if utils.is_client_active "emmet_ls" then
```
PasiBergman, line 87: As discussed, please check if emmet LSP client is attached to the current buffer/filetype.

PasiBergman (changes requested):

lua/core/compe.lua
rebuilt: I think moving check_backspace before is_emmet_active will have the same effect and avoid calling check_back_space twice.  If there's a space, it will return a tab and never reaches the emmet check.

PasiBergman (approved):

PR #1336, opened 2021-08-16
[Feature]: Join lines improvement

andremeireles:
# Join line

This commit change the default behavior of jump the cursor to the first word from joined line.
Now the cursor keeps your position.


Tastyep:
After a discussion with other contributors, we decided that we won't merge this.

Feel free however to add it to your own config.

``` lua
lvim.keys.normal_mode["<S-j>"] = "mzJ`z",
```

Issue #1214, opened 2021-08-02
More reasonable install dir ? 


midnightexigent:
The current installation lives in `~/.local/share/lunarvim` which is kind of weird

Any reason it can't be somewhere more common like `~/.lunarvim`, `~/.lvim` or even `~/.config/lvim` with the user config in `~/.config/lvimrc` ?

ChristianChiarulli:
I wouldn't recommend putting any config or data directories in the home directory. I'm not sure why you would would want it collocated with your repo since the core of lunarvim should be left unedited by the user, unless you want to fork it or something.

rebuilt:
`~/.local/share/` is a standard directory to place these kind of files.  It conforms to the xdg specification 

https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

midnightexigent:
I guess it's just not common for config frameworks to be in located in `~/.local/share` (ex. `doom-emacs`, `zsh-znap`, `oh-my-zsh` etc.) 

`~/.local/share/lunarvim/site` definitely belongs there tho

rebuilt:
The config folder is designed as a place to put user-editable files.  LunarVim has separated out the configuration to `.config/lvim`.  Data files live in $XDG_DATA_HOME which is $HOME/.local/share.  The application is nvim so it gets placed somewhere else.  The way I see it, LunarVim is adhering to the xdg standard.

ChristianChiarulli:
They should all probably start conforming to the xdg standard.

midnightexigent:
The standard says that `~/.local/share/` is the location where "user-specific data files should be stored". Which can honestly mean anything  (hell, some apps store the log files there), tho, **usually, apps tend to use this directory to store some sort of state that can be rebuilt whenever deleted** which is why I believe that `~/.local/share/lunarvim/site` definitely belongs there but `lvim` doesn't 

The way it currently is, is probably just fine, I am just saying that a `~/.lvim` makes it look more familiar, has less side effects since the binary can be there (`~/.lvim/bin/` which is also a common pattern) and is probably more flexible for the future

ChristianChiarulli:
I understand the familiarity aspect to some degree but the way I see it there are 4 directories I can use (before I use a package manager to install in the future)

`~` <- this is a terrible option because it clutters up your home directory and should be avoided at all costs
`~/.config` <- this is for a users config something they can check into version control and have complete control over
[...]

PR #1338, opened 2021-08-17
[Feature] Encapsulate config logic

Tastyep:
# Description

This provides encapsulation for lvim configuration.
This also defines `lua/core/builtins/init.lua`, but leaves the builtins in `lua/core/`
I replaced the `file.open() + file.close()` with a call to `uv.fs_stat()`

## How Has This Been Tested?

Renamed `config.lua` to `lv-config.lua` - it does fallback
Renamed `config.lua` to `toto.lua` - an error is printed
Modify `config.lua` - the config is updated
Update `vim.opt.relativenumber` from `config.lua` - it hot reloads


abzcoding:
i don't know why, but setting `vim.opt.relativenumber = true` doesn't work with this new change
Screen Shot 2021-08-17 at 12 13 45 PM

abzcoding (approved):
this makes the code look much cleaner, but before merging it, let some other people test it as well

abzcoding:
I should also mention that this would cause a lot of conflicts for other PRs, so we might want to merge a couple of them before this.

rebuilt:
I'm testing it now

lua/core/builtins/init.lua
```diff
@@ -0,0 +1,25 @@
...
+  require "core.telescope",
+  require "core.treesitter",
+  require "core.nvimtree",
+  require "core.rooter",
+  require "core.bufferline",
+}
+
+function M.config(config)
+  for _, builtin in ipairs(builtins) do
+    builtin.config(config)
```
rebuilt, line 21: It's the teeniest bit cleaner to remove the require statements from above and change this line to ` require(builtin).config(config)`

rebuilt (approved):
The lv-config.lua fallback works as expected.  This still works with sym-linked files.  Callbacks work.  Reloading still works. This opens up the possibility of loading a configuration from a user-provided path.

kylo252 (commented):
I understand the idea generally but I'm not convinced with the naming choices. Take a look at what the structure looks like now:

|
|-- config
|-- core
|-- default-config.lua
|-- interface
|-- keymappings.lua
|-- lsp
|-- plugin-loader.lua
|-- plugins.lua
|-- spacegray
|-- utils

What are we exactly configuring? are the rest of the files not configuration?

Tastyep:
This made me realize that I forgot to move the `default-config` file
The structure now is:
``` bash
config/
core/
interface/
lsp/
spacegray/
utils/
keymappings.lua
plugin-loader.lua
plugins.lua
```

`keymappings` should also be part of config, but logic and configuration got merged a few weeks ago.
`plugins` and `plugin-loader` should be moved somewhere else as well.
I'll apply this change in a future PR where I'll move builtins in the `builtin/` directory.

kylo252:
> `keymappings` should also be part of config, but logic and configuration got merged a few weeks ago.
> `plugins` and `plugin-loader` should be moved somewhere else as well.
> I'll apply these changes in a future PR where I'll move builtins in the `builtin/` directory.

It might be semantics, but I'm still not convinced this the a good move naming-wise. Even referring to Telescope and co as "builtins" is probably misleading.

I would much prefer a clearer name since the end-goal is maintainability vs obfuscation. I think `bootstrap.lua` conveys the intent more clearly, instead of having a "configuration" folder that may or may not be related to "lvim/config.lua" depending on what you're doing.

Tastyep:
> It might be semantics, but I'm still not convinced this the a good move naming-wise. Even referring to Telescope and co as "builtins" is probably misleading.

I picked `builtins` because it's the name we currently use (also in the config). If I remember correctly, Chris did a quick 'poll' during a stream and `builtin` was picked. `plugin` was another option, if you ask me, this would be a better choice.

> I would much prefer a clearer name since the end-goal is maintainability vs obfuscation. I think `bootstrap.lua` conveys the intent more clearly, instead of having a "configuration" folder that may or may not be related to "lvim/config.lua" depending on what you're doing.

I know we don't have the same vision on this, so we will have to deal with not agreeing with each other.
That's how I see it: 
LunarVim is not just a configuration, it used to be, but now it's an IDE and has its own executable `lvim`.
What we call `builtins` or `plugins` are nothing more than `libraries`, some are more generic like `plenary`, others are more behavior specific like`telescope`.
Based on this, `lvim` is an application using libraries that provides functionalities to its users. Because we want these functionalities to be configurable, a `config` module defines all the available settings.

So when you ask
> What are we exactly configuring? are the rest of the files not configuration?

I would say that no, they are not configurations. To me, the `core` files compose the domain layer of the application, ie, the core logic of lunarvim. Also, In these files there are many small functions that we defined because we wanted to provide additional behavior.
Because we didn't have to implement everything, doesn't mean that we are just a configuration on top of these plugins. On the opposite, we are lucky to have those, as they allow us to move faster and focus on the bigger picture.
To support my view, a good example would be the `interface` module that I added a few days ago. It defines the interface layer of the application, it doesn't implement any logic, but only uses the domain capabilities and displays available information.

That's how I see it, I'm also interested reading what the other contributors' vision is.

abzcoding:
> > It might be semantics, but I'm still not convinced this the a good move naming-wise. Even referring to Telescope and co as "builtins" is probably misleading.
> 
> I picked `builtins` because it's the name we currently use (also in the config). If I remember correctly, Chris did a quick 'poll' during a stream and `builtin` was picked. `plugin` was another option, if you ask me, this would be a better choice.
> 
> > I would much prefer a clearer name since the end-goal is maintainability vs obfuscation. I think `bootstrap.lua` conveys the intent more clearly, instead of having a "configuration" folder that may or may not be related to "lvim/config.lua" depending on what you're doing.
> 
> I know we don't have the same vision on this, so we will have to deal with not agreeing with each other.
> That's how I see it:
> LunarVim is not just a configuration, it used to be, but now it's an IDE and has its own executable `lvim`.
> What we call `builtins` or `plugins` are nothing more than `libraries`, some are more generic like `plenary`, others are more behavior specific like`telescope`.
> Based on this, `lvim` is an application using libraries that provides functionalities to its users. Because we want these functionalities to be configurable, a `config` module defines all the available settings.
> 
> So when you ask
> 
> > What are we exactly configuring? are the rest of the files not configuration?
> 
> I would say that no, they are not configurations. To me, the `core` files compose the domain layer of the application, ie, the core logic of lunarvim. Also, In these files there are many small functions that we defined because we wanted to provide additional behavior.
> Because we didn't have to implement everything, doesn't mean that we are just a configuration on top of these plugins. On the contrary, we are lucky to have those, as they allow us to move faster and focus on the bigger picture.
> To support my view, a good example would be the `interface` module that I added a few days ago. It defines the interface layer of the application, it doesn't implement any logic, but only uses the domain capabilities and displays available information.
> 
> That's how I see it, I'm also interested reading what the other contributors' vision is.

I do agree with your vision of LunarVim

kylo252:
> I know we don't have the same vision on this, so we will have to deal with not agreeing with each other.
> That's how I see it:
> LunarVim is not just a configuration, it used to be, but now it's an IDE and has its own executable `lvim`.

Do you know what that executable is? It's literally just `nvim -u ~/.local/share/lunarvim/lvim/init.lua`

> What we call `builtins` or `plugins` are nothing more than `libraries`, some are more generic like `plenary`, others are more behavior specific like`telescope`.
> Based on this, `lvim` is an application using libraries that provides functionalities to its users. Because we want these functionalities to be configurable, a `config` module defines all the available settings.

How is `lvim` an application based on this? It's more appropriate to call it a "wrapper" or a "layer" around neovim. So _in my opinion_, starting to obfuscate the fact that it's not a standalone application will only serve to hurt the project and the ecosystem in the long-run. Because when anything breaks in your "black-box" then it will get reported here instead of the appropriate plugin issue-tracker or possibly neovim's. 

> To support my view, a good example would be the `interface` module that I added a few days ago. It defines the interface layer of the application, it doesn't implement any logic, but only uses the domain capabilities and displays available information.

I don't see how that's related to the points above. I've always been a big advocate for modularization because it eases maintainability. So please don't confuse my points or views regarding that with my goal/hope of trying to avoid overly complicated structures and naming-schemes that _may or may not_ mislead users to think that lunarvim has implemented some crazy functionality that's not readily available outside of it.
 
> That's how I see it, I'm also interested reading what the other contributors' vision is.

I would love to see those as well.

Tastyep:
> Do you know what that executable is

I'm pretty sure I did review [this PR](https://github.com/LunarVim/LunarVim/pull/1052), what I was saying is that a separate executable shows a certain intent.

> How is lvim an application based on this

It's an application because it defines logic and manages dependencies ? Speaking of "layer", an operating system is a layer, and it's a big ass one.

> Because when anything breaks in your "black-box"

I hoped that my work on the project showed that I always intend to implement better interfaces and testable code, that makes me quite sad :disappointed: 

> So please don't confuse my points or views regarding that with my goal/hope of trying to avoid overly complicated structures and naming-schemes that may or may not mislead users to think that lunarvim has implemented some crazy functionality that's not readily available outside of it.

I really don't get how you would expect a lambda user to go through the whole code base because he wants to learn how to use lvim. Seriously `wc -l lua/**/*` outputs `5069 total`, don't you think it would be better for the user to operate lvim through abstracted interfaces ?
An example: We could implement popups for the user to configure lvim completely, that's way better than browsing code I think.

kylo252:
> It's an application because it defines logic and manages dependencies ? Speaking of "layer", an operating system is a layer, and it's a big ass one.

Check #1354 btw.

> > Because when anything breaks in your "black-box"
> 
> I hoped that my work on the project showed that I always intend to implement better interfaces and testable code, that makes me quite sad 😞

Sorry if it came across like that. 

What I'm trying to say is that we should not make it appear as a black-box if it doesn't have to be. I feel like there's a balance that we either don't agree on, or that I'm not able to explain this more clearly.

> I really don't get how you would expect a lambda user to go through the whole code base because he wants to learn how to use lvim. Seriously `wc -l lua/**/*` outputs `5069 total`, don't you think it would be better for the user to operate lvim through abstracted interfaces ?

You can abstract a lot of the inner interactions between the components while still keeping a transparent API to the underlying plugins. Would you like to spend your time trying to explain to a user why XYZ is happening? or would you like to tell them, that's because which-key is configured this way. You can also set its other options like this. 

Do you want to take the burden of troubleshooting common plugins issues because the user needs go RTFM for small adjustments? Are we planning to re-implement the options for every single "builtin"? How is that focusing on the big picture?

> An example: We could implement popups for the user to configure lvim completely, that's way better than browsing code I think.

I agree but I'm also not sure how that contradicts anything I said. You can have both a nice API and whatever else you want.

Tastyep:
> I feel like there's a balance that we either don't agree on, or that I'm not able to explain this more clearly.

I understand what you are saying, I just don't agree :smile:. 

You suggest transforming `config/init.lua` into `bootstrap.lua` and moving back everything to the top level because you are concerned the `config`  would obfuscate code. I don't understand how grouping related files into a module obfuscates things, and again, for whom ?

> Would you like to spend your time trying to explain to a user why XYZ is happening

The user config is in `~/.config/lvim/`, that's all he needs to know about, we split code and config for a reason ?
Again, the user should not need to browse lvim's code, the config already acts as an interface for the user to interact with.

Finally, the only thing this PR does about the plugin's configuration is moving the `config()` calls away from default-config, so I don't see how all that is related to this work

ahmedkhalf:
I agree with both of you guys, but allow me to share my philosophy:

- FAST. Gotta go fast. If a feature is going to slow down lvim, better think three times before adding it.
- Maintainable. Right now, lvim is not very maintainable as we essentially have a copy of every single plugin's setting in our repo, this is bad for many reasons one of which is if the plugin decides to change its config, or add to it.
- Good for the newbie and the vim veteran. For the newbie, the idea is clear, we want to set him up efficiently, and spoonfed him as much as possible so he doesn't have to bump his head 50 times on the wall. For the veteran, the same thing applies, we want to set him up quickly, but the problem is if we add too much abstraction he will have to learn many things which go against setting him up quickly.

Right now though, the argument seems to be about semantics and naming conventions... we can talk about that later and have a discussion over an issue or over discord. But as a concept, I like this PR.

> You can abstract a lot of the inner interactions between the components while still keeping a transparent API to the underlying plugins

Agreed.

ahmedkhalf:
Also speaking of setting up quickly, the current doc website is an absolute trash dumpster, and I believe it should be rehauled and/or improved upon.

rebuilt:
Based on the kind of questions we get on Discord, I don't think most of our users are reading the LunarVim source.  Half don't even read our documentation.

 I think the days of most users being able to understand the LunarVim codebase are long gone.  That's partly because the config is so big but also because it does so much.  Even Chris has said he doesn't know every bit of the configuration anymore.  
 
 I have some opinions about what LunarVim is which I wrote in [this post: ](https://github.com/LunarVim/LunarVim/issues/1214#issuecomment-891395204)
 To summarize, I think it's more than a config but less than an application.  
 
 But arguing semantics isn't productive.  The question is, how do we organize the code?  There's a lot of code and we have to be able to keep working with it.    
 
So we have to abstract some behavior and keep refactoring the code.  

I prefer bootstrap.lua as a name because 'config' is overloaded with meaning.  I'm not sure what other concrete improvements Hatsu wants to see happen with this pr.  I get the impression there's just a general aversion to the additional abstraction.  But the abstraction from Tasty's past pr's have also come with benefits.  I'm thinking of the plugin_loader abstraction in particular.  

I like what this pr does.  I think it should be merged.

PR #1342, opened 2021-08-17
[Feature]: LSP Fortran

### LINKED ISSUE #1334: [Feature]: LSP support Fortran
### Feature motivation

Fortran LSP can be installed with `LSPInstall fortran` however it's not attached when Fortran file is opened `ft=fortran`.

LSP Inactive -> LSP Fortran

_No response_

### Additional context

_No response_

lkhphuc:
# Description

Add LSP support for Fortran filetype.

Fixes #1334 

## How Has This Been Tested?

- Open any `*.f90/.f95` file.
- Check logs: `LspInfo` attached `fortls`

### DIFF
--- a/lua/default-config.lua
+++ b/lua/default-config.lua
@@ -488,6 +488,21 @@ lvim.lang = {
       },
     },
   },
+  fortran = {
+    formatters = {},
+    linters = {},
+    lsp = {
+      provider = "fortls",
+      setup = {
+        cmd = {
+          DATA_PATH .. "/lspinstall/fortran/venv/bin/fortls",
+        },
+        on_attach = common_on_attach,
+        on_init = common_on_init,
+        capabilities = common_capabilities,
+      },
+    },
+  },
   go = {
     formatters = {
       -- {
--- a/ftplugin/fortran.lua
+++ b/ftplugin/fortran.lua
@@ -1,0 +1,1 @@
+require("lsp").setup "fortran"
--- a/lua/plugins.lua
+++ b/lua/plugins.lua
@@ -3,7 +3,7 @@ return {
   { "wbthomason/packer.nvim" },
   { "neovim/nvim-lspconfig" },
   { "tamago324/nlsp-settings.nvim" },
-  { "jose-elias-alvarez/null-ls.nvim", commit = "341b726b910e0dce0d7da98a2b4c47f1629e8339" },
+  { "jose-elias-alvarez/null-ls.nvim" },
   {
     "kabouzeid/nvim-lspinstall",
     event = "VimEnter",
--- a/lua/plugins.lua
+++ b/lua/plugins.lua
@@ -3,7 +3,7 @@ return {
   { "wbthomason/packer.nvim" },
   { "neovim/nvim-lspconfig" },
   { "tamago324/nlsp-settings.nvim" },
-  { "jose-elias-alvarez/null-ls.nvim" },
+  { "jose-elias-alvarez/null-ls.nvim", commit = "341b726b910e0dce0d7da98a2b4c47f1629e8339" },
   {
     "kabouzeid/nvim-lspinstall",
     event = "VimEnter",


lua/default-config.lua
```diff
@@ -488,6 +488,35 @@ lvim.lang = {
       },
     },
   },
+  fortran = {
+    formatters = {
+      {
+        -- @usage can be clang_format or uncrustify
+        exe = "",
+        args = {},
+        stdin = true,
```
abzcoding, line 497: plz remove this line, we are using null-ls as formatting/linting provider
so just make it look like 
```lua
formatters = {
  -- {
  --  exe = "",
  --   args = {},
  -- },
},
```

Tastyep: Since there are no formatters available, I would even say to put it like this
`formatters = {}`

lkhphuc: Fixed. I was copying from a different language.

Tastyep: These are small changes done on the rolling branch, you might not be up-to-date with that one :)

lua/default-config.lua
```diff
@@ -488,6 +488,28 @@ lvim.lang = {
...
+        },
+        on_attach = common_on_attach,
+        on_init = common_on_init,
+        capabilities = common_capabilities,
+        root_dir = function(fname)
+          local util = require "lspconfig/util"
+          return util.root_pattern(".fortls", ".git")(fname)
+        end,
+        settings = {
+          nthreads = 2,
```
Tastyep, line 508: Out of curiosity, why setting it to 2 when the default is 1 ?

lua/default-config.lua
```diff
@@ -488,6 +488,28 @@ lvim.lang = {
...
+      setup = {
+        cmd = {
+          DATA_PATH .. "/lspinstall/fortran/venv/bin/fortls",
+        },
+        on_attach = common_on_attach,
+        on_init = common_on_init,
+        capabilities = common_capabilities,
+        root_dir = function(fname)
+          local util = require "lspconfig/util"
+          return util.root_pattern(".fortls", ".git")(fname)
```
Tastyep, line 505: Is `.git` necessary ?
[Defaults](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/fortls.lua#L8)

abzcoding: yes if you want it to activate on `fortran` files without the existence of a `.fortls` setting

Tastyep: I agree, but shouldn't this change be on lspconfig ?

lkhphuc: Sorry I'm not familiar with the project. Would you want me to delete `.git` or move it to somewhere else?

lua/default-config.lua
lkhphuc: Not thing in particular. The default of `fortls` is actually 4 threads, but the default of the `lspconfig` is 1 somehow. In practice I found 1 thread is usually quite slow for large codebase, but some (really) old machine may not have 4 threads so I opt for 2.

lua/default-config.lua
Tastyep: It would be perfect it you could remove it from here and add it [there](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/fortls.lua#L8)
Example: https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/sumneko_lua.lua#L10

lkhphuc: Actually I would also like it to default to current directory like this? Would that make sense here?
```suggestion
          return util.root_pattern(".fortls", ".git")(fname) or vim.fn.getcwd()
```

Tastyep: Probably not, but this would be something you could put in your config file :)

lkhphuc: Ok thank. I have remove all non-default setting.
All my custom config are now in my personal `lvim/config.lua` like:
```
lvim.lang.fortran.lsp.setup.root_dir = function(fname)
  local util = require"lspconfig.util"
  return util.root_pattern(".git", ".fortls")(fname) or vim.fn.getcwd()
end
```
Will try to add this upstream to `lspconfig`.

lkhphuc: https://github.com/neovim/nvim-lspconfig/pull/1159

Tastyep: Thank you very much for working on this and for your flexibility on the implementation !

Tastyep (approved):

PR #1353, opened 2021-08-19
[Bugfix] fix formatter/linter nil issue

abzcoding:
# Description

Fixes two issues

1. linters for `vim` files was set to `{""} ` in `default-config.lua` which is wrong.
Screen Shot 2021-08-19 at 6 16 32 PM

2. we didn't check for the existence of `lvim.lang[filetype]` before using it, which cause this error
Screen Shot 2021-08-19 at 6 17 52 PM

## How Has This Been Tested?

opened a sample `.vim` file and a sample `.rmd` file without any errors

### DIFF
--- a/lua/lsp/null-ls/formatters.lua
+++ b/lua/lsp/null-ls/formatters.lua
@@ -63,7 +63,7 @@ function M.list_configured(formatter_configs)
 end
 
 function M.setup(filetype, options)
-  if formatters_by_ft[filetype] and not options.force_reload then
+  if (formatters_by_ft[filetype] and not options.force_reload) or not lvim.lang[filetype] then
     return
   end
 
--- a/lua/lsp/null-ls/formatters.lua
+++ b/lua/lsp/null-ls/formatters.lua
@@ -63,7 +63,7 @@ function M.list_configured(formatter_configs)
 end
 
 function M.setup(filetype, options)
-  if (formatters_by_ft[filetype] and not options.force_reload) or not lvim.lang[filetype] then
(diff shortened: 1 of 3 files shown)


abzcoding:
@JulianTao can you confirm this fixes your issues?

lua/lsp/null-ls/formatters.lua
```diff
@@ -63,7 +63,7 @@ function M.list_configured(formatter_configs)
 end
 
 function M.setup(filetype, options)
-  if formatters_by_ft[filetype] and not options.force_reload then
+  if (formatters_by_ft[filetype] and not options.force_reload) or not lvim.lang[filetype] then
```
Tastyep, line 66: Can you reverse the condition's order please :)

abzcoding: sure

Tastyep (commented):
There is something I don't understand, null-ls is setup in `common_on_attach`, which means that `lvim.lang[filetype]` has to be defined. Where is the real problem ?

abzcoding:
> There is something I don't understand, null-ls is setup in `common_on_attach`, which means that `lvim.lang[filetype]` has to be defined. Where is the real problem ?

`rmd.lua` is symlinked to `r.lua`
Screen Shot 2021-08-19 at 6 27 05 PM

maybe that is the problem 🤷

Tastyep (approved):
Alright then :+1:

JulianTao:
> @JulianTao can you confirm this fixes your issues?

Yes, I confirm that the issue is fixed. Thank you!

PR #1354, opened 2021-08-19
fix: set runtime directories correctly

### LINKED ISSUE #1352: [BUG]: Runtime directories are not set correctly
# Description

There are inconsistencies in how `runtimepath` and `packpath` are. There neovim's defaults that weren't removed in some cases, and in other instances, lunarvim's directory are missing.

That's why the shim that's used to launch lunarvim contains a command for modifying `runtimepath`

https://github.com/LunarVim/LunarVim/blob/0bf6fd68c625396d83eb8fca8d7248c5543ae6f1/utils/bin/lvim#L3

Here's what lunarvim gives me for `&packpath` and `&rtp`
```lua
> rtp list:
"/home/hatsu/.local/share/lunarvim/lvim/after"
"/home/hatsu/.local/share/nvim/site"
"/home/hatsu/.local/share/nvim/site/after"
"/home/hatsu/.local/share/lunarvim/lvim"
"/home/hatsu/.config/lvim"

> pack list:
"/home/hatsu/.config/lvim"
"/home/hatsu/.local/share/lunarvim/site"
"/home/hatsu/.local/share/nvim/site/after"
```

Here's what the default one should be:
```lua
> rtp list:
"/home/hatsu/.config/nvim"
"/home/hatsu/.local/share/nvim/site"
"/home/hatsu/.local/share/nvim/site/after"
"/home/hatsu/.config/nvim/after"

> pack list:
"/home/hatsu/.config/nvim"
"/home/hatsu/.local/share/nvim/site"
"/home/hatsu/.local/share/nvim/site/after"
"/home/hatsu/.config/nvim/after"
```

kylo252:
# Description

This also simplifies the way to invoke LunarVim to just be:
`nvim -u "$LUNARVIM_RUNTIME_DIR/lvim/init.lua"`

Fixes #1352

## How Has This Been Tested?

Add this to test the rtp variable more easily

```lua
print("-- rtp list: ")
local rtp_list = vim.opt.rtp:get()
local user_rtp_list = {}
for _, path in pairs(rtp_list) do
    if vim.startswith(path, "/home") and not path:match(".*packer.*") then
        table.insert(user_rtp_list, path)
        print(vim.inspect(path))
    end
end

print("-- pack list: ")
local pack_list = vim.opt.packpath:get()
local user_pack_list = {}
for _, path in pairs(pack_list) do
    if vim.startswith(path, "/home") and not path:match(".*packer.*") then
        table.insert(user_pack_list, path)
        print(vim.inspect(path))
    end
end
```

You can now start LunarVim by just using `nvim -u "$LUNARVIM_RUNTIME_DIR/lvim/init.lua"` :)

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -1,24 +1,21 @@
---    __                          _    ___
---   / /   __  ______  ____ _____| |  / (_)___ ___
---  / /   / / / / __ \/ __ `/ ___/ | / / / __ `__ \
--- / /___/ /_/ / / / / /_/ / /   | |/ / / / / / / /
---/_____/\__,_/_/ /_/\__,_/_/    |___/_/_/ /_/ /_/
+-- {{{ Bootstrap
+local home_dir = vim.loop.os_homedir()
 
---- Bootstrap
-vim.opt.rtp:append(os.getenv "HOME" .. "/.local/share/lunarvim/lvim")
+vim.opt.rtp:append(home_dir .. "/.local/share/lunarvim/lvim")
 
-vim.opt.rtp:remove(os.getenv "HOME" .. "/.config/nvim")
-vim.opt.rtp:remove(os.getenv "HOME" .. "/.config/nvim/after")
-vim.opt.rtp:append(os.getenv "HOME" .. "/.config/lvim")
-vim.opt.rtp:append(os.getenv "HOME" .. "/.config/lvim/after")
+vim.opt.rtp:remove(home_dir .. "/.config/nvim")
+vim.opt.rtp:remove(home_dir .. "/.config/nvim/after")
+vim.opt.rtp:append(home_dir .. "/.config/lvim")
+vim.opt.rtp:append(home_dir .. "/.config/lvim/after")
 
-vim.opt.rtp:remove(os.getenv "HOME" .. "/.local/share/nvim/site")
-vim.opt.rtp:remove(os.getenv "HOME" .. "/.local/share/nvim/site/after")
-vim.opt.rtp:append(os.getenv "HOME" .. "/.local/share/lunarvim/site")
-vim.opt.rtp:append(os.getenv "HOME" .. "/.local/share/lunarvim/site/after")
+vim.opt.rtp:remove(home_dir .. "/.local/share/nvim/site")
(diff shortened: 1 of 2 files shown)


init.lua
```diff
@@ -1,16 +1,24 @@
...
+--/_____/\__,_/_/ /_/\__,_/_/    |___/_/_/ /_/ /_/
 
-  set runtimepath-=~/.config/nvim
-  set runtimepath-=~/.config/nvim/after
-  set runtimepath+=~/.config/lvim
-  set runtimepath^=~/.local/share/lunarvim/lvim/after
-]]
--- vim.opt.rtp:append() instead of vim.cmd ?
+--- Bootstrap
+vim.opt.rtp:append(os.getenv "HOME" .. "/.local/share/lunarvim/lvim")
```
Tastyep, line 8: local home = os.getenv "HOME"

init.lua
```diff
@@ -1,16 +1,24 @@
-vim.cmd [[
-  set packpath-=~/.config/nvim
-  set packpath-=~/.config/nvim/after
-  set packpath-=~/.local/share/nvim/site
-  set packpath^=~/.local/share/lunarvim/site
-  set packpath^=~/.config/lvim
+--    __                          _    ___
```
Tastyep, line 1: Do we need this ?

Tastyep (changes requested):

init.lua
vlcinsky: or
```
local home = vim.loop.os_homedir()
```
what shall simplify future modifications for MS Windows environment.

Anyway, it seems usable as is with the repeated `os.getenv`

init.lua
vlcinsky: I would remove this banner.

vlcinsky (approved):
The PR is very well usable as is.

I like the care of configuring all the paths not leaving so much space for unexpected side effects.

There are only minor things which are not affecting usability.

init.lua
kylo252: rip banner 😢

Tastyep (approved):

vlcinsky (approved):

PR #1175, opened 2021-07-30
[Refactor/Bugfix] move on_config_done callbacks to relevant setup()

devtoi:
# Description

An issue has been detected with some plugins having trouble when having `require` called twice on them. This would result in weird issues when you are using the on_config_done callback on a troublesome plugin. This PR moves the relevant on_config_done callbacks to the setup function of the related plugins. This way there will be no double `require` and should solve several hard to detect issues.

Includes/depends on #1174 and #1171 
Fixes #1348

## How Has This Been Tested?

Added this to my `lv-config.lua`:
```lua
lvim.builtin.autopairs.active = true
lvim.builtin.compe.enabled = true
lvim.builtin.dap.active = true
lvim.builtin.dashboard.active = true
lvim.builtin.terminal.active = true
lvim.builtin.galaxyline.active = true
lvim.builtin.gitsigns.active = true
lvim.builtin.rooter.active = true
lvim.builtin.telescope.active = true
lvim.builtin.which_key.active = true

lvim.builtin.lspinstall.on_config_done = function(module)
	vim.schedule(function()
		print("lspinstall: " .. vim.inspect(module ~= nil))
	end)
end

lvim.builtin.telescope.on_config_done = function(module)
	vim.schedule(function()
		print("telescope: " .. vim.inspect(module ~= nil))
	end)
end

lvim.builtin.compe.on_config_done = function(module)
	vim.schedule(function()
		print("compe: " .. vim.inspect(module ~= nil))
	end)
end

lvim.builtin.autopairs.on_config_done = function(module)
	vim.schedule(function()
		print("autopairs: " .. vim.inspect(module ~= nil))
	end)
end

lvim.builtin.treesitter.on_config_done = function(module)
	vim.schedule(function()
		print("treesitter: " .. vim.inspect(module ~= nil))
	end)
end

lvim.builtin.nvimtree.on_config_done = function(module)
	vim.schedule(function()
		print("nvimtree: " .. vim.inspect(module ~= nil))
	end)
end

lvim.builtin.gitsigns.on_config_done = function(module)
	vim.schedule(function()
		print("gitsigns: " .. vim.inspect(module ~= nil))
	end)
end

lvim.builtin.which_key.on_config_done = function(module)
	vim.schedule(function()
		print("which_key: " .. vim.inspect(module ~= nil))
	end)
end

lvim.builtin.comment.on_config_done = function(module)
	vim.schedule(function()
		print("comment: " .. vim.inspect(module ~= nil))
	end)
end

[...]

PR #1366, opened 2021-08-20
[fix]: don't overwrite dashboard config

kylo252:
# Description

Don't overwrite dashboard config from the project setup function. Also, allow the rest of the settings to be configurable.

## How Has This Been Tested?

It was always setting the second slot in the dashboard to be "Recent Projects", that is not the case anymore.

### DIFF
--- a/lua/core/dashboard.lua
+++ b/lua/core/dashboard.lua
@@ -31,7 +31,10 @@ M.config = function()
         description = { "  Find File          " },
         command = "Telescope find_files",
       },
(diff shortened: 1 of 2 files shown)


abzcoding:
seems to be fine, I'm just waiting for https://github.com/LunarVim/LunarVim/pull/1175 to be merged
cause it would cause conflicts with that one

Tastyep (approved):

rebuilt (approved):

PR #1367, opened 2021-08-21
[Feature] Add Vi Mode name instead of just bar with color in lualine

razzkumar:
# Description

Add Vi Mode name instead of just bar with color in lualine

pic-selected-210821-0635-20

## How Has This Been Tested?
Is show mode name instead of just color


razzkumar:
oh got thank you

Issue #1015, opened 2021-07-18
[Feature]: Make testing great again


abzcoding:
we need to test a couple of things after the installation process:

- [ ] Install workflow runs only if lint and formatting was passed
- [ ] LV can install `treesitter` stuff without issue
- [x] LV can run `:PackerSync` without issues
- [ ] LV can install a LSP without issues
- [ ] LV can successfully load and format a sample file

PR #1369, opened 2021-08-21
[Feature]: Add some very basic unit-tests

kylo252:
# Description

A first attempt at trying some generic unit-tests. Helps towards #1015.

TODO: 
[ ] being able to parse the output of `checkhealth`
[ ] being able to see the status of Packer commands
[ ] fix installing TreeSitter grammar on the CI

## How Has This Been Tested?

`make test`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,32 +1,33 @@
-
 SHELL := /bin/bash
 
 install:
 	@echo Starting LunarVim Installer
-	bash ./utils/installer/install.sh
-
-uninstall:
-	@echo TODO: this is currently not supported 
+	./utils/installer/install.sh
 
 install-neovim-binary:
 	@echo Installing Neovim from github releases
-	bash ./utils/installer/install-neovim-from-release
+	./utils/installer/install-neovim-from-release
 
-test:
-	bash ./utils/bin/test_runner.sh "$(UNIT)"
+uninstall:
+	@echo TODO: this is currently not supported
 
 lint: lint-lua lint-sh
 
-style: style-lua style-sh
-
 lint-lua:
 	luacheck *.lua lua/* tests/*
+
 lint-sh:
 	shfmt -f . | grep -v jdtls | xargs shellcheck
 
-style-lua:
-	stylua --check .
-style-sh:
-	shfmt -f . | grep -v jdtls | xargs shfmt -i=2 -ci -w
+style: style-lua style-sh
 
-.PHONY: test lint style install install-neovim-binary uninstall
+style-lua:
+	stylua --config-path .stylua.toml --check .
+
+style-sh:
+	shfmt -f . | grep -v jdtls | xargs shfmt -i 2 -ci -l -d
+
+test:
+	./utils/bin/test_runner.sh "$(TEST)"
+
+.PHONY: install install-neovim-binary uninstall lint style test
--- a/Makefile
+++ b/Makefile
@@ -1,19 +1,21 @@
 test:
 	nvim --headless -u ./init.lua -c "PlenaryBustedDirectory tests/ { minimal_init = './init.lua' }"
-#	nvim --headless -u tests/minimal_init.lua -c "PlenaryBustedDirectory tests/ { minimal_init = './tests/minimal_init.lua' }"
+
+test-minimal:
+	nvim --headless -u tests/minimal_init.lua -c "PlenaryBustedDirectory tests/ { minimal_init = './tests/minimal_init.lua' }"
 
 lint: lint-lua lint-sh
 
-lint-lua:
-	pre-commit run --all-files luacheck
-lint-sh:
-	pre-commit run --all-files shellcheck
-
 style: style-lua style-sh
 
-style-lua:
-	pre-commit run --all-files stylua
-style-sh:
-	pre-commit run --all-files shfmt
+lint-lua:
+	luacheck *.lua lua/* tests/*
+lint-sh:
+	shfmt -f . | grep -v jdtls | xargs shellcheck
 
-.PHONY: test lint lint-sh lint-lua style style-lua style-sh
+style-lua:
+	stylua --check .
+style-sh:
+	shfmt -f . | grep -v jdtls | xargs shfmt -i=2 -ci -w
+
+.PHONY: test test-minimal lint lint-sh lint-lua style style-lua style-sh
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,19 @@
-test:
-	nvim --headless -u ./init.lua -c "PlenaryBustedDirectory tests/ { minimal_init = './init.lua' }"
 
-test-minimal:
-	nvim --headless -u tests/minimal_init.lua -c "PlenaryBustedDirectory tests/ { minimal_init = './tests/minimal_init.lua' }"
+SHELL := /bin/bash
+
+install:
+	@echo Starting LunarVim Installer
+	bash ./utils/installer/install.sh
+
+uninstall:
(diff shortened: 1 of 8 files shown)


rebuilt:
I'm not sure why the treesitter test is failing.  I run `:TSInstall lua` from neovim but the test for it fails
```
Fail	||	initial start should be able to run treesitter without errors	
            ...packer/start/plenary.nvim/lua/plenary/async_lib/util.lua:320: Blocking on future failed ...acker/start/plenary.nvim/lua/plenary/async_lib/async.lua:14: The coroutine failed with this message: ...lson/.local/share/lunarvim/lvim/tests/bootstrap_spec.lua:19: Vim:E492: Not an editor command: TSInstall lua
```

Makefile
```diff
@@ -0,0 +1,19 @@
+test:
+	nvim --headless -u ./init.lua -c "PlenaryBustedDirectory tests/ { minimal_init = './init.lua' }"
+#	nvim --headless -u tests/minimal_init.lua -c "PlenaryBustedDirectory tests/ { minimal_init = './tests/minimal_init.lua' }"
+
+lint: lint-lua lint-sh
+
+lint-lua:
+	pre-commit run --all-files luacheck
```
Tastyep, line 8: I think pre-commit should rely on the Makefile more than the other way around

tests/bootstrap_spec.lua
```diff
@@ -0,0 +1,34 @@
+local a = require "plenary.async_lib.tests"
+
+a.describe("initial start", function()
```
Tastyep, line 3: Can you add newlines and make the code `breathable` ?

tests/bootstrap_spec.lua
```diff
@@ -0,0 +1,34 @@
+local a = require "plenary.async_lib.tests"
```
Tastyep, line 1: Why not using `describe` and `it` directly ?

tests/bootstrap_spec.lua
```diff
@@ -0,0 +1,34 @@
...
+
+a.describe("initial start", function()
+  local uv = vim.loop
+  local home_dir = uv.os_homedir()
+  local lvim_config_path = home_dir .. "/.config/lvim"
+  local lvim_runtime_path = home_dir .. "/.local/share/lunarvim/lvim"
+  a.it("should not be reading default neovim directories in the home directoies", function()
```
Tastyep, line 8: Check https://github.com/Olivine-Labs/luassert#luassert to work the asserters you are using in the file

tests/bootstrap_spec.lua
```diff
@@ -0,0 +1,34 @@
...
+    local found_illegal_path = vim.tbl_contains(rtp_list, vim.fn.stdpath "config")
+    assert.same(found_illegal_path, false)
+  end)
+  a.it("should be able to read lunarvim directories", function()
+    local rtp_list = vim.opt.rtp:get()
+    assert.is_true(vim.tbl_contains(rtp_list, lvim_runtime_path))
+    assert.is_true(vim.tbl_contains(rtp_list, lvim_config_path))
+  end)
+  a.it("should be able to run treesitter without errors", function()
+    vim.cmd [[TSInstall lua]]
```
Tastyep, line 19: Would it work to do a pcall(vim.cmd ....) and check the return value ?

Tastyep (changes requested):

Makefile
kylo252: I was just being lazy 😅

tests/bootstrap_spec.lua
kylo252: It's `luacheck` or `sumneko` that kept complaining. do I have to add these keywords to the globals list?

tests/bootstrap_spec.lua
kylo252: can you expand on that? do you mean that we should add extra asserts or what exactly? Isn't this library included with plenary anyway?

tests/bootstrap_spec.lua
Tastyep: That's what I would do if possible

tests/bootstrap_spec.lua
Tastyep: For instance, `assert.same(found_illegal_path, false)` could be `assert.False(found_illegal_path)`
I would probably also switch `assert.is_true` with `assert.True`

tests/bootstrap_spec.lua
kylo252: That should work to check if the command exists or not, but it doesn't look too reliable to me, examples:

```lua
local status, msg = pcall(vim.cmd, "TSInstall lua") -- status -> true
local status, msg = pcall(vim.cmd, "TSInstall gibberish") -- status -> false
local status, msg = pcall(vim.cmd, "checkhealth treesitter") -- status -> true
local status, msg = pcall(vim.cmd, "checkhealth gibberish") -- status -> true ... wtf!
```

Makefile
```diff
@@ -0,0 +1,19 @@
+test:
```
Tastyep, line 1: It would be nice to be able to specify a `file` argument to run only specific tests.
That's how I would do it:
https://github.com/Tastyep/structlog.nvim/blob/main/Makefile#L2
https://github.com/Tastyep/structlog.nvim/blob/main/script/test.sh

kylo252:
> I'm not sure why the treesitter test is failing. I am able to run `:TSInstall lua` manually but the test for it fails. It's failing at this statement ` vim.cmd [[TSInstall lua]]` not the following pcall or assertion.
> 
> ```
> Fail	||	initial start should be able to run treesitter without errors	
>             ...packer/start/plenary.nvim/lua/plenary/async_lib/util.lua:320: Blocking on future failed ...acker/start/plenary.nvim/lua/plenary/async_lib/async.lua:14: The coroutine failed with this message: ...lson/.local/share/lunarvim/lvim/tests/bootstrap_spec.lua:19: Vim:E492: Not an editor command: TSInstall lua
> ```

It looks like TreeSitter failed to install or load.

Makefile
kylo252: Yeah for sure, I can just add additional targets.

tests/bootstrap_spec.lua
kylo252: I was able to get `luacheck` to recognize the library, but I could not do the same for the language server..

Makefile
```diff
@@ -0,0 +1,32 @@
+
```
Tastyep, line 1: blankline

Tastyep (changes requested):

Makefile
```diff
@@ -0,0 +1,32 @@
...
+lint: lint-lua lint-sh
+
+style: style-lua style-sh
+
+lint-lua:
+	luacheck *.lua lua/* tests/*
+lint-sh:
+	shfmt -f . | grep -v jdtls | xargs shellcheck
+
+style-lua:
```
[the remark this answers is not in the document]
kylo252: I updated both `shfmt` and `luacheck` to use `make` since they're a bit manual, and switched to a pre-made one for `stylua` which already has all the rules in a file independent of `make` anyway.

Makefile
```diff
@@ -0,0 +1,32 @@
+
+SHELL := /bin/bash
+
+install:
+	@echo Starting LunarVim Installer
+	bash ./utils/installer/install.sh
```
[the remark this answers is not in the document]
kylo252: so reverting it again since github didn't like it.. 😞

Tastyep (approved):

PR #1370, opened 2021-08-21
fix: limit shfmt scope to utils to avoid surprises

kylo252:
# Description

Limit `shfmt` scope to `./utils` because it was somehow wanting to run on `lunarvim/site`

## How Has This Been Tested?

`pre-commit run shfmt`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,14 +7,14 @@ repos:
         language: system
         types: [shell]
         entry: bash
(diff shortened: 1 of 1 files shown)


Tastyep (changes requested):
As discussed, we should find what the issue is instead of working it around

kylo252:
We'll leave this alone for now since the original culprit was a faulty packer installation on my side :)

PR #1373, opened 2021-08-22
[Refactor]: reload of modules

ahmedkhalf:
# Description

Stop using vimL and use lua instead for hot reloading the plugins and settings.

## How Has This Been Tested?

I pressed `:w` when in the `config.lua`

### DIFF
--- a/lua/utils/init.lua
+++ b/lua/utils/init.lua
@@ -88,11 +88,16 @@ function utils.toggle_autoformat()
 end
 
 function utils.reload_lv_config()
+  -- Hot reload of statusline
   require("core.lualine").config()
-  vim.cmd "source ~/.local/share/lunarvim/lvim/lua/settings.lua"
+
+  -- Remove package cache
+  package.loaded["settings"] = nil
+  package.loaded["plugins"] = nil
+
+  require "settings"
   vim.cmd("source " .. USER_CONFIG_PATH)
   require("keymappings").setup() -- this should be done before loading the plugins
-  vim.cmd "source ~/.local/share/lunarvim/lvim/lua/plugins.lua"
   local plugins = require "plugins"
   local plugin_loader = require("plugin-loader").init()
   utils.toggle_autoformat()


lua/utils/init.lua
```diff
@@ -88,11 +88,16 @@ function utils.toggle_autoformat()
 end
 
 function utils.reload_lv_config()
+  -- Hot reload of statusline
   require("core.lualine").config()
-  vim.cmd "source ~/.local/share/lunarvim/lvim/lua/settings.lua"
+
+  -- Remove package cache
+  package.loaded["settings"] = nil
+  package.loaded["plugins"] = nil
```
Tastyep, line 96: Not necessary, the file won't change

lua/utils/init.lua
```diff
@@ -88,11 +88,16 @@ function utils.toggle_autoformat()
 end
 
 function utils.reload_lv_config()
+  -- Hot reload of statusline
   require("core.lualine").config()
-  vim.cmd "source ~/.local/share/lunarvim/lvim/lua/settings.lua"
+
+  -- Remove package cache
+  package.loaded["settings"] = nil
```
Tastyep, line 95: Also not necessary + this does nothing since no code is evaluated at module loading time.
IMO, `require "settings"` is not useful

lua/utils/init.lua
```diff
@@ -88,11 +88,16 @@ function utils.toggle_autoformat()
 end
 
 function utils.reload_lv_config()
+  -- Hot reload of statusline
```
Tastyep, line 91: Comment is not accurate, it doesn't hot reload anything, it configures it, and since the code is explicit enough, the comment is not useful

Tastyep (changes requested):

Tastyep:
We should probably first discuss the outcome of #1338 and maybe even make the changes on that one instead

lua/utils/init.lua
ahmedkhalf: Will delete comment

lua/utils/init.lua
ahmedkhalf: Agreed

lua/utils/init.lua
ahmedkhalf: Actually no, plugins here is necessary in case the user changes the `active` options of a builtin.

Tastyep: Indeed, we should probably wrap data that can be reloaded into functions, another thing to add to my notes :)

kylo252:
Fixed in https://github.com/LunarVim/LunarVim/pull/1699

PR #1377, opened 2021-08-23
[Refactor]: Disable sumneko_lua telemetry by default

ahmedkhalf:
# Description

Disable's telemetry from sumneko lua lsp by default.

## How Has This Been Tested?

N/A, but this setting is also in folke/luadev so it must work.


Tastyep:
It's already `false` [by default](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/sumneko_lua.lua#L13)

PR #1381, opened 2021-08-23
refactor: use more flexible paths

kylo252:
# Description

Remove most of the hard-coded paths that can lead to trouble on other platforms. 

NOTE: We are not able to `require` any modules before adjusting the `&rtp`. So any utility functions that are to be used need to be manually defined in `init.lua`.

TODO: 
[ ] make it work on the CI
[ ] fix `runtimepath` for windows, possible use it unmodified

## How Has This Been Tested?

It should start as normal.

### DIFF
--- a/lua/bootstrap.lua
+++ b/lua/bootstrap.lua
@@ -1,0 +1,82 @@
+local M = {}
+-- It's not safe to require 'utils' without adjusting the runtimepath
+function _G.join_paths(...)
+  local uv = vim.loop
+  local path_sep = uv.os_uname().version:match "Windows" and "\\" or "/"
+  local result = table.concat({ ... }, path_sep)
+  return result
+end
+
+function _G.get_runtime_dir()
+  local lvim_runtime_dir = os.getenv "LUNARVIM_RUNTIME_DIR"
+  if not lvim_runtime_dir then
+    -- when nvim is used directly
+    return vim.fn.stdpath "config"
+  end
+  return lvim_runtime_dir
+end
+
+function _G.get_config_dir()
+  local lvim_config_dir = os.getenv "LUNARVIM_CONFIG_DIR"
+  if not lvim_config_dir then
+    return vim.fn.stdpath "config"
+  end
+  return lvim_config_dir
+end
+
+function _G.get_cache_dir()
+  local lvim_cache_dir = os.getenv "LUNARVIM_CACHE_DIR"
+  if not lvim_cache_dir then
+    return vim.fn.stdpath "cache"
+  end
+  return lvim_cache_dir
+end
+
+function M:init()
+  self.runtime_dir = get_runtime_dir()
+  self.config_dir = get_config_dir()
+  self.cache_path = get_cache_dir()
+
+  self.pack_dir = join_paths(self.runtime_dir, "site", "pack")
+
+  if os.getenv "LUNARVIM_RUNTIME_DIR" then
+    vim.opt.rtp:remove(join_paths(vim.fn.stdpath "data", "site"))
+    vim.opt.rtp:remove(join_paths(vim.fn.stdpath "data", "site", "after"))
+    vim.opt.rtp:prepend(join_paths(self.runtime_dir, "site"))
+    vim.opt.rtp:append(join_paths(self.runtime_dir, "site", "after"))
+
+    vim.opt.rtp:remove(vim.fn.stdpath "config")
+    vim.opt.rtp:remove(join_paths(vim.fn.stdpath "config", "after"))
+    vim.opt.rtp:prepend(self.config_dir)
+    vim.opt.rtp:append(join_paths(self.config_dir, "after"))
+    -- TODO: we need something like this: vim.opt.packpath = vim.opt.rtp
+
+    vim.cmd [[let &packpath = &runtimepath]]
+    vim.cmd("set spellfile=" .. join_paths(self.config_dir, "spell", "en.utf-8.add"))
+  end
+
+  -- FIXME: currently unreliable in unit-tests
+  if not os.getenv "LVIM_TEST_ENV" then
+    require("impatient").setup {
+      path = vim.fn.stdpath "cache" .. "/lvim_cache",
+      enable_profiling = true,
+    }
+  end
+
+  local config = require "config"
+  config:init {
+    path = join_paths(self.config_dir, "config.lua"),
+  }
+
+  require("plugin-loader"):init {
+    cache_path = self.cache_path,
+    runtime_dir = self.runtime_dir,
+    config_dir = self.config_dir,
+    package_root = join_paths(self.runtime_dir, "site", "pack"),
+    compile_path = join_paths(self.config_dir, "plugin", "packer_compiled.lua"),
+  }
+
+  return self
+end
+
+return M
--- a/lua/core/telescope.lua
+++ b/lua/core/telescope.lua
@@ -1,5 +1,7 @@
 local M = {}
 
+local utils = require "utils"
+
 function M.config()
   -- Define this minimal config so that it's available if telescope is not yet available.
   lvim.builtin.telescope = {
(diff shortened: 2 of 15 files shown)


rebuilt:
I'll be using this as my branch today

lua/bootstrap.lua
```diff
@@ -0,0 +1,48 @@
...
+  vim.opt.rtp:remove(join_paths(default_runtime_dir, "site"))
+  vim.opt.rtp:remove(join_paths(default_runtime_dir, "site", "after"))
+
+  vim.opt.rtp:append(join_paths(LUNARVIM_RUNTIME_DIR, "site"))
+  vim.opt.rtp:append(join_paths(LUNARVIM_RUNTIME_DIR, "site", "after"))
+
+  -- TODO: we need something like this: vim.opt.packpath = vim.opt.rtp
+  vim.cmd [[let &packpath = &runtimepath]]
+
+  vim.cmd("set spellfile=" .. require("utils").join_paths(LUNARVIM_CONFIG_DIR, "spell", "en.utf-8.add"))
```
Tastyep, line 45: What is this for ?

lua/core/telescope.lua
```diff
@@ -96,7 +96,7 @@ function M.find_lunarvim_files(opts)
     },
     prompt = ">> ",
     prompt_title = "~ LunarVim files ~",
-    cwd = CONFIG_PATH,
+    cwd = require("utils").join_paths(LUNARVIM_RUNTIME_DIR, "lvim"),
```
Tastyep, line 99: Can we store `utils` in a local variable ? I don't mind when we call a single function, but here it's used multiple times.
https://github.com/luarocks/lua-style-guide#modules

lua/lsp/init.lua
```diff
@@ -153,4 +143,29 @@ function M.setup(lang)
   end
 end
 
+function M.global_setup()
```
Tastyep, line 146: I like this, to make naming more consistent with builtins and my changes in 1406, are you ok renaming this to `setup` ?

lua/utils/init.lua
```diff
@@ -127,6 +127,12 @@ function utils.is_file(filename)
   return stat and stat.type == "file" or false
 end
 
+function utils.join_paths(...)
+  local path_sep = vim.loop.os_uname().version:match "Windows" and "\\" or "/"
+  local result = table.concat(vim.tbl_flatten { ... }, path_sep):gsub(path_sep .. "+", path_sep)
```
Tastyep, line 132: Why do you need to wrap `...` into a table and then flatten it ?

utils/installer/install.ps1
```diff
@@ -0,0 +1,312 @@
+$ErrorActionPreference = "Stop" # exit when command fails
```
Tastyep, line 1: What do we do with https://github.com/LunarVim/LunarVim/pull/1261 ?

utils/installer/install.sh
```diff
@@ -2,24 +2,37 @@
 set -eo pipefail
 
 #Set branch to master unless specified by the user
-declare -r LV_BRANCH="${LV_BRANCH:-rolling}"
-declare -r LV_REMOTE="${LV_REMOTE:-lunarvim/lunarvim.git}"
-declare -r INSTALL_PREFIX="${INSTALL_PREFIX:-"$HOME/.local"}"
+declare LV_BRANCH="${LV_BRANCH:-"rolling"}"
```
Tastyep, line 5: Outer quotes are unnecessary for variables between braces
`LV_BRANCH=${LV_BRANCH:-"rolling"}`

utils/installer/install.sh
```diff
@@ -247,6 +264,7 @@ function clone_lvim() {
 }
 
 function setup_shim() {
+
```
Tastyep, line 267: newline

utils/installer/install.sh
```diff
@@ -262,6 +280,13 @@ EOF
 }
 
 function setup_lvim() {
+  # no need to install in shared_mode
+  if [ -n "$SHARE_NEOVIM_CONFIG" ]; then
+    echo "You can now add 'require \"lvim\"' in your ~/.config/init.lua"
```
Tastyep, line 285: I think this is incompatible with my changes in 1406, also how would it work if the config requires lvim and lvim loads the config, don't we have an import loop ?
Since we made lvim a "separate instance", I wonder if we want to support this configuration

Tastyep (changes requested):

utils/installer/install.ps1
kylo252: I just wanted the installer here to test the branch. I'll port the changes back to #1261 and remove the script from here before merging it.

utils/installer/install.sh
kylo252: It was actually the other way around, but I thought you liked them to be consistent 😄

utils/installer/install.sh
kylo252: I was afraid they might be incompatible. 

I actually think that this is the only reliable way to use lunarvim on Windows, it's also probably the easiest way to use it with a GUI.

There's no `lvim` binary anymore in this mode, you just start `nvim` like usual and require `lvim` at some point.

lua/lsp/init.lua
kylo252: yeah I just wanted to avoid confusion about what this is actually doing, because it's just setting `nlsp`, `null-ls` and some configurations. There's still the actual function `setup()` that's language specific and needs to be explicitly called.

so what do you think then?

lua/bootstrap.lua
```diff
@@ -0,0 +1,48 @@
...
+end
+
+function M:init()
+  local default_config_dir = vim.fn.stdpath "config"
+  local default_runtime_dir = vim.fn.stdpath "data"
+  local default_cache_dir = vim.fn.stdpath "cache"
+
+  DATA_PATH = default_runtime_dir
+  CACHE_PATH = default_cache_dir
+  TERMINAL = vim.fn.expand "$TERMINAL"
```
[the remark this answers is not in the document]
kylo252: That would be nice. I'll double check and delete them if that's the case.

lua/bootstrap.lua
kylo252: I have no idea but it broke neovim when I didn't have it set. It was already defined from before, just shorter and more hard-coded 😃

kylo252: possibly related to https://github.com/LunarVim/LunarVim/issues/1445

utils/installer/install.sh
Tastyep: I'm wondering how cna you require lvim using nvim when it is not in the runtime path, I think my reasons for saying this is incompatible with my changes would be the same for the current implementation. So maybe this works and I'm missing something :)

Note that the path `~/.config/init.lua"` seems incorrect

rebuilt:
It was looking for the config in the nvim folder.  
```lua                                          
Vim(luafile):E5112: Error while creating lua chunk: cann
ot open /home/nelson/.config/nvim/lv-config.lua: No such
 file or directory        
 ```
 So I copied my config to the nvim folder.  It starts without errors but I can't install any plugins.  
 I get this erro I get this error when running `:PackerSync`.  (I've already removed the packer_compiled folder)r when running `:PackerSync`.  (I've already removed the packer_compiled folder)
 ```lua
 [packer.nvim] [ERROR 14:36:35] async.lua:20: Error in co
routine: Vim:E944: Reverse range in character class 
 ```

kylo252:
> So I copied my config to the nvim folder. It starts without errors but I can't install any plugins.

Do you mean no plugins at all? or user-plugins? 

What worked for me so far is having lunarvim repo inside `~/.config/nvim` and created an empty file `config.lua` in the same directory. I'll try to see where else I can move it.

rebuilt:
I tried on my main machine and it worked fine.  I went back to my other machine to figure out why that one was failing. 

The difference is, I was using the old lvim launcher in `/usr/local/bin/lvim`.   User's with older versions of lunarvim will not be able to `git pull` this update.  They will need to do a full reinstall or manually edit the lvim launcher.

kylo252:
> The difference is, I was using the old lvim launcher in `/usr/local/bin/lvim`. User's with older versions of lunarvim will not be able to `git pull` this update. They will need to do a full reinstall or manually edit the lvim launcher.

The idea is that you start without the using `lvim`. But for now I'm not able to let it run from anywhere else on the `&rtp` other than `~/.config/nvim`. Maybe I should just leave it at that for the time being, since we can always add `config.lua` to `.gitignore`.

Tastyep (approved):

rebuilt (approved):

PR #1406, opened 2021-08-26
[Refactor] update config structure

Tastyep:
# Description

https://github.com/LunarVim/LunarVim/projects/1#card-67485846

### Feature
- Adapt the user configuration to return a table instead of editing in-place (same for default-config)
- Make lvim completely reloadable.
- Validate configuration entries by type, see `config:merge`
- Add state to builtins (removed from global `lvim`)
- Allow builtins to define which-key mappings

### improvement
- Use `loadfile` instead of vim command.
- Trim down config-no-ts <- Does it need to be documented like the other one ?
- Improve the documentation of config.example
- Split keymappings between configuration and logic files
- Move Keymap logic in `core.service.keymap`
- Move builtins in `core.builtins`
- Swap plugins' config and setup method names (setup first, configure later)
- Remove settings/keymappings/autocmds loading from config:init
- Rely on the builtins' state instead of global lvim (ex: `local active = require("core.builtins.xyz").config.active`)

### Bugfix
- Fix possible conflicts between plugin's configuration and our own
- Remove our plugins' default configurations when only duplicating information. <- This will require quality reviews
- Correctly handle the second argument  (option) when registering a mapping
- Rename builtin to builtins
- Fix function style issue `M.func = function()` -> `function M.func()`
- Remove lot of ugly code

**WARNING**: 

Because the format of the configuration has changed, please back up your `config.lua`. To test this PR, create a new `config.lua` with `return {}` if you don't want to adapt your config yet. Once/If this gets merged, you will have to convert from the old format to the new one.

Don't forget to regenerate the packer_compiled file


abzcoding:
quick question, do we want to merge this before next release, or after releasing a version first?

Tastyep:
> quick question, do we want to merge this before next release, or after releasing a version first?

That's up to us, I would do it before the major release as it's such a big change.

lua/config/autocmds.lua
```diff
@@ -0,0 +1,103 @@
...
+      { "BufNewFile", "*.zsh", "setlocal filetype=sh" },
+    },
+    -- _solidity = {
+    --     {'BufWinEnter', '.sol', 'setlocal filetype=solidity'}, {'BufRead', '*.sol', 'setlocal filetype=solidity'},
+    --     {'BufNewFile', '*.sol', 'setlocal filetype=solidity'}
+    -- },
+    -- _gemini = {
+    --     {'BufWinEnter', '.gmi', 'setlocal filetype=markdown'}, {'BufRead', '*.gmi', 'setlocal filetype=markdown'},
+    --     {'BufNewFile', '*.gmi', 'setlocal filetype=markdown'}
+    -- },
```
kylo252, line 64: I think it's safe to remove this since it can just live in the user-config.lua.

lua/config/autocmds.lua
```diff
@@ -0,0 +1,103 @@
+local autocmds = {}
+
+function autocmds.setup(config)
+  config:extend {
```
kylo252, line 4: Why is `git` not treating this as a rename? If it's because of this line, maybe try 

```lua
local autocmd_groups = {
  _general_settings = {
  --
  }
}
  
function autocmds.setup(config)
  config:extend(autocmd_groups)
end
```

lua/core/gitsigns.lua
```diff
@@ -1,63 +1,21 @@
...
-        noremap = true,
-        buffer = true,
-      },
-      watch_index = { interval = 1000 },
-      sign_priority = 6,
-      update_debounce = 200,
-      status_formatter = nil, -- Use default
-    },
-  }
+    config = {},
```
kylo252, line 5: Why did this suddenly get deleted?

lua/core/lualine/init.lua
```diff
@@ -1,46 +1,29 @@
-local M = {}
-M.config = function()
-  lvim.builtin.lualine = {
+local M = {
+  defaults = {
     active = true,
-    style = "lvim",
```
kylo252, line 5: Same comment as gitsigns. This will remove the autocomplete "feature".

lua/core/project.lua
```diff
@@ -1,50 +1,29 @@
...
-  lvim.builtin.project = {
+local M = {
+  defaults = {
     ---@usage set to false to disable project.nvim.
     --- This is on by default since it's currently the expected behavior.
     active = true,
 
     on_config_done = nil,
 
-    ---@usage set to true to disable setting the current-woriking directory
```
kylo252, line 11: Same as others, this will break autocomplete.

lua/main.lua
```diff
@@ -0,0 +1,79 @@
...
+
+  local lsp_settings_status_ok, lsp_settings = pcall(require, "nlspsettings")
+  if lsp_settings_status_ok then
+    local home_dir = vim.loop.os_homedir()
+    lsp_settings.setup {
+      config_home = home_dir .. "/.config/lvim/lsp-settings",
+    }
+  end
+
+  local keymap = require "keymappings"
```
kylo252, line 75: Since we're on the subject, this will actually always overwrite any plugins bindings with lunarvim's defaults.

lua/plugins.lua
```diff
@@ -1,193 +1,198 @@
...
+
+    { "nvim-lua/popup.nvim" },
+    { "nvim-lua/plenary.nvim" },
+    -- Telescope
+    {
+      "nvim-telescope/telescope.nvim",
+      config = function()
+        require("core.telescope"):config()
+      end,
+      disable = not config:get "telescope.active",
```
kylo252, line 27: how about creating a utility function for this? e.g: 
```lua
disable = config:condition "telescope"
``` 
also, isn't `not nil == true` ? is that behavior desirable? because all the plugins must now have the "active" attribute.

kylo252 (commented):
I quite like the idea, my main notes:

1) first reverting to `builtin` will simplify reviewing these, and less headaches when telling people what to use.
2) I don't get why some files are duplicated now. (`keymappings.lua` and `autcocmd.lua`)
3) I have an idea for a naming scheme that I think will actually make the most sense:

image

we can discuss it in more details over discord.

README.md
```diff
@@ -90,7 +90,7 @@ lvim.format_on_save = true
 lvim.lint_on_save = true
 lvim.colorscheme = "onedarker"
 
-lvim.builtin.compe.autocomplete = true
+lvim.builtins.compe.autocomplete = true
```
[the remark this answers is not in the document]
Tastyep: I don't mind moving the commit to a new PR, however, we should agree that we need to be consistent with naming, we can't have variables named in different ways

lua/config/init.lua
```diff
@@ -1,49 +1,117 @@
...
+
+  return entries
+end
+
+--- Merge recursively the given entries with our own.
+-- @param overrides The entries overrides to merge
+-- @param opts Optional parameters
+-- @param opts.force Use the given value, default: True
+-- @param opts.log TODO
+function M:extend(overrides, opts)
```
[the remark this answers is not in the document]
Tastyep: I/We should stop naming non module objects `M`, it would help on the naming part.
WDYT of `Config:merge(overrides, opts)` ?

lua/config/init.lua
```diff
@@ -1,49 +1,117 @@
...
+  opts = opts or {}
+  local keep = not opts.force or false
```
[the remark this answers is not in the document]
Tastyep: 1. Ok
2. I'm waiting on your PR that I reviewed and also asked if I could help you make it move forward.

lua/config/autocmds.lua
Tastyep: I don't mind using the same pattern as for plugins

lua/core/gitsigns.lua
Tastyep: I'm sorry if these were not defaults

lua/core/lualine/init.lua
Tastyep: Are you talking about the auto-completion when writing `lvim.` ?

lua/plugins.lua
Tastyep: > how about creating a utility function for this? e.g:
> 
> ```lua
> disable = config:condition "telescope"
> ```

The configuration should not share the kind of knowledge, especially if you suggest that the function should check the active field.

> also, isn't `not nil == true` ? is that behavior desirable? because all the plugins must now have the "active" attribute.

That's something I said for quite a while :)

lua/core/lualine/init.lua
kylo252: Yeah basically, otherwise I had wanted to remove them ages ago.

lua/plugins.lua
kylo252: > The configuration should not share the kind of knowledge, especially if you suggest that the function should check the active field.

You can make it a local wrapper instead.

lua/main.lua
```diff
@@ -0,0 +1,79 @@
...
+
+    path = lv_config
+  end
+  config.path = path
+
+  local settings = require "config.settings"
+  settings.load_options()
+
+  local keymappings = require "config.keymappings"
+  keymappings.setup(config:sub "keys")
```
kylo252, line 22: Is this calling the `keymappings` module to make it do something or is it the `config` module that's loading the `keymappings` ? If it's the former, there's no reason it should have side-effects like this inside "load" config. If it's the latter, then that syntax is not great.

kylo252 (changes requested):
I think you re-evaluate the some of the architecture details with the config loader. 

Some food for thoughts about this module:
_Is it a singleton? What does it do exactly? What should it be responsible for? why is it [not] handling the defaults internally? where does `main` fit between `init.lua` and the `config-loader`?_

init.lua
```diff
@@ -17,45 +17,14 @@ vim.opt.rtp:append(home_dir .. "/.local/share/lunarvim/site/after")
...
-  lsp_settings.setup {
-    config_home = home_dir .. "/.config/lvim/lsp-settings",
-  }
-end
-
-require("keymappings").setup()
+-- GLOBAL configuration
+-- This varibale is required as packer cannot capture variables using string.dump
+lvim = {}
+require "main"()
```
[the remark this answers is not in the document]
Tastyep: I still need to be able to reload the whole config, I'll make the necessary updates once we merge your PR

lua/core/lualine/init.lua
Tastyep: Please read this https://github.com/LunarVim/LunarVim/projects/1#card-67485846

lua/plugins.lua
Tastyep: I would then prefer to have some kind of factory function handle this job, but yep, we can do that

lua/main.lua
Tastyep: I agree that we should distinguish both, open the files to answer your questions :)

Tastyep:
The goal is to provide a configuration object, agnostic of the core. Preferably, this configuration should be injected everywhere instead of using the global lvim to make the code testable. 

> _Is it a singleton?

That's exactly what I want to avoid, defining a function in terms of input and output will allow us to inject dependencies such as config entries and make our code testable.

> What does it do exactly ? What should it be responsible for ?

The configuration object is responsible for loading / providing (and maybe validating) config entries.

> why is it [not] handling the defaults internally? 

Because doing that you couple core logic with the configuration which should be agnostic. It means that doing tests, you will have to work with a configuration defaulted with lvim's entries. It also makes testing other behaviors impossible.

> where does `main` fit between `init.lua` and the `config-loader`?_

Main sets up lvim, the config loads lvim's configuration entries , init.lua calls main (maybe it could call bootstrap once your PR is merged)

kylo252:
> The goal is to provide a configuration object, agnostic of the core. Preferably, this configuration should be injected everywhere instead of using the global lvim to make the code testable.
> That's exactly what I want to avoid, defining a function in terms of input and output will allow us to inject dependencies such as config entries and make our code testable.

You have inadvertently created a singleton, at least on the surface. Why else would you use a _single_ instance in `main.lua`?   

> The configuration object is responsible for loading / providing (and maybe validating) config entries.
> Because doing that you couple core logic with the configuration which should be agnostic. It means that doing tests, you will have to work with a configuration defaulted to lvim's entries. It also makes testing other behaviors impossible.
> Main sets up lvim, the config loads lvim's configuration entries , init.lua calls main (maybe it could call bootstrap once your PR is merged)

This all sounds good, but it sadly doesn't quite add up. This will basically extract the ugliness from the config-loader and dump it into `main` which now needs to maintain order.

`Config:load()` makes it look a lot less agnostic and decoupled when it actually knows how to: a) rely on some default path b) [de-]serialize data c) decide that we're always gonna prefer "overwrite".

`Config:sub()` looks like it can mutate the data because it's using a direct reference instead of a copy to `self.entries`.
It also has hard requirements about how and where to look for data.

Try dumping the output of `load_config()` and see how massive that table has gotten which leads to `Config:get()` which is either: a) useless because we don't need a function to access the the first level b) used in a very inefficient way because it needs to travers the entire aforementioned giant table over and over for every call.

lua/core/project.lua
Tastyep: I'm closing this as it's an obvious consequence of the task https://github.com/LunarVim/LunarVim/projects/1#card-67485846
I don't see a way for benefiting from the autocompletion while not predefining the whole structure, which should be avoided.

lua/core/lualine/init.lua
Tastyep: Closing for the same reason as below

Tastyep:
> You have inadvertently created a singleton, at least on the surface. Why else would you use a _single_ instance in `main.lua`?   

No and No, I can't retrieve the same instance of config, how is it a singleton ? 
I use a single instance because I only need one, why would you want to use many instances ?
Actually, I do use many 'sub' instances by using the `sub` method

> This all sounds good, but it sadly doesn't quite add up. This will basically extract the ugliness from the config-loader and dump it into `main` which now needs to maintain order.

The 'ugliness' you describe, and its fixed order have to be somewhere, do you have a suggestion ?

> `Config:load()` makes it look a lot less agnostic and decoupled when it actually knows how to: a) rely on some default path b) [de-]serialize data c) decide that we're always gonna prefer "overwrite".

a) It relies on its default path and not a hard coded one, I don't see the issue with that.
b) That's usually what a config loader does, is supports a specific format and parses it, in our case it's lua.
c) should we provide additional options ?

I am open to constructive criticism, but please don't forget to provide suggestions to support your points.

> `Config:sub()` looks like it can mutate the data because it's using a direct reference instead of a copy to `self.entries`.
It also has hard requirements about how and where to look for data.

That's the whole point.

> Try dumping the output of `load_config()` and see how massive that table has gotten which leads to `Config:get()` which is either: a) useless because we don't need a function to access the the first level b) used in a very inefficient way because it needs to travers the entire aforementioned giant table over and over for every call.

That dump represents all the default entries set by lvim, this has always been this big.
a) Please suggest an alternative, using `config.entries` directly maybe ?
b) Wrong, please re-review the function

lua/core/gitsigns.lua
Tastyep: I double-checked, only the signs are different:
- our: https://github.com/LunarVim/LunarVim/blob/master/lua/core/gitsigns.lua#L8
- defaults: https://github.com/lewis6991/gitsigns.nvim/blob/main/lua/gitsigns/config.lua#L83

I think the defaults are better, I wonder if ours are old defaults or if we decided to have them.
Anyway, I'm obviously open to modifying them.

lua/main.lua
Tastyep: The problem comes from configuring key mappings at multiple places.
We can't register one before the other, as the latter will override the former.

I don't know yet how to fix that :shrug:

kylo252:
All of these notes are from the version I last reviewed. I can't currently run `lvim` no matter what I try. I need to stress that I'm not saying or claiming in any shape or form that this way of handling this massive config table is a result of this. I know that you've been working hard to mitigate that.

---

> I use a single instance because I only need one, why would you want to use many instances ?

Is this not the definition of a singleton?

> The 'ugliness' you describe, and its fixed order have to be somewhere, do you have a suggestion ?

We can start loading each config in a new loader instance and validate it separately. Instead of trying to keep track of the entire table. Doing that would allow us to see the dependency chain more clearly.

> I am open to constructive criticism, but please don't forget to provide suggestions to support your points.

I don't think that `load()` should internally know anything regarding the user-config. The user-config path is needed in other places as well, so handling it or storing any metadata about it should be done outside of this module. We can also have/keep another function that does the file loading from a specified path and returns a table. 
So `load()` should instead be designed to handle merging two tables in an agnostic way. Basically a wrapper for `vim.tb_deep_extend` with some a priori info about lvim, and not much more. 

Maybe you have spent enough time with this that this workflow still looks obvious enough, but I'll give some notes about how I see what's happening:

```lua
local function load_config()
  local defaults = require "config.defaults"
  local Config = require "config"
  local config = Config(defaults)
  --
  lvim = config.entries
  config:load() -- load what? that's why it should say config:load_or_merge(user-config)
  --
  settings.load_commands(config) -- maybe this is just a typo but it still looks odd
  -- 
  return config
end
local function main()
	local config = load_config()
	--
	vim.g.colors_name = config:get "colorscheme" -- why does it look like we can't get it with config.entries.colorscheme? 
	-- wouldn't it be better if it was `vim.g.colors_name = config:get(userconfig.colorscheme, _default)` 
	--
end
```

In my opinion, at the current state of the PR, I don't think it's clear enough how it's supposed to work or solve the most pressing issue (loading hierarchy). But maybe I'm just being dense.

Tastyep:
> > I use a single instance because I only need one, why would you want to use many instances ?
>
> Is this not the definition of a singleton?

My point was that the instance is not globally accessible, which is what a singleton would provide, but I see your point so let's not battle about semantics for now

> > The 'ugliness' you describe, and its fixed order have to be somewhere, do you have a suggestion ?
> 
> We can start loading each config in a new loader instance and validate it separately. Instead of trying to keep track of the entire table. Doing that would allow us to see the dependency chain more clearly.

I agree with you, I've been trying to delocalize content from the `lvim` global to builtin and lsp states, so this is more or less what I've introduced in the last commits.
There is still some effort to do, but I'm seeing the end of the tunnel.

> 
> > I am open to constructive criticism, but please don't forget to provide suggestions to support your points.
> 
> I don't think that `load()` should internally know anything regarding the user-config. The user-config path is needed in other places as well, so handling it or storing any metadata about it should be done outside of this module. We can also have/keep another function that does the file loading from a specified path and returns a table.

I agree with you, however, it's quite convenient seeing where the config path is used to attach it to the config object.
I'll try to see if I can move it somewhere else (without making it global obviously)

> So `load()` should instead be designed to handle merging two tables in an agnostic way. Basically a wrapper for `vim.tb_deep_extend` with some a priori info about lvim, and not much more.

That would mean removing `load()` and calling `merge()` instead, I'm fine with that too :+1: 

> 
> Maybe you have spent enough time with this that this workflow still looks obvious enough, but I'll give some notes about how I see what's happening:
> 
> ```lua
> local function load_config()
>   local defaults = require "config.defaults"
>   local Config = require "config"
>   local config = Config(defaults)
>   --
>   lvim = config.entries
>   config:load() -- load what? that's why it should say config:load_or_merge(user-config)

Maybe smth like that
``` lua
local user_config = loadfile(config_path)() 
config:merge(user_config)
```

>   --
>   settings.load_commands(config) -- maybe this is just a typo but it still looks odd

I know this needs reworking, but I'm trying to finish the big picture before because I'll very soon be limited in time.

>   -- 
>   return config
> end
> local function main()
> 	local config = load_config()
> 	--
> 	vim.g.colors_name = config:get "colorscheme" -- why does it look like we can't get it with config.entries.colorscheme? 
> 	-- wouldn't it be better if it was `vim.g.colors_name = config:get(userconfig.colorscheme, _default)` 
> 	--
> end
> ```

> In my opinion, at the current state of the PR, I don't think it's clear enough how it's supposed to work or solve the most pressing issue (loading hierarchy). But maybe I'm just being dense.

I think localizing the state by moving it away from the `lvim` variable tries to achieve that, but I could be wrong. Can you expand on that ?

kylo252:
> I agree with you, however, it's quite convenient seeing where the config path is used to attach it to the config object.
> I'll try to see if I can move it somewhere else (without making it global obviously)

The thing is that the user-config path is closely coupled to the `&rtp`. So these three paths [config|runtime|cache] should probably have their own module, maybe `bootstrap` similar to what I have in #1381. We can then provide some getter functions that can be used by all the plugins that need it. Alternatively it can live as a private entry in the lvim table, where only a util function is used to access it.

> Maybe smth like that
> 
> ```lua
> local user_config = loadfile(config_path)() 
> config:merge(user_config)
> ```

yeah exactly, it can even generate a new instance one the fly since it's probably a variadic function

```lua
local res = config:merge(defaults, user_config, ... )
```

> I think moving the state away from the `lvim` variable tries to achieve that, but I could be wrong. Can you expand on that ?

I'm not sure what `get()` is supposed to do. Do you have to specify the depth? because that would make it redundant. We can just call `config.entries.colorscheme`.
And if you don't specify the depth, then how will it be efficient? how many levels does it traverse before giving up? See how big those Telescope tables can get.

It's also not doing anything particularly complicated with the defaults, since you either get all the keys or the defaults. It might be that I'm missing the use-case here, but I feel like it should probably be tweaked or called something else.

Tastyep:
It's been a year, we can probably close this one.

PR #1426, opened 2021-08-30
use `<leader>gd` to see git diff with previous head

### LINKED ISSUE #1363: [Feature]: Mapping to view git diff of current file
### Feature motivation

This is something I do a lot and was surprised to find it wasn't in the built in mappings provided for git operations.

A mapping like \<Leader\>gd to open a diff view between the current file and the file at HEAD

Could just leave it up to each user to configure this, maybe I'm overestimating how common this need is.

### Additional context

_No response_

abzcoding:
# Description

Add a simple git diff with prev header

Fixes #1363

## How Has This Been Tested?

open a file and use `<leader>gd` to see diff with `HEAD~1`

### DIFF
--- a/lua/core/which-key.lua
+++ b/lua/core/which-key.lua
@@ -130,6 +130,10 @@ M.config = function()
           "<cmd>Telescope git_bcommits<cr>",
           "Checkout commit(for current file)",
         },
+        d = {
+          "<cmd>Gitsigns diffthis HEAD<cr>",
+          "Git Diff",
+        },
       },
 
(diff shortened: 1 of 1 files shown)


abzcoding:
@marcusbuffett  is this what you wanted?

marcusbuffett:
I was thinking `HEAD` instead of `HEAD~1`. So like changes since last commit (staged or otherwise).

marcusbuffett:
Thank you!

PR #1430, opened 2021-08-30
Update README.md

Moelf:
update Julia link


abzcoding:
thank you for your contribution, but i just updated all of those links

PR #1440, opened 2021-09-01
[Feature]: Add new which-key command for switching to the last buffer

genesis-algorithms:
# Description

Adds ability to switch back to the previous buffer (like alt+tab in windows). Which-key command is `<leader> -> b -> b`

https://user-images.githubusercontent.com/5767281/131635300-ea41deb7-3af2-4ccc-829c-57f62d913da6.mp4

## How Has This Been Tested?

No need for testing, it's just a new which-key command.

### DIFF
--- a/lua/core/which-key.lua
+++ b/lua/core/which-key.lua
@@ -72,26 +72,26 @@ M.config = function()
       ["h"] = { "<cmd>nohlsearch<CR>", "No Highlight" },
       b = {
         name = "Buffers",
-        j = { "<cmd>BufferPick<cr>", "jump to buffer" },
-        f = { "<cmd>Telescope buffers<cr>", "Find buffer" },
-        b = { ":b#<cr>", "go back to previous buffer" },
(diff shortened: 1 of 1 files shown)


Tastyep (approved):

chase:
@Tastyep This also changed the key bindings from `<Leader>bh` to `<Leader>bl` and `<Leader>bl` to `<Leader>br`

I think this change should be reverted as it is unrelated to the rest of the PR and the change is not idiomatic to the rest of the default keybindings.

Tastyep:
> @Tastyep This also changed the key bindings from `<Leader>bh` to `<Leader>bl` and `<Leader>bl` to `<Leader>br`
> 
> I think this change should be reverted as it is unrelated to the rest of the PR and the change is not idiomatic to the rest of the default keybindings.

I am not opposed to rolling the change back, but I'd like to understand your reasons first.
I matched the key with the corresponding action like for other key bindings, why do you think the original mappings are better ?

Tastyep:
@chase Do you validate https://github.com/LunarVim/LunarVim/pull/1458 ?

PR #1443, opened 2021-09-01
[Bug]: fix runtime path order

### LINKED ISSUE #1442: [Bug]: Runtimepath not set correctly, some user plugins are not loaded
### Problem description

After #1354, the order of the runtimepath is not set correctly anymore. The */after paths are expected to come at the end. But _~/.local/share/lunarvim/site_ instead comes after _~/.config/lvim/after_, which causes some users plugins to be not loaded correctly.

### LunarVim version

rolling

### Neovim version (>= 0.5)

v0.6.0

### Relevant log output

```shell
packpath=/etc/xdg/xdg-ubuntu/nvim,/etc/xdg/nvim,/usr/share/ubuntu/nvim/site,/usr/loc
al/share/nvim/site,/usr/share/nvim/site,/var/lib/snapd/desktop/nvim/site,/usr/local/sh
are/nvim/runtime,/usr/local/lib/nvim,/var/lib/snapd/desktop/nvim/site/after,/usr/share
/nvim/site/after,/usr/local/share/nvim/site/after,/usr/share/ubuntu/nvim/site/after,/e
tc/xdg/nvim/after,/etc/xdg/xdg-ubuntu/nvim/after,~/.local/share/lunarvim/lvim,~/.confi
g/lvim,~/.config/lvim/after,~/.local/share/lunarvim/site,~/.local/share/lunarvim/site/
after
```

### Screenshots

_No response_

### I have

- [X] Read the readme
- [X] Read the wiki

aaronsms:
# Description

Fixes #1442

## How Has This Been Tested?

- Run command `set rtp?` and `set packpath?`
- Results shows that */after paths indeed comes after paths without.
- Testing on a user plugin "vimtex" now loads correctly.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -3,15 +3,15 @@ local home_dir = vim.loop.os_homedir()
 
 vim.opt.rtp:append(home_dir .. "/.local/share/lunarvim/lvim")
 
-vim.opt.rtp:remove(home_dir .. "/.config/nvim")
-vim.opt.rtp:remove(home_dir .. "/.config/nvim/after")
-vim.opt.rtp:append(home_dir .. "/.config/lvim")
-vim.opt.rtp:append(home_dir .. "/.config/lvim/after")
-
 vim.opt.rtp:remove(home_dir .. "/.local/share/nvim/site")
 vim.opt.rtp:remove(home_dir .. "/.local/share/nvim/site/after")
-vim.opt.rtp:append(home_dir .. "/.local/share/lunarvim/site")
+vim.opt.rtp:prepend(home_dir .. "/.local/share/lunarvim/site")
 vim.opt.rtp:append(home_dir .. "/.local/share/lunarvim/site/after")
+
+vim.opt.rtp:remove(home_dir .. "/.config/nvim")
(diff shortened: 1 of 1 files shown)


kylo252:
So the problem is basically solved by using `prepend` instead of `append` for the "after" directory.

But why are you mixing up the removal, or the operations, order in general?

It's easier to maintain and reason about which directory replaces which when they are in order. e,g. fix the "config" ones first and then fix the "data" directories.

kylo252 (changes requested):
As I mentioned, it's easier to track if the "listing order" of the modifications is maintained as we have it right now. 
Readability takes precedence over removing a few extra of calls in this case for me.

aaronsms:
> So the problem is basically solved by using `prepend` instead of `append` for the "after" directory.
> 
> But why are you mixing up the removal, or the operations, order in general?
> 
> It's easier to maintain and reason about which directory replaces which when they are in order. e,g. fix the "config" ones first and then fix the "data" directories.
> 

ok fair enough, but the "data" directories changes would have to come before "config" ones to preserve the following order.

1. config 
2. data
3. data (after)
4. config (after)

kylo252 (approved):

PR #1444, opened 2021-09-01
refactor: use a static lvim binary template

kylo252:
# Description

Use a static `lvim` binary instead of generating it from the installer script.

## How Has This Been Tested?

- Should pass the CI
- You can also trigger it manually  in `install.sh`

```shell
# main "$@" comment this out
setup_shim
```

then run the installer again.


ahmedkhalf (dismissed):
Please call `setup_shim` in the update lvim function. So the shim gets updated too.

utils/installer/install.sh
```diff
@@ -250,15 +250,28 @@ function setup_shim() {
...
+  if [ -e "$dst" ]; then
+    if cmp --quiet "$src" "$dst"; then
+      rm "$dst"
+    fi
+  fi
+
+  cp "$src" "$dst"
+
+  # replace the default path with the new INSTALL_PREFIX
+  sed s"@LUNARVIM_RUNTIME_DIR=.\(.*\)@LUNARVIM_RUNTIME_DIR=\"$LUNARVIM_RUNTIME_DIR\"@"g \
```
Tastyep, line 267: As suggested in the other PR, I would still prefer that it is known that these variables are replaced (also that the file is to be modified by another one).

`export LUNARVIM_RUNTIME_DIR="${LUNARVIM_RUNTIME_DIR:-{ LUNARVIM_RUNTIME_DIR }}"`

This would also allow to write a simpler sed command.

Tastyep (dismissed):

utils/installer/install.sh
kylo252: That also means you can't copy paste the "default" binary, since it's a template now. I'm not sure if anyone expects it to be there. I don't really mind either way.

kylo252:
> Please call `setup_shim` in the update lvim function. So the shim gets updated too.

I think I understand the reason, but it's slightly tricky to achieve. 

a) we need to deduct the `$INSTALL_PREFIX`, probably with `which lvim`
b) what if the user installed it to a directory that requires admin privileges'? how do you overwrite that?

so can you try writing the full use-case so we can see how to manage it?

ahmedkhalf:
> a) we need to deduct the `$INSTALL_PREFIX`, probably with `which lvim`

I agree, and furthermore if `which` returns code other than 0 then we use the default destination.

> b) what if the user installed it to a directory that requires admin privileges'? how do you overwrite that?

Consider the following lines that are currently present in the install script:

```sh
local dst="$INSTALL_PREFIX/bin/lvim"
# remove outdated installation so that `cp` doesn't complain
rm -f "$dst"
```

What if the user set `$INSTALL_PREFIX` in a root directory? As you can see, the installation is already flawed, and already suffer's from the problem described in b.

A suggested fix is to parse the output of:
```
ls -l file
```

or just try to delete without then if it fails delete with sudo:
```
rm file||sudo rm file
```

kylo252:
> > b) what if the user installed it to a directory that requires admin privileges'? how do you overwrite that?
> 
> Consider the following lines that are currently present in the install script:
> 
> ```shell
> local dst="$INSTALL_PREFIX/bin/lvim"
> # remove outdated installation so that `cp` doesn't complain
> rm -f "$dst"
> ```
> 
> What if the user set `$INSTALL_PREFIX` to a root directory? As you can see, the installation is already flawed, and already suffers from the problem described in b.

I think you're forgetting about the context. You can require `sudo` to install software but not for running it.
```shell
sudo apt install cowsay
cowsay "I don't need sudo every time I run this"
```

> or just try to delete without sudo, then if it fails delete with sudo:
> 
> ```
> rm file||sudo rm file
> ```

That's a hammer and not a fix. I've specifically removed any `sudo` elevation from the installer because it's dangerous. 

_OBS: you should definitely avoid having `sudo rm <anything>` in any script you even run or write._

ahmedkhalf:
> I think you're forgetting about the context. You can require sudo to install software but not for running it.

Yea, I am aware, but I fail to understand how this statement relates to my remark.

> That's a hammer and not a fix. I've specifically removed any sudo elevation from the installer because it's dangerous.

You are correct, perhaps one could fix this by using `ln` instead? Like right now all we need is for a way to also update the lvim bin; so when using `ln`, this update would occur without having to call `setup_shim` again.

chase (approved):
Much easier to manage :+1:

PR #1446, opened 2021-09-01
[Feature]: Add restart functionality

ahmedkhalf:
# Description

The following pr adds restart functionality to lunarvim which is an alternative
way of reloading.

## How Has This Been Tested?

I launched nvim with the new lvim bin, then I restarted my neovim using the
function defined in the pr.


kylo252:
Hot-reloading has had a significant refactor recently. Please let me know if there's still something we're missing there.

PR #1447, opened 2021-09-01
feat: support wildcard filetypes for null-ls providers

### LINKED ISSUE #1196: Replace all `-` and ` ` with `_` for null-ls
This way we don't confuse the user.

Tastyep:
# Description

Some builtins from null-ls have '*' for file type, this is now supported on our side as well.
I also added some utility functions for tables, `any_of` and `find_if` and replaced parts of the code to use them.
Some key-mappings for compe were hardcoded, I moved them to the configuration and added support for `select_mode` key-mappings.

Convert - to _ when resolving builtins command to name.

Fixes #1196 

## How Has This Been Tested?

`:LvimInfo` will now show spelling linters.
I checked that tab and stab were still working in the completion menu.
Configure cmake-format and open a CMakeLists.txt file

### DIFF
--- a/lua/core/info.lua
+++ b/lua/core/info.lua
@@ -11,6 +11,7 @@ local M = {
 
 local fmt = string.format
 local text = require "interface.text"
+local lsp_utils = require "lsp.utils"
 
 local function str_list(list)
   return fmt("[ %s ]", table.concat(list, ", "))
@@ -67,7 +68,7 @@ local function tbl_set_highlight(terms, highlight_group)
 end
 
 local function make_client_info(client)
-  local client_enabled_caps = require("lsp.utils").get_ls_capabilities(client.id)
+  local client_enabled_caps = lsp_utils.get_client_capabilities(client.id)
   local name = client.name
   local id = client.id
   local document_formatting = client.resolved_capabilities.document_formatting
@@ -89,8 +90,7 @@ local function make_client_info(client)
 end
 
 function M.toggle_popup(ft)
-  local lsp_utils = require "lsp.utils"
-  local clients = lsp_utils.get_active_client_by_ft(ft)
+  local clients = lsp_utils.get_active_clients_by_ft(ft)
   local client_names = {}
 
   local header = {
--- a/lua/lsp/null-ls/services.lua
+++ b/lua/lsp/null-ls/services.lua
@@ -4,7 +4,7 @@ local function find_root_dir()
   local util = require "lspconfig/util"
   local lsp_utils = require "lsp.utils"
 
-  local ts_client = lsp_utils.get_active_client "typescript"
+  local ts_client = lsp_utils.is_client_active "typescript"
   if ts_client then
     return ts_client.config.root_dir
   end
--- a/lua/lsp/utils.lua
+++ b/lua/lsp/utils.lua
@@ -2,21 +2,26 @@ local M = {}
 
 local tbl = require "utils.table"
 
-function M.get_active_client(name)
+function M.is_client_active(name)
   local clients = vim.lsp.get_active_clients()
   return tbl.find_first(clients, function(client)
     return client.name == name
   end)
 end
 
-function M.get_active_client_by_ft(filetype)
+function M.get_active_clients_by_ft(filetype)
+  local matches = {}
   local clients = vim.lsp.get_active_clients()
-  return tbl.find_first(clients, function(client)
-    return client.name == lvim.lang[filetype].lsp.provider
-  end)
+  for _, client in pairs(clients) do
+    local supported_filetypes = client.config.filetypes or {}
+    if client.name ~= "null-ls" and vim.tbl_contains(supported_filetypes, filetype) then
+      table.insert(matches, client)
+    end
+  end
+  return matches
 end
 
-function M.get_ls_capabilities(client_id)
+function M.get_client_capabilities(client_id)
   if not client_id then
     local buf_clients = vim.lsp.buf_get_clients()
     for _, buf_client in ipairs(buf_clients) do


abzcoding:
i get this error while using it, not sure why 
Screen Shot 2021-09-01 at 11 39 08 PM

lua/core/compe.lua
```diff
@@ -50,10 +50,18 @@ M.config = function()
           ["<C-e>"] = { "compe#close('<C-e>')", { silent = true, noremap = true, expr = true } },
           ["<C-f>"] = { "compe#scroll({ 'delta': +4 })", { silent = true, noremap = true, expr = true } },
           ["<C-d>"] = { "compe#scroll({ 'delta': -4 })", { silent = true, noremap = true, expr = true } },
+
+          ["<Tab>"] = { "v:lua.tab_complete()", { expr = true } },
+          ["<S-Tab>"] = { "v:lua.s_tab_complete()", { expr = true } },
+        },
+        select_mode = {
+          ["<Tab>"] = { "v:lua.tab_complete()" },
```
abzcoding, line 58: maybe because this line and the one below it

abzcoding (approved):

kylo252 (changes requested):
You can use `vim.tbl_filter` for this. That way we can avoid defining what appears to be a new "type" of table 😅

Tastyep:
> You can use `vim.tbl_filter` for this. That way we can avoid defining what appears to be a new "type" of table sweat_smile

But I don't want a filtered table, I want to find the matching element

kylo252:
> > You can use `vim.tbl_filter` for this. That way we can avoid defining what appears to be a new "type" of table sweat_smile
> 
> But I don't want a filtered table, I want to find the matching element

How are these two not the same? 

There's also `vim.tbl_map` if you need something a little bit different.

Tastyep:
> > > You can use `vim.tbl_filter` for this. That way we can avoid defining what appears to be a new "type" of table sweat_smile
> > 
> > 
> > But I don't want a filtered table, I want to find the matching element
> 
> How are these two not the same?
> 
> There's also `vim.tbl_map` if you need something a little bit different.

One returns a table, the other one returns the entry. `tbl_map` doesn't do what I want either.

kylo252:
> One returns a table, the other one returns the entry. `tbl_map` doesn't do what I want either.

You can just take the first element of the filtered table and discard the rest if you want. what am I missing here?

I believe this should do it

```lua
function M.get_active_client_by_ft(filetype)
  if not lvim.lang[filetype] or not lvim.lang[filetype].lsp then
    return nil
  end

  local clients = vim.lsp.get_active_clients()
  return vim.tbl_filter(function(client)
    return client.name == lvim.lang[filetype].lsp.provider
  end, clients)
end
```

I even have a FIXME on that function because it's only picking up the first client :(

Tastyep:
> > One returns a table, the other one returns the entry. `tbl_map` doesn't do what I want either.
> 
> You can just take the first element of the filtered table and discard the rest if you want. what am I missing here?
> 
> I believe this should do it
> 
> ```lua
> function M.get_active_client_by_ft(filetype)
>   if not lvim.lang[filetype] or not lvim.lang[filetype].lsp then
>     return nil
>   end
> 
>   local clients = vim.lsp.get_active_clients()
>   return vim.tbl_filter(function(client)
>     return client.name == lvim.lang[filetype].lsp.provider
>   end, clients)
> end
> ```
> 
> I even have a FIXME on that function because it's only picking up the first client :(

Feel free to make a PR to support returning a list once you've merged your other PRs

kylo252:
> > I even have a FIXME on that function because it's only picking up the first client :(
> 
> Feel free to make a PR to support returning a list once you've merged your other PRs

As I said, you can just take first value and ignore the rest if you want one value. If you want to see if the predicate worked then you just see if there are elements in the return value. It really doesn't make sense to add a function with a slightly different name that does the same exact thing. 

> _In functional programming, filter is a higher-order function that processes a data structure (usually a list) in some order to produce a new data structure containing exactly those elements of the original data structure for which a given predicate returns the boolean value true._

You can create a lambda, call it `if_any` and pass it to `vim.tbl_filter`. Now you have the big advantage that people already know how the library function work, so when there's something broken. It's one less thing to think about. How often are we even going to use that?

Tastyep:
Sorry, but I don't agree with your suggestions.
``` lua
  local client = tbl.find_if(clients, function(client)
    return client.name == lvim.lang[filetype].lsp.provider
  end)
```

``` lua
  local client = vim.tbl_filter(function(client)
    return client.name == lvim.lang[filetype].lsp.provider
  end, clients)[1]
```

Note that additionally to better semantic, my function doesn't need to iterate over the whole array in case the element is found, neither does it need to create a new table.

> how the library function work

Is a  well named, documented, < 10lines function is too hard to understand, really ?
I come from a c++ background, so to me these functions are standard, but still.

> So when there's something broken

Luckily, we will be able to implement tests in a near future.

Really, I don't understand your reason for preferring accommodating tbl_filter to match your usage instead of having a well-defined function that does one job and is explicit about it.
It seems to me that you are against adding new code more than against this function, maybe I'm wrong.

kylo252:
I feel like you missed the most important point "_How often are we even going to use that?_"

> Note that additionally to better semantic, my function doesn't need to iterate over the whole array in case the element is found, neither does it need to create a new table.

How many filetypes are there? what's the longest possible list of filetypes where you think this optimization is gonna matter, or is this future-proofing?

> Luckily, we will be able to implement tests in a near future.

Testing and rationalizing about a specialized function is easier every day of the week. And we don't need to test library calls.

> Is a well named, documented, < 10lines function too hard to understand, really ?
> I come from a c++ background, so to me these functions are standard, but still I would have expected that to be easy to understand, I even matched the variable names `Table` -> `tbl`.

It's about avoiding unnecessary complexity. I've learned to always try to use well validated and tested library calls unless I have a very good reason not to. And so far I see zero evidence how these offer any benefits. And again, one of them is literally used in a buggy function anyway.

> Really, I don't understand your reason for preferring accommodating tbl_filter to match your usage instead of having a well-defined function that does one job and is explicit about it.
> It seems to me that you are against adding new code more than against this function, maybe I'm wrong.

I think it's fair to say we're not writing a library.


Since I already gave an example for `get_active_clients_by_ft()`, let me suggest this for the other use-case: 


```lua
  function M.list_available(filetype)
    local linters = {}
    local tbl = require "utils.table"
    for _, provider in pairs(null_ls.builtins.diagnostics) do
      if vim.tbl_contains(provider.filetypes or {}, ft) then
        table.insert(linters, provider.name)
      end
    end
    return linters
  end
```

I'm not really sure you can give any argument against it compared to what you have here:

```lua
function M.list_available(filetype)
  local linters = {}
  local tbl = require "utils.table"
  for _, provider in pairs(null_ls.builtins.diagnostics) do
    if tbl.any_of(provider.filetypes or {}, function(ft)
      return ft == "*" or ft == filetype
    end) then
      table.insert(linters, provider.name)
    end
  end
  return linters
end

```

Tastyep:
>I'm not really sure you can give any argument against it compared to what you have here

Only that it doesn't support wildcards, which was the goal of the PR.
I'm not answering anymore, feel free to merge or delete the PR.

kylo252:
Let's leave the tie-breaker to @rebuilt 😃

rebuilt:
As the dumbest person in the room, I'm glad you left it to me to decide.  My abilities better match those of the average person reading this code.   😅 

@Tasty I think the functions you wrote need better names so anyone reading the code can better infer the arguments and return values.  The functions you wrote are short and well written but they will never be as easy to understand as functions coders already know.  

When I read `local client = vim.tbl_filter`, I instantly know what arguments this function is supposed to take and what will likely be returned.  I do not have to read the source code to understand what it does. 

When I read 'local client = tbl.find_if`, I can make NO assumptions about the arguments the function takes or what it returns.  I am forced to go to the definition and read what it does.  

`tbl.find_if` acts a lot like `tbl.find_first`.  I'm not sure if this name is clearer or not.  Can we come up with a better name?  I'd rather have a longer name that describes the intent than a short but vague name.

I wasn't quite sure what `any_of` should return.  Once I read the source I mentally changed the name to `contains_any_of` which made the return of boolean seem more obvious.

abzcoding:
btw in the new workflow, we don't need this, we can just use the normal null-ls
for example instead of setting prettierd 10 times, we can just use
Screen Shot 2021-10-09 at 12 23 46 PM

kylo252:
> btw in the new workflow, we don't need this

Yes and no, this is solving a few different issues at once. I've been able to verify at least one of them, this actually works now:

```lua
lvim.lang.markdown.linters = { { exe = "write-good" } }
```
The other issue it solves is providers that will return `{'*'}` as their filetype, but this PR is done before I had refactored null-ls. I'm not sure this will break anything because we've since been using a null-ls utility:
https://github.com/LunarVim/LunarVim/blob/5fef931d8720f2ef0d7a0fca18b96f31fd36689a/lua/lsp/null-ls/services.lua#L53-L55

> we can just use the normal null-ls for example instead of setting prettierd 10 times

I can already do this with our API. Just need to add like 2 lines of code, so I'm gonna fix it quickly.

PR #1457, opened 2021-09-03
Add exclude_dirs as a configuration option

reisnobre:
Add exclude_dirs to lvim.builtin.project to allow the user to exclude folders from the project list

# Description

A new configuration option was provided by the developer of the package project.nvim through the commit https://github.com/ahmedkhalf/project.nvim/commit/21986b90407064d3f20daff3ec241bb8790c8565
and this PR adds this configuration option to the project.

## How Has This Been Tested?

Tested locally with the config `lvim.builtin.project.exclude_dirs = { "~/Archive/*" }` and projects that before where being listed disappeared. 
Removing the configuration above, reintroduced projects under the folder as projects again.

### DIFF
--- a/lua/core/project.lua
+++ b/lua/core/project.lua
@@ -20,6 +20,10 @@ function M.config()
     --- can also delete or rearangne the detection methods.
     detection_methods = { "lsp", "pattern" },
 
+    ---@usage Methods of detecting patterns for excluding from the project list
+    --- Ex: { "~/.cargo/*", ... }
+    exclude_dirs = {},
+
     ---@usage patterns used to detect root dir, when **"pattern"** is in detection_methods
     patterns = { ".git", "_darcs", ".hg", ".bzr", ".svn", "Makefile", "package.json" },
 


Tastyep:
We should actually avoid duplicating these config entries, not having them here don't prevent you from having them in your configuration.
I actually removed all of them [here](https://github.com/LunarVim/LunarVim/pull/1406/files#diff-79d96c5df9ecc67e16f86c94130ceace7410de214e7cb995b4c966c31a337fb2)

ahmedkhalf:
I agree, however, I have not tried your pr yet, but I am curious as to whether the autocomplete works when the user types the following in his config:

```
lvim.builtin.project.<c-space>
```

reisnobre:
Yes it does 😄 .

This pr is not **required** because the functionality works event without those lines, but the autocomplete tooltip is good enoght to justify the addition.
Screen Shot 2021-09-03 at 18 18 27

Tastyep:
Closing as the goal is for us not to maintain duplicated configurations for all plugins

PR #1458, opened 2021-09-03
[Bugfix] Revert back buffer bindings

Tastyep:


chase (commented):
This fixes the behavior.

They also changed the labels from the original config, but that's nitpicking at this point.
I think that `Close all to the left` and `Close all to the right` may be more clear.

chase (approved):

chase:
Looks good @Tastyep

PR #1463, opened 2021-09-04
Fix: Install Nodejs

ZepbachVNTM:
[Bugfix]: Fix install NodeJs

### DIFF
--- a/utils/installer/install.sh
+++ b/utils/installer/install.sh
@@ -164,7 +164,7 @@ function install_nodejs_deps() {
   for dep in "${__npm_deps[@]}"; do
     if ! npm ls -g "$dep" &>/dev/null; then
       printf "installing %s .." "$dep"
-      npm install -g "$dep"
+      sudo npm install -g "$dep"
     fi
   done
   echo "All NodeJS dependencies are succesfully installed"


kylo252:
Hey, any reason why your `npm install` is not running on a user directory?

ChristianChiarulli (approved):

Uzaaft:
Most likely, (just a hypothesis), author has not set up npm to use a folder in /home/$USER, which in turn requires sudo to install global packages. I had the same issue while setting up arch the first time.

Uzaaft:
A couple of reasons why not to use sudo can be found here(behind paywall, I think): https://medium.com/@ExplosionPills/dont-use-sudo-with-npm-still-66e609f5f92

abzcoding (changes requested):
we should not use sudo, we can warn users to install those separately or fix their nodejs

Uzaaft:
@abzcoding You can close the PR. It does not fix anything.

kylo252:
> Most likely, (just a hypothesis), author has not set up npm to use a folder in /home/$USER, which in turn requires sudo to install global packages. I had the same issue while setting up arch the first time.

What I wanted to know is if there there's a reason why he can't change it, since I figured it's the default way on his system probably. I should have framed my question better 😄.

And I 100% agree with the rest of the your comment, thank you for the useful links, much appreciated!

kylo252:
Closed due to lack of a response from the author. Please let me know if you still need help with it later on.

Issue #1422, opened 2021-08-29
[Feature]: Call lvim.builtin.which_key.on_config_done after all core logic plugins have been loaded


nahojer:
### Feature motivation

You have this very nice `on_config_done` function when customizing which-key. However, it serves no purpose when trying to configure your own LSP mappings. (Also, I want to remove the ["\<leader\>gg"] mapping because I don't use Lazygit and I find that it bloats my which-key "popup.") Can I suggest moving below code from the setup function of the which-key config (`lua/core/which-key.lua`) to `lua/core/builtins/init.lua`?

```lua
if lvim.builtin.which_key.on_config_done then
    lvim.builtin.which_key.on_config_done(which_key)
end
```

`lua/core/builtins/init.lua` could then look something like this:

```lua
local M = {}

local builtins = {
  "keymappings",
  "core.which-key",
  "core.gitsigns",
  "core.compe",
  "core.dashboard",
  "core.dap",
  "core.terminal",
  "core.telescope",
  "core.treesitter",
  "core.nvimtree",
  "core.project",
  "core.bufferline",
  "core.autopairs",
  "core.comment",
  "core.lspinstall",
  "core.lualine",
}

function M.config(config)
  for _, builtin_path in ipairs(builtins) do
    local builtin = require(builtin_path)
    builtin.config(config)
  end

  if lvim.builtin.which_key.on_config_done then
      lvim.builtin.which_key.on_config_done(which_key)
  end
end

return M
```

What do you think? Am I missing something that prevents us from doing this?

### Describe the expected solution

Want to be able to override all of the keymappings.

nahojer:
On top of this, maybe it's also possible to switch the order in how you call `lvim.lsp.on_attach_callback`? So instead of

```go
function M.common_on_attach(client, bufnr)
  if lvim.lsp.on_attach_callback then
    lvim.lsp.on_attach_callback(client, bufnr)
    Log:debug "Called lsp.on_init_callback"
  end
  lsp_highlight_document(client)
  add_lsp_buffer_keybindings(bufnr)
  require("lsp.null-ls").setup(vim.bo.filetype)
end
```

One could do

```go
function M.common_on_attach(client, bufnr)
  lsp_highlight_document(client)
  add_lsp_buffer_keybindings(bufnr)
  require("lsp.null-ls").setup(vim.bo.filetype)
  if lvim.lsp.on_attach_callback then
    lvim.lsp.on_attach_callback(client, bufnr)
[...]

PR #1472, opened 2021-09-04
[Refactor] Remap whichkey Explorer NvimTreeToggle

jorgeOmurillo:
# Description

Fixes #(issue)
- Remap whichkey Explorer to lua command

## How Has This Been Tested?

- Run command `:mycommand`
- Check logs
- ...

### DIFF
--- a/lua/core/nvimtree.lua
+++ b/lua/core/nvimtree.lua
@@ -80,8 +80,6 @@ function M.setup()
     }
   end
 
-  lvim.builtin.which_key.mappings["e"] = { "<cmd>NvimTreeToggle<CR>", "Explorer" }
-
   local tree_view = require "nvim-tree.view"
 
   -- Add nvim_tree open callback
--- a/lua/core/which-key.lua
+++ b/lua/core/which-key.lua
@@ -70,6 +70,7 @@ M.config = function()
       ["c"] = { "<cmd>BufferClose!<CR>", "Close Buffer" },
       ["f"] = { "<cmd>Telescope find_files<CR>", "Find File" },
       ["h"] = { "<cmd>nohlsearch<CR>", "No Highlight" },
+      ["e"] = { "<cmd>NvimTreeToggle<CR>", "Explorer" },
       b = {
         name = "Buffers",
         j = { "<cmd>BufferPick<cr>", "Jump" },


abzcoding:
there is only a single issue, what happens if someone has disabled nvim-tree
do we still want to show that in which-key?

i'm fine with it though

rebuilt:
I'm fine with it as is.  It shouldn't take setting an nvimtree callback to remove this mapping.

In the past we've done something like this to set conditional mappings.

```lua
if(lvim.builtin.nvimtree.active)
  lvim.builtin.which_key.mappings["e"] = { "<cmd>NvimTreeToggle<CR>", "Explorer" }
end
```

jorgeOmurillo:
@abzcoding @rebuilt I can push that condition if so. What do you guys think?

Tastyep:
Related: https://github.com/LunarVim/LunarVim/pull/1406#discussion_r702261627

I wouldn't merge this as it is, because we have other files where which-key mappings are defined in builtins' config.

To be frank, I don't know how this works in the first place, we edit `lvim.builtin.which_key.mappings` which is used in which-key `setup` (soon `config`) function. How does that work, is which-key supposed to be configured last, how does it work if it is configured first ?

nahojer:
@Tastyep this is related to https://github.com/LunarVim/LunarVim/issues/1422 which I posted a few days ago. Fixing this issue would make me fall in love with LunarVim so much more. This is the only thing that bugs me with LunarVim at the moment.

What about providing a callback function for each builtin plugin that is called after after all plugins have been loaded? Similar to, or instead of, the existing callback function `on_config_done = nil` that is invoked as the last thing in `M.setup`, maybe one could have a callback function `on_config_loaded` that is called after all other plugins have been loaded? For instance, in core/builtins/init.lua, one could have:

```lua
function M.config(config)
  for _, builtin_path in ipairs(builtins) do
    local builtin = require(builtin_path)
    builtin.config(config)
  end
  for _, builtin_path in ipairs(builtins) do
    local plugin_conf = plugin configuration from some path trickery?
    local plugin = require from some path trickery?
    lvim[plugin_conf].on_config_loaded(plugin)
  end
end
```

Tastyep:
> What about providing a callback function for each builtin plugin that is called after after all plugins have been loaded? Similar to, or instead of, the existing callback function `on_config_done = nil` that is invoked as the last thing in `M.setup`.

I think it is also something that we need on our side, we need to make sure dependencies are loaded before executing some logic, so I'm not against implementing this kind of mechanism.
Just give it some time as I'm already making big changes on #1406

nahojer:
Absolutely, thanks!

rebuilt:
I don't think the leader key table really belongs to whichkey.  It's used by other plugins and the end user.  

Maybe it belongs in `lvim.keys.leader_keys` ?

Tastyep:
Some updates on this, that's how I think it should be done

https://github.com/LunarVim/LunarVim/pull/1406/commits/257add693e17d9ff738590c5eea8ee953f9a0f3d#diff-1db1be37fe64f29eeec18cb160ed468f80f648bbf4d9f6d80441975f3326b0a1R103-R104

My reasons are that doing it this way, we don't leak information into the `which-key` builtin and we only register the mapping if the builtin is active.

abzcoding:
tyvm for your effort, this has been fixed in another PR, 
so now you can change it to whatever you want in your config.lua
```lua
lvim.builtin.which_key.mappings["e"] = { "<cmd>RnvimrToggle<CR>", "Ranger" }
```

PR #1503, opened 2021-09-08
[Feature]: Add better text objects in the keymaps

sethigeet:
# Description

Now people have more text objects to use in visual and operator pending mode!
- You can now select inside many more common characters(*, #, :, ;, etc) using `vi<char>` or  `i<char>` after any operator(d, y, etc.)
- The same text objects also work with 'around' selection by using `a` instead of `i` ie. `va<char>` or `a<char>` after any operator(d, y, etc.)
- Select all the text in the buffer using `vae` or `ae` after any operator(d, y, etc.)

### DIFF
--- a/lua/keymappings.lua
+++ b/lua/keymappings.lua
@@ -1,14 +1,14 @@
 local M = {}
 local Log = require "core.log"
 
-local generic_opts_any = { noremap = true, silent = true }
-
-local generic_opts = {
-  insert_mode = generic_opts_any,
-  normal_mode = generic_opts_any,
-  visual_mode = generic_opts_any,
-  visual_block_mode = generic_opts_any,
-  command_mode = generic_opts_any,
+local generic_opts = { noremap = true, silent = true }
+local opts = {
+  insert_mode = generic_opts,
+  normal_mode = generic_opts,
+  visual_mode = generic_opts,
+  visual_block_mode = generic_opts,
+  command_mode = generic_opts,
+  operator_mode = generic_opts,
   term_mode = { silent = true },
 }
 
@@ -16,9 +16,10 @@ local mode_adapters = {
   insert_mode = "i",
   normal_mode = "n",
   term_mode = "t",
-  visual_mode = "v",
-  visual_block_mode = "x",
+  visual_and_select_mode = "v",
+  visual_mode = "x",
   command_mode = "c",
+  operator_mode = "o",
 }
 
 -- Append key mappings to lunarvim's defaults for a given mode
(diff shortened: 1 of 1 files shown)


lua/keymappings.lua
```diff
@@ -1,24 +1,25 @@
...
   term_mode = { silent = true },
 }
 
 local mode_adapters = {
   insert_mode = "i",
   normal_mode = "n",
   term_mode = "t",
-  visual_mode = "v",
-  visual_block_mode = "x",
+  visual_and_select_mode = "v",
```
Tastyep, line 19: Why renaming to visual_and_select_mode ? I would prefer that we keep the came terminology as vim.
Also, select mode is a different mode, see `:help vim-modes`

lua/keymappings.lua
```diff
@@ -152,8 +153,22 @@ function M.config()
       ["<C-j>"] = { 'pumvisible() ? "\\<C-n>" : "\\<C-j>"', { expr = true, noremap = true } },
       ["<C-k>"] = { 'pumvisible() ? "\\<C-p>" : "\\<C-k>"', { expr = true, noremap = true } },
     },
+
+    operator_mode = {},
   }
 
+  -- Add some text objects
```
Tastyep, line 160: Apart from being horribly difficult to read, what does this achieve ?
(Note that I didn't understand the PR's description)

Tastyep (changes requested):

lua/keymappings.lua
sethigeet: From the vim docs:

lua/keymappings.lua
sethigeet: Just how we can select inside parentheses(`(, ), {, }, [, ]`), paragraphs(`p`), etc., with the code from this PR, we can now also select inside some special characters such as *, #, etc.

lua/keymappings.lua
Tastyep: Alright, I think 's' is missing, do you mind adding it ?
I would also prefer `visual_and_select_mode` -> `visual_select_mode`

kylo252 (changes requested):
This seems very interesting, thank you!

There are some issues, and they are all connected to the interaction with which-key.

- This is showing inside which-key for some reason, which is usually fine on its own, but it's only showing these mapping and not all the other 'inside' delimiters. So this will be pretty misleading for a user that doesn't know the defaults.
- This currently only works for `v`, but not the other motions. After some digging it turns out that I needed to comment-out this line:
https://github.com/LunarVim/LunarVim/blob/390f277b702fe7877f0d5fabbece230528b58cb1/lua/core/which-key.lua#L15

I'm not sure if either of the above are bugs in which-key, but that's what we have to deal with, so we need to figure it out.

kylo252:
Let's leave this to the user to configure manually, as this is still not without issues.

PR #1506, opened 2021-09-08
[Refactor]: Improve the plugin loader

sethigeet:
# Description

Now we no longer will have to keep writing 
```lua
{
  "autor/plugin-name",
  config = function()
    require("core.plugin-config-file-name").setup()
  end
}
```
and can instead just write
```lua
{
  "autor/plugin-name",
  config = "core.plugin-config-file-name"
}
```

Also, I know that we can also avoid writing 'core.' but I kept that so that we can distinguish between the core plugin config and the custom user configs so that when the user puts in a config name and it does not contain 'core.' we would not apply any special logic for it!


Tastyep (commented):
I am against this change as it couples lvim's implementation with the fact that packer uses `strings.dump`. This is an implementation detail of packer which, IMO, should not be relied on.

kylo252 (commented):
> I am against this change as it couples lvim's implementation with the fact that packer uses `strings.dump`. This is an implementation detail of packer which, IMO, should not be relied on.

I guess he can add a wrapper for that in plugin-loader? then it's gonna be something like:
```lua
{
  "autor/plugin-name",
  config = execute("core.plugin-config-file-name")
}
```

That still looks a bit odd to me though. I get it if it's just loading a file, but if it's gonna be hiding the fact that it's a function, then that sounds like a recipe for headaches.

Tastyep:
> > I am against this change as it couples lvim's implementation with the fact that packer uses `strings.dump`. This is an implementation detail of packer which, IMO, should not be relied on.
> 
> I guess he can add a wrapper for that in plugin-loader? then it's gonna be something like:
> 
> ```lua
> {
>   "autor/plugin-name",
>   config = execute("core.plugin-config-file-name")
> }
> ```

That's unfortunately not possible because the parameter passed to `execute` will need to be captured in the returned lambda, which is a limitation of `strings.dump`.

> That still looks a bit odd to me though. I get it if it's just loading a file, but if it's gonna be hiding the fact that it's a function, then that sounds like a recipe for headaches.

I agree with you, I also dislike that it only works for the config field and that it's not done for the config one.

kylo252:
Superseded by #1163

PR #1515, opened 2021-09-09
[Feature]: Add lunarvim binary for windows 10

maxxnino:
# Description

This add start-up script for lunarvim on windowns 10
Still need update some doc how to use, manually install language server, and [lvim config](https://github.com/maxxnino/lvim)

## How Has This Been Tested?

Been using it on my local computer for months without any proplem


abzcoding:
cc: @MarcSchaetz

utils/bin/lvim.ps1
```diff
@@ -0,0 +1,8 @@
+$env:HOME = "$HOME"
+$env:USER = "$HOME"
+$env:SHELL = "cmd"
```
kylo252, line 3: I'm not sure this one is a good idea since PowerShell exists.

utils/bin/lvim.ps1
```diff
@@ -0,0 +1,8 @@
+$env:HOME = "$HOME"
+$env:USER = "$HOME"
```
kylo252, line 2: This will end up with a path that looks like this `$HOME/$HOME/...`.

utils/bin/lvim.ps1
```diff
@@ -0,0 +1,8 @@
...
+$env:SHELL = "cmd"
+$env:TERMINAL = "$env:SHELL"
+$env:VIMRUNTIME = "$(scoop prefix neovim-nightly)/share/nvim/runtime";
```
kylo252, line 5: I'm not sure everyone is installing with scoop necessarily, so it should not be a default value.

utils/bin/lvim.ps1
```diff
@@ -0,0 +1,8 @@
...
+Set-Variable -name "LUNARVIM_RUNTIME_DIR" -Value "$HOME/.local/share/lunarvim"
+Set-Variable -name "LUNARVIM_CONFIG_DIR" -Value "$HOME/.config/lvim"
+nvim -u "$LUNARVIM_RUNTIME_DIR/lvim/init.lua" $args
```
kylo252, line 8: I suggest using some error handling since PowerShell allows it.

utils/bin/lvim.ps1
maxxnino: The $HOME and $env:HOME are two different thing. Neovim can't find $HOME if $env:HOME not set

utils/bin/lvim.ps1
maxxnino: Powershell can't open with floatterm. Only cmd do

utils/bin/lvim.ps1
maxxnino: It's just a template for the reference

utils/bin/lvim.ps1
maxxnino: That would be nice. But sadly, I know nothing about error handling in powershell

rallyrabbit:
I may try to convert this to a batch file.  The problem is, many organizations lock down the ability to run powershell scripts while granting administrator or similar access to install things.  There has to be a better way and make all this independent of needing a lot of tools just to constantly compile rolling updates of a nice IDE.

kylo252:
Now available in https://github.com/LunarVim/LunarVim/blob/3c5a6d06b966eedac74a01d0fcc3a6a06631128e/utils/bin/lvim.ps1

PR #1523, opened 2021-09-10
[Feature] Allow cmp confirm options to be configured and jump after confirm if available

chase:
# Description

This allows users to change the options of the complete action' mapped to `<CR>` for cmp.
For example, I prefer it not to `select`, so I would add `lvim.builtin.cmp.confirm_opts.select = false` to my `config.lua`

This also make it so that if there is a snippet currently in place, it completes the current completion and then jumps instead of falling back to inserting a newline.

## How Has This Been Tested?
- Add `lvim.builtin.cmp.confirm_opts.select = false` to `config.lua`, enter a snippet and press `<CR>`

### DIFF
--- a/lua/core/cmp.lua
+++ b/lua/core/cmp.lua
@@ -30,6 +30,10 @@ M.config = function()
     return
   end
   lvim.builtin.cmp = {
+    confirm_opts = {
+      behavior = cmp.ConfirmBehavior.Replace,
+      select = true,
+    },
     formatting = {
       format = function(entry, vim_item)
         local icons = require("lsp.kind").icons
@@ -107,10 +111,15 @@ M.config = function()
 
       ["<C-Space>"] = cmp.mapping.complete(),
       ["<C-e>"] = cmp.mapping.close(),
-      ["<CR>"] = cmp.mapping.confirm {
-        behavior = cmp.ConfirmBehavior.Replace,
-        select = true,
-      },
+      ["<CR>"] = cmp.mapping(function(fallback)
+        if not require("cmp").confirm(lvim.builtin.cmp.confirm_opts) then
+          if luasnip.jumpable() then
+            vim.fn.feedkeys(T "<Plug>luasnip-jump-next", "")
+          else
+            fallback()
+          end
+        end
+      end),
     },
   }
 end


abzcoding:
can't you just use this ?
```lua
lvim.builtin.cmp.mapping["<CR>"] = require("cmp").mapping(function(fallback)
  if not require("cmp").confirm(lvim.builtin.cmp.confirm_opts) then
    if require("luasnip").jumpable() then
      vim.fn.feedkeys(vim.api.nvim_replace_termcodes("<Plug>luasnip-jump-next", true, true, true))
    else
      fallback()
    end
  end
end)
```
instead of passing `confirm_opts` to the cmp setup function?

abzcoding:
what i meant was that we are sending `lvim.builtin.cmp` to nvim_cmp setup
```lua
require("cmp").setup(lvim.builtin.cmp)
```
and you are sending `confirm_opts` to that setup as well, which i'm not sure is a good idea

abzcoding:
right now these are the accepted values in the cmp setup
Screen Shot 2021-09-16 at 12 47 07 PM

chase:
Oh, I see what you mean now, my apologies.
It wouldn't work if it wasn't set and doing it that way will still end up with `confirm_opts` getting passed to setup if it is in the user config.

If you really think it's a problem, could just do:
```lua
M.setup = function()
  local confirm_opts = vim.builtin.cmp.confirm_opts
  require("luasnip/loaders/from_vscode").lazy_load()
  vim.builtin.cmp.confirm_opts = nil
  require("cmp").setup(lvim.builtin.cmp)
  vim.builtin.cmp.confirm_opts = confirm_opts
end
```

abzcoding:
> Oh, I see what you mean now, my apologies.
> It wouldn't work if it wasn't set and doing it that way will still end up with `confirm_opts` getting passed to setup if it is in the user config.
> 
> If you really think it's a problem, could just do:
> 
> ```lua
> M.setup = function()
>   local confirm_opts = vim.builtin.cmp.confirm_opts
>   require("luasnip/loaders/from_vscode").lazy_load()
>   vim.builtin.cmp.confirm_opts = nil
>   require("cmp").setup(lvim.builtin.cmp)
>   vim.builtin.cmp.confirm_opts = confirm_opts
> end
> ```

😅 but that way you are still passing `confirm_opts = nil` and also ` vim.builtin.cmp.confirm_opts = confirm_opts` doesn't do anything
i'm not sure man, i'll let others decide on it
cc: @ChristianChiarulli  @kylo252

chase:
> sweat_smile but that way you are still passing `confirm_opts = nil`
Do a `:lua print(vim.inspect({x=nil}))`, note the table is empty.
Setting a value in a table to nil is the accepted way of deleting a key from a table in Lua.

> doesn't do anything bcz it gets called after the setup has been finished already
Incorrect.
It does do something: it makes it available for the function to be called when pressing CR and makes it available for autocomplete in `config.lua`

abzcoding:
> > sweat_smile but that way you are still passing `confirm_opts = nil`
> > Do a `:lua print(vim.inspect({x=nil}))`, note the table is empty.
> 
> Setting a value in a table to nil is the accepted way of deleting a key from a table in Lua.
> 
> > doesn't do anything bcz it gets called after the setup has been finished already
> 
> Incorrect.
> It does do something: it makes it available for the function to be called when pressing CR and makes it available for autocomplete in `config.lua`
> Tables are passed by reference, so the value being set does in fact affect:
> 
> ```lua
> if not require("cmp").confirm(lvim.builtin.cmp.confirm_opts) then
> ```

can you test the code you wrote in your config plz, 
setup is called after config (in this instance)
Screen Shot 2021-09-16 at 1 17 17 PM

chase:
> setup is called after config (in this instance)

It isn't, I wrote that run method, and just like the comment says: that is only run once on the initial Packer install.

Config is always run before setup:
https://github.com/LunarVim/LunarVim/blob/rolling/lua/bootstrap.lua#L69-L77

abzcoding:
ngl i'm a bit confused here
`setup is called after setup`
`config is run before setup`
Screen Shot 2021-09-16 at 1 23 51 PM

abzcoding:
anyway test it locally and update the PR, and let @kylo252  and @ChristianChiarulli decide on it,
i'm not either for or against it,

chase:
> ngl i'm a bit confused here
> `setup is called after config`
> `config is run before setup`
> Screen Shot 2021-09-16 at 1 23 51 PM

I think you're mixing up terminology and how the code actually loads.

`config()` in https://github.com/LunarVim/LunarVim/blob/rolling/lua/core/cmp.lua#L23 is not the same as `config` in https://github.com/LunarVim/LunarVim/blob/rolling/lua/plugins.lua#L30-L32

This is why the `run` function here is necessary: https://github.com/LunarVim/LunarVim/blob/rolling/lua/plugins.lua#L41-L46

---

Breaking it down:

https://github.com/LunarVim/LunarVim/blob/rolling/lua/bootstrap.lua#L69-L72
eventually runs:
https://github.com/LunarVim/LunarVim/blob/rolling/lua/core/builtins/init.lua#L1
calling:
`config()` in https://github.com/LunarVim/LunarVim/blob/rolling/lua/core/cmp.lua#L23 

https://github.com/LunarVim/LunarVim/blob/rolling/lua/bootstrap.lua#L74-L77
eventually runs:
https://github.com/LunarVim/LunarVim/blob/rolling/lua/plugin-loader.lua#L53-L61
running Packer.use(...) on:
https://github.com/LunarVim/LunarVim/blob/rolling/lua/plugins.lua#L28-L47
which runs the _plugin config's setup_ on load:
https://github.com/LunarVim/LunarVim/blob/rolling/lua/plugins.lua#L30-L32

A plugin's config method is not the same, that is handled by Packer.

Packer calls `run = function()...` only on install or update.

abzcoding:
again , just update the PR, test your code and let other devs decide on it 😅

chase:
> again , just update the PR, test your code and let other devs decide on it sweat_smile

Will do when I'm back at my computer :+1: 
Thought you'd appreciate having a better understanding of the code base.

abzcoding:
> > again , just update the PR, test your code and let other devs decide on it sweat_smile
> 
> Will do when I'm back at my computer 👍
> Thought you'd appreciate having a better understanding of the code base.

haha, good one

abzcoding:
at first that was my issue, which you tried to fix using that extra code in setup
this is my current issue with it, 
right now you can customize any of those mappings, but you need to write n lines of code in your config.lua
instead you are trying to add n+x lines of code for it in lv core so that users just need to add a single line of code
so what happens if someone wanted to change the behavior of `TAB` mapping?
do we add another option and make it nil and add x lines of code to cmp.lua so that users can write another oneline ?
or do we just let users edit everything they want the way it is right now?

chase:
> this is my current issue with it,
> right now you can customize any of those mappings, but you need to write n lines of code in your config.lua
> instead you are trying to add n+x lines of code for it in lv core so that users just need to add a single line of code in their config
> so what happens if someone wants to change the behavior of `TAB` mapping?
> do we add another option and make it nil and add x lines of code to cmp.lua so that users can write another oneliner ?
> or do we just let users edit everything they want the way it is right now?

It solves a [problem that occurs by default #1563](https://github.com/LunarVim/LunarVim/issues/1563), while still maintaining the default behavior of inserting non-selected items. To opt out of the problem, it is a one line config change.

This was also what I meant to solve in a way that was backwards compatible.

We could add to the website those 10 lines to make it so that the problem doesn't occur, if you think that's better?

abzcoding:
that is why i said i'm neither for or against it ( and the reason why i mentioned this PR in that issue)
i can live without it, but i'll understand if we want to merge it , so that users who probably would not bother looking at our documentation even if we provide those 10 lines can avoid the issue

kylo252 (approved):
Looks good to me.

I just wish that `cmp` had a little bit nicer looking API than this 😄

PR #1527, opened 2021-09-11
refactor: add 32bit support for impatient

kylo252:
# Description

- Add support for running impatient on 32bit systems
- Add a setup function for impatient with basic options

## How Has This Been Tested?

Runs fine on a RPi 4


chase:
I've made #1529, since there was a bug in this PR I wanted to fix and some simplification :shrug:

PR #1531, opened 2021-09-12
refactor(ci): better path handling

kylo252:
# Description

- Make Packer installation path more robust
- Remove the need to use hombrew for sha256sum

## How Has This Been Tested?

Should pass the CI tests

### DIFF
--- a/.github/workflows/install.yaml
+++ b/.github/workflows/install.yaml
@@ -23,16 +23,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      # sha256sum is not available by default
-      - name: Installl dependencies for OSX
-        if: matrix.os == 'osx'
-        run: |
-          echo "HOMEBREW_NO_AUTO_UPDATE=1" >> $GITHUB_ENV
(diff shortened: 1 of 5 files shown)


argapost:
After this commit lvim cannot find my .config/lvim and it falls back to .config/nvim where there is no config.lua etc. so I get an error.
From the code I see that you get the config_dir from:

```lua
function _G.get_config_dir()
 local lvim_config_dir = os.getenv "LUNARVIM_CONFIG_DIR"
```
 but there is no such env var in my system, should I define it myself?

rebuilt:
It's possible you have an older version of the launcher installed.  What is the content of `cat ~/.local/bin/lvim` ?

argapost:
Yeap just realised it, I git pull so I don't run the launcher therefore the bin/lvim stays the same.

kylo252:
@argapost, so it still an issue?

argapost:
Yes this has been fixed but now it cannot find the require("lsp").global_setuo function even though i can see it in the file. I am working on it

argapost:
Ok I had to :LuaCacheClear first, all good now thanks.

PR #1537, opened 2021-09-13
[Bugfix] common capabilities returning nothing if no cmp_nvim_lsp

luong-komorebi:
# Description

While we heavily depend on nvim-cmp for completion with lsp, it should not block other capabilities of the lsp.

Given a use case when user tries to unload or opt for an alternative to nvim-cmp-lsp, the line of code that I fixed will return basic capabilities instead of nothing at all (no capabilities for ls)


kylo252:
Thank you!

PR #1547, opened 2021-09-14
[Refactor]: Reflect deprecated lualine options

dylantjb:
# Description

Replaced deprecated lualine options with their counterparts.

## How Has This Been Tested?

Deprecation message does not show on lualine.


abzcoding:

PR #1551, opened 2021-09-14
fix: errors caused by deprecated options in lualine config.

juanlopez4691:
# Description

This fixes errors caused by use of deprecated settings in lualine config file.

When LunarVim starts, the following error appears:
```
lualine: There are some issues with your config. Run :LualineNotices for details
```

Errors reported when executing `:LualineNotices`:
```
### option.left_padding

Option `left_padding` has been deprecated.
Please use `padding` option to set left/right padding.

You have some thing like this in your config:

'''lua
  left_padding = 0,
'''

You'll have to change it to this to retain old behavior:

'''lua
  padding = { left = 0 }
'''
if you've set both left_padding and right_padding for a component
you'll need to have something like
'''lua
  padding = { left = x, right = y }
'''
When you set `padding = x` it's same as `padding = {left = x, right = x}`

### option.left_padding

Option `left_padding` has been deprecated.
Please use `padding` option to set left/right padding.

You have some thing like this in your config:

'''lua
  left_padding = 0,
'''

You'll have to change it to this to retain old behavior:

'''lua
  padding = { left = 0 }
'''
if you've set both left_padding and right_padding for a component
you'll need to have something like
'''lua
  padding = { left = x, right = y }
'''
When you set `padding = x` it's same as `padding = {left = x, right = x}`

### option.condition
condition option has been renamed to `cond`. Please use `cond` instead in your config
for diagnostics component.

### option.condition
condition option has been renamed to `cond`. Please use `cond` instead in your config
for function component.

### option.condition
condition option has been renamed to `cond`. Please use `cond` instead in your config
for function component.

### option.condition
condition option has been renamed to `cond`. Please use `cond` instead in your config
for filetype component.

### diff.options.colors
Previously colors in diff section was set with color_added, color_modified..
separate options . They've been unified under diff_color option.
Now it should be something like:
'''lua
{ 'diff',
  diff_color = {
    added = color_added,
    modified = color_modified,
    removed = color_removed,
  }
}
'''

### option.condition
condition option has been renamed to `cond`. Please use `cond` instead in your config
for function component.

### option.condition
condition option has been renamed to `cond`. Please use `cond` instead in your config
for function component.
```

Fixes #(issue)

## How Has This Been Tested?

Once the fixes are applied, lualine complains no more.

### DIFF
--- a/lua/core/lualine/components.lua
+++ b/lua/core/lualine/components.lua
@@ -17,31 +17,33 @@ return {
     function()
       return " "
     end,
-    left_padding = 0,
-    right_padding = 0,
+    padding = { left = 0 },
+    padding = { right = 0 },
     color = {},
-    condition = nil,
+    cond = nil,
   },
   branch = {
     "b:gitsigns_head",
     icon = " ",
     color = { gui = "bold" },
-    condition = conditions.hide_in_width,
+    cond = conditions.hide_in_width,
   },
   filename = {
     "filename",
     color = {},
-    condition = nil,
+    cond = nil,
   },
   diff = {
     "diff",
     source = diff_source,
     symbols = { added = "  ", modified = "柳", removed = " " },
-    color_added = { fg = colors.green },
-    color_modified = { fg = colors.yellow },
-    color_removed = { fg = colors.red },
+    diff_color = {
+      added = colors.green,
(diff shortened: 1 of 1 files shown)


abzcoding:

PR #1552, opened 2021-09-14
Fix lualine deprecation warnings

seik:
Lualine was giving deprecation warnings about some settings. This changes adapt the lualine configuration to use the new settings.


abzcoding:

PR #1564, opened 2021-09-16
fix: use the correct API for the log-viewer

### LINKED ISSUE #1498: [Bug]: which-key bindings for LunarVim logs are invalid
### Problem description

After #1405 was introduced, the [which-key bindings for logging](https://github.com/LunarVim/LunarVim/blob/rolling/lua/core/which-key.lua#L195-L212) are no longer valid.
The lower-case bindings are all invalid and produce an error, the upper-case bindings no longer make sense since the logs seem to be unified now

### LunarVim version

rolling#151684bba11b702907e58d2acdd41ebc69c27809

### Neovim version (>= 0.5)

0.5

### Relevant log output

_No response_

### Screenshots

_No response_

### I have

- [X] Read the readme
- [X] Read the wiki

kylo252:
# Description

Updates the internal call in `log_viewer` to reflect the latest changes.

Fixes #1498 

## How Has This Been Tested?

All the options in `<leader>L` should work correctly.


abzcoding:
does this fix #1498 as well?

PR #1565, opened 2021-09-16
[Feature]: use yarn to install nodejs dependencies

ssbanerje:
## Description

This PR adds `yarn` support for the installer script.

**Edit:** I have also added non-interactive install options to the CLI. This was after the original PR. Not changing the title.

## Specific changes
In `utils/installer/install.sh`
- `check_dep` can now check for multiple commands and "return" the value of the first command that it found.
- `install_nodejs_deps` uses `yarn` and falls back to `npm`. 
- `parse_arguments` to parse all the arguments to the script.
    - Added a `-h, --help` 
    - Added a `-y, --yes`. Yes to all interactive options.
    - Added a `-n, --no`. No to all interactive options.

## Discussion
The PR changes behavior to prefer `yarn` over `npm` for installing the nodejs dependencies. In the case that `yarn` is not found, it will fall back to `npm`. 

I can update it if there is a preference for `npm` over `yarn`. In that case, maybe there should be a CLI argument or environment variable to signal a preference for yarn.

## How Has This Been Tested?
- Installs on `archlinux:latest` container.
- Installs on the `utils/docker` container image in the repository.


abzcoding:
cc: @kylo252

utils/installer/install.sh
```diff
@@ -1,4 +1,12 @@
 #!/usr/bin/env bash
+#/ Usage: install.sh [<options>]
```
kylo252, line 2: It's easier and more standard if you make this into a function `usage()`.

utils/installer/install.sh
```diff
@@ -31,13 +39,36 @@ declare -a __pip_deps=(
   "pynvim"
 )
 
+function parse_arguments() {
+  while (("$#")); do
+    case "$1" in
+      -y | --yes)
+        ARGS_INSTALL_NONINTERACTIVE="y"
+        ;;
+      -n | --no)
```
kylo252, line 48: `--no` is redundant and will break compatibility. It's enough by declaring `ARGS_INSTALL_NONINTERACTIVE=''` and checking if it's empty or not.

utils/installer/install.sh
```diff
@@ -61,17 +92,25 @@ EOF
 
   __add_separator "80"
 
-  echo "Would you like to install lunarvim's NodeJS dependencies?"
-  read -p "[y]es or [n]o (default: no) : " -r answer
-  [ "$answer" != "${answer#[Yy]}" ] && install_nodejs_deps
+  if [[ -z "$ARGS_INSTALL_NONINTERACTIVE" ]]; then
```
kylo252, line 95: Single brackets for consistency.

utils/installer/install.sh
```diff
@@ -80,17 +119,17 @@ EOF
...
   __add_separator "80"
 
-  case "$@" in
-    *--overwrite*)
-      echo "!!Warning!! -> Removing all lunarvim related config \
-        because of the --overwrite flag"
+  if [[ -n "$ARGS_OVERWRITE" ]]; then
+    echo "!!Warning!! -> Removing all lunarvim related config \
+      because of the --overwrite flag"
+    if [[ -z "$ARGS_INSTALL_NONINTERACTIVE" ]]; then
```
kylo252, line 125: Single brackets for consistency.

utils/installer/install.sh
```diff
@@ -138,8 +177,16 @@ function print_missing_dep_msg() {
 }
 
 function check_dep() {
-  if ! command -v "$1" &>/dev/null; then
-    print_missing_dep_msg "$1"
```
kylo252, line 142: I get the idea, but it has a few problems.
- it's used once, and pretty unlikely to be needed in the future
- the function name doesn't even reflect the new behaviour
- it created an implicit global variable `$check_dep_return` but you're ignoring the return value, maybe it was the one actually intended to be used.

Overall I actually think it's better to retain the original function since this is not gonna save us anything. It will be more clear if you just check each of them individually.

kylo252 (changes requested):
Thank you for adding the help function, it's been on my radar for a while. 

I'm not sure about choosing `yarn` over `npm`. I have not been able to reliably use `yarn add global` with `fnm`. But I'm leaving this particular choice for others to decide on the pros/cons of each tool. I also don't use either nearly enough to tell 😶

utils/installer/install.sh
ssbanerje: I will move it into a `usage` funciton.

utils/installer/install.sh
ssbanerje: I am not sure why you say `--no` is redundant. It is answering now for all interactive questions, which needs some mechanism to be signalled.

Also internally, the `ARGS_INSTALL_NONINTERACTIVE` has to have three values:
1.  Non-interactive and "yes to all". Here we have `ARGS_INSTALL_NONINTERACTIVE=y`
2. Non-interactive and  "no to all". Here we have `ARGS_INSTALL_NONINTERACTIVE=n`
3. Interactive. Here we have `ARGS_INSTALL_NONINTERACTIVE` not set
 
Maybe I do not understand how this can be achieved with two states.

utils/installer/install.sh
ssbanerje: Sure, I can make the change, maybe it would be nice to have consensus on the order of dependencies -- `npm` as first choice or `yarn`. That way it could be one commit.

utils/installer/install.sh
kylo252: I guess I just don't see the use-case for it. I know that `-y` is pretty common, but it looks odd to me even looking at it from a pure logical point of view: `non-interactive=no` -> `not non-interactive` -> `interactive?`.
From a user-experience point of view: you pass `-y` if you trust the installer wouldn't do anything weird, but you wouldn't pass `-n` because you probably don't know what that means/does yet.
Maybe you didn't notice that the answers to these questions are set to `no` by default? you can just quickly press `enter` and move on with the installation.

Since this is user-experience oriented, let's see what the others think
cc. @abzcoding, @Tastyep, @chaesngmin

utils/installer/install.sh
```diff
@@ -31,13 +32,46 @@ declare -a __pip_deps=(
...
+  echo ""
+  echo "Options:"
+  echo "    -h, --help    Print this help message"
+  echo "    -y, --yes     Yes for all choices (Install NodeJS, Python, Rust dependencies)"
+  echo "    -n, --no      No for all choices (Non interactive install)"
+  echo "    --overwrite   Overwrite previous lvim configuration"
+}
+
+function parse_arguments() {
+  while (("$#")); do
```
Tastyep, line 46: To avoid evaluating `$#` you can instead write
`while [[ $# -gt 0 ]]; do`

utils/installer/install.sh
Tastyep: IMO, the interactive option isn't necessary. I see the usage for `apt` where you can do `apt install --yes`, but for the setup script since I'm only executing it once or twice I don't really see a use case.

utils/installer/install.sh
Tastyep: IMO the whole function serves no purpose, check_dep is already achieved by `command -v`.
You would just need to move the loop in `install_nodejs_deps` and print a message if it returns a non 0 code
Also, the function should not exit

utils/installer/install.sh
```diff
@@ -156,14 +206,22 @@ function check_system_deps() {
 }
 
 function install_nodejs_deps() {
-  check_dep "npm"
-  echo "Installing node modules with npm.."
-  for dep in "${__npm_deps[@]}"; do
-    if ! npm ls -g "$dep" &>/dev/null; then
+  check_dep "yarn" "npm"
+  if [ "$check_dep_return" == "npm" ]; then
```
Tastyep, line 210: Using globals for means of communication is bad practice

Tastyep (changes requested):

utils/installer/install.sh
ssbanerje: @Tastyep Ok. I can change it to that.

I think the current behavior is to fail (exit) if npm is not installed and the user answer was yes. Should I retain that?

utils/installer/install.sh
ssbanerje: I think it's useful because it allows running the setup script in non-interactive mode as part of a script.

I think this is useful to package the editor setup as a container. I do this to make my editor/dotfiles portable across machines. Having this in the installer script allows the install to be consistent, as opposed doing a manual install in the dockerfile. 

It seems that in the current container setup (i.e., `utils/docker`) will fail because of the interactive input requirement of `install.sh`. (**In fact, it currently fails because it can't detect Neovim. But once that is fixed, it will fail because of the input requirement.**)

In any case, I can get rid of the `--no` if you guys think that it isn't useful.

utils/installer/install.sh
ssbanerje: Thats the only thing you can do in the shell. No return values on the functions :(

Tastyep: To 'return' values, you have to echo "value" and capture it back in the caller.

Anyway, since I think check_dep should be merged inside this function, it won't be an issue  anymore.
I have a very similar function on one of my projects if you want to draw inspiration:
https://github.com/Tastyep/Pi-OpenCast/blob/master/setup.sh#L30-L49

@kylo252 Are you ok with this ?

utils/installer/install.sh
Tastyep: https://github.com/LunarVim/LunarVim/pull/1565/files#r711616459

utils/installer/install.sh
kylo252: I had an entire https://github.com/LunarVim/LunarVim/pull/1565#discussion_r711380560 about it where I concluded that we should get rid of it.

> Overall I actually think it's better to retain the original function since this is not gonna save us anything. It will be more clear if you just check each of them individually.

Tastyep: I read your comment and agree with it.
In my comment I said we could go even further and I showed how I would do it.

Uzaaft:
> Thank you for adding the help function, it's been on my radar for a while.
> 
> I'm not sure about choosing `yarn` over `npm`. I have not been able to reliably use `yarn add global` with `fnm`. But I'm leaving this particular choice for others to decide on the pros/cons of each tool. I also don't use either nearly enough to tell

Struggled a little bit with this myself. Turns out, its not 
```yarn add global``` but ```yarn global add```
[Reference](https://classic.yarnpkg.com/en/docs/cli/global/)
Just thought I`d help xD

**Comment by ssbanerje** [comment]
Hi @kylo252 @Tastyep... I have made the changes you had requested. Can you have a look at it now?

Also, since I am doing this... should I also fix the docker container? It fails to run (which was partially the reason I got started with this). Also it seems to do a lot of redundant things (like install fonts and X11 packages like ueberzug). Both are unlikely to make any difference in the container. I guess both have to be set up on the host which runs the container. Could you give me any context about this?

File: utils/installer/install.sh [hunk 8]
```diff
@@ -187,18 +246,19 @@ function __attempt_to_install_with_cargo() {
...
 
 # we try to install the missing one with cargo even though it's unlikely to be found
 function install_rust_deps() {
-  if ! command -v fd &>/dev/null; then
-    __attempt_to_install_with_cargo "fd-find"
-  fi
-  if ! command -v rg &>/dev/null; then
-    __attempt_to_install_with_cargo "ripgrep"
-  fi
+  local -A deps=(["fd"]="fd-find" ["rg"]="ripgrep")
```
kylo252, line 256: This will sadly now work on macOS which uses an ancient bash by default. It's quite unfortunate because I was also using these nice associative arrays and had to change them.

utils/installer/install.sh
```diff
@@ -61,17 +91,25 @@ EOF
...
+    [ "$answer" != "${answer#[Yy]}" ] && install_python_deps
 
-  echo "Would you like to install lunarvim's Rust dependencies?"
-  read -p "[y]es or [n]o (default: no) : " -r answer
-  [ "$answer" != "${answer#[Yy]}" ] && install_rust_deps
+    echo "Would you like to install lunarvim's Rust dependencies?"
+    read -p "[y]es or [n]o (default: no) : " -r answer
+    [ "$answer" != "${answer#[Yy]}" ] && install_rust_deps
+  else
+    if [ "$ARGS_INSTALL_NONINTERACTIVE" == y ]; then
```
kylo252, line 107: There's no need to validate this again since it can only take a pre-set value. You can declare as read-only if you want.

kylo252 (changes requested):

kylo252:
> Also, since I am doing this... should I also fix the docker container? It fails to run (which was partially the reason I got started with this). Also it seems to do a lot of redundant things (like install fonts and X11 packages like ueberzug). Both are unlikely to make any difference in the container. I guess both have to be set up on the host which runs the container. Could you give me any context about this?


That dockerfile is pretty rudimentary, and hasn't been really maintained since it was first introduced.
I would say feel free to write a new one from scratch, if it's easier, with hopefully some proper multi-stage builds if possible.

kylo252:
> Struggled a little bit with this myself. Turns out, its not
> `yarn add global` but `yarn global add`
> [Reference](https://classic.yarnpkg.com/en/docs/cli/global/)
> Just thought I`d help xD

wtf! how is this even a thing? thanks for solving that mystery for me 😄

ssbanerje:
> > Also, since I am doing this... should I also fix the docker container? It fails to run (which was partially the reason I got started with this). Also it seems to do a lot of redundant things (like install fonts and X11 packages like ueberzug). Both are unlikely to make any difference in the container. I guess both have to be set up on the host which runs the container. Could you give me any context about this?
> 
> That dockerfile is pretty rudimentary, and hasn't been really maintained since it was first introduced.
> I would say feel free to write a new one from scratch, if it's easier, with hopefully some proper multi-stage builds if possible.

Ok while doing this I just figured out that the install script has a bug in the cargo installation:

https://github.com/LunarVim/LunarVim/blob/65392c553ea6a69053a99fea57c3dee6c48afa75/utils/installer/install.sh#L186-L191

That `!` should not be there. I will fix it in my next push.

utils/docker/Dockerfile
```diff
@@ -1,11 +1,9 @@
...
-SHELL ["/bin/bash", "-c"]
+RUN apk upgrade --no-cache && \
+    apk add --update --no-cache bash sudo git nodejs npm fzf ranger curl python3-dev py3-pip neovim cargo lazygit g++
 
-ENV DEBIAN_FRONTEND=noninteractive
-
-COPY script.sh /tmp/script.sh
-
-RUN bash -c "/tmp/script.sh"
+RUN LVBRANCH=master curl -s https://raw.githubusercontent.com/lunarvim/lunarvim/rolling/utils/installer/install.sh | bash -s -- -y
```
Tastyep, line 6: Shouldn't we try to match the version ran in the container with the current branch ? I think that's why the setup script was copied

utils/installer/install.sh
```diff
@@ -167,6 +203,29 @@ function install_nodejs_deps() {
...
 
+function __install_nodejs_deps_yarn() {
+  echo "Installing node modules with yarn.."
+  yarn global add "${__npm_deps[@]}"
+  echo "All NodeJS dependencies are succesfully installed"
+}
+
+function install_nodejs_deps() {
+  local -a pkg_managers=("yarn" "npm")
+  for pkg in "${pkg_managers[@]}"; do
```
Tastyep, line 214: `for pkg_manager in "${pkg_managers[@]}"`

utils/docker/Dockerfile
ssbanerje: Well it did not do that previously. The script just installed whatever was in the `master` branch. Only the script was copied in and not the actual LVIM git repo.

I just pushed a commit which uses a container build argument to point it to correct `LVBRANCH`. This way whoever is building the container will be able to point it at a branch on GitHub.

The real question here is this meant to be a development container? Then copying in the local git repo makes sense.

Or is it a user container? Then just pointing it to a branch on GitHub makes sense.

utils/installer/install.sh
ssbanerje: fixed in 2d28c78

utils/installer/install.sh
ssbanerje: fixed in 2d28c78

utils/installer/install.sh
ssbanerje: fixed in 2d28c78

utils/docker/Dockerfile
kylo252: Would you mind moving the changes to the dockerfile to a separate PR? I have a feeling it's gonna take some elbow grease 😶

ssbanerje:
@kylo252 @Tastyep I have moved the container commit out of this PR.

Moved the Dockerfile stuff to #1583

kylo252:
Thank you, @ssbanerje ❤️

PR #1581, opened 2021-09-19
[Bugfix]: key mappings of nvim-cmp doesn't work

jtr860830:
Relate to: https://github.com/hrsh7th/nvim-cmp/issues/221

# Description

Because of this https://github.com/hrsh7th/nvim-cmp/commit/7c8876330df6e11cf91ca42ab6e52f37d81cc003, mappings of nvim-cmp's configuration need updating to all caps.

## How Has This Been Tested?

- Use `lvim` to edit a file. e.g. `test.js`
- Check `<Tab>` and `<S-Tab>` can work as expect when completion menu is visible.

### DIFF
--- a/lua/core/cmp.lua
+++ b/lua/core/cmp.lua
@@ -73,10 +73,10 @@ M.config = function()
       { name = "crates" },
     },
     mapping = {
-      ["<C-d>"] = cmp.mapping.scroll_docs(-4),
-      ["<C-f>"] = cmp.mapping.scroll_docs(4),
+      ["<C-D>"] = cmp.mapping.scroll_docs(-4),
+      ["<C-F>"] = cmp.mapping.scroll_docs(4),
       -- TODO: potentially fix emmet nonsense
-      ["<Tab>"] = cmp.mapping(function()
+      ["<TAB>"] = cmp.mapping(function()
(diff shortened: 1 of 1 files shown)


lua/core/cmp.lua
```diff
@@ -73,10 +73,10 @@ M.config = function()
       { name = "crates" },
     },
     mapping = {
-      ["<C-d>"] = cmp.mapping.scroll_docs(-4),
-      ["<C-f>"] = cmp.mapping.scroll_docs(4),
+      ["<C-D>"] = cmp.mapping.scroll_docs(-4),
```
rebuilt, line 76: I understand mapping tab and space but `C-d` is a completely different mapping than `C-D`.  I'm not sure the letter mappings need to be uppercased.

jtr860830: You are right. It is my mistake.

jtr860830:
This issue has been fixed by https://github.com/hrsh7th/nvim-cmp/commit/13e8b61371925ebcea4069a67f16315e96c8f935.
Just do `:PackerSync` or `<leader>ps` and these mapping work as expect.

rebuilt:
Fixed upstream

Issue #945, opened 2021-07-14
[Bug]: No Diagnostics on Ruby


jonathan-raines:
### Problem description

- Using rbenv as Ruby version manager
- Using 2.7.1
- gem install solargraph, gem install rails
- LspInstall solargraph

I no longer have diagnostics on Ruby files. I have not modified any of the Ruby settings in default-config.lua. Also how could I switch back to using Rubocop? I have a rubocop.yml file that used to be sourced for linting and formatting and all of this has stopped working around the time Rufo was added as the Ruby formatter.

I am up to date with Neovim and I'm on the Rolling LunarVim release.

### Check the following

- [X] I am in a git managed directory
- [X] I restarted Neovim after running :LspInstall
- [X] I have a valid language configuration (~/.config/nvim/ftplugin/)
- [X] I checked the options (~/.config/nvim/lv-config.lua)
- [X] The programs specified by the config are installed

### LunarVim version

0.4.8

### Neovim version (>= 0.5)

NVIM v0.6.0-dev+42-geece0735f

### Screenshots
![image](https://user-images.githubusercontent.com/5840124/125691332-b443cd7c-b9f1-4750-b218-218a1eb3fd75.png)

### I have

- [X] Read the readme
- [X] Read the wiki
- [X] Searched for similar issues

abzcoding:
for changing the formatter, you can do this 
```lua
-- lv-config.lua
O.lang.ruby.formatter.exe = "robocop"
O.lang.ruby.formatter.args =  {}
```

jonathan-raines:
Running rubocop in that fashion runs the linter and checks for errors. I used to receive diagnostics from the LSP depending on the `.rubocop.yml` file for the project. E.g. line too long. Format on save used to fix most, if not all of those errors for me. Now I don't receive any type of diagnostics from the LSP.

`:echom &filetype = ruby`

abzcoding:
> Running rubocop in that fashion runs the linter and checks for errors. I used to receive diagnostics from the LSP depending on the `.rubocop.yml` file for the project. E.g. line too long. Format on save used to fix most, if not all of those errors for me. Now I don't receive any type of diagnostics from the LSP.
> 
> `:echom &filetype = ruby`

can you provide your `~/.cache/nvim/lsp.log` so i can dig deeper

jonathan-raines:
[lsp.log](https://github.com/ChristianChiarulli/LunarVim/files/6824508/lsp.log)
``

[...]

PR #1584, opened 2021-09-19
refactor: auto-generate language configuration

kylo252:
# Description

Break the monolithic `lvim.lang` design into a more modular approach.

**IMPORTANT**: run `:LvimUpdate` in order to generate the new ftplugins template files..

<details>
<summary>Interactive demo</summary>
<a href="https://asciinema.org/a/m4ZtzKIkqOXZLkJFgbKeBOj39" target="_blank"></a>

</details>

---

TODO: 
- [ ] add documentation
- [x] add unit-tests
- [x] fix lualine
- [x] make sure LvimInfo still works
- [x] make sure no providers are missing
- [x] benchmarks startup time (optional)
- [x] make the switch to `nvim-lsp-installer`
- [x] remove all entries that only contain the `server` command
- [x] provide a mechanism to update lvim


lua/lsp/providers/typescriptreact.lua
```diff
@@ -0,0 +1,11 @@
+local opts = {
```
rebuilt, line 1: lvim.lang.javascript and lvim.lang.typescript are available through autocomplete but their react analogues are not: lvim.lang.javascriptreact and lvim.lang.typescriptreact.  And any attempt to use typescriptreact results in a config error.

lua/lsp/providers/ruby.lua
```diff
@@ -0,0 +1,27 @@
+local opts = {
```
rebuilt, line 1: Ruby is a special case.  

Currently, installing the language server using `:LspInstall` does not work correctly.  The language server, solargraph, doesn't like being launched from a different directory from the codebase and cannot find the project files.  

Solargraph needs to be gem installed `gem install solargraph` from the root directory of the project.  To get it started correctly, I have this in my configuration, which no longer works because the field `lsp` is no longer available on `lvim.lang.ruby.lsp`


```lua


local util = require("lspconfig/util")
lvim.lang.ruby.lsp.setup = {
	cmd = { "solargraph", "stdio" },
	filetypes = { "ruby" },
	init_options = {
		formatting = true,
	},
	root_dir = util.root_pattern("Gemfile", ".git"),
	settings = {
		solargraph = {
			diagnostics = true,
		},
	},
}
```

lua/lsp/providers/python.lua
```diff
@@ -0,0 +1,11 @@
+local opts = {
```
rebuilt, line 1: I used to be able to place my null-ls formatter configurations in my ftplugin folder in `~/.config/lvim`  but that is no longer possible.  I have to place `lvim.lang.python.formatters` directly in my config.lua

rebuilt (changes requested):

lua/lsp/providers/ruby.lua
kylo252: So my plan is to port any generic changes upstream to lspconfig instead of maintaining them here, and only keep any edge-cases.
If you take a look here, you can see that you're basically want to use a generic setup: 
https://github.com/neovim/nvim-lspconfig/blob/aa0b9fd746d73a5ebbc72c732c645e96423d4504/lua/lspconfig/solargraph.lua

Can you try deleting the file entirely and see if solargraph works for you?

lua/lsp/providers/python.lua
kylo252: I didn't really do anything to the autocmd logic except deleting the folder `ftplugin` only because I didn't actually check if there are non-lsp related files. So can you try creating an empty `ftplugin` folder and see if that was reason?

kylo252:
By the way, you guys don't need to actually review the files in 'lsp/providers' individually. These are basically auto-generated from the old defaults. So I didn't really change anything in them, at least not intentionally. So the bulk of the number of the changed files is just me deleting them from `ftplugin` and temporarily having some of them in `lsp/providers`

lua/lsp/providers/typescriptreact.lua
kylo252: Oh I forgot to document this, I added a new variable temporarily, to set up the common languages quickly: https://github.com/LunarVim/LunarVim/blob/71aa17810261b0a41130de19df692c3ad675b75f/lua/config/defaults.lua#L34

I'm not yet settled if we should keep it or not. 

Can you append the values you want to it and if it solves the issue?

lua/utils/init.lua
```diff
@@ -146,6 +144,18 @@ function utils.join_paths(...)
   return result
 end
 
+function utils.write_file(path, txt, flag)
```
Tastyep, line 147: is this used ?

Tastyep (changes requested):
I think this goes in the right direction, the config is finally split into multiple parts, however, the negative point is that at setup time it is loaded all at once.

IIUC, [formatters and linters](https://github.com/LunarVim/LunarVim/pull/1584/files#diff-80ab79c7feae2125a0e751b385e4689bf2c92e0cdb374fd9ba810a453602b985R55) can still be configured, but not the LSP setup anymore.

I don't want to shift the subject to why having the global `lvim` is more of a constraint than smth useful, but I'm sure you can see why it's annoying here.

lua/lsp/init.lua
```diff
@@ -115,41 +82,51 @@ function M.common_on_attach(client, bufnr)
...
   lsp_highlight_document(client)
   add_lsp_buffer_keybindings(bufnr)
-  require("lsp.null-ls").setup(vim.bo.filetype)
 end
 
-function M.setup(lang)
-  local lsp_utils = require "lsp.utils"
-  local lsp = lvim.lang[lang].lsp
-  if (lsp.active ~= nil and not lsp.active) or lsp_utils.is_client_active(lsp.provider) then
+function M.setup_by_ft(lang, opts)
```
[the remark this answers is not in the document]
kylo252: It's annoying/wrong to have the opts connected to the language instead of the provider. If that's what you mean, then we're on the same page. 

Also, nearly all of those files in providers are gonna disappear as well, so we can freely change that struct.

lua/lsp/init.lua
```diff
@@ -115,41 +82,51 @@ function M.common_on_attach(client, bufnr)
...
+  local lsp_settings_status_ok, lsp_settings = pcall(require, "nlspsettings")
+  if lsp_settings_status_ok then
+    lsp_settings.setup {
+      config_home = utils.join_paths(get_config_dir(), "lsp-settings"),
+    }
+  end
+end
+
+local function is_overrided(lang)
   local overrides = lvim.lsp.override
```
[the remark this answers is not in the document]
kylo252: I haven't seen it used, but it feels like it will turn off the autosetup of the language-server. It should be removable if we end up going with some variant of `ensure_configured`. Since you can just exclude any language that you may want to setup manually.

lua/lsp/init.lua
```diff
@@ -158,20 +135,24 @@ function M.global_setup()
...
-  local utils = require "utils"
+  local configured_languages = lvim.ensure_configured
+  Log:debug("Loading " .. #configured_languages .. " language(s): " .. vim.inspect(configured_languages))
 
-  local lsp_settings_status_ok, lsp_settings = pcall(require, "nlspsettings")
-  if lsp_settings_status_ok then
-    lsp_settings.setup {
-      config_home = utils.join_paths(get_config_dir(), "lsp-settings"),
-    }
+  for _, lang in ipairs(configured_languages) do
```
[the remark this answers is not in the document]
kylo252: If you mean that I should wait for a `BufReadPre` or something similar, that sounds like a good idea.

lua/lsp/manager.lua
```diff
@@ -0,0 +1,82 @@
+local M = {}
+
+function M.get_all_supported_servers()
```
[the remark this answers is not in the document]
kylo252: Yeah this list isn't even used currently, I had it generated from lspconfig. The functions you're looking at is a bit deceiving as it will actually only give you servers _you have already configured_.

lua/utils/init.lua
kylo252: Not in this iteration, but I used it to write all the files in lsp/providers.

lua/lsp/null-ls/formatters.lua
```diff
@@ -62,12 +45,13 @@ function M.list_configured(formatter_configs)
   return { supported = formatters, unsupported = errors }
 end
 
-function M.setup(filetype, options)
-  if not lvim.lang[filetype] or (formatters_by_ft[filetype] and not options.force_reload) then
+function M.setup(formatter_configs, filetype)
+  if vim.tbl_isempty(formatter_configs) then
```
Tastyep, line 49: An empty table should be a valid value, I think we can remove this condition

lua/lsp/null-ls/services.lua
```diff
@@ -45,4 +45,19 @@ function M.find_command(command)
   return nil
 end
 
+function M.list_registered_providers_names(filetype)
+  local u = require "null-ls.utils"
+  local c = require "null-ls.config"
```
Tastyep, line 50: Please don't use one character variables.
https://github.com/luarocks/lua-style-guide#variable-names

lua/lsp/null-ls/services.lua
```diff
@@ -45,4 +45,19 @@ function M.find_command(command)
...
+  local registered = {}
+  for method, source in pairs(c.get()._methods) do
```
Tastyep, line 52: We should never rely on private attributes.

lua/lsp/utils.lua
```diff
@@ -10,19 +10,51 @@ function M.is_client_active(name)
...
+    local supported_filetypes = client.config.filetypes or {}
+    if client.name ~= "null-ls" and vim.tbl_contains(supported_filetypes, filetype) then
+      table.insert(matches, client)
+    end
+  end
+  return matches
+end
+
+function M.get_ls_capabilities(client_id)
+  local client
```
Tastyep, line 32: Merge with L47
https://github.com/luarocks/lua-style-guide#variable-scope

lua/lsp/utils.lua
```diff
@@ -10,19 +10,51 @@ function M.is_client_active(name)
...
-  return nil
+  if not client_id then
+    error "Unable to determine client_id"
+    return
+  end
+
+  client = vim.lsp.get_client_by_id(tonumber(client_id))
+
+  local enabled_caps = {}
+
```
Tastyep, line 50: no newline

lua/lsp/utils.lua
```diff
@@ -10,19 +10,51 @@ function M.is_client_active(name)
...
+  for k, v in pairs(client.resolved_capabilities) do
```
Tastyep, line 51: k, v -> capability, status ?

lua/lsp/null-ls/formatters.lua
```diff
@@ -1,29 +1,12 @@
...
-  return list_names(formatters_by_ft[filetype], { filter = "supported" })
-end
-
-function M.list_unsupported_names(filetype)
-  if not formatters_by_ft[filetype] then
-    return {}
-  end
-  return list_names(formatters_by_ft[filetype], { filter = "unsupported" })
+  local registered_providers = services.list_registered_providers_names(filetype)
+  return registered_providers["NULL_LS_FORMATTING"] or {}
```
Tastyep, line 9: Hard-coded values "NULL_LS_FORMATTING" and "NULL_LS_DIAGNOSTICS" should be local constants

lua/lsp/null-ls/init.lua
```diff
@@ -1,44 +1,26 @@
...
 
-  local formatters = require "lsp.null-ls.formatters"
-  local linters = require "lsp.null-ls.linters"
-
-  formatters.setup(filetype, options)
-  linters.setup(filetype, options)
+  null_ls.config()
+  require("lspconfig")["null-ls"].setup {}
+  for _, filetype in pairs(lvim.lang) do
+    if filetype.formatters then
```
Tastyep, line 17: I think that once it is autogenerated (which you have already partially done) we can remove these conditions

Tastyep (changes requested):
First, it's really nice to finally get rid of all these structures.

Null-ls integration needs to be fixed, linters and formatters don't work for me.
As commented, we should not rely on private attributes from null-ls, I understand that it's handy, but it's preferable that we duplicate providers on our side.

lua/lsp/init.lua
```diff
@@ -115,63 +80,45 @@ function M.common_on_attach(client, bufnr)
...
-    if not lsp.setup.on_init then
-      lsp.setup.on_init = M.common_on_init
-    end
-    if not lsp.setup.capabilities then
-      lsp.setup.capabilities = M.common_capabilities()
-    end
-
-    lspconfig[lsp.provider].setup(lsp.setup)
-  end
+function M.get_common_opts()
```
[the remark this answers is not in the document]
kylo252: I was thinking of creating a `commons` table instead. The problem is that I currently need it in `lv-reload-config` but shouldn't be necessary if we switch to templates.

lua/lsp/utils.lua
```diff
@@ -10,19 +10,51 @@ function M.is_client_active(name)
...
-  if not lvim.lang[filetype] or not lvim.lang[filetype].lsp then
-    return nil
-  end
+function M.disable_formatting_capability(client)
+  -- FIXME: figure out a reasonable way to do this
+  client.resolved_capabilities.document_formatting = false
+  require("core.log"):debug(string.format("Turning off formatting capability for language server [%s] ", client.name))
+end
 
+function M.get_active_client_by_ft(filetype)
```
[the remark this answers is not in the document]
kylo252: It's because you can "filter" the list by other attributes, so this sort of a core neovim idiom, hence `by_ft`
I wanted to actually remove this function in favor of the one I added in https://github.com/neovim/nvim-lspconfig/blob/9ca394ec3a9305a6c30018ee6f731c9b23a4b25e/lua/lspconfig/util.lua#L345, but I need your help cleaning up the null-ls module 😄

lua/config/init.lua
```diff
@@ -22,6 +22,9 @@ function M:init(opts)
 
   local settings = require "config.settings"
   settings.load_options()
+
+  lvim.lsp = require "lsp.config"
```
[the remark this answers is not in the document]
kylo252: Sounds good 👍 
I was basically trying to piece-out the logic from #1406 to see what I can re-use.

lua/lsp/null-ls/formatters.lua
```diff
@@ -1,29 +1,12 @@
...
-end
-
 function M.list_supported_names(filetype)
-  if not formatters_by_ft[filetype] then
-    return {}
-  end
-  return list_names(formatters_by_ft[filetype], { filter = "supported" })
-end
-
-function M.list_unsupported_names(filetype)
```
kylo252, line 22: I'm not sure if we can, or want to, restore this function, @Tastyep

kylo252: Do keep in mind that I'm currently using the info from null-ls directly that a provider is working, instead of us assuming it works just because we found the executable.

Tastyep: It would probably be easier to make a PR on null-ls to expose a method.
I think it was nice being able to see incorrectly configured providers.
If the data is exposed publicly from null-ls we could compare it to the configured providers and get the list from here.

lua/lsp/utils.lua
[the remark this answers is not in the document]
Tastyep: `list_acive_clients_by_ft` or `filter_active_clients_by_ft`

lua/core/info.lua
```diff
@@ -65,44 +66,55 @@ local function tbl_set_highlight(terms, highlight_group)
...
     fmt("Treesitter active:      %s", tostring(next(vim.treesitter.highlighter.active) ~= nil)),
   }
 
-  local text = require "interface.text"
   local lsp_info = {
     "Language Server Protocol (LSP) info",
-    fmt("* Associated server:    %s", client_name),
-    fmt("* Active:               %s (id: %d)", tostring(is_client_active), client_id),
-    fmt("* Supports formatting:  %s", tostring(document_formatting)),
+    fmt "* Associated server(s):",
```
Tastyep, line 103: You removed the parameter, the server is not displayed anymore.
Also, it might be useful to print whether the server is active or not.

rebuilt:
I have not been able to get solargraph to work with this pull request.  

For context solargraph has never worked using the regular LunarVim setup.  That's why a workaround was required.  There was a lot of headache to get it working properly in LunarVim.  https://github.com/LunarVim/LunarVim/issues/945

## LvimInfo
```bash
                          Detected filetype:      ruby
                          Treesitter active:      true
                          
                          Language Server Protocol (LSP) info
                          * Associated server(s):
                          
                          Formatters and linters
                          * Configured providers: 
                          
                          
                           HINT 
                          
                          * List of supported formatters: [ rubocop, rufo ]
      
```

## LspInfo

```bash
 Language client log: /home/nelson/.cache/nvim/lsp.log
 Detected filetype:   ruby
 
 0 client(s) attached to this buffer: 
 
 Configured servers list: emmet_ls, null-ls

```
`which solargraph` returned the one installed locally in the project so I uninstalled it and installed using `:LspInstall`

The `:LspInstall solargraph` installed to:
/home/user/.local/share/nvim/lsp_servers/solargraph 
That directory is full of files and folders but the language server just does not start.

:LspStart solargraph returns:
`E5108: Error executing lua .../site/pack/packer/start/nvim-lspconfig/lua/lspconfig.lua:29:
 attempt to index a nil value`
 
 The lsp error log does not contain any entries for solargraph(at least not after switching to this branch).

rebuilt:
Well, maybe this isn't a ruby problem.  I just tried some other languages and I don't get LSP for anything.  I've run out of time today to test though.

kylo252:
> Well, maybe this isn't a ruby problem. I just tried some other languages and I don't get LSP for anything. I've run out of time today to test though.

Oh I should edit the instructions for this. You need to run the generator at least once, or define them manually. I'm gonna make a quick update function to make this easier for everyone.

rebuilt:
Ok, so running make update worked.  Ruby works and solargraph is attached correctly.  Woohoo!

The only wrinkle now is that some of the other languages don't work correctly in that the language server still doesn't work or too many language servers are getting installed and attached.

`:LspInstall pyright` 
Pyright is correctly installed and attached but for unexplained reasons, so is the jedi language server and pylsp.  I thought it might just be incorrectly reporting the attachment but all the servers do seem to be correctly installed and attached.  The servers are all installed to /home/user/.local/share/nvim/lsp_servers.  

```lua
 3 client(s) attached to this buffer: 
 
 Client: pyright (id: 21, pid: 203554, bufnr: [12])
 	filetypes:       python
 	autostart:       true
 	root directory:  /home/nelson/workspace/canary
 	cmd:             /home/nelson/.local/share/nvim/lsp_servers/python/node_modules/.bin/pyright-langserver --stdio
 
 Client: pylsp (id: 20, pid: 203553, bufnr: [12])
 	filetypes:       python
 	autostart:       true
 	root directory:  /home/nelson/workspace/canary
 	cmd:             /home/nelson/.local/share/nvim/lsp_servers/pylsp/venv/bin/pylsp
 
 Client: jedi_language_server (id: 19, pid: 203552, bufnr: [12])
 	filetypes:       python
 	autostart:       true
 	root directory:  /home/nelson/workspace/canary
 	cmd:             /home/nelson/.local/share/nvim/lsp_servers/jedi_language_server/venv/bin/jedi-language-server

```
Cases where too many servers get installed:
tsserver:  also installs and attaches: stylelint_lsp, rome, ember, denols, tailwindcss
vuels: also installs tailwindcss, stylelint_lsp

Cases where the server does not get installed or attached correctly:
Omnisharp, html, cssls, 

Cases where everything works as expected
clangd

kylo252:
> Ok, so running make update worked.  Ruby works and solargraph is attached correctly.  Woohoo!
> 
> The only wrinkle now is that some of the other languages don't work correctly.  Either too many language servers are getting installed and attached or the language server does not attach at all.  
> 
> `:LspInstall pyright` 
> Pyright is correctly installed and attached but for unexplained reasons, so is the jedi language server and pylsp.  I thought it might just be incorrectly reporting the attachment but all the servers do seem to be correctly installed and attached.  The servers are all installed to /home/user/.local/share/nvim/lsp_servers.  
> 
> ```lua
>  3 client(s) attached to this buffer: 
>  
>  Client: pyright (id: 21, pid: 203554, bufnr: [12])
>  	filetypes:       python
>  	autostart:       true
>  	root directory:  /home/nelson/workspace/canary
>  	cmd:             /home/nelson/.local/share/nvim/lsp_servers/python/node_modules/.bin/pyright-langserver --stdio
>  
>  Client: pylsp (id: 20, pid: 203553, bufnr: [12])
>  	filetypes:       python
>  	autostart:       true
>  	root directory:  /home/nelson/workspace/canary
>  	cmd:             /home/nelson/.local/share/nvim/lsp_servers/pylsp/venv/bin/pylsp
>  
>  Client: jedi_language_server (id: 19, pid: 203552, bufnr: [12])
>  	filetypes:       python
>  	autostart:       true
>  	root directory:  /home/nelson/workspace/canary
>  	cmd:             /home/nelson/.local/share/nvim/lsp_servers/jedi_language_server/venv/bin/jedi-language-server
> 
> ```
> Cases where too many servers get installed:
> tsserver:  also installs and attaches: stylelint_lsp, rome, ember, denols, tailwindcss
> vuels: also installs tailwindcss, stylelint_lsp
> 
> Cases where the server does not get installed or attached correctly:
> Omnisharp, html, cssls, 
> 
> Cases where everything works as expected
> clangd

Yeah we need some sort of whitelist method to select the desired server if we still wanna go with the auto option. Alternatively, we can disable the auto installation part and leave it to be manual.

lua/core/info.lua
kylo252: I'm not sure if this is exactly what you mean, but "active" parameter is redundant in this case since there's no a priori info about the server because that info is not in the global state anymore. 
So the way it works [now] is that if it's active, then it'll show up.

Tastyep: Alright, then only the name of the LS is missing

Makefile
```diff
@@ -11,6 +11,10 @@ install-neovim-binary:
 uninstall:
 	@echo TODO: this is currently not supported
 
+update:
+	@echo Starting LunarVim Updateer
```
Tastyep, line 15: Updateer

lua/utils/init.lua
```diff
@@ -163,7 +163,25 @@ function utils.lvim_cache_reset()
...
+      assert(not write_err, write_err)
+      uv.fs_close(fd, function(close_err)
+        assert(not close_err, close_err)
+      end)
+    end)
+  end)
+end
+
+--FIXME: this is still broken
+function utils.lvim_update()
```
Tastyep, line 179: Can we rename functions to always have the action first:
- update_lvim
- reset_cache

PR #1589, opened 2021-09-20
Rollback default branch name to master

travelerspb:
A few days ago the default installation branch was changed to `rolling` however everywhere in the documentation its still stated as `master`. Also I found no remarks about why it was changed, so I think it might me unwanted.

So I reverted it to `master`

### DIFF
--- a/utils/installer/install.sh
+++ b/utils/installer/install.sh
@@ -2,7 +2,7 @@
 set -eo pipefail
 
 #Set branch to master unless specified by the user
-declare LV_BRANCH="${LV_BRANCH:-"rolling"}"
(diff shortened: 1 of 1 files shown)


kylo252:
Thank you.
This was done at the time because the master branch was behind that caused the installer to fail. So we need to update master before we do this.

travelerspb:
@kylo252 do you think it might worth to update the documentation instead? (at least temporary) Because it states by default `master` will be pulled

kylo252:
@abzcoding, what do you think?

abzcoding:
i think updating the documentation is the better choice, until we've merge the new changes into master and release a new vesion.

PR #1593, opened 2021-09-21
[Bugfix]: Fix Dockerfile

ssbanerje:
## Description

This was created by splitting #1565 into two.

The docker file in `utils/docker` was outdated and failed to build. This was updated to work with the `utils/install/install.sh` script in #1565.

## Changes
1. The container now adds two build arguments:
    - `LVBRANCH`: Branch of lvim to use in installation (Defaults to `rolling`)
    - `NODE_RELEASE`: Version of Neovim to use (Defaults to `v0.5.0`)
2. The default docker command now launches `lvim` directly.
3. Removed dependencies that installed fonts, X11 components. These should be in the host and not the container

## How Has This Been Tested?
The build works and lvim launches when the container launches?
The current system has no tests for the container.


kylo252:
Thank you!

PR #1617, opened 2021-09-25
update nvim-tree based on the refactor

abzcoding:
# Description

Update nvim-tree configurations based on the changes in  https://github.com/kyazdani42/nvim-tree.lua/commit/a864b80bafeba6ea17d1a55e4c6f5e73f189f6bb

### DIFF
--- a/lua/core/nvimtree.lua
+++ b/lua/core/nvimtree.lua
@@ -57,16 +57,12 @@ function M.setup()
   end
   local g = vim.g
 
-  for opt, val in pairs(lvim.builtin.nvimtree) do
-    g["nvim_tree_" .. opt] = val
-  end
-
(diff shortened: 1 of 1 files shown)


Tastyep (approved):

lua/core/nvimtree.lua
```diff
@@ -57,16 +57,12 @@ function M.setup()
   end
   local g = vim.g
 
-  for opt, val in pairs(lvim.builtin.nvimtree) do
```
chase, line 60: Icons aren't part of the changes introduced into nvim-tree, so removing this causes the icon customizations done above to fail

abzcoding: yes 😅

PR #1628, opened 2021-09-26
[Bugfix]: Use local install in Github actions

### LINKED ISSUE #1621: [Bug]: Install script cannot be used in Github actions (outside this project)
### Problem description

The installer script in `utils/installer/install.sh` cannot be used in any external Github actions workflows. 

For example, I run tests for my dotfiles on Github actions, and those fail because of the following lines in the script:
https://github.com/LunarVim/LunarVim/blob/64a5964b545ec7f3f78f725ec96cba553a496d9f/utils/installer/install.sh#L83-L88

From what I could make out it is only used here:
https://github.com/LunarVim/LunarVim/blob/64a5964b545ec7f3f78f725ec96cba553a496d9f/.github/workflows/install.yaml#L34-L37

IMO this check and exit criteria should be moved off to an internal file (it is pretty small) or enabled by a CLI argument, which is explicitly called in the `.github/workflows/install.yaml`.

### LunarVim version

rolling (64a5964b545ec7f3f78f725ec96cba553a496d9f)

### Neovim version (>= 0.5)

0.5.0

### Relevant log output

```shell
See here -- https://github.com/ssbanerje/dotfiles/runs/3709959451#step:4:360
```

### Screenshots

_No response_

### I have

- [X] Read the readme
- [X] Read the wiki

ssbanerje:
## Description

This PR adds a CLI argument `-l` or `--local` in the `install.sh` script to install the local repo of lunar vim instead of cloning one from Github.

This option is enabled in the the `.github/workflows/install.yaml` action. This allows the the install script to be used in GitHub workflows outside this repository.

Fixes #1621 

## How Has This Been Tested?

This repos `.github/workflows/install.yaml` action.


kylo252 (changes requested):
I don't get what this is supposed to accomplish, since we're already using a local copy of the repo by adding a symlink to the entire directory:
https://github.com/LunarVim/LunarVim/blob/94a166594e9e4913a0b89f6bfd400c536a596dd4/.github/workflows/install.yaml#L36
Also do keep in mind that `setup_lvim` and `clone_lvim` are two separate commands for this specific reason.

A better change would be to remove that hard-coded symlink into the new local switch and figure out if you're having problems with the `LV_BRANCH` assignment for `GITHUB_ACTIONS`.

ssbanerje:
When running this script in Github Actions in another repo (other than this one). The script doesn't find the local copy copy and errors out. 

It seems that the assumption for the file is that this script will only be run in the Github actions/workflow of this repo. It only runs the the `install_packer` and `setup_lvim` actions. Which totally works in this repos `install.yaml` workflow. But as the other functions calls from the default install script is never done, it doesn't work outside this repo.

See here:
https://github.com/LunarVim/LunarVim/blob/64a5964b545ec7f3f78f725ec96cba553a496d9f/utils/installer/install.sh#L83-L88

This PR essentially makes that behavior explicit. It only does this when the `-l` option is enabled. Otherwise it goes for the default install script behavior.

kylo252:
> @kylo252:
> 
> When running this script in Github Actions in another repo (other than this one). The script doesn't find the local copy of Lvim and errors out. The problem is that the install script sort of assumes that all the commands in `install.yaml` will be done if the script runs in Github actions. Which is not the case when this script is run in another repo's Github actions.

Are you sure about that? Here a test I tried, see commit https://github.com/LunarVim/LunarVim/commit/f6c1ed1fa6fdd47829433c7313e3ea208b2ddec9,  
and the corresponding log: https://github.com/kylo252/LunarVim/runs/3747362126#step:4:19


> This PR essentially makes that behavior explicit. It only does this (behavior described above) when the -l option is enabled. Otherwise it goes for the default install script behavior. So it works outside.

It's not explicit until you actually move that `ln -s ..` inside the installer.

Again, I think these are two separate problems.

ssbanerje:
Yeah I did try that. The thing is it errors out like so -- https://github.com/ssbanerje/dotfiles/runs/3709959451#step:4:360. This is a trace from me using the installer script for my dotfiles.

 > It's not explicit until you actually move that `ln -s ..` inside the installer.

I can definitely do that. I can move in the logic from the `install.yaml` to the install script. That way the script will be the only thing one needs to run even after a manual `git clone`.

ssbanerje:
@kylo252 I have brought in the `ln` logic into the install script and added a Dockerfile to test the local install

utils/installer/install.sh
```diff
@@ -82,9 +90,6 @@ EOF
 
   if [ -n "$GITHUB_ACTIONS" ]; then
     LV_BRANCH="${GITHUB_REF##*/}"
```
kylo252, line 92: We can delete this as it's no longer needed.

utils/installer/install.sh
```diff
@@ -38,6 +38,8 @@ function usage() {
   echo "Options:"
   echo "    -h, --help    Print this help message"
   echo "    -y, --yes     Yes for all choices (Install NodeJS, Python, Rust dependencies)"
+  echo "    -n, --no      No for all choices (Install NodeJS, Python, Rust dependencies)"
```
kylo252, line 41: Let's rename these to `--[no]-install-dependencies` where `$INSTALL_DEPDENDENCIES` is set to true by default. I think it would be a lot less confusing for everyone. Then you can call the scripts like so:
`bash install.sh --local --no-install-dependencies`

utils/installer/install.sh
```diff
@@ -303,6 +312,16 @@ function clone_lvim() {
...
 }
 
+function link_local_lvim() {
+  echo "Linking local LunarVim repo"
+  mkdir -p "$LUNARVIM_RUNTIME_DIR"
+  local BASEDIR
+  BASEDIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)"
+  BASEDIR="$(dirname -- "$(dirname -- "$BASEDIR")")"
+  echo "   - $BASEDIR -> $LUNARVIM_RUNTIME_DIR/lvim"
+  ln -s "$BASEDIR" "$LUNARVIM_RUNTIME_DIR/lvim"
```
kylo252, line 322: You might want to add a `-f` flag since this can now be used outside of the CI.

kylo252 (changes requested):

ssbanerje:
@kylo252 Made the changes you asked for.

This commit linter is annoying! Cant use upper case :(

kylo252:
> This commit linter is annoying! Cant use upper case :( I imagine everything will be squashed in the end anyways.

Yeah I enabled just this morning, and this is one of the first PRs to go through it :)
We had a discussion about the lint rules and decided to align them with Neovim Core: https://github.com/LunarVim/LunarVim/blob/rolling/CONTRIBUTING.md#pull-requests-prs

> @kylo252 Made the changes you asked for.

Thank you! I have made some minor refactoring since I figured it might be faster than explaining. I hope it was okay with you.

Only thing that seems iffy is that I needed to run `PackerSync` manually, not sure why that happened. Here's the test: 

```shell
LUNARVIM_RUNTIME_DIR="/tmp/lunarvim-test" bash utils/installer/install.sh --local --no-install-dependencies
```

utils/docker/Dockerfile.remote
```diff
@@ -1,25 +1,27 @@
+# To run this file execute:
+# docker build -f Dockerfile.remote . -t Lunarvim:remote
+
 FROM ubuntu:latest
 
-ARG NEOVIM_RELEASE=v0.5.0
+# Build argument to point to correct branch on GitHub
 ARG LVBRANCH=rolling
```
kylo252, line 7: Let's rename this to `LV_BRANCH` to align with the installer.

utils/docker/Dockerfile.remote
```diff
@@ -1,25 +1,27 @@
...
-  chmod u+x /tmp/nvim.appimage && \
-  /tmp/nvim.appimage --appimage-extract && \
-  mv squashfs-root /usr/local/neovim && \
-  ln -s /usr/local/neovim/usr/bin/nvim /usr/bin/nvim && \
   apt clean && rm -rf /var/lib/apt/lists/* /tmp/* && \
-  LVBRANCH=${LVBRANCH} curl -s https://raw.githubusercontent.com/lunarvim/lunarvim/rolling/utils/installer/install.sh | bash -s -- -y
-
-ENV PATH="/root/.local/bin:/root/.cargo/bin:/root/.npm-global/bin${PATH}"
+  curl -s https://raw.githubusercontent.com/lunarvim/lunarvim/${LVBRANCH}/utils/installer/install-neovim-from-release | bash && \
+  LVBRANCH=${LVBRANCH} curl -s https://raw.githubusercontent.com/lunarvim/lunarvim/${LVBRANCH}/utils/installer/install.sh | bash -s -- -y
```
kylo252, line 21: Use `curl -LSs`

kylo252, line 21: Set `--no-install-dependencies` and install them manually if you need them. It will have failed on the rust ones anyway.

kylo252 (changes requested):

utils/installer/install.sh
```diff
@@ -104,10 +112,6 @@ EOF
     echo "Would you like to install lunarvim's Rust dependencies?"
     read -p "[y]es or [n]o (default: no) : " -r answer
     [ "$answer" != "${answer#[Yy]}" ] && install_rust_deps
-  elif [ "$ARGS_INSTALL_NONINTERACTIVE" = "y" ]; then
```
ssbanerje: @kylo252: Shouldn't `--install-dependencies` always install the dependencies and not ask the user. 

I think it might be nice to still keep the non-interactive "install dependency" option like before.

utils/docker/Dockerfile.remote
ssbanerje: Fixed in 7549c02

utils/docker/Dockerfile.remote
ssbanerje: Fixed in 7549c02

ssbanerje: The container installs the default `rustc` and `cargo` that comes with the `ubuntu:latest`'s package manager. That way it can compile those packages. 

In any case I think ubuntu has a `ripgrep` package (https://packages.ubuntu.com/hirsute/ripgrep) and `fd` (https://packages.ubuntu.com/hirsute/fd-find). Those can be used in the container.

But I could not find ones for `lazygit`. Seems like the authors have stopped maintaining the PPA. 

Maybe the solution is to use an Arch image instead of ubuntu for the container. That way the packages can stay up to date

utils/installer/install.sh
kylo252: As I mentioned earlier, these dependencies are not that hard to manage from the outside, so I don't see the need to complicated things just for the sake of installing them.

kylo252:
Thank you!

PR #1634, opened 2021-09-27
feat: add LvimUpdate command

kylo252:
# Description

Add a convenient way to update LunarVim.

- use it directly `:LvimUpdate` 
- use it from the commandline `lvim +LvimUpdate`


lua/core/commands.lua
```diff
@@ -12,6 +12,8 @@ M.defaults = {
   ]],
   -- :LvimInfo
   [[command! LvimInfo lua require('core.info').toggle_popup(vim.bo.filetype)]],
+  [[ command! LvimCacheReset lua require('utils').lvim_cache_reset() ]],
```
Tastyep, line 15: Do you agree renaming these two to `reset_lvim_cache` and `update_lvim` ?

lua/utils/init.lua
```diff
@@ -152,7 +152,13 @@ function utils.lvim_cache_reset()
   require("plugin-loader"):cache_reset()
 end
 
-vim.cmd [[ command! LvimCacheReset lua require('utils').lvim_cache_reset() ]]
+--FIXME: this is still broken
```
Tastyep, line 155: What is wrong ?

utils/installer/update_lvim.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+LUNARVIM_RUNTIME_DIR="${LUNARVIM_RUNTIME_DIR:-"$HOME/.local/share/lunarvim"}"
+LUNARVIM_CONFIG_DIR="${LUNARVIM_CONFIG_DIR:-"$HOME/.config/lvim"}"
+
+function msg() {
```
Tastyep, line 6: Let's not implement the same function multiple times, create a new file and source it here

utils/installer/update_lvim.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+LUNARVIM_RUNTIME_DIR="${LUNARVIM_RUNTIME_DIR:-"$HOME/.local/share/lunarvim"}"
```
Tastyep, line 3: Since these variables are 'defined' when running the setup script, maybe we should instead generate this file

utils/installer/update_lvim.sh
```diff
@@ -0,0 +1,32 @@
...
+
+  if ! git -C "$LUNARVIM_RUNTIME_DIR/lvim" diff-index --quiet "@{upstream}" 2>/dev/null; then
+    if ! git -C "$LUNARVIM_RUNTIME_DIR/lvim" merge --ff-only --progress 2>/dev/null; then
+      msg "Error: unable to guarantee data integrity while updating your branch.\nPlease pull the changes manually instead."
+      exit 1
+    fi
+  fi
+
+  msg "Clearing up old startup cache"
+
```
Tastyep, line 26: 26 and 28 are unnecessary newlines

Tastyep (changes requested):

utils/installer/update_lvim.sh
kylo252: It's a bit tricky to source it in the installer when it's being curl'd in 😅

lua/utils/init.lua
kylo252: all good now :)

utils/installer/update_lvim.sh
kylo252: I was able to remove that file entirely now and implement the functionality in Lua.

PR #1637, opened 2021-09-28
[Feature] Handlers cleanup

abzcoding:
# Description

Cleanup the way we handle diagnostics, needs to be tested before the merge

### DIFF
--- a/lua/lsp/handlers.lua
+++ b/lua/lsp/handlers.lua
@@ -3,72 +3,42 @@
 local M = {}
 
 function M.setup()
-  local config = { -- your config
-    virtual_text = lvim.lsp.diagnostics.virtual_text,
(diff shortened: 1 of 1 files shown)


kylo252:
I have a suggestion, we should just use: 
```lua
["gl"] = {
  "<cmd>lua vim.diagnostic.show_position_diagnostics({source = 'always'}) <CR>",
  "Show position diagnostics",
},
```

And maybe revert the rest of the changes to pre-0.6, we can also keep the compatibility even more minimal by having a fallback to the default `show_line_diagnostics()` without any extra modifications.

abzcoding:
sounds good to me 👍

abzcoding:
i'm closing this in favor of getting rid of our custom implementation at once!

PR #1638, opened 2021-09-28
fix: can't open lvim log file cause by `lvim.log.viewer.cmd`

younger-1:
When I try to open log file by pressing `<leader>Lld`, nothing happens.

I found it was caused by this line:

https://github.com/LunarVim/LunarVim/blob/238bbf18ac4e35152be06736c75e94a81e3382e8/lua/config/defaults.lua#L23

`lvim.log.viewer.cmd` is not executable in this case. Use `less` instead of `lnav` will solve it

### DIFF
--- a/lua/config/defaults.lua
+++ b/lua/config/defaults.lua
@@ -20,7 +20,7 @@ lvim = {
     level = "warn",
     viewer = {
       ---@usage this will fallback on "less +F" if not found
-      cmd = "less +F",
+      cmd = "lnav",
       layout_config = {
         ---@usage direction = 'vertical' | 'horizontal' | 'window' | 'float',
         direction = "horizontal",
--- a/lua/config/defaults.lua
+++ b/lua/config/defaults.lua
@@ -20,7 +20,7 @@ lvim = {
     level = "warn",
     viewer = {
       ---@usage this will fallback on "less +F" if not found
-      cmd = "less",
+      cmd = "less +F",
       layout_config = {
         ---@usage direction = 'vertical' | 'horizontal' | 'window' | 'float',
         direction = "horizontal",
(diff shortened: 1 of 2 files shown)


abzcoding:
cc @kylo252

lua/config/defaults.lua
```diff
@@ -20,7 +20,7 @@ lvim = {
     level = "warn",
     viewer = {
       ---@usage this will fallback on "less +F" if not found
-      cmd = "lnav",
+      cmd = "less",
```
kylo252, line 23: This is not the correct place to change it. 

There's supposed to be a fallback in `toggle_log_viewer` but it disappeared somehow. It should also be `less +F` or `tail -f`.

kylo252 (changes requested):

lua/config/defaults.lua
```diff
@@ -20,7 +20,7 @@ lvim = {
     level = "warn",
     viewer = {
       ---@usage this will fallback on "less +F" if not found
-      cmd = "lnav",
+      cmd = "less +F",
```
kylo252, line 23: Sorry I didn't make it clear. It's better to set it like this:

```lua
local log_viewer = lvim.log.viewer.cmd
if vim.fn.executable(log_viewer) ~= 0 then
	log_viewer = 'less +F'
end
local cmd = log_viewer .. " " .. logfile
local term_opts = vim.tbl_deep_extend("force", lvim.builtin.terminal, {
    cmd = cmd,
	--- rest of the arguments
```

in here:
https://github.com/LunarVim/LunarVim/blob/6b79e00cb04bf889fc30b84482d6db6e951ca9e9/lua/core/terminal.lua#L115-L116

kylo252 (changes requested):

lua/config/defaults.lua
younger-1: Sorry for my silly mistake. Solved now.

kylo252:
Thank you!

PR #1648, opened 2021-09-30
[Feature]: Conditional loading of `which_key` keymaps

### LINKED ISSUE #1422: [Feature]: Call lvim.builtin.which_key.on_config_done after all core logic plugins have been loaded

### LINKED ISSUE #1755: [Bug]:  Can't overwrite whichkey binding for nvim tree
### Problem description

In config.lua:

` lvim.builtin.which_key.mappings["e"] = { "<cmd>RnvimrToggle<CR>", "Ranger" }`

Does not work, the default keybind `"<cmd>NvimTreeToggle<CR>", "Explorer"` is not overwritten. 
I noticed that this binding is defined in `nvimtree.lua` while most keybinds are defined in `which-key.lua`. Moving this definition to `which-key.lua` fixes the problem.

It seems that there are a few other default which key bindings being defined outside `which-key.lua`, namely dashboard and dap bindings.

### LunarVim version

rolling

### Neovim version (>= 0.5)

NVIM v0.5.1. LuaJIT 2.0.5

### Relevant log output

_No response_

### Screenshots

_No response_

### I have

- [X] Read the readme
- [X] Read the wiki

ssbanerje:
## Description

This PR makes the `<leader>d` keymap consistent with the other `<leader>...` keymaps defined by default in LunarVim.

The current implementation loads the `<leader>d` which-key map in the "DAP" plugin config. Hence those mappings are not defined before the users `config.lua` is loaded and cannot be modified in the same way as other keymaps.

The PR follows a discussion with @abzcoding on Matrix.

Fixes #1472
Fixes #1422
Fixes #1755

## Discussion

Edit: See https://github.com/LunarVim/LunarVim/pull/1648#discussion_r720327204 & https://github.com/LunarVim/LunarVim/pull/1648#issuecomment-932540744 for updated proposal.

~~There is a corner case where the user modifies these keymaps, but the corresponding core plugin is inactive.~~

~~Currently, in this PR (as of 7b3c62f), this is being checked by using the `name` of the mapping. For example, the condition for deleting the DAP keymap is:~~

```lua
not lvim.builtin.dap.active and lvim.builtin.which_key.mappings.d.name = "Debug"
```

~~That condition might not capture all the edits that the user might have done to this mapping entry. Any suggestions for a better condition are welcome.~~

## How Has This Been Tested?

Toggling `active` on `dap` and `nvimtree` removes the corresponding keybindings.


lua/core/which-key.lua
```diff
@@ -236,6 +249,19 @@ M.config = function()
   }
 end
 
+-- Clean up mappings for inactive core plugins
+local function clean_up_inactive_mappings()
+  -- DAP
+  if not lvim.builtin.dap.active and lvim.builtin.which_key.mappings.d.name == "Debug" then
+    lvim.builtin.which_key.mappings.d = nil
+  end
```
kylo252, line 257: This will backfire sadly, because you're restricting what `d` and `e` can used for. You might end up removing more than necessary, or even removing the user's own configuration.

A better approach is either:
a) fetch the config from each of those modules, but only assign it if the plugin is active.
b) keep them in this file but don't add them unless the plugin is installed.

kylo252 (changes requested):

lua/core/which-key.lua
ssbanerje: @kylo252 

Yeah, I did try that first. But it seems that the problem is loading the keymaps in the plugin's `config()` does not work universally. 

Here is an example: if the user remaps the `<leader>d` to be a set of mappings dealing with say docker, then loading `d` selectively if `lvim.builtin.dap.active == true` might not be the correct outcome.

Maybe going forward, it might make sense to have some kind of metatable hacks so that each of the mappings gets a lambda to check when a mapping should be enabled or not. Then when they are loaded, we create a wrapper around the `register()` to check and register.

ssbanerje:
@kylo252 Have a look c558167.

Here is a prototype implementation of conditionally enabling the which_key bindings. This might need some work before its completely ready.

This keeps backward compatibility, in that all user config changes to `lvim.builtin.which_key.(v)mapping` will work as it did before. Those are assumed to be always enabled remaps.

New "conditional" remaps can be created using a new function `lvim.builtin.which_key.remap`. This essentially sets the metatable entries for the mapping so that we can transparently check the enable condition in the `clean_up_inactive_mappings()` function.

ssbanerje:
Hi @kylo252! I think all the changes for this one are in place now. Could you have a look at it?

Tastyep:
Hello,

The way I see it, this is a good example of how it should **not** be done.
Doing it like this you strongly couple which-key with all other plugins. For example, if you want to remove one of the builtins, as a consequence of this PR you would now have to update which-key's file.

Also, the conditional logic for loading a bultin is already handled by it's `setup` function (which, BTW, should be named `configure`). Here you end up duplicating that logic with the conditional functions.

What I think is the right solution is to expose a method from which-key (core/which-key) to register mappings.
This way, builtins would conditionally call this method from their `setup` function (again `configure`). I say conditionally because that method is only called when the builtin is loaded.
From a dependency point of view, I think it's more logical and maintainable to have a "many-to-one" than a "one-to-many" dependency.

ssbanerje:
@Tastyep @abzcoding @ChristianChiarulli
Bump. Did you guys have a chance to look at this one?

abzcoding:
tyvm for your efforts on this, the keybindings has changed a lot, so i don't think this is relevant anymore

PR #1649, opened 2021-09-30
fix(installer): fix ARGS_LOCAL condition

WhatTheFar:
# Description

`install.sh` doesn't work properly. It always link local lvim regardless of the `ARGS_LOCAL` value. 

This PR fixes the if-else condition to correctly check `ARGS_LOCAL`.

## How Has This Been Tested?

- Uninstall lvim 
- Run the installer

Uninstall lvim
```
$ rm -rf ~/.local/share/lunarvim

$ sudo rm /usr/local/bin/lvim

$ rm ~/.local/bin/lvim

$ rm -rf ~/.local/share/applications/lvim.desktop
```

Run the installer
```
$ ./utils/installer/install.sh
```

### DIFF
--- a/utils/installer/install.sh
+++ b/utils/installer/install.sh
@@ -135,7 +135,7 @@ EOF
     echo "Updating LunarVim"
     update_lvim
   else
-    if [ -n "$ARGS_LOCAL" ]; then
+    if [ "$ARGS_LOCAL" -eq 1 ]; then
       link_local_lvim
     else
       clone_lvim


ssbanerje:
Yup this got introduced in #1628.

I had told @kylo252 about this in the other thread.

kylo252:
> @kylo252 @WhatTheFar Just repeating my last comment from the other thread. Maybe it makes sense to add the remote install (to the current local install) into the CI so that this kind of bug gets caught there.

Yeah for sure, we need to test full install separately.

PR #1670, opened 2021-10-03
fix: add deprecation notice for lvim.lang.foo.lsp

kylo252:
# Description

`lvim.lang.FOO.lsp` is no longer supported after #1584.
 You can either use `:NlspConfig` for most of the settings you might need, or override the setup by adding an entry to `lvim.lsp.override = { "FOO" }`.

## How Has This Been Tested?

You should see a warning if you have any `lvim.lang.FOO.lsp` settings in your `config.lua`.

### DIFF
--- a/lua/config/init.lua
+++ b/lua/config/init.lua
@@ -124,6 +124,20 @@ function M:init(opts)
   require("lsp.manager").init_defaults(supported_languages)
 end
 
+local function deprecation_notice()
+  for lang, entry in pairs(lvim.lang) do
+    if not vim.tbl_isempty(entry["lsp"]) then
+      local msg = string.format(
+        "Deprecation notice: [lvim.lang.%s.lsp] setting is no longer supported. See https://github.com/LunarVim/LunarVim#breaking-changes",
+        lang
+      )
+      vim.schedule(function()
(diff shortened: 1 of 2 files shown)


abzcoding:
this is the first thing i see after using this branch
Screen Shot 2021-10-03 at 11 20 49 PM

lua/config/init.lua
```diff
@@ -124,6 +124,20 @@ function M:init(opts)
   require("lsp.manager").init_defaults(supported_languages)
 end
 
+local function deprecation_notice()
+  for lang, entry in pairs(lvim.lang) do
```
abzcoding, line 128: what happens if `lvim.lang` is nil ?

kylo252:
> this is the first thing i see after using this branch Screen Shot 2021-10-03 at 11 20 49 PM

It looks like the list of languages that I'm using is slightly off. I'm gonna add a default value to it so it doesn't panic.

lua/config/init.lua
kylo252: That shouldn't happen because I'm setting it earlier in the `init` function.

abzcoding (approved):
thank you

PR #1679, opened 2021-10-04
feat: better kubernetes support

abzcoding:
# Description

in order to get these:
Screen Shot 2021-10-04 at 11 26 29 AM
Screen Shot 2021-10-04 at 11 27 04 AM

### DIFF
--- a/lua/lsp/providers/yamlls.lua
+++ b/lua/lsp/providers/yamlls.lua
@@ -1,58 +1,30 @@
 local opts = {
-    settings = {
-      yaml = {
-        hover = true,
-        completion = true,
-        validate = true,
-        schemaStore = {
-          enable = true,
-          url = "https://www.schemastore.org/api/json/catalog.json",
-        },
-        schemas = {
-          kubernetes = {
-            "daemon.yaml",
-            "manager.yaml",
-            "restapi.yaml",
-            "role.yaml",
-            "role_binding.yaml",
-            "*onfigma*.yml",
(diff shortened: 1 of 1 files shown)


lua/lsp/providers/yamlls.lua
```diff
@@ -0,0 +1,58 @@
...
+          "inventories/*.yml",
+          "*nventori*/*.yml",
+          "staging.yml",
+          "production.yml",
+        },
+        ["https://json.schemastore.org/ansible-playbook.json"] = {
+          "playbook.yml",
+          "playbook.yaml",
+          "main.yml",
+        },
```
kylo252: I'm not sure if it was available before, but there's now support for [ansiblels](https://github.com/williamboman/nvim-lsp-installer/blob/2f9a3800b976fb88f99a805dbb53f6ff6290a7ab/lua/nvim-lsp-installer/servers/ansiblels/init.lua). 

It would be good if you can give it a try if you already use ansible just to make sure there's conflict, since yamlls is set to validate the schemas.

kylo252 (approved):

lua/lsp/providers/yamlls.lua
abzcoding: yeah it was not there beforehand, let me check, we need the kubernetes stuff though

abzcoding: you are right, we don't need the ansible stuff anymore

PR #1689, opened 2021-10-04
fix: disable unintended luasnip jump-to-prev

### LINKED ISSUE #1645: [Bug]: LuaSnip prevent jumping out of snippet when jumping-next-prev
### Problem description

When we don't complete a snippet, exit `insert` mode, open new line with `o`, press `tab`, it's jump to the next location in the old snippet. We can prevent this with `region_check_events` option. But I don't know how to add this. hope someone can fix it

https://github.com/L3MON4D3/LuaSnip/issues/116

https://github.com/L3MON4D3/LuaSnip/issues/170

### LunarVim version

lastest

### Neovim version (>= 0.5)

NVIM v0.6.0-dev+176-g4b3ffde20

### Relevant log output

_No response_

### Screenshots

_No response_

### I have

- [X] Read the readme
- [X] Read the wiki

kylo252:
- pressing enter will no longer activate luasnip.
- exiting insert-mode will unlink the current snippet.

Fixes #1645

# Description

Fixes #(issue)

## How Has This Been Tested?

- Run command `:mycommand`
- Check logs
- ...

### DIFF
--- a/lua/core/cmp.lua
+++ b/lua/core/cmp.lua
@@ -4,7 +4,6 @@ local check_backspace = function()
   local col = vim.fn.col "." - 1
   return col == 0 or vim.fn.getline("."):sub(col, col):match "%s"
 end
-
 local function T(str)
   return vim.api.nvim_replace_termcodes(str, true, true, true)
 end
@@ -139,11 +138,7 @@ M.config = function()
       ["<C-e>"] = cmp.mapping.close(),
       ["<CR>"] = cmp.mapping(function(fallback)
         if not require("cmp").confirm(lvim.builtin.cmp.confirm_opts) then
-          if luasnip.jumpable() then
-            vim.fn.feedkeys(T "<Plug>luasnip-jump-next", "")
-          else
-            fallback()
-          end
+          fallback()
         end
       end),
     },
--- a/lua/core/autocmds.lua
+++ b/lua/core/autocmds.lua
@@ -14,6 +14,11 @@ lvim.autocommands = {
       "nnoremap <silent> <buffer> q :q<CR>",
     },
     {
+      "InsertLeave",
+      "*",
+      "lua require'luasnip'.unlink_current()",
+    },
+    {
       "FileType",
       "lsp-installer",
       "nnoremap <silent> <buffer> q :q<CR>",


lua/core/cmp.lua
```diff
@@ -139,11 +138,7 @@ M.config = function()
       ["<C-e>"] = cmp.mapping.close(),
       ["<CR>"] = cmp.mapping(function(fallback)
         if not require("cmp").confirm(lvim.builtin.cmp.confirm_opts) then
-          if luasnip.jumpable() then
-            vim.fn.feedkeys(T "<Plug>luasnip-jump-next", "")
-          else
-            fallback()
-          end
+          fallback()
```
abzcoding, line 141: users can disable this part themselves (more info at https://github.com/LunarVim/LunarVim/pull/1523)
so i don't think we should remove it

abzcoding (changes requested):

lua/core/cmp.lua
kylo252: It should definitely not be on by default, it doesn't even come from LuaSnip's own suggested configuration.

Take a look at what this causes and how the cursor is behaving:

Cmpw2MKp0Y

abzcoding: i know that, but this is what @ChristianChiarulli added, maybe we should ask him why

chase: LuaSnip shouldn't be flagged as jumpable if the snippet has been exited, in my opinion.

The reason for this implementation was due to this default:
If you are in a snippet and select something from autocomplete and hit enter it will select the item, then you need to hit tab again in order to jump to the next part of the snippet. I thought the correct behavior would be to select the item and move to the next part of the snippet.

chase: It seems that `unlink_current()` isn't actually the right function for this use case, since some snippets aren't linear (like function in the example above). Instead what we probably want is almost: `exit_out_of_region(require("luasnip").session.current_nodes[vim.api.nvim_get_current_buf()])`, which [should already be handled by config](https://github.com/L3MON4D3/LuaSnip/blob/94bf4e7cc56d03d82c4058df7090f9d982aa264d/lua/luasnip/config.lua#L102).

The problem is that you can't exit out of the region using this method, [because it does a bounds check to see if it is not inside the region of the node first](https://github.com/L3MON4D3/LuaSnip/blob/7ae2627a8088f879272f85b9ac97d346ec554521/lua/luasnip/init.lua#L341-L347). If it didn't, one could simply use this:

``` lua
require("luasnip").config.set_config({
  history = true,
  region_check_events = 'InsertLeave'
})
```

The reason why `InsertEnter` is a poor choice for `region_check_events` here is because a lot of text manipulation happens outside of Insert mode. The reason why `Cursor*` events are probably poor choices is because this would then run every time the cursor moves, but it works.

With all that, we return to the problem of not being able to return to the snippet using tab while _within the bounds of the snippet_: https://github.com/LunarVim/LunarVim/issues/1645#issuecomment-930980836

I think the correct solution is to add a lines bounds check for expand_or_jumpable and jumpable, either by adding methods to upstream like `in_region_jumpable` and `in_region_expandable` and `in_region_expand_or_jumpable`, or just wrapping them ourselves and using those instead.

Untested, but this is roughly what you'd write:
```lua

local ls = require("luasnip")
local current_buf = vim.api.nvim_get_current_buf
local win_get_cursor = vim.api.nvim_win_get_cursor

local function inside_snippet()
  local node = ls.session.current_nodes[current_buf()]
  if not node then
    return false
  end

  local snip_begin_pos, snip_end_pos = node.parent.snippet.mark:pos_begin_end()
  local pos = win_get_cursor(0)
  return pos[1] >= snip_begin_pos[1] and pos[1] <= snip_end_pos[1]
end

local function in_region_jumpable()
  return inside_snippet() and ls.jumpable()
end

local function in_region_expand_or_jumpable()
  return inside_snippet() and ls.expand_or_jumpable()
end
```

kylo252:
Superseded by #1708

PR #1692, opened 2021-10-04
[Feature]: Add possibility to override plugin configuration

epontan:
# Description

It can be useful be able to override the default plugin configuration
from the user configuration. One example is to be able to change at what
event a plugin should be loaded on.

Merge all the plugin configurations by keying on the plugin name and let
the configurations down the chain take precedence by overriding the
configuration. The order of which the plugins are loaded is kept in the
order they are specified to keep it deterministic and consistent.

## How Has This Been Tested?

My use case was to make `gitsigns.nvim` load on startup when not starting vim with a file as argument (not only when BufRead event is triggered).
I use the gitsigns to review all the changes of last commit by changing its base.

In my `config.lua` I override the default plugin configuration like this:
```lua
lvim.plugins = {
  {
    "lewis6991/gitsigns.nvim",
    config = function()
      require("core.gitsigns").setup()
    end,
    disable = not lvim.builtin.gitsigns.active,
  },
}
```
This makes the plugin available at start without having to open a file first.


kylo252:
I like the idea of having that ability, and I think the use-case is valid, but I'm not sure that the workflow is that intuitive. 

It might be better to expose the default plugins entries instead, we can index them by the name of the plugin itself. Alternatively, we can expose the `event` attribute, but that also leaves other ones that are supported by packer.

cc: @Tastyep

Issue #1644, opened 2021-09-29
[Bug]: Automatic config reload does not always work


ssbanerje:
## Problem 1
Running `:lua require("utils").reload_lv_config()` resets some user configuration settings to the LunarVim defaults.
For example, user configuration to the lualine builtin is removed.

## Problem 2
This doesn't work when `config.lua` is a sym link.
https://github.com/LunarVim/LunarVim/blob/94a166594e9e4913a0b89f6bfd400c536a596dd4/lua/core/autocmds.lua#L41

Maybe something like this would work better: 
```lua
  {
    "BufWritePost",
    vim.fn.resolve(require("config").path),
    "lua require('utils').reload_lv_config()"
  },
```

## Problem 3
This file does not exits in the default config:
https://github.com/LunarVim/LunarVim/blob/94a166594e9e4913a0b89f6bfd400c536a596dd4/lua/core/autocmds.lua#L80-L83

### LunarVim version

rolling @ 87d3610

### Neovim version (>= 0.5)

0.5.1

### Relevant log output

```shell
» lvim -v
NVIM v0.5.1
Build type: Release
LuaJIT 2.1.0-beta3
Compilation: clang -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -DNVIM_TS_HAS_SET_MATCH_LIMIT -O2 -DNDEBUG -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/tmp/neovim-20210927-1441-demfsz/neovim-0.5.1/build/config -I/tmp/neovim-20210927-1441-demfsz/neovim-0.5.1/src -I/usr/local/include -I/Library/Developer/CommandLineTools/SDKs/MacOSX11.sdk/usr/include -I/usr/local/opt/gettext/include -I/tmp/neovim-20210927-1441-demfsz/neovim-0.5.1/build/src/nvim/auto -I/tmp/neovim-20210927-1441-demfsz/neovim-0.5.1/build/include
Compiled by brew@BigSur

[...]

PR #1699, opened 2021-10-05
feat: add an independent lvim namespace

### LINKED ISSUE #1230: [Feature]: Completely reload all of LunarVim config as if Neovim restarted
### Feature motivation

There is no way to completely reload all of the LunarVim config without restarting Neovim, sequentially (sequence is important).

For example, if a new plugin is activated, then after the `PackerCompile` it is necessary to restart Neovim to see the effects, finally proceeding with `PackerInstall`.

More in general, what I'm looking for is a way to completely reset and refresh LunarVim settings as done during its startup.

Maybe something like this could work?:

Have a `util.lua` file that defines the following:

````lua
_G.load = function(file)
    require("plenary.reload").reload_module(file, true)
    return require(file)
end
````

Then, instead of writing `require(...)`, write `load(...)` and the only file to require is `util.lua`, which is also the first one.

Then have the following mapping defined:

````lua
util.nnoremap("<leader>sv", "<cmd>luafile $MYVIMRC<CR>")
````

Which reloads the whole config.

https://stackoverflow.com/questions/68415300/how-can-i-reload-my-neovim-config-written-in-lua-with-a-shortcut

_No response_

### Additional context

_No response_

### LINKED ISSUE #1479: [Bug]: Avoiding naming collisions with LunarVim Lua files
### Problem description

I recently ran into an issue that took me quite a while to figure out.

The configuration I'm importing from NeoVim is somewhat large, and I've always liked to split it into several files. Here's what my `config.lua` file looked like:

``` lua
require("settings")
require("keys")
require("plugins")
```

What I didn't realize the time was that there's already a `plugins.lua` file present in LunarVim. When I tried loading LunarVim, I'd get an obscure error because my `plugins.lua` file was earlier in the `runtimepath`. This would cause LunarVim to load the wrong file and fail.

For now, I renamed my plugins file to `local_plugins.lua` to avoid the problem.

Is there any way to avoid this type of naming collision? I understand this is likely a larger problem with NeoVim's Lua imports, but I'm wondering if there's a pragmatic solution. Would it make sense for LunarVim to namespace its files to avoid collisions?

### LunarVim version

Latest

### Neovim version (>= 0.5)

v0.5.0

### Relevant log output

```shell
N/A
```

### Screenshots

N/A

### I have

- [X] Read the readme
- [X] Read the wiki

kylo252:
# Description

- avoid name collisions with user's configuration
- fixes a lot of issues with reloading

Fixes #1479, #1644
Partial fixes #1230


kylo252:
> you might've forgotten these
> 
> https://github.com/LunarVim/LunarVim/blob/7fec97a59044a5fb6001b9f9f9365ee24b9bcca2/lua/lvim/lsp/config.lua#L33-L36

yeah, I've updated them in the structlog branch and forgot to backport the changes 😄 
I'm not sure if I should include the logger change yet.

kylo252:
> also there is this other thing, which i'm not sure about, i think sometimes the `lvim` namespace gets confused with the `lvim` object (the one in `lvim.*.*`) by the language server this is what happens after migrating to this branch Screen Shot 2021-10-10 at 8 51 56 PM

Strange, I thought that updating the workspace library is enough. Can you make sure you've updated this in your `lua-dev` setup?
https://github.com/LunarVim/LunarVim/pull/1699/files#diff-072b5a568c83434dda5a33ca67ecdcb46358ba94308236464a9dc8b1578bbff7

Also feel free to push that change here if you want. As I said, I think everyone would like it anyway.

abzcoding:
> Also feel free to push that change here if you want. As I said, I think everyone would like it anyway.

yeah my bad, i forgot to update this part in my config
```lua
            [require("lvim.utils").join_paths(get_runtime_dir(), "lvim", "lua")] = true,
```

abzcoding:
w8 don't merge it 😅

abzcoding:
@kylo252 we are not using the impatient plugin, all of them need to be namespaced as well

abzcoding:
Screen Shot 2021-10-10 at 9 03 05 PM

abzcoding:
also i get this if i run `LvimCacheReset`
Screen Shot 2021-10-10 at 9 07 22 PM

PR #1707, opened 2021-10-07
feat(lsp): handle user configuration in setup()

### LINKED ISSUE #1697: [Feature]: User override of LSP configurations
### Feature motivation

It would be nice to have a mechanism to use the `lsp.manager` to load a user-provided table into the LSP configuration.

The current mechanism of using the `lvim.lsp.override` does two things:
1. Removes the default lazy installation of the LS.
2. Removes the default configurations from `providers`.

For some user configuration changes, it might be good to retain (1) and just change (2). The current override mechanism sort of just nukes everything.

1. Define a user config table like `lvim.lsp.user_ls_config.<servername>`.
2. Have `lsp.manager.setup()` accept a user config table and force `tbl_deep_extend()` the  config like it does below for the default config: https://github.com/LunarVim/LunarVim/blob/7a939ad771403e20341c9465af66573d225300bb/lua/lsp/manager.lua#L56-L65

Some plugins also provide configuration parameters to the LSs to work (e.g., https://github.com/jakewvincent/texmagic.nvim). It might be nice to have (2) happen after the plugins load, and use lambdas in `lvim.lsp.user_ls_config.<servername>` to configure the servers correctly.

The only alternative now is to install the overridden servers manually. This is what I am using right now (based on  @abzcoding's overridden LSP loading style from [here](https://github.com/abzcoding/lvim/blob/8cb8e7e2bb3d2f65c8b6e6bcff24eb35ae993711/config.lua#L53-L55)). But this installs the server the first time the `config.lua` file is read. 

It might be possible to combine this with `autocmd`s to get the necessary result. But that seems like replicating effort, as the `lsp.manager.setup()` and the code above is almost identical. Also, for servers that are reused across multiple FTs (e.g. tailwind), this is just not ergonomic as it would mean having to find the relevant FTs. 

### Additional context

_No response_

kylo252:
# Description

- allow passing user configuration directly when calling `require('lsp.manager').setup("myserver", myopts)`. 
These options will take precedence over any pre-existing default configuration.
- allow calling the setup function for servers not supported yet by "nvim-lsp-installer"

Fixes #1697

## How Has This Been Tested?

You can try this simple test

```lua
lvim.lsp.on_init_callback = function ()
  print("hello")
end

require("lsp.manager").setup("stylelint_lsp", {
	on_attach = function()
		print("world")
	end,
})
```

You can also test a server that's not supported by "nvim-lsp-installer"

### DIFF
--- a/lua/bootstrap.lua
+++ b/lua/bootstrap.lua
@@ -1,6 +1,7 @@
 local M = {}
 
-local in_headless = #vim.api.nvim_list_uis() == 0
+package.loaded["utils.hooks"] = nil
+local _, hooks = pcall(require, "utils.hooks")
 
 ---Join path segments that were passed as input
 ---@return string
@@ -42,12 +43,23 @@ function _G.get_cache_dir()
   return lvim_cache_dir
 end
 
+---Get the full path to the currently installed lunarvim repo
+---@return string
+local function get_install_path()
+  local lvim_runtime_dir = os.getenv "LUNARVIM_RUNTIME_DIR"
+  if not lvim_runtime_dir then
+    -- when nvim is used directly
+    return vim.fn.stdpath "config"
+  end
+  return join_paths(lvim_runtime_dir, "lvim")
+end
+
 ---Get currently installed version of LunarVim
 ---@param type string can be "short"
 ---@return string
 function _G.get_version(type)
   type = type or ""
-  local lvim_full_ver = vim.fn.system("git -C " .. get_runtime_dir() .. "/lvim describe --tags")
+  local lvim_full_ver = vim.fn.system("git -C " .. get_install_path() .. " describe --tags")
 
   if string.match(lvim_full_ver, "%d") == nil then
     return nil
@@ -65,7 +77,7 @@ function M:init()
   self.runtime_dir = get_runtime_dir()
   self.config_dir = get_config_dir()
   self.cache_path = get_cache_dir()
-  self.repo_dir = join_paths(self.runtime_dir, "lvim")
+  self.install_path = get_install_path()
 
   self.pack_dir = join_paths(self.runtime_dir, "site", "pack")
   self.packer_install_dir = join_paths(self.runtime_dir, "site", "pack", "packer", "start", "packer.nvim")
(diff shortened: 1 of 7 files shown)


xeluxee:
I tried to put this in `config.lua` but it doesn't work

``` lua
require("lsp.manager").setup("clangd", {
  on_attach = function() print("clangd on_attach") end,
})
```

abzcoding:
> I tried to put this in `config.lua` but it doesn't work
> 
> ```lua
> require("lsp.manager").setup("clangd", {
>   on_attach = function() print("clangd on_attach") end,
> })
> ```

can you try this instead?
```lua
lvim.lsp.on_init_callback = function ()
  require("lsp.manager").setup("clangd", {
    on_attach = function() print("clangd on_attach") end,
  })
end
```

xeluxee:
> 
> can you try this instead?
> 
> ```lua
> lvim.lsp.on_init_callback = function ()
>   require("lsp.manager").setup("clangd", {
>     on_attach = function() print("clangd on_attach") end,
>   })
> end
> ```

@abzcoding it works but two language servers get attached to a single buffer

abzcoding:
> > can you try this instead?
> > ```lua
> > lvim.lsp.on_init_callback = function ()
> >   require("lsp.manager").setup("clangd", {
> >     on_attach = function() print("clangd on_attach") end,
> >   })
> > end
> > ```
> 
> @abzcoding it works but two language servers get attached to a single buffer

can you run `:LvimInfo` and `:LspInfo`

xeluxee:
Ok I've blacklisted clangd but now no client can't be attached.
My config in `config.lua`:
``` lua
lvim.lsp.automatic_servers_installation = false
lvim.lsp.override = {"clangd"}
lvim.lsp.on_init_callback = function()
  require("lsp.manager").setup("clangd", {
    cmd = {"clangd", "--background-index", "--header-insertion=never", "--cross-file-rename"},
  })
end
```

abzcoding:
maybe there is a better way for blacklisting, i'm a bit out of the loop though
 @kylo252 might be able to help

kylo252:
Oops I forgot that this isn't configurable: 
https://github.com/LunarVim/LunarVim/blob/85e7b7804a2c9d6847ff978c82bb3a3ac16d757f/lua/lsp/templates.lua#L18-L24

Feel free to open a ticket about it, and hopefully a PR!

Otherwise just delete that folder/file or point the `lvim.lsp.templates_dir` variable somewhere where you can have it version controlled so you don't lose your changes.

PR #1708, opened 2021-10-07
fix(cmp/autopairs): prevent out of bounds jump and re-enable jump after confirm

### LINKED ISSUE #1645: [Bug]: LuaSnip prevent jumping out of snippet when jumping-next-prev

chase:
# Description

Prevents LuaSnip from hijacking the cursor when pressing `<Tab>`, `<S-Tab>` or `<CR>` when snippets are left incomplete after InsertLeave by putting up a boundary guard. This still allows re-entering snippets using Tab or S-Tab.

By default autopairs completely destroys the `<CR>` handler setup in `cmp.lua`, so the configuration options to setup cmp (which we are doing already) are overridden.

This also switches `select = false` as the default, because I don't think most people actually want auto-selection on complete, but again, this can be configured as usual to do the opposite and is a single line change if maintainers disagree. The current behavior expects select = false.

Fixes #1645

## How Has This Been Tested?

- Open a lua file with sumneko LSP installed
- Enter Insert mode
- Type `for` and select the `for ... ipairs` snippet from LSP using `<CR>`.
- Type i and press `<CR>` it will not select from complete and will jump to the next part of the snippet
- Exit Insert mode
- Move above or below the snippet and insert a new line (`o`, `O`, etc.)
- Press `<Tab>`, `<S-Tab>`, and `<CR>` and note it does not enter the snippet
- Exit Insert mode
- Move the cursor back inside the region of the snippet, press `i`
- Press `<Tab>`, `<S-Tab>` and note that it navigates through the snippet

### DIFF
--- a/lua/core/cmp.lua
+++ b/lua/core/cmp.lua
@@ -127,8 +127,8 @@ M.config = function()
       select = false,
     },
     experimental = {
-      ghost_text = false,
-      native_menu = true,
+      ghost_text = true,
+      native_menu = false,
     },
     formatting = {
       kind_icons = {
@@ -203,8 +203,8 @@ M.config = function()
       ["<C-f>"] = cmp.mapping.scroll_docs(4),
       -- TODO: potentially fix emmet nonsense
       ["<Tab>"] = cmp.mapping(function()
-        if vim.fn.pumvisible() == 1 then
-          vim.fn.feedkeys(T "<down>", "n")
+        if cmp.visible() then
+          cmp.select_next_item()
         elseif luasnip.expandable() then
           luasnip.expand()
         elseif inside_snippet() and seek_luasnip_cursor_node() and luasnip.jumpable() then
@@ -221,8 +221,8 @@ M.config = function()
         "s",
       }),
       ["<S-Tab>"] = cmp.mapping(function(fallback)
-        if vim.fn.pumvisible() == 1 then
-          vim.fn.feedkeys(T "<up>", "n")
+        if cmp.visible() then
+          cmp.select_prev_item()
         elseif inside_snippet() and luasnip.jumpable(-1) then
           luasnip.jump(-1)
         else
@@ -236,7 +236,7 @@ M.config = function()
       ["<C-Space>"] = cmp.mapping.complete(),
       ["<C-e>"] = cmp.mapping.close(),
       ["<CR>"] = cmp.mapping(function(fallback)
-        if vim.fn.pumvisible() ~= 0 and cmp.confirm(lvim.builtin.cmp.confirm_opts) then
+        if cmp.visible() and cmp.confirm(lvim.builtin.cmp.confirm_opts) then
           return
         end
 


chase:
@maxxnino Could you give this branch a try to see if this works as you'd expect?

chase:
@kylo252 I understand this is frustrating, but we're much closer to a nice user experience than we were before. `<CR>` is what users expect after using autocomplete in a lot of other software.

This implementation not behaving the way you expect is my fault, not LuaSnip's. No need to scrap everything just yet.

> Hell, we can't even see an indicator that there are more place holders in the snippet.

LuaSnip uses the extmark system under the hood, if it doesn't already support highlighting the remaining placeholders it probably should be possible to assign them a highlight group.

That said, in my LunarVim it displays a preview of the snippet along with showing the number of stops.

---

It took me a few views to figure out why you felt the GIF represented incorrect behavior, it goes by quite fast. If you are able to record/post MP4, I could more easily pause and rewind.
If I understand correctly, instead of pressing tab or CR twice to jump through the snippet to get to the place you want, you leave the snippet and make a new line.
You then press CR but instead of making a new line like you expect, it jumps to the arguments position (which is the next section of the snippet)

The "if I leave, don't do that" option might be one approach, where we simply clear the current node on InsertLeave (not unlink, just set to nil). This means you can't re-enter the snippet, this is probably the quickest fix.

There are less destructive approaches as well, but of course there may be edge cases that we'll discover.
One approach I think best fits: further bounds checking within a snippet. Basically, add a relative jump and jumpable that advances to the section before or after the current cursor position. If it is after the bounds of the last section, <CR> should complete the snippet and fallback.
There might be some nested snippets where seeking the next jump goes backwards, like `${5: out ${4: in}}`. Handling this might take some extra work, but I'll consider that an uncommon edge case unless someone can bring up a commonly used snippet that does that.

Another approach is maybe having CR only jump if the PUM is visible, but for placeholders the user doesn't change this means they will have to use Tab instead.

kylo252:
Why is github painful sometimes 🥵 
Can "requesting changes" be a little be more flexible please?

chase:
@abzcoding Since I'm touching Tab/S-Tab behavior here anyway, should I finish up the bindings for the popup-window version of cmp in this PR?

abzcoding:
> @abzcoding Since I'm touching Tab/S-Tab behavior here anyway, should I finish up the bindings for the popup-window version of cmp in this PR?

that would be much appreciated <3

kylo252:
Let's append any tweaks later on.

PR #1711, opened 2021-10-08
fix(telescope): prompt position when find/grep LunarVim files

younger-1:
# Description

With this [commit](https://github.com/nvim-telescope/telescope.nvim/commit/be600b5421c652c84f94c0c75ba6d309165ed0ef) from `telescope.nvim`, `prompt_position` is provided for `bottom_pane` and `vertical` layout strategy.
~~In order to keep `prompt` in the upper place, we should set this key explicitly in LunarVim.~~
In order to keep `prompt` in the upper place, we should remove this key in LunarVim's default layout and just let every layout strategy use their default prompt position.

### DIFF
--- a/lua/core/telescope.lua
+++ b/lua/core/telescope.lua
@@ -26,7 +26,6 @@ function M.config()
       layout_strategy = "horizontal",
       layout_config = {
         width = 0.75,
-        prompt_position = "bottom",
         preview_cutoff = 120,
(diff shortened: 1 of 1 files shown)


kylo252:
Thank you! I've been wondering why `get_ivy` has been looked weird lately.

younger-1:
Yeah, my pleasure

PR #1717, opened 2021-10-08
[Refactor] Make preselect behavior configurable

### LINKED ISSUE #1716: [Feature]: Change auto completion behavior back
### Feature motivation

Recently, I've noticed that the autocompletion behavior in LunarVim has changed. Before, when I'd tab through my auto completions, they'd automatically expand. Now, I have to hit enter.

It took me a while to track down why this was occurring, but I finally figure out this was due to a change @ChristianChiarulli made a little while back (f0b30f0). I only found it because I stumbled across nvim-cmp#225 after an hour of digging through settings and source code.

I think there are a couple of solutions to this:

* **Make this behavior configurable.** The behavior I'm seeking is the default in nvim-cmp. I can understand that others might have a different preference, but at the very least this should be easily configurable, and I'm not sure the current setting makes much sense as the default.
* **Revert this change.** Personally, I don't like this setting. I don't know how others feel about it. I'd be happy if this change were dropped.

I've tried out this behavior, and I just don't like it. It's cumbersome and slows down my development.

### Additional context

_No response_

LandonSchropp:
# Description

This makes the preselection behavior for auto-completion configurable. 

This *does not* address changes from f0b30f0 in keymappings.lua. Does it make sense for LunarVim to use `lvim.builtin.cmp.preselect` in that file?

Also, does this need to be documented anywhere?

Fixes #1716

## How Has This Been Tested?

I used this locally and confirmed it works. To repro, try typing something that's autocompletable, and then tabbing through the results.

### DIFF
--- a/lua/core/cmp.lua
+++ b/lua/core/cmp.lua
@@ -106,13 +106,15 @@ M.config = function()
       { name = "treesitter" },
       { name = "crates" },
     },
+    ---@usage enable or disable preselection of autocompletion results when tabbing
+    preselect = true,
     mapping = {
       ["<C-d>"] = cmp.mapping.scroll_docs(-4),
       ["<C-f>"] = cmp.mapping.scroll_docs(4),
       -- TODO: potentially fix emmet nonsense
       ["<Tab>"] = cmp.mapping(function()
         if vim.fn.pumvisible() == 1 then
-          vim.fn.feedkeys(T "<down>", "n")
+          vim.fn.feedkeys(T(lvim.builtin.cmp.preselect and "<C-n>" or "<down>"), "n")
         elseif luasnip.expand_or_jumpable() then
           vim.fn.feedkeys(T "<Plug>luasnip-expand-or-jump", "")
         elseif check_backspace() then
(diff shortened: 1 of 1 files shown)


LandonSchropp:
CC @abzcoding and @ChristianChiarulli.

ChristianChiarulli:
after rebasing I think this will be useful, I'll check it out on stream later after the rebase

LandonSchropp:
I started to go down the path of updating this to work with the refactored cmp.lua file. However, I noticed that the behavior has reversed, and we're back to tabbing by default. Is this intentional? Should I close this PR?

If not, I agree with @kylo252 that it'd be easy to modify that function's behavior. I found [this](https://github.com/hrsh7th/nvim-cmp/blob/ae42ce7b05d435a0eae22ebcde5fb9aac90de86b/lua/cmp/init.lua#L93-L97) in the nvim-cmp source code. It looks like the `behavior` option does the same thing my original PR was going to do.

kylo252:
Closing this since it seems to have been resolved. Please let me know if you're still having issues with it.

PR #1718, opened 2021-10-08
feat(cmp): make popupmenu formatting options configurable

xeluxee:
# Description
Settings related to popupmenu as duplicate entries source names are configurable now.
Why? Due to defaults duplicate setting many elements were missing from popupmenu (for example some snippets provided by luasnip) because they had equal names (as for snippet conflicting with for keyword).

**Before merging** I'd like to know if maintainers agree with setting all duplicate settings to 1 by default, so elements won't get hidden. If so I'll make another commit.

## How Has This Been Tested?
Put this in your `config.lua`
``` lua
lvim.builtin.cmp.formatting.source_names = {
	nvim_lsp = "(LSP)",
	emoji = "(Emoji)",
	path = "(Path)",
	calc = "(Calc)",
	cmp_tabnine = "(Tabnine)",
	vsnip = "(Snippet)",
	luasnip = "(Snippet)",
	buffer = "(Buffer)",
}

lvim.builtin.cmp.formatting.duplicates = {
	buffer = 1,
	path = 1,
	nvim_lsp = 0,
}

lvim.builtin.cmp.formatting.duplicates_default = 0
```

### DIFF
--- a/lua/core/cmp.lua
+++ b/lua/core/cmp.lua
@@ -66,23 +66,27 @@ M.config = function()
         Value = " ",
         Variable = " ",
       },
+      source_names = {
+        nvim_lsp = "(LSP)",
+        emoji = "(Emoji)",
+        path = "(Path)",
+        calc = "(Calc)",
+        cmp_tabnine = "(Tabnine)",
+        vsnip = "(Snippet)",
+        luasnip = "(Snippet)",
+        buffer = "(Buffer)",
+      },
+      duplicates = {
+        buffer = 1,
+        path = 1,
+        nvim_lsp = 0,
+      },
+      duplicates_default = 0,
       format = function(entry, vim_item)
         vim_item.kind = lvim.builtin.cmp.formatting.kind_icons[vim_item.kind]
-        vim_item.menu = ({
-          nvim_lsp = "(LSP)",
-          emoji = "(Emoji)",
-          path = "(Path)",
-          calc = "(Calc)",
-          cmp_tabnine = "(Tabnine)",
-          vsnip = "(Snippet)",
-          luasnip = "(Snippet)",
-          buffer = "(Buffer)",
-        })[entry.source.name]
-        vim_item.dup = ({
-          buffer = 1,
-          path = 1,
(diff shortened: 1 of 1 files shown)


xeluxee:
> **Before merging** I'd like to know if maintainers agree with setting all duplicate settings to 1 by default, so elements won't get hidden. If so I'll make another commit.

I think setting duplicates values to 1 by default should be better, because things like snippets or extra completion items could get hidden. Thoughts?

abzcoding:
> > **Before merging** I'd like to know if maintainers agree with setting all duplicate settings to 1 by default, so elements won't get hidden. If so I'll make another commit.
> 
> I think setting duplicates values to 1 by default should be better, because things like snippets or extra completion items could get hidden. Thoughts?

i don't think setting lsp to show duplicates is a good idea, the other two i'm fine with

xeluxee:
Ok, so what about setting `lvim.builtin.cmp.formatting.duplicates_default` to 1 by default?

abzcoding:
here is what i think, the current config is good for the default behavior, but we should document it somewhere and let users know

abzcoding:
> I agree with documenting it somewhere, but in my case using clangd many snippets from luasnip were hidden because they had the same name of entries provided by LSP (for example `for`, `while`, `printf`, `struct`)
> That's why I think that setting it to 1 by default would prevent this issue

sure that is why thanks to your PR, you can configure it for your own use case

xeluxee:
I came across another solution, I think we should set this as default:
``` lua
lvim.builtin.cmp.formatting.duplicates = {
	buffer = 1,
	path = 1,
	nvim_lsp = 0,
	luasnip = 1,
}
```

This prevents snippets from being hidden

xeluxee:
Ok so should I make a PR?
Since this would be a 1-line PR maybe someone with write access could directly edit the interested file?

abzcoding:
> Ok so should I make a PR?
> Since this would be a 1-line PR maybe someone with write access could directly edit the interested file?

yeah i will edit it asap, thank you

PR #1719, opened 2021-10-08
feat: support nvim-cmp floating menu highlights

abzcoding:
# Description

Support cmp new floating window in onedarker colorscheme
Screen Shot 2021-10-09 at 1 14 56 AM

# How to Test
you need to enable the `custom_menu` and also change the `Tab` and `S-Tab`

https://github.com/abzcoding/lvim/blob/main/lua/user/builtin.lua

### DIFF
--- a/lua/onedarker/highlights.lua
+++ b/lua/onedarker/highlights.lua
@@ -94,6 +94,14 @@ local highlights = {
   TabLine = { fg = C.light_gray, bg = C.alt_bg },
   TabLineSel = { fg = C.fg, bg = C.alt_bg },
   TabLineFill = { fg = C.fg, bg = C.alt_bg },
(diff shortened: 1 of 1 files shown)


abzcoding:
btw @kylo252 do we want to use the new nvim-cmp floating window?
if yes then we need to make some changes
https://github.com/abzcoding/lvim/blob/23224802ce5002fe51c637c1ce61e687812caaec/lua/user/builtin.lua#L60
Screen Shot 2021-10-09 at 2 38 57 PM

chase:
> I think we should try to somehow find a way to modularize the configuration for `cmp`. e.g: enable/disable sources, configure keybindings.

I think as-is, the keybindings are complex enough that exposing them to the users in a modular way might be kind of difficult. Other stuff would probably be a good idea though.

PR #1722, opened 2021-10-09
feat: support select and operator mode keymaps

kylo252:
# Description

- support select mode keymap
- support operator mode keymap
- update which-key settings

### DIFF
--- a/lua/core/which-key.lua
+++ b/lua/core/which-key.lua
@@ -37,9 +37,25 @@ M.config = function()
         height = { min = 4, max = 25 }, -- min and max height of the columns
         width = { min = 20, max = 50 }, -- min and max width of the columns
         spacing = 3, -- spacing between columns
+        align = "left", -- align columns left, center or right
       },
       hidden = { "<silent>", "<cmd>", "<Cmd>", "<CR>", "call", "lua", "^:", "^ " }, -- hide mapping boilerplate
       show_help = true, -- show help message on the command line when the popup is visible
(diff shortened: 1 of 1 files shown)


abzcoding:
> Why are visual mode and visual block mode renamed? 'x' has definitively been Visual Block Mode as far as I've known vim, same with 'v' being Visual Mode.

i do agree with you though, it feels a bit weird

kylo252:
I was also skeptical the first time I saw it, but it's actually in the friendly [manual](https://neovim.io/doc/user/map.html#mapping):
```doc
1.4 LISTING MAPPINGS					*map-listing*

When listing mappings the characters in the first two columns are:

      CHAR	MODE	
     <Space>	Normal, Visual, Select and Operator-pending
	n	Normal
	v	Visual and Select
	s	Select
	x	Visual
	o	Operator-pending
	!	Insert and Command-line
	i	Insert
	l	":lmap" mappings for Insert, Command-line and Lang-Arg
	c	Command-line
	t	Terminal-Job
```

chase:
> I was also skeptical the first time I saw it, but it's actually in the friendly [manual](https://neovim.io/doc/user/map.html#mapping):

Huh, I just associated it with the mode and not the mapping intent!

Although this is a breaking change, maybe keep the originals around so as not to break things but only keep the changes around in documentation?

kylo252:
Let's leave this alone for now, we can revisit it later.

PR #1725, opened 2021-10-09
feat: configure multiple filetypes per provider

### LINKED ISSUE #1724: [Feature]: support setting a null-ls provider once for selected languages
# Motiviation

Avoid having to set `prettier` 10 times when it's using the same options:

```lua
lvim.lang.typescript.formatters = { { exe = "prettier" } }
lvim.lang.javascript.formatters = { { exe = "prettier" } }
---
```

While still being able to set a different formatter for `json` or different arguments for json:

```lua
lvim.lang.json.formatters = { { exe = "prettier", args = {"--no-editorconfig"} } }
lvim.lang.typescriptreact.formatters = { { exe = "prettier_d" } }
```

kylo252:
# Description

Allow setting the same provider for multiple filetypes.

Fixes #1724

## How Has This Been Tested?

Try this

```lua
local configs = {
  {exe = "prettier", filetypes = {"javascript", "typescript"} }
}
require("lsp.null-ls.formatters").setup(configs)
```

### DIFF
--- a/lua/lsp/null-ls/formatters.lua
+++ b/lua/lsp/null-ls/formatters.lua
@@ -23,7 +23,7 @@ function M.list_available(filetype)
   return formatters
 end
 
-function M.list_configured(formatter_configs, filetype)
+function M.list_configured(formatter_configs)
   local formatters, errors = {}, {}
 
   for _, fmt_config in ipairs(formatter_configs) do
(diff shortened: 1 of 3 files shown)


kylo252:
@abzcoding, notice the nested table in the setup function. 
We can make it accept a single table if you think it's more intuitive.

PR #1733, opened 2021-10-10
[Bugfix] anything not in nvim-tree setup should be set using 1/0

abzcoding:
# Description

https://github.com/LunarVim/LunarVim/commit/c44550249b85307f668aff0d356c3461856c0e49 caused some problems
anything not included in the setup part should be set using 0/1
for example, this doesn't show icons
Screen Shot 2021-10-10 at 2 20 00 PM

but this does
Screen Shot 2021-10-10 at 2 19 36 PM

because we are setting them using `vim.g` or `g["nvim_tree_" .. opt] = val`

### DIFF
--- a/lua/core/nvimtree.lua
+++ b/lua/core/nvimtree.lua
@@ -23,18 +23,18 @@ function M.config()
       },
     },
     show_icons = {
-      git = true,
-      folders = true,
-      files = true,
-      folder_arrows = true,
+      git = 1,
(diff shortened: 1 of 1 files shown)


rebuilt:
Damn.  Sorry about that.

abzcoding:
> Damn. Sorry about that.

no worries mate, I had totally missed it in my review as well 😅

PR #1737, opened 2021-10-10
[Feature] use structlog and notify.nvim

kylo252:
# Description

A much better logging experience all around.

### DIFF
--- a/lua/lvim/config/defaults.lua
+++ b/lua/lvim/config/defaults.lua
@@ -28,5 +28,7 @@ return {
         float_opts = {},
       },
     },
+    ---@usage set to false to restore the default behavior of vim.notify
+    override_notify = true,
   },
 }
--- a/lua/lvim/core/log.lua
+++ b/lua/lvim/core/log.lua
@@ -85,11 +85,13 @@ function Log:init()
 
   local logger = structlog.get_logger "lvim"
 
-  -- Overwrite vim.notify to use the logger
-  vim.notify = function(msg, vim_log_level, opts)
-    nvim_notify_params = opts or {}
-    -- https://github.com/neovim/neovim/blob/685cf398130c61c158401b992a1893c2405cd7d2/runtime/lua/vim/lsp/log.lua#L5
-    logger:log(vim_log_level + 1, msg)
+  if lvim.log.override_notify then
+    -- Overwrite vim.notify to use the logger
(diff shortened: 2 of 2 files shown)


Tastyep:
> I was also wondering how easy would it be to _optionally_ relegate the console responsibility to this plugin: https://github.com/rcarriga/nvim-notify
> 
> notify
> 
> Which I believe that @abzcoding will be very much in favor of smile

This looks great ! It should be easily achievable by providing a new sink.
I could either provide it directly from structlog or you could implement it yourself here. Since the logger is already specialized for nvim I don't mind adding it to structlog.

kylo252:
> This looks great ! It should be easily achievable by providing a new sink. I could either provide it directly from structlog or you could implement it yourself here. Since the logger is already specialized for nvim I don't mind adding it to structlog.

Oh I didn't even think about adding it a sink. I set out to add it but got stuck in the step with specifiying the log level. 

<details>
<summary>Notify handler</summary>

```lua
--- Sends log entries to Notify.

--- The Notify sink class.
-- @type Notify
local Notify = {}

--- Create a new Notify handler
-- @function Notify
-- @param opts Optional parameters
-- @param opts.timeout Number of milliseconds to show the window, default: 5000
-- @param opts.stages Animation stages, default "fade_in_slide_out"
-- @param opts.background_colour Highlight group, default: Normal
-- @param opts.icons Icons used for earch log level
setmetatable(Notify, {
  __call = function(cls, ...)
    return cls:new(...)
  end,
})

function Notify:new(opts)
  opts = opts or {}

  local notify = require("notify").setup(opts)

  Notify.__index = Notify
  setmetatable(notify, Notify)

  return Notify
end

function Notify:send(msg, level, opts)
end

return Notify
```

</details>

I don't think it's gonna work as a sink with the current abstraction level that you have in Structlog, because it needs more metadata and should be higher up in the pipeline anyway.

> A sink is the final output of your logs entries. Once they have been through the pipeline of log processors and formatter, the final result is sent to the sink.

abzcoding:
am i supposed to get this when there is an error in my config?
Screen Shot 2021-10-20 at 8 26 40 PM

abzcoding:
> > other than that it looks very good tyvm 👍
> 
> Did you update structlog? I had another issue that was still related to that.

let me try it again

PR #1741, opened 2021-10-11
fix: code action module core.telescope not found error

schwartx:


abzcoding:
https://github.com/LunarVim/LunarVim/blob/fdec599954dcc093e56522c4f9dcb126ad869979/lua/lvim/core/telescope.lua#L160-L178

PR #1766, opened 2021-10-13
feat(cmp): add option for supertab snippets precedence

xeluxee:
# Description
If this option is enabled when pressing `<tab>` the cursor jumps to the next snippet position instead of selecting next item in completion menu.
Sometimes is annoying to press `<C-e>` to close completion menu and then press `<tab>` to jump to the next snippet location.
Also I've added `<C-j>` and `<C-k>` keybindings to navigate in completion menu, since these keybindings are widely used.

## How Has This Been Tested?
Put this in your config.lua
``` lua
lvim.builtin.cmp.supertab_snippets_precedence = true
```

### DIFF
--- a/lua/lvim/core/cmp.lua
+++ b/lua/lvim/core/cmp.lua
@@ -203,7 +203,7 @@ M.config = function()
       { name = "treesitter" },
       { name = "crates" },
     },
-    supertab_snippets_precedence = false,
+    tab_prefer_snippets = false,
     mapping = {
       ["<C-k>"] = cmp.mapping.select_prev_item(),
       ["<C-j>"] = cmp.mapping.select_next_item(),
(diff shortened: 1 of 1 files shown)


chase:
> Yeah but since I've put this line in my config (I want first menu item to be preselected) I can't use `<cr>` to jump forward because it would complete the selected item in menu
> 
> ```lua
> lvim.builtin.cmp.completion = {completeopt = 'menu,menuone,noinsert'}
> ```

I would say just don't do that, but I can understand why you might want that behavior.

Overall it sounds like you replace a lot of the behavior setup by default, so I'm wondering if maybe we're not offering the right level of abstraction for things to be configured.
If we exported the helper methods used in the keybindings (ex. `inside_snippet`, `seek_luasnip_cursor_node`, `is_emmet_active`, ...) so that you could use them without copying and pasting, would that resolve your issue?

xeluxee:
> The "secret sauce" isn't exposed for super tab, things like `inside_snippet`, `seek_luasnip_cursor_node`, are all private and would require you to copy them entirely to get working. I think that exposing these in the module would allow you to redefine without much code duplication.

Ok, let's go with this

chase:
> Before proceeding what about moving cmp setup settings under `lvim.builtin.cmp.setup` to avoid confusion with LunarVim settings (such as kind icons, source names, duplicates settings, functions that will be exposed in the future...)?

Moving things under `.setup` would be non-standard to the rest, so I feel like that's not the best approach.

This was a mistake I introduced after a lengthy conversation with another maintainer:
```lua
    confirm_opts = {
      behavior = cmp.ConfirmBehavior.Replace,
      select = false,
    },
```

Now this option has to be supported to some extent going forward, rather than simplifying how the confirm code works using a better abstraction.

As pointed out in my PR introducing confirm_opts (#1523), there are a lot of downsides to introducing an unofficial parameter instead of delegating responsibility to the user's config and the plugin itself, but I felt like I had a good fix that other people said they wanted in Discord.

I'm open to other suggestions, but I'm hesitant to change how options are passed and introducing new options that abstract behavior.

PR #1767, opened 2021-10-13
feat: Add find_files_relative function

zalegrala:
Without this function, there isn't a clean way to use telescope to find
files that are relative to the current buffer.  Here we include a
function to help.


kylo252:
I think this might be better reserved as a user configuration since it's overriding the local theme, even though I get why you're doing that.

PR #1773, opened 2021-10-14
[Bugfix]: require lspconfig for a language server only if it's not already loaded

vadimi:
# Description

After running `:LvimUpdate` when .go or .yaml file is open language features stop working. I was able to track it down to the ftplugin regen issue - basically `:LvimUpdate` removes all files from `~/.local/share/lunarvim/site/after/ftplugin` and regenerates them, but it doesn't do it for the actively running lsp server. The fix is to require specific language server only if it's not already loaded.

## How Has This Been Tested?

- Run `lvim main.go`
- Run command `:LvimUpdate`
- Restart lvim
- Run `lvim main.go`
- Check that language features are working (`gd` for example)


kylo252:
Superseded by #1836

Issue #1229, opened 2021-08-04
[Feature]: Support for VimR and neovide


vlaw:
### Feature motivation

Since LunarVim uses a custom directory (`~/.local/share/lunarvim/lvim`), the LunarVim configuration cannot be loaded when opening [vimr](https://github.com/qvacua/vimr) (a Neovim GUI for macOS) directly.
Even I copy it into `~/.config/nvim`, it doesn't work, 
because the path has been remove from `packpath` and `runtimepath`

> <https://github.com/ChristianChiarulli/LunarVim/blob/7cd03ff4e332e1427497b54d5ca8b037afbc0982/init.lua#L2>
> <https://github.com/ChristianChiarulli/LunarVim/blob/7cd03ff4e332e1427497b54d5ca8b037afbc0982/init.lua#L8>

PS. the command `vimr --nvim -u ~/.local/share/lunarvim/lvim/init.lua --cmd "set runtimepath+=~/.local/share/lunarvim/lvim" ` works fine.

### Describe the alternatives you've considered

an installation switch option to choose directory between `~/.local/share/lunarvim/lvim` and `~/.config/nvim` ?

sa-mendez:
I second this.  I use neovide, and would like to be able to just run with the lunarvim setup by default.

I actually ended up getting this to work by changing a few paths in a few of the files (in lunarvim 0.5.1 release), but the moment I get the next release, I know I will have to do some merges.  It would be nice to have user configurable root, so that some of us that just want to use lunarvim full-time, can do so without running a different executable (lvim) and/or aliasing nvim.

rebuilt:
> I actually ended up getting this to work by changing a few paths in a few of the files (in lunarvim 0.5.1 release), but the moment I get the next release, I know I will have to do some merges. It would be nice to have user configurable root, so that some of us that just want to use lunarvim full-time, can do so without running a different executable (lvim) and/or aliasing nvim.

Does this launch neovide correctly for you (without the altering of paths?)

```
neovide -- -u ~/.local/share/lunarvim/lvim/init.lua --cmd "set runtimepath+=~/.local/share/lunarvim/lvim"

```

kylo252:
> PS. the command `vimr --nvim -u ~/.local/share/lunarvim/lvim/init.lua --cmd "set runtimepath+=~/.local/share/lunarvim/lvim" ` works fine.

By the way, if that command works for you then you can add this shell function to your bashrc/zshrc

```shell
function lvimr(){
[...]

PR #1777, opened 2021-10-15
fix(bootstrap): support for using default neovim config and data directories

sa-mendez:
# Description

Fixes bootstrap.lua global function get_runtime_dir() to set the proper default directory for direct neovim usage (or GUIs such as neovide or VimR).  This is simply `stdpath("data")`.  It resolves a lingering problem where packer could not find your packages during startup.

Small fix to finalize #1229

## How Has This Been Tested?

Tested on:

- Manjaro Linux (using nvim in terminal, and neovide)

- Windows 10 VM (using nvim-qt)

Steps to test:
- Clone Lunarvim directly into ~/.config/nvim
-  Ensure none of the Lunarvim variables ($LUNARVIM_CONFIG_DIR, $LUNARVIM_RUNTIME_DIR, $LUNARVIM_CACHE_DIR) are defined in your environment.
- (Optional): Copy over (or sym link) config.lua  to ~/.config/nvim
-  Run `nvim` command in terminal, or launch neovide/VimR/nvim-qt, or your preferred neovim GUI.
-  Run `:PackerSync` inside neovim editor.

### DIFF
--- a/lua/lvim/bootstrap.lua
+++ b/lua/lvim/bootstrap.lua
@@ -1,5 +1,4 @@
 local M = {}
-local lvim_base_dir
 
 package.loaded["lvim.utils.hooks"] = nil
 local _, hooks = pcall(require, "lvim.utils.hooks")
@@ -45,20 +44,21 @@ function _G.get_cache_dir()
   return lvim_cache_dir
 end
 
-function _G.get_lvim_base_dir()
-  return lvim_base_dir
-end
-
 ---Initialize the `&runtimepath` variables and prepare for startup
 ---@return table
 function M:init(base_dir)
   self.runtime_dir = get_runtime_dir()
   self.config_dir = get_config_dir()
   self.cache_path = get_cache_dir()
-  lvim_base_dir = base_dir
   self.pack_dir = join_paths(self.runtime_dir, "site", "pack")
   self.packer_install_dir = join_paths(self.runtime_dir, "site", "pack", "packer", "start", "packer.nvim")
   self.packer_cache_path = join_paths(self.config_dir, "plugin", "packer_compiled.lua")
+
+  ---Get the full path to LunarVim's base directory
+  ---@return string
+  function _G.get_lvim_base_dir()
+    return base_dir
+  end
 
   if os.getenv "LUNARVIM_RUNTIME_DIR" then
     -- vim.opt.rtp:append(os.getenv "LUNARVIM_RUNTIME_DIR" .. path_sep .. "lvim")


luxus:
would it be possible to still have $LUNARVIM_CONFIG_DIR set to .config/lvim and have the config there instead of symlink into .config/nvim?

sa-mendez:
> I've made the same change on my fork for the same reason.
> 
> @sa-mendez I think your PR includes more commits than you intended?

Yeah, I had noticed something was off when I updated the commit comment earlier.  Thanks @kylo252  for fixing this.

sa-mendez:
> @sa-mendez @kylo252 see #1791 for an updated PR of my "global-install" technique. It includes the same change done here, but shows how we can support running lunarvim as "nvim" or "lvim" where "lvim" can be run from the user's local path, or alternatively a global-system wide install.

@jameswalmsley I read the earlier comments from you and @kylo252, and I agree, having a new variable LUNARVIM_INSTALL_DIR is the right way to go.  I am planning to update this PR to incorporate that, leveraging some of the changes you made in your referenced PR.

IMHO though, I think it best to keep out your global install support. Perhaps once this PR is finalized and incorporated into rolling you can update your PR to just include the global install support.

jameswalmsley:
@sa-mendez That would be fine with me.. 
I'm applying feedback from @kylo252 at the moment, and minimising the scope of the PR at the moment.

kylo252:
> Dont you mean "**can use it**"?

No I did mean you can **not** use it directly if you don't use `~/.config/nvim`. You're gonna have to mess with the runtimepath for this to work, or symlink some files or use an `lvim` binary or something. And we don't or shouldn't need any of that, because I know it nearly works already. We just need to decouple these three directories: {runtimepath, installation, packpath}

sa-mendez:
> Does this means we can clone it manually to .config/nvim ? And how do I set the lvim config dir?

Yes, this works now.  Just tested from rolling.  Thanks @jameswalmsley  and @kylo252  for making this happen.

The lvim config dir in my setup is not set at all, so it is defaulting to .config/nvim (e.g. it will expect config.lua to reside in that directory, among other configs).

If you want to override the default, you should simply be able to just set the LUNARVIM_CONFIG_DIR environment variable to whatever you want.

kylo252:
> is it possible to prioritize `.config/lvim/config.lua` over `.config/nvim/config.lua`? so it would still use the standard `LUNARVIM_CONFIG_DIR` that the standard installation uses

It should be looking for `config.lua` in a single place only, so there's no priority involved there:
https://github.com/LunarVim/LunarVim/blob/03e5760e59b5dd420956908b6f8b57c33f70f2cd/lua/lvim/config/init.lua#L6

PR #1790, opened 2021-10-16
fix(cmp): fix <C-j> and <C-k> keybindings not working

xeluxee:
# Description
When `nvim-cmp` experimental menu was introduced in LunarVim `<C-j>` and `<C-k>` stopped working.
This PR restores them.

## How Has This Been Tested?
Try opening a completion menu and press `<C-j>` to go to next element and `<C-k>` to go to previous element.
Of course you can use `<tab>`, `<s-tab>`, `<C-n>` and `<C-p>` too

### DIFF
--- a/lua/lvim/core/cmp.lua
+++ b/lua/lvim/core/cmp.lua
@@ -204,6 +204,8 @@ M.config = function()
       { name = "crates" },
     },
     mapping = {
+      ["<C-k>"] = cmp.mapping.select_prev_item(),
+      ["<C-j>"] = cmp.mapping.select_next_item(),
       ["<C-d>"] = cmp.mapping.scroll_docs(-4),
       ["<C-f>"] = cmp.mapping.scroll_docs(4),
       -- TODO: potentially fix emmet nonsense
--- a/lua/lvim/keymappings.lua
+++ b/lua/lvim/keymappings.lua
@@ -81,10 +81,6 @@ function M.config()
       ["<A-Down>"] = "<C-\\><C-N><C-w>j",
       ["<A-Left>"] = "<C-\\><C-N><C-w>h",
       ["<A-Right>"] = "<C-\\><C-N><C-w>l",
-      -- navigate tab completion with <c-j> and <c-k>
-      -- runs conditionally
-      ["<C-j>"] = { 'pumvisible() ? "\\<down>" : "\\<C-j>"', { expr = true, noremap = true } },
-      ["<C-k>"] = { 'pumvisible() ? "\\<up>" : "\\<C-k>"', { expr = true, noremap = true } },
     },
 
     ---@usage change or add keymappings for normal mode


chase:
If nvim-cmp deprecated these bindings by default, I'm not sure we're obligated to support them moving forward.
This seems like a personal preference and isn't difficult to setup in `config.lua`, as I've stated before, but I'll defer to the other maintainers.

abzcoding:
i don't think we need to add more this, when you can easily add it to your own config
```lua
local cmp = require "cmp"
lvim.builtin.cmp.mapping["<C-k>"] = cmp.mapping.select_prev_item()
lvim.builtin.cmp.mapping["<C-j>"] = cmp.mapping.select_next_item()
```

xeluxee:
> If nvim-cmp deprecated these bindings by default, I'm not sure we're obligated to support them moving forward.

These keybindings weren't deprecated, simply they remained there but they didn't worked (see changed files). So I'm not adding anything new, I'm fixing an existent keybinding


> i don't think we need to add more this, when you can easily add it to your own config

Yeah it's easy to add this to config but I'm wondering why not keeping these keymaps
I think these keybindings are pretty standard, especially in vim where `j` and `k` are probably the most used keys, on which a user always keep his fingers

chase:
> > If nvim-cmp deprecated these bindings by default, I'm not sure we're obligated to support them moving forward.
> 
> These keybindings weren't deprecated, simply they remained there but they didn't worked (see changed files). So I'm not adding anything new, I'm fixing an existent keybinding

Ah, my bad, I missed that. It's not quite the same since it doesn't check for `cmp.visible` and fallback otherwise, fix that and this is good to go.

abzcoding:
> > > If nvim-cmp deprecated these bindings by default, I'm not sure we're obligated to support them moving forward.
> > 
> > 
> > These keybindings weren't deprecated, simply they remained there but they didn't worked (see changed files). So I'm not adding anything new, I'm fixing an existent keybinding
> 
> Ah, my bad, I missed that. It's not quite the same since it doesn't check for `cmp.visible` and fallback otherwise, fix that and this is good to go.

i don't think he needs to ( `cmp.mapping.select_prev_item` has that builtin )
https://github.com/hrsh7th/nvim-cmp/blob/0f28030aef5b7b15708aaed8ed90b94d206aa712/lua/cmp/config/mapping.lua#L64

chase:
> i don't think he needs to ( `cmp.mapping.select_prev_item` has that builtin ) https://github.com/hrsh7th/nvim-cmp/blob/0f28030aef5b7b15708aaed8ed90b94d206aa712/lua/cmp/config/mapping.lua#L64

You're right, it returns false if not visible:
https://github.com/hrsh7th/nvim-cmp/blob/0f28030aef5b7b15708aaed8ed90b94d206aa712/lua/cmp/init.lua#L118

PR #1876, opened 2021-10-30
Fixed lualine icon

Ali-Aref:
Lualine edit icon that was "柳", fixed to edit icon.

### DIFF
--- a/lua/lvim/core/lualine/components.lua
+++ b/lua/lvim/core/lualine/components.lua
@@ -35,7 +35,7 @@ return {
   diff = {
     "diff",
     source = diff_source,
-    symbols = { added = "  ", modified = "柳", removed = " " },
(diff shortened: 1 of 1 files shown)


kylo252:
This is probably a problem with the your font, because it displays correctly for me. I don't really have an opinion on which is the better icon though, so I'll leave it to someone else :)

https://user-images.githubusercontent.com/59826753/139590856-d58eec95-44d9-47eb-9e8f-dd98d667a713.mp4

abzcoding:
the issue is probably from your font,
using Firacode to show the changes you made,
Screen Shot 2021-10-31 at 7 02 49 PM

abzcoding (commented):
i don't think this change is necessary
i would suggest installing and using a nerd font with full glyph support

PR #1883, opened 2021-10-31
feat: better error handling for packer

kylo252:
# Description

- better error handling for packer
- use packer in protected mode

### DIFF
--- a/.github/workflows/install.yaml
+++ b/.github/workflows/install.yaml
@@ -33,12 +33,6 @@ jobs:
         run: |
           ./utils/installer/install.sh --local --no-install-dependencies
 
-      - name: Test LunarVim PackerCompile
-        run: if "$HOME"/.local/bin/lvim --headless +PackerCompile -c ':qall' 2>&1|grep -q 'Error'; then false; fi
-
-      - name: Test LunarVim Health
-        run: if "$HOME"/.local/bin/lvim --headless +checkhealth -c ':qall' 2>&1|grep -q 'Error'; then false; fi
-
       - name: Run unit-tests
         # NOTE: make sure to adjust the timeout if you start adding a lot of tests
         timeout-minutes: 4
--- a/init.lua
+++ b/init.lua
@@ -10,7 +10,7 @@ require("lvim.bootstrap"):init(base_dir)
 require("lvim.config"):load()
 
 local plugins = require "lvim.plugins"
-require("lvim.plugin-loader"):load { plugins, lvim.plugins }
+require("lvim.plugin-loader").load { plugins, lvim.plugins }
 
 local Log = require "lvim.core.log"
 Log:debug "Starting LunarVim"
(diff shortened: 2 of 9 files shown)


.github/workflows/install.yaml
```diff
@@ -33,12 +33,6 @@ jobs:
         run: |
           ./utils/installer/install.sh --local --no-install-dependencies
 
-      - name: Test LunarVim PackerCompile
-        run: if "$HOME"/.local/bin/lvim --headless +PackerCompile -c ':qall' 2>&1|grep -q 'Error'; then false; fi
-
-      - name: Test LunarVim Health
-        run: if "$HOME"/.local/bin/lvim --headless +checkhealth -c ':qall' 2>&1|grep -q 'Error'; then false; fi
```
abzcoding, line 40: is there a reason we are removing health check and packercompile test from our CI?

abzcoding:
the only issue that i get is this
Screen Shot 2021-10-31 at 7 07 55 PM
Screen Shot 2021-10-31 at 7 08 03 PM

kylo252:
> the only issue that i get is this Screen Shot 2021-10-31 at 7 07 55 PM Screen Shot 2021-10-31 at 7 08 03 PM

This means it actually intercepted an error, in theory at least. Do you see any warnings from Packer?

abzcoding:
> This means it actually intercepted an error, in theory at least. Do you see any warnings from Packer?

i've not seen any errors 🤔 
maybe i need to `LvimCacheReset` or sth 🤷

kylo252:
> i've not seen any errors 🤔 maybe i need to `LvimCacheReset` or sth 🤷

when are you getting that error exactly? btw, I finally fixed/added `LvimReload` 😄

abzcoding:
> > i've not seen any errors 🤔 maybe i need to `LvimCacheReset` or sth 🤷
> 
> when are you getting that error exactly? btw, I finally fixed/added `LvimReload` 😄

basically the Packer floating window comes up and does its thing, and i get that harmless error in the side 😅

kylo252:
> basically the Packer floating window comes up and does its thing, and i get that harmless error in the side 😅

make sure you have set

```lua
lvim.log.level = "trace"
```

some tests to try:
- `PackerSync`, does it produce a different output than `LvimReload`?
- `PackerCompile`, does it cause any errors?
- `:lua require("lvim.plugin-loader"):sync()` <-- this will probably trigger an error

kylo252:
I think it should be solved now, I was doing some dumb things that we shall not make any mention of..

abzcoding:
> I think it should be solved now, I was doing some dumb things that we shall not make any mention of..

yeah its fixed now

kylo252:
> yeah its fixed now

I decided to re-adjust the scope of the PR, only expose/use the protected mode for core-plugins. That's why I reverted the which-key bindings. 

Now I need to figure out how to fix the race-condition between `PackerSync` and `PackerCompile` for the installer.

PR #1890, opened 2021-11-01
fix(installer): use symbolic link for local `lvim.ps1`

younger-1:
# Description

Use symbolic link for local `lvim.ps1` to sync with repository's changes to `lvim.ps1`

### DIFF
--- a/utils/installer/install.ps1
+++ b/utils/installer/install.ps1
@@ -209,7 +209,8 @@ function setup_shim() {
         New-Item "$INSTALL_PREFIX\bin" -ItemType Directory
     }
     # Copy-Item "$env:LUNARVIM_RUNTIME_DIR\lvim\utils\bin\lvim.ps1" -Destination "$INSTALL_PREFIX\bin\lvim.ps1" -Force
-    New-Item -ItemType SymbolicLink -Path "$INSTALL_PREFIX\bin\lvim.ps1" -Target "$env:LUNARVIM_RUNTIME_DIR\lvim\utils\bin\lvim.ps1" -Force
+    # New-Item -ItemType SymbolicLink -Path "$INSTALL_PREFIX\bin\lvim.ps1" -Target "$env:LUNARVIM_RUNTIME_DIR\lvim\utils\bin\lvim.ps1" -Force
+    & "$env:COMSPEC" /c "mklink /h `"$INSTALL_PREFIX\bin\lvim.ps1`" `"$env:LUNARVIM_RUNTIME_DIR\lvim\utils\bin\lvim.ps1`"" | Out-Null
 }
 
 function setup_lvim() {
--- a/utils/installer/install.ps1
+++ b/utils/installer/install.ps1
@@ -213,7 +213,7 @@ function setup_shim() {
     if ((Test-Path "$INSTALL_PREFIX\bin\lvim.ps1") -eq $true) {
         Remove-Item -Force "$INSTALL_PREFIX\bin\lvim.ps1"
     }
-    & "$env:COMSPEC" /c "mklink /h `"$INSTALL_PREFIX\bin\lvim.ps1`" `"$env:LUNARVIM_RUNTIME_DIR\lvim\utils\bin\lvim.ps1`"" | Out-Null
+    & "$env:COMSPEC" /c "mklink `"$INSTALL_PREFIX\bin\lvim.ps1`" `"$env:LUNARVIM_RUNTIME_DIR\lvim\utils\bin\lvim.ps1`""
 }
 
 function setup_lvim() {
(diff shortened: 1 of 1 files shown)


younger-1:
@kylo252 You are pretty right. And symbolic link require extra `sudo` privilege in `PowerShell` which is diffent from `pwsh` or `PowerShell Core` I use :sweat_smile: . I think the majority of `lunarvim` users of windows may accept the responsibility for preparing their basic developing environment if they really want to use `lunarvim` to achieve all daily editing tasks. Otherwise they could use out of box tools like `idea` and `vscode`, or just vanilla neovim. We should trust that those `lunarvimer` could overcome all the troubles for using `lunarvim` in windows. Maybe we could add more help info for setup `lunarvim` in windows to lunarvim website. To me, the basic developing environment for windows includes [git](https://gitforwindows.org), [pwsh](https://github.com/PowerShell/PowerShell) with developer option enable, [sudo](https://github.com/lukesampson/psutils), [nodejs](https://nodejs.org), `C/C++/make` provided by [winlibs-mingw](https://winlibs.com/) or [llvm-mingw](https://github.com/mstorsjo/llvm-mingw), searching provided by [rg](https://github.com/BurntSushi/ripgrep) and [fd](https://github.com/sharkdp/fd). Without one of above tools, I could hardly benefit from `treesitter`, `telescope`, `lsp` those basic feature in neovim ecosystem, let alone to setup a useable `lvim` editor. All in all, I want `lunarvim` to be bettter and friendly to real developer.

younger-1:
> 
> 
> @kylo252 You are pretty right. And symbolic link require extra `sudo` privilege in `PowerShell` which is diffent from `pwsh` or `PowerShell Core` I use 😅 . I think the majority of `lunarvim` users of windows may accept the responsibility for preparing their basic developing environment if they really want to use `lunarvim` to achieve all daily editing tasks. Otherwise they could use out of box tools like `idea` and `vscode`, or just vanilla neovim. We should trust that those `lunarvimer` could overcome all the troubles for using `lunarvim` in windows. Maybe we could add more help info for setup `lunarvim` in windows to lunarvim website. To me, the basic developing environment for windows includes [git](https://gitforwindows.org), [pwsh](https://github.com/PowerShell/PowerShell) with developer option enable, [sudo](https://github.com/lukesampson/psutils), [nodejs](https://nodejs.org), `C/C++/make` provided by [winlibs-mingw](https://winlibs.com/) or [llvm-mingw](https://github.com/mstorsjo/llvm-mingw), searching provided by [rg](https://github.com/BurntSushi/ripgrep) and [fd](https://github.com/sharkdp/fd). Without one of above tools, I could hardly benefit from `treesitter`, `telescope`, `lsp` those basic feature in neovim ecosystem, let alone to setup a useable `lvim` editor. All in all, I want `lunarvim` to be bettter and friendly to real developer.

All of those tools could be installed by [scoop](scoop.sh/) and official [winget](https://github.com/microsoft/winget-cli) will support them eventually

younger-1:
As a workaround, creating hard link by `mklink` may work as expect.

kylo252:
Is it only for maintaining changes? There's an easier way to achieve that if it's what you're looking for. We just need to add the required logic to the post-update hook which I specifically created for such cases:

https://github.com/LunarVim/LunarVim/blob/32ca5afa4ad21f1a616cc30323c272191e7548c1/lua/lvim/utils/hooks.lua#L25-L42

younger-1:
Yes, this is only for maintaining changs. The post-update hook looks great. But how to get the `$INSTALL_PREFIX` specified by user, since we can't assume it as `~/.local/`

kylo252:
> Yes, this is only for maintaining changs. The post-update hook looks great. But how to get the `$INSTALL_PREFIX` specified by user, since we can't assume it as `$HOME\.local`

How about using something like `:! which lvim`?

younger-1:
I will try. But I think it may not the responsibility of the reload function of `lunarvim` runtime's to regenerate local shim.

younger-1:
Since we are unable to use `which` in `powershell` and `cmd` in windows. Do you mean vim function `exepath('lvim')`?

And I can't find other helper function to get the path of an executable in lua module `vim` or `vim.loop`. 

`exepath('lvim.ps1')` `exepath("lvim")` return `nil` because in windows `cmd.exe` shell can only load `exe` or `cmd` file.

I have commit in #1893 how to setup `lvim.cmd` to solve this

kylo252:
I have this snippet in my `profile.ps1` that I probably copied from SO

```pwsh
function which ($command) {
    Get-Command -Name $command -ErrorAction SilentlyContinue |
        Select-Object -ExpandProperty Path -ErrorAction SilentlyContinue
}
```

younger-1:
It still look weird to maintain changes of `lvim` and `lvim.ps1` in post-update hook which may effect other files in system path containing `lvim` or `lvim.ps1`.

younger-1:
I am so silly that until now I realize that neither developer options nor administrator's privilege is required to creat symbolic link by using `mklink` without extra options. Just using this:
```
mklink Link Target
```

PR #1897, opened 2021-11-03
fix(lualine): add space to diff components

younger-1:
# Description

Add proper space between glyphs and number for diff components.

Before:
图片

After:
图片

### DIFF
--- a/lua/lvim/core/lualine/components.lua
+++ b/lua/lvim/core/lualine/components.lua
@@ -35,7 +35,7 @@ return {
   diff = {
     "diff",
     source = diff_source,
-    symbols = { added = "  ", modified = "柳", removed = " " },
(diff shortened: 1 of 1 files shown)


Tastyep:
Out of curiosity, what's the setup to reproduce this ? I can't reproduce it on my machine

younger-1:
> 
> 
> Out of curiosity, what's the setup to reproduce this ? I can't reproduce it on my machine

I am pretty sure the problem happens in whatever configuration since I install `lunarvim` in another computer just now and take `config.example.lua` as `config.lua`.

kylo252:
I think that the diff symbol is actually 2 characters-wide instead of 1, so that's probably the issue. We should replace it instead.

PR #1904, opened 2021-11-04
feat: enable context commentstring by default

ChristianChiarulli:
Simple PR to enable context comment string for treesitter by default, A good amount of people use lunarvim for webdev and I don't think you should need to think about installing a plugin to make your comments work correctly.

### DIFF
--- a/lua/lvim/core/treesitter.lua
+++ b/lua/lvim/core/treesitter.lua
@@ -16,7 +16,7 @@ M.config = function()
       disable = { "latex" },
     },
     context_commentstring = {
-      enable = false,
+      enable = true,
       config = { css = "// %s" },
     },
     -- indent = {enable = true, disable = {"python", "html", "javascript"}},
--- a/lua/lvim/core/treesitter.lua
+++ b/lua/lvim/core/treesitter.lua
@@ -17,7 +17,16 @@ M.config = function()
     },
     context_commentstring = {
       enable = true,
-      config = { css = "// %s" },
+      config = {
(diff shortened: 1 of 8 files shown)


lua/lvim/plugins.lua
```diff
@@ -75,6 +75,7 @@ return {
       require("lvim.core.treesitter").setup()
     end,
   },
+  { "JoosepAlviste/nvim-ts-context-commentstring" },
```
kylo252, line 78:
```suggestion
  { 
  	"JoosepAlviste/nvim-ts-context-commentstring",
  	event = "BufReadPost",
  },
```

lua/lvim/core/treesitter.lua
```diff
@@ -16,7 +16,7 @@ M.config = function()
       disable = { "latex" },
     },
     context_commentstring = {
-      enable = false,
+      enable = true,
       config = { css = "// %s" },
```
kylo252, line 20:
```suggestion
      config = {
            -- Languages that have a single comment style
            typescript = '// %s',
            css = '/* %s */',
            scss = '/* %s */',
            html = '<!-- %s -->',
            svelte = '<!-- %s -->',
            vue = '<!-- %s -->',
            json = ''
      },
```
do you think we should add these? although it has a different style for css, so I'll leave it to you since I don't use any of these languages.

ChristianChiarulli: sure, looks good to me, if there are complaints one way or the other we can update this list

PR #1910, opened 2021-11-05
fix: use an indepdent shadafile from neovim

kylo252:
# Description

Use an independent [shadafile](https://neovim.io/doc/user/starting.html#shada-file) from base Neovim.

Fixes mingled jumplists when using both Neovim and LunarVim.

### DIFF
--- a/lua/lvim/config/settings.lua
+++ b/lua/lvim/config/settings.lua
@@ -51,6 +51,8 @@ M.load_options = function()
 
   vim.opt.shortmess:append "c"
 
+  vim.opt.shadafile = utils.join_paths(get_cache_dir(), "lvim.shada")
+
   for k, v in pairs(default_options) do
     vim.opt[k] = v
   end
--- a/lua/lvim/bootstrap.lua
+++ b/lua/lvim/bootstrap.lua
@@ -46,7 +46,7 @@ end
 function M:init(base_dir)
   self.runtime_dir = get_runtime_dir()
   self.config_dir = get_config_dir()
-  self.cache_path = get_cache_dir()
+  self.cache_dir = get_cache_dir()
   self.pack_dir = join_paths(self.runtime_dir, "site", "pack")
   self.packer_install_dir = join_paths(self.runtime_dir, "site", "pack", "packer", "start", "packer.nvim")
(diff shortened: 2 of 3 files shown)


lua/lvim/config/settings.lua
```diff
@@ -51,6 +51,23 @@ M.load_options = function()
...
   vim.opt.shortmess:append "c"
 
+  -- Defaults:
+  --   Neovim: !,'100,<50,s10,h
+  --
+  -- - ! save/restore global variables (only all-uppercase variables)
+  -- - '100 save/restore marks from last 100 files
+  -- - <50 save/restore 50 lines from each register
+  -- - s10 max item size 10KB
+  -- - h do not save/restore 'hlsearch' setting
```
chase, line 61: I actually think that these two defaults are good.
Restoring the highlights of search on a reload probably loses context.

For reference `:help sd`

lua/lvim/config/settings.lua
```diff
@@ -51,6 +51,23 @@ M.load_options = function()
...
+  --
+  -- Our overrides (credit wincent):
+  -- - '50 save/restore marks from last 100 files
+  -- - <0 don't save registers
+  -- - f0 don't store file marks
+  -- - n: store in $LUNARVIM_CACHE_DIR/lvim.shada
+  --
+  vim.opt.shada = "'50,<0,f0,n" .. utils.join_paths(get_cache_dir(), "lvim.shada")
```
chase, line 69: Instead of using the `n` option, I would recommend setting `shadafile` instead since it overrides the `n` option
`:help sdf`

chase (changes requested):
Overall seems like a good idea, but see my comments

lua/lvim/config/settings.lua
kylo252: I thought it was the other way around with `h` command. It's a bit confusing how they named these variables 🥲

lua/lvim/config/settings.lua
kylo252: I decided to use this option and leave the rest as default, since they have different values on Windows and I don't feel like digging through that.

chase (approved):

PR #1922, opened 2021-11-07
fix: packersync issue when you have large number of plugins

abzcoding:
# Description

if you have large number of plugins ( more than 70) running `:PackerSync` wouldn't finish
and gets stuck, this fixes that by forcing at most 50 concurrent git jobs for packer

more info:
https://github.com/wbthomason/packer.nvim/issues/120

### DIFF
--- a/lua/lvim/plugin-loader.lua
+++ b/lua/lvim/plugin-loader.lua
@@ -23,6 +23,7 @@ function plugin_loader.init(opts)
     compile_path = compile_path,
     log = { level = "warn" },
     git = { clone_timeout = 300 },
+    max_jobs = 50,
     display = {
       open_fn = function()
         return require("packer.util").float { border = "rounded" }


kylo252:
Wouldn't that spawn 50 threads? I think it's meant to be something a little bit more manageable 😄 
Maybe try 4 and see if works for well.

abzcoding:
> Wouldn't that spawn 50 threads? I think it's meant to be something a little bit more manageable 😄
> Maybe try 4 and see if works for well.

50 is still  better than the current situation (which is unlimited 😅 )

abzcoding:
Screen Shot 2021-11-07 at 7 27 58 PM

chase:
Maybe I should put in a PR for Packer to do something sane like using a tarball for github repos under a config flag.
That would probably make downloads significantly faster. I'd argue that unless a repo is explicitly set to not do that, there's no reason to even get the git objects.

kylo252 (approved):

PR #1934, opened 2021-11-09
fix(lsp): correct map and prevent highlight leak of diagnostic signs in 0.6

### LINKED ISSUE #1880: [Bug]: Error executing vim.schedule lua callback: Vim:E849: Too many highlight and syntax groups
### Problem description

Open c++ project.
Code as normal, after 10min, this error show on every keystoke

I have to close neovim, and start again. But It happen again and agian every 10 min

### LunarVim version

lastest rolling 377cab434c79aa537820ba21bbcc44261304ed09

### Neovim version (>= 0.5)

NVIM v0.6.0-dev+493-g035d82e0d

### Relevant log output

```shell
:message
Error executing vim.schedule lua callback: Vim:E849: Too many highlight and syntax groups
Error executing vim.schedule lua callback: Vim:E849: Too many highlight and syntax groups
```

### Screenshots

Untitled

### I have

- [X] Read the readme
- [X] Read the wiki

chase:
# Description

For Neovim nightly builds (versions greater than 0.5.1), every time the LSP server published diagnostics (quite frequently in my case), it would define and re-define a sign, leaving hundreds of cleared highlight groups.

Additionally, because position was being used instead of mapping by name, the highlight groups were incorrect.

This PR fixes both of these problems.

Fixes #1880

## How Has This Been Tested?

Use neovim nightly, open a decently sized document that has a supported and setup LSP.
Make an error that triggers a lot of diagnostic errors (for example an `{` in the middle of a function), then trigger cmp autocomplete and scroll through. This error should no longer occur while using this PR.

### DIFF
--- a/lua/lvim/lsp/init.lua
+++ b/lua/lvim/lsp/init.lua
@@ -136,6 +136,13 @@ function M.get_common_opts()
   }
 end
 
+local LSP_DEPRECATED_SIGN_MAP = {
+  ["LspDiagnosticsSignError"] = "DiagnosticSignError",
+  ["LspDiagnosticsSignWarning"] = "DiagnosticSignWarn",
+  ["LspDiagnosticsSignHint"] = "DiagnosticSignHint",
+  ["LspDiagnosticsSignInformation"] = "DiagnosticSignInfo",
+}
+
 function M.setup()
   Log:debug "Setting up LSP support"
 
@@ -144,7 +151,13 @@ function M.setup()
     return
   end
 
+  local is_neovim_nightly = vim.fn.has "nvim-0.5.1" > 0
+
   for _, sign in ipairs(lvim.lsp.diagnostics.signs.values) do
+    local lsp_sign_name = LSP_DEPRECATED_SIGN_MAP[sign.name]
+    if is_neovim_nightly and lsp_sign_name then
+      vim.fn.sign_define(lsp_sign_name, { texthl = lsp_sign_name, text = sign.text, numhl = lsp_sign_name })
+    end
     vim.fn.sign_define(sign.name, { texthl = sign.name, text = sign.text, numhl = sign.name })
   end
 
--- a/lua/lvim/lsp/handlers.lua
+++ b/lua/lvim/lsp/handlers.lua
@@ -36,16 +36,6 @@ function M.setup()
         if not vim.api.nvim_buf_is_loaded(bufnr) then
           return
         end
-
-        local sign_names = {
-          "DiagnosticSignError",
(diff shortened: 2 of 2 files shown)


chase:
All failed checks are from previous commits, FYI

chase:
@maxxnino can you reproduce the error you were having before (#1880) while using the branch for this PR?

maxxnino:
@chase i can confirm. it solved the problem

lua/lvim/lsp/init.lua
```diff
@@ -136,6 +136,13 @@ function M.get_common_opts()
   }
 end
 
+local NVIM_05_TO_06_SIGN_NAMES = {
```
kylo252, line 139:
```suggestion
local LSP_SIGN_MAP = {
```

lua/lvim/lsp/init.lua
```diff
@@ -144,7 +151,13 @@ function M.setup()
...
 
+  local isNvim06 = vim.fn.has "nvim-0.5.1" > 0
+
   for _, sign in ipairs(lvim.lsp.diagnostics.signs.values) do
+    local nvim06SignName = NVIM_05_TO_06_SIGN_NAMES[sign.name]
+    if isNvim06 and nvim06SignName then
+      vim.fn.sign_define(nvim06SignName, { texthl = nvim06SignName, text = sign.text, numhl = nvim06SignName })
+    end
     vim.fn.sign_define(sign.name, { texthl = sign.name, text = sign.text, numhl = sign.name })
   end
```
kylo252, line 162:
```suggestion
  local is_neovim_nightly = vim.fn.has "nvim-0.5.1" > 0

  for _, sign in ipairs(lvim.lsp.diagnostics.signs.values) do
    local  lsp_sign_name = LSP_SIGN_MAP[sign.name]
    if is_neovim_nightly and lsp_sign_name then
      vim.fn.sign_define(lsp_sign_name, { texthl = lsp_sign_name, text = sign.text, numhl = lsp_sign_name })
    else
      vim.fn.sign_define(sign.name, { texthl = sign.name, text = sign.text, numhl = sign.name })
    end
  end
```

kylo252 (changes requested):

lua/lvim/lsp/init.lua
kylo252: Don't we need the `else` here?

chase: No, the deprecated names for signs still need to be defined for plugins that haven't been updated yet.

kylo252 (approved):

PR #1935, opened 2021-11-09
fix(cmp): if possible jump in the snippet after completion

chase:
# Description

cmp changed some behaviors so that when pressing `<CR>` for a completion, if it is within a snippet it no longer jumps to the next field if possible.

This PR fixes this behavior so that if it is within a snippet, it _will_ jump to the next field if possible.
It also fixes replacement/insertion bugs in cmp by moving the pinned commit to a fixed version, without which causes completions and snippets to misbehave.

## How Has This Been Tested?

In a lua file, while using the sumneko_lua LSP server, type `for` and select the `for .. ipairs` snippet.
Press: `x` `<CR>` `y` `<CR>`
Type: `vi` and select `vim`
Press: `<CR>`
It should jump into the body of the snippet, between `for` and `end`


chase:
All failed checks are from previous commits, FYI

Tastyep (approved):

Issue #1684, opened 2021-10-04
[Feature]: Option to set formatting timeout duration


nalbury:
### Feature motivation

The terraform-ls language server has a built in formatter but its so slow that it times out on most files.

The null-ls formatter works, but forces the user to choose which formatter to use on each save if its used alongside the language server.

Using the language server's formatter by itself fails silently, but I noticed it logs:
```
[ ERROR ] 2021-10-04T06:11:47-0700 ] ...llar/neovim/0.5.0/share/nvim/runtime/lua/vim/lsp/rpc.lua:462 ]	"rpc"	"/Users/nalbury/.local/share/nvim/lsp_servers/terraform/terraform-ls/terraform-ls"	"stderr"	'2021/10/04 06:11:47 rpc_logger.go:45: Error for "textDocument/formatting" (ID 3): [-32800] error code -32800: Execution of "Format" canceled\n'
[ ERROR ] 2021-10-04T06:11:47-0700 ] ...llar/neovim/0.5.0/share/nvim/runtime/lua/vim/lsp/rpc.lua:462 ]	"rpc"	"/Users/nalbury/.local/share/nvim/lsp_servers/terraform/terraform-ls/terraform-ls"	"stderr"	"2021/10/04 06:11:47 server.go:248: Completed 1 requests [1.205960166s elapsed]\n"
```
This seems somewhat benign but I noticed the duration time (1.2s) and I believe the underlying function being called defaults to 1s(1000ms). 

I looked at the source [here](https://github.com/LunarVim/LunarVim/blob/d01ba08eaec1640ac2d038893525b3ba0af25813/lua/utils/init.lua#L61) and then tried disabling the auto format option, and manually configured a `BufWritePre` with much longer timeout:
```
lvim.autocommands.custom_groups = {
  { "BufWritePre", "*.tf", "lua vim.lsp.buf.formatting_sync(nil, 6000)" },
}
```
Which worked beautifully, but it looks like it would require a configuration for each filetype I want auto formatting on(?), so I was hoping for a way to set that globally.

### Describe the expected solution

Something like `lvim.format_on_save_timeout = 6000` would be sufficient for me. But I also would be happy with being able to enable the global option and override the BufWritePre cmd on a case by case basis ([at the moment they're only used if format on save is disabled](https://github.com/LunarVim/LunarVim/blob/d01ba08eaec1640ac2d038893525b3ba0af25813/lua/utils/init.lua#L76)). 

### Describe the alternatives you've considered

[...]

PR #1937, opened 2021-11-10
refactor: more configurable format-on-save

### LINKED ISSUE #1561: [Feature]: Disable format_on_save for specific file types
### Feature motivation

I use Intelephense for PHP development and by default it's set to format on save. I confirmed this by disabling any other PHP linters and formatters and enabled only Intelephense. When I save the file it's automatically formatted.

The easiest solution seems to be to allow people to disable autoformat for specific file types, e.g. via a `lvim.lang.php.format_on_save = false`.

I still want format on save enabled for other languages, e.g. JS/TS, so disabling `format_on_save` for the every type of file is a bit too excessive.

It's also a bit counter-intuitive to have no formatters configured in `lvim.lang.php.formatters` and my php files are still formatted on save.

Allow me to disable `format_on_save` for specific file types.

### Manually configure the language server

I tried changing the settings by adding this in my `.config/lvim/config.lua`:

```lua
lvim.lang.php.lsp.settings = {
  intelephense = {
    format = {
      enable = false,
    },
  },
}
```

But that didn't work. I tried other variations of this, e.g. `lvim.lang.php.lsp.settings.intelephense.format.enable = false`, but that didn't work either.

### Additional context

_No response_

kylo252:
# Description

- add `lvim.format_on_save.pattern`
- add `lvim.format_on_save.timeout_ms`
- add `LvimToggleFormatOnSave`

Fixes #1561, #1684

## How Has This Been Tested?

```lua
lvim.format_on_save = {
	pattern = "*.lua"
}
```

You can also try `:LvimToggleFormatOnSave` to allow saving files without formatting.

### DIFF
--- a/lua/lvim/config/defaults.lua
+++ b/lua/lvim/config/defaults.lua
@@ -3,7 +3,12 @@ return {
   colorscheme = "onedarker",
   line_wrap_cursor_movement = true,
   transparent_window = false,
-  format_on_save = true,
+  format_on_save = {
+    ---@usage pattern string pattern used for the autocommand (Default: '*')
+    pattern = "*",
+    ---@usage timeout number timeout in ms for the format request (Default: 1000)
+    timeout = 1000,
+  },
   keys = {},
 
   builtin = {},
--- a/lua/lvim/config/init.lua
+++ b/lua/lvim/config/init.lua
@@ -20,7 +20,7 @@ end
 -- Define lvim global variable
 function M:init()
   if vim.tbl_isempty(lvim or {}) then
-    lvim = require "lvim.config.defaults"
+    lvim = vim.deepcopy(require "lvim.config.defaults")
     local home_dir = vim.loop.os_homedir()
     lvim.vsnip_dir = utils.join_paths(home_dir, ".config", "snippets")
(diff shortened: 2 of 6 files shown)


chase:
@kylo252 Needs a rebase to rolling, then I'll review

kylo252:
@rebebop, feel free to experiment with the timeout length, as well as excluding the filetype where the formatter is slow.

chase (approved):
I think this is an improvement, but I think it is unclear that `format_on_save_pattern` is actually an `autocmd-pattern`: https://neovim.io/doc/user/autocmd.html#autocmd-pattern

This should be documented more clearly in the README that this is the case, since it is a special syntax with lots of possibilities, but some unexpected behavior other than regular globs. For example you can annotate two file extensions using `*.c,*.lua` but you can also use full paths with vim variables and other things `$VIMRUNTIME/doc/*.txt`

dylan-chong:
Is there a way to enable format on save for all files except a particular type?

He1pa:
> Is there a way to enable format on save for all files except a particular type?

I've encountered the same problem, do you have any solution now?

dylan-chong:
just manually list the file types i want to format

```
lvim.format_on_save.enabled = true
lvim.format_on_save.pattern = '*.ts,*.tsx,*.js,*.jsx,*.lua'
```

PR #1941, opened 2021-11-10
fix(keymap): unset on reload, load correct order, add unset default

chase:
# Description

Prior to this PR:

- A key mapping remains after reloading even after removing the line setting it
- Key mappings were loading using the user's config before being reloaded
- There was no way to easily unmap a default key mapping

After this PR:

- Removing the line setting a key mapping and triggering a reload will unbind that key mapping
- Key mappings are loaded after the user's config is loaded
- Setting a key mapping to false will explicitly unset that binding

## How Has This Been Tested?

Open `config.lua`, set the following and save:
```
lvim.keys = {
  normal_mode = {
      ["<C-b>"] = "<cmd>lua print('hi!')<cr>",
      ["<A-j"] = false
  }
}
```
`:nmap <A-j>` will show that `<A-j>` has been unset
`:nmap <C-b>` will show that `<C-b>` has been set

Set to the following (remove the two normal mode binding lines):
```
lvim.keys = {
  normal_mode = {
  }
}
```
`:nmap <A-j>` will show that `<A-j>` is the default binding
`:nmap <C-b>` will show that `<C-b>` has been unset

### DIFF
--- a/lua/lvim/config/init.lua
+++ b/lua/lvim/config/init.lua
@@ -32,9 +32,6 @@ function M:init()
   local settings = require "lvim.config.settings"
   settings.load_options()
 
-  local default_keymaps = require("lvim.keymappings").get_defaults()
-  lvim.keys = apply_defaults(lvim.keys, default_keymaps)
-
   local autocmds = require "lvim.core.autocmds"
   lvim.autocommands = apply_defaults(lvim.autocommands, autocmds.load_augroups())
 
@@ -89,6 +86,9 @@ function M:load(config_path)
   autocmds.define_augroups(lvim.autocommands)
 
   vim.g.mapleader = (lvim.leader == "space" and " ") or lvim.leader
+
+  local default_keymaps = require("lvim.keymappings").get_defaults()
+  lvim.keys = apply_defaults(lvim.keys, default_keymaps)
   require("lvim.keymappings").load(lvim.keys)
 
   local settings = require "lvim.config.settings"
@@ -98,6 +98,8 @@ end
 --- Override the configuration with a user provided one
 -- @param config_path The path to the configuration overrides
 function M:reload()
+  require("lvim.keymappings").clear(lvim.keys)
+
   local lvim_modules = {}
   for module, _ in pairs(package.loaded) do
     if module:match "lvim.core" then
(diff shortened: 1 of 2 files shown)


lua/lvim/config/init.lua
```diff
@@ -98,6 +98,8 @@ end
 --- Override the configuration with a user provided one
 -- @param config_path The path to the configuration overrides
 function M:reload()
+  require("lvim.keymappings").clear(lvim.keys)
```
kylo252, line 101: I think you missed a critical use-case where the keymaps are overridden by a plugin and not the user. What's the plan then? Unless you reload _all_ the plugins, then I'm not sure if it's even possible.

lua/lvim/keymappings.lua
```diff
@@ -24,13 +24,27 @@ local mode_adapters = {
...
   end
 end
 
+-- Unsets all keybindings defined in keymaps
+-- @param keymaps The table of key mappings containing a list per mode (normal_mode, insert_mode, ..)
+function M.clear(keymaps)
+  for mode, mappings in pairs(keymaps) do
+    mode = mode_adapters[mode] and mode_adapters[mode] or mode
+    for key, _ in pairs(mappings) do
+      pcall(vim.api.nvim_del_keymap, mode, key)
```
kylo252, line 43: Actually, that could be solved here, we can just compare the values and only delete it if it's the default one.

kylo252 (changes requested):

lua/lvim/keymappings.lua
chase: This only clears what has been set in `lvim.keys` not _all_ keybindings, so that's more or less the current behavior.

lua/lvim/config/init.lua
kylo252: It's more like I don't want to define the plugin's mapping manually _if possible_. My use case is specifically about `<C-hjkl>` as they're usually used by tmux-related plugins.

chase: Okay, so your specific use case is that you are using a plugin that overrides a default which you aren't defining in your `lvim.keys`

So basically the fix here would be to _not_ unset any default key mapping

kylo252: Still the same problem

https://user-images.githubusercontent.com/59826753/141161515-2910d313-ba04-4c7f-90f5-507279b1293b.mp4

kylo252 (approved):
It's an improvement for sure, even if it doesn't solve all the issues. Thank you!

PR #1945, opened 2021-11-11
feat: add some messages in uninstall.sh

Spaxly:
A simple, basic uninstall message

### DIFF
--- a/utils/installer/uninstall.sh
+++ b/utils/installer/uninstall.sh
@@ -55,4 +55,5 @@ function main() {
   rm -f "$LVIM_BIN"
 }
 
+echo "Uninstalled LunarVim!"
 main "$@"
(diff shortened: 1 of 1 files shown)


utils/installer/uninstall.sh
```diff
@@ -55,4 +55,5 @@ function main() {
   rm -f "$LVIM_BIN"
 }
 
+echo "Uninstalled LunarVim!"
```
kylo252, line 58: It should be moved inside the main function, otherwise it will always print that, even for `--help` 
```suggestion
  rm -f "$LVIM_BIN"
  echo "Uninstalled LunarVim!"
}

```

kylo252 (changes requested):

kylo252 (approved):

PR #1949, opened 2021-11-12
Fix example of unmapping an lvim default

linsomniac:
The example config file has a sample for unmapping a LunarVim key by
setting it to "".  When I tried that, the result was that the key became
a no-op.  If I instead set it to nil, the key returned to default vim
operation.  In particular, I was trying to unmap normal mode L and H.  I
got to nil by searching for how to delete a entry in a dictionary in
lua, but I'm not familiar with lua.

# Description

Fixes #(issue)

## How Has This Been Tested?

- Run command `:mycommand`
- Check logs
- ...

### DIFF
--- a/utils/installer/config.example.lua
+++ b/utils/installer/config.example.lua
@@ -18,7 +18,7 @@ lvim.leader = "space"
 -- add your own keymapping
 lvim.keys.normal_mode["<C-s>"] = ":w<cr>"
 -- unmap a default keymapping
--- lvim.keys.normal_mode["<C-Up>"] = nil
+-- lvim.keys.normal_mode["<C-Up>"] = false
 -- edit a default keymapping
 -- lvim.keys.normal_mode["<C-q>"] = ":q<cr>"
 
--- a/utils/installer/config.example.lua
+++ b/utils/installer/config.example.lua
@@ -18,7 +18,7 @@ lvim.leader = "space"
 -- add your own keymapping
 lvim.keys.normal_mode["<C-s>"] = ":w<cr>"
 -- unmap a default keymapping
--- lvim.keys.normal_mode["<C-Up>"] = ""
+-- lvim.keys.normal_mode["<C-Up>"] = nil
 -- edit a default keymapping
 -- lvim.keys.normal_mode["<C-q>"] = ":q<cr>"
 


chase:
@linsomniac `nil` is also incorrect (which is frustrating!)

This is a very recent change (just yesterday in https://github.com/LunarVim/LunarVim/commit/627fdd60493860c9c5194d8ec0a70ec5aaa45bf8) but it should be set to `false`.
Setting a table in lua to `nil` actually unsets that entry instead of having it hold a null-ish value.

For an example see:
https://github.com/LunarVim/LunarVim/pull/1941

kylo252:
This is not necessary after #1965, sorry for the inconvenience! 😅

chase:
@kylo252 I think this is still necessary, because the existing example still sets a no-op instead of clearing, right?

kylo252:
> @kylo252 I think this is still necessary, because the existing example still sets a no-op instead of clearing, right?

They're in a local table now, so you'd have to overwrite the binding either way, to no-op or something actually useful.

chase:
> > @kylo252 I think this is still necessary, because the existing example still sets a no-op instead of clearing, right?
> 
> They're in a local table now, so you'd have to overwrite the binding either way, to no-op or something actually useful.

The point of this is to undo a default such that the original keybinding is restored. Setting to a no-op in this case is not desired, in this case, unsettling it is.

kylo252:
> The point of this is to undo a default such that the original keybinding is restored. Setting to a no-op in this case is not desired, in this case, unsettling it is.

In that case, using `unmap` or `nvim_del_keymap()` should do the trick. You can test it with `:unmap H`.

chase:
> In that case, using `unmap` or `nvim_del_keymap()` should do the trick. You can test it with `:unmap H`.

Correct, however this is not reflected in the README or in the config template. The existing setting/example which sets a no-op should be updated, as this PR attempted to do.

utils/installer/config.example.lua
```diff
@@ -18,7 +18,7 @@ lvim.leader = "space"
 -- add your own keymapping
 lvim.keys.normal_mode["<C-s>"] = ":w<cr>"
 -- unmap a default keymapping
--- lvim.keys.normal_mode["<C-Up>"] = ""
+-- lvim.keys.normal_mode["<C-Up>"] = false
```
kylo252, line 21:
```suggestion
-- vim.cmd [[ unmap <C-Up> ]]
```

kylo252 (changes requested):

utils/installer/config.example.lua
chase: Setting to false also functions properly still, so I guess do we want to advocate a Lua-based or a Vimscript-based solution like the above?

linsomniac: Since it's an example config, I'd say there could be a case for including both so that the user can pick the one that works best for them.  But, there's also a case for picking the most lvim-best-practices way.

kylo252:
Thank you so much, @linsomniac!

I went ahead and fixed the commit messages so that it passes the CI.

PR #1955, opened 2021-11-13
feat: support new null-ls

ChristianChiarulli:


ChristianChiarulli:
@kylo252 Is this working now?

kylo252:
I think it should be good now, but I had to pin back a commit that was causing `NullLsInfo` to fail when registering a source with args.

PR #1965, opened 2021-11-17
refactor: load the default keymaps once

### LINKED ISSUE #1548: [Feature]: Keymapping Tables
### Feature motivation

This is a follow up to #1525. The problem I was experiencing in that issue was that some of my keyboard mappings were randomly not applied 50% of the time. Here are a few examples of settings that were only working half the time:

``` lua
lvim.keys.normal_mode["H"] = "gT"
lvim.keys.normal_mode["L"] = "gt"
lvim.builtin.telescope.defaults.mappings.i["<cr>"] = actions.file_tab
```

After a ton of digging, I finally figured out what was going on. Internally, LunarVim was using different forms of these mappings:

[keymappings.lua](https://github.com/LunarVim/LunarVim/blob/rolling/lua/keymappings.lua#L105-L106):

 ``` lua
["<S-l>"] = ":BufferNext<CR>",
["<S-h>"] = ":BufferPrevious<CR>",
```

[telescope.lua](https://github.com/LunarVim/LunarVim/blob/rolling/lua/core/telescope.lua#L58):

``` lua
["<CR>"] = actions.select_default + actions.center
```

When my config is loaded by LunarVim and converted to NeoVim mappings, the iteration order seems to be non-deterministic, which causes my mappings to be randomly applied.

I'd propose a special subclass of Lua's table that can handle keymappings. I'd expect this class to pick a standard for keys (both in terms of casing and modifiers), and to be used anywhere keymappings are used. The only two places that I've encountered are in the normal `lvim.keys.<mode>` maps, as well as `telescope.defaults.mappings`, but there could be others.

Since mappings are not case (or form) sensitive in Neovim, this would follow the editor's normal behavior.

I would expect this to be able to handle the following cases (and maybe more I'm not considering):

``` lua
-- These should all be assigned to the *same value* inside the table. They should override each other.
lvim.keys.normal_mode["B"] = ":echo Banana<cr>"
lvim.keys.normal_mode["<S-b>"] = ":echo Basketball<cr>"
lvim.keys.normal_mode["<s-b>"] = ":echo Binoculars<cr>"

-- Each one of these should return the exact same value.
lvim.keys.normal_mode["B"] -- ":echo Binoculars<cr>"
lvim.keys.normal_mode["<S-b>"] -- ":echo Binoculars<cr>"
lvim.keys.normal_mode["<s-b>"] -- ":echo Binoculars<cr>"

-- When iterating over the keys, the mapping should only be selected once. I don't think it matters
-- which format is used for key by LunarVim, as long as it's programmatically consistent. For
-- example, the following should output something like this:
--
-- B: :echo Binoculars<cr>
for key, value in pairs(lvim.keys.normal_mode) do
  print(key .. ": " .. value)
end
```

I could update my config to match the internal implementations (which is probably how I'll resolve my immediate issue). However, I don't think anyone should have to know how the keymappings are defined in order to override them. Furthermore, the keymappings are not defined consistently internally (and they shouldn't need to be.) This will likely be a confusing mess for anyone else who tries to override LunarVim case-sensitive or modifier mappings and happens to pick the wrong form.

### Additional context

This took me several days of debugging to identify the root cause. 😔 Had I not found the root cause, I would have had to drop LunarVim as my editor to maintain my productivity.

kylo252:
# Description

- `lvim.keys` no longer holds the default values
- only load the default keymaps once
- `lvim.keys` is always loaded after the default values, so it will always take precedence
- configurable which-key registration for `dashboard` and `nvimtree`

Fixes #1548, #1755, 

## How Has This Been Tested?

- These should all be assigned to the *same value* inside the table. They should override each other, so the result will always be `Binoculars`

```lua
lvim.keys.normal_mode["B"] = ":echo 'Banana'<cr>"
lvim.keys.normal_mode["<S-b>"] = ":echo 'Basketball'<cr>"
lvim.keys.normal_mode["<s-b>"] = ":echo 'Binoculars'<cr>"
```

- it's also possible to override NvimTree's binding directly now

```lua
lvim.keys.normal_mode["<space>e"] = ":echo 'hello'<cr>"
```

### DIFF
--- a/lua/lvim/config/init.lua
+++ b/lua/lvim/config/init.lua
@@ -26,7 +26,7 @@ function M:init()
     lvim.database = { save_location = utils.join_paths(home_dir, ".config", "lunarvim_db"), auto_execute = 1 }
   end
 
-  lvim.keys = apply_defaults(lvim.keys, require("lvim.keymappings").get_defaults())
+  require("lvim.keymappings").load_defaults()
 
   local builtins = require "lvim.core.builtins"
   builtins.config { user_config_file = user_config_file }
@@ -89,8 +89,6 @@ function M:load(config_path)
 
   vim.g.mapleader = (lvim.leader == "space" and " ") or lvim.leader
 
-  local default_keymaps = require("lvim.keymappings").get_defaults()
-  lvim.keys = apply_defaults(lvim.keys, default_keymaps)
   require("lvim.keymappings").load(lvim.keys)
 
   local settings = require "lvim.config.settings"
@@ -100,8 +98,6 @@ end
 --- Override the configuration with a user provided one
 -- @param config_path The path to the configuration overrides
 function M:reload()
-  require("lvim.keymappings").clear(lvim.keys)
-
   local lvim_modules = {}
   for module, _ in pairs(package.loaded) do
     if module:match "lvim.core" then
--- a/lua/lvim/core/bufferline.lua
+++ b/lua/lvim/core/bufferline.lua
@@ -5,17 +5,13 @@ M.config = function()
     active = true,
     on_config_done = nil,
     keymap = {
-      normal_mode = {
-        ["<S-l>"] = ":BufferNext<CR>",
-        ["<S-h>"] = ":BufferPrevious<CR>",
-      },
+      normal_mode = {},
     },
   }
 end
 
 M.setup = function()
-  local keymap = require "lvim.keymappings"
-  keymap.append_to_defaults(lvim.builtin.bufferline.keymap)
+  require("lvim.keymappings").load(lvim.builtin.bufferline.keymap)
 
   if lvim.builtin.bufferline.on_config_done then
     lvim.builtin.bufferline.on_config_done()
--- a/lua/lvim/core/dashboard.lua
+++ b/lua/lvim/core/dashboard.lua
@@ -58,6 +58,7 @@ M.config = function(config)
 
     footer = { "lunarvim.org" },
   }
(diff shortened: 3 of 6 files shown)


LandonSchropp:
@kylo252 This is a much better solution than what I was thinking with keymapping tables. 🙂

kylo252:
> This is a much better solution than what I was thinking with keymapping tables. 🙂

I started looking into your suggestion, but it requires translating the keycodes back and forth to make sure that the keys aren't actually defined already. Furthermore, these APIs aren't implemented in lua just yet, so we need to call vimL function (`vim.fn`) for every single item in the list, and that needs to be done at startup. 

You can see that there is a sizeable overhead for implementing that kind of validation, and without any real benefit. So I figured it's better to extract the defaults from the table altogether, and instead only set them up once. This means that any subsequent assignment to the same key will always take effect, and there's no chance of the defaults interfering again 😃

abzcoding (approved):
it seems to be working fine, tyvm <3

chase (approved):

gonzaloserrano:
Works fine here too, thanks!

meijieru:
I meet some problems with the commit.
I couldn't not disable the internal kemaps via `lvim.keys.normal_mode = {}` as from the [doc](https://www.lunarvim.org/configuration/02-keybindings.html#lunarvim-keybindings).

kylo252:
@meijieru, I don't see how the docs mention that what you're trying would work, but either way, please keep an eye on #1949.

PR #1967, opened 2021-11-18
fix(windows): autocmd requires forward slashes

### LINKED ISSUE #1959: [Bug]: autocmd for reloading `config.lua` do not work in windows
#### Steps

1. Open `config.lua`
2. Make some changes and save it 

#### Expected behaviour

LunarVim will be reloaded through the autocmd:
```lua
      { "BufWritePost", user_config_file, "lua require('lvim.config'):reload()" },
```

#### Actural behaviour
Reloading didn't happen

### LunarVim version

0.5.1

### Neovim version (>= 0.5)

0.5.1

### Relevant log output

_No response_

### Screenshots

_No response_

### I have

- [X] Read the readme
- [X] Read the wiki

kylo252:
Fixes #1959

### DIFF
--- a/lua/lvim/core/autocmds.lua
+++ b/lua/lvim/core/autocmds.lua
@@ -3,7 +3,12 @@ local Log = require "lvim.core.log"
 
 --- Load the default set of autogroups and autocommands.
 function M.load_augroups()
-  local user_config_file = vim.fn.resolve(require("lvim.config"):get_user_config_path())
+  local user_config_file = require("lvim.config"):get_user_config_path()
+
+  if vim.loop.os_uname().version:match "Windows" then
+    -- autocmds require forward slashes even on windows
+    user_config_file = vim.fn.resolve(user_config_file:gsub("\\", "/"))
+  end
 
   return {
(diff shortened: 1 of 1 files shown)


kylo252:
@younger-1, can you please test this?

younger-1:
It works on my machine 😄

younger-1:
@kylo252 Sorry I didn't make it clear. It's better not to use `vim.fn.resolve` and set it like this:

```lua
  if vim.loop.os_uname().version:match "Windows" then  
    -- autocmds require forward slashes even on windows
    user_config_file = user_config_file:gsub("\\", "/")
  end
```

Otherwise it will cause error as `string.gsub()` has two return values.
```
E5113: Error while calling lua chunk: Vim:E118: Too many arguments for function: resolve
```

kylo252:
> Otherwise it will cause error as `string.gsub()` has two return values.
> 
> ```
> E5113: Error while calling lua chunk: Vim:E118: Too many arguments for function: resolve
> ```

I'm not getting those errors for this specific usage.

PR #1977, opened 2021-11-20
feat: add support for fsharp

grenaad:
# Description

This add's support for the language F#. The LSP is called `fsautocomplete` and is available in [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#fsautocomplete) and in [nvm-lsp-install](https://github.com/williamboman/nvim-lsp-installer/blob/main/lua/nvim-lsp-installer/servers/init.lua#L56)
This should be available with `LspInstall`. 

## How Has This Been Tested?

Added the auto command to a clean lvim installation and opened a fsharp project work with. The language type was correctly identified.


lua/lvim/config/supported_languages.lua
```diff
@@ -26,6 +26,7 @@ return {
   "fennel",
   "fish",
   "fortran",
+  "fsautocomplete",
```
kylo252, line 29:
```suggestion
  "fsharp",
```

kylo252 (changes requested):
Sorry for the late review, I had to update the installer dependencies before this could work.

It's sufficient with adding the `ftdetect/fsharp.lua` file, but it doesn't hurt to add `fsharp` to the list of supported languages as well, even though it's mostly redundant now.

lua/lvim/config/supported_languages.lua
grenaad: No worries. I made the suggested change. I'm glad to hear I was not going crazy because I could not see how the `supported_languages.lua` are been used.

kylo252:
All the PRs should go into rolling instead of master, I'll grab your commits and merge them cleanly while also fixing the complaints from the CI.

grenaad:
> All the PRs should go into rolling instead of master, I'll grab your commits and merge them cleanly while also fixing the complaints from the CI.

Thanx 💯

PR #1981, opened 2021-11-21
fix(lsp): avoid installing an overridden server

### LINKED ISSUE #1972: Give ability to use external lsp server via just config override (ex: "cmd") without the server from Lsp-installer
### Problem description

if you don't have the server installed via lsp-installer then you can't override the config, but we are using a external server via the modified config

I know we can directly config servers without using `lvim.lsp.manager`
but if someone tried this way i thought it would work, they would be just stuck.

if we config it like..
```
lvim.lsp.automatic_servers_installation = false
vim.list_extend(lvim.lsp.override, { "sumneko_lua", "tsserver" })
```
add config override to -> `lua/lvim/lsp/providers/sumneko_lua.lua`

run `:LvimCacheReset`

what's happening in these conditions is it ends up here 
`Log:debug(requested_server.name...` 

> https://github.com/LunarVim/LunarVim/blob/fe99ee6e42ddec225c46a9d1fcda5c5d76396601/lua/lvim/lsp/manager.lua#L62-L85

,where there's nothing being done to config the server
since we aren't using auto installer, 
but it just wait until the requested server is ready, which is never going to happen since we aren't installing it

So Docs should include that if you aren't using lsp-installer then only use 
`require("lspconfig")["pyright"].setup(opts)` type of manual setup calls..

Hope this help someone save their time...

### LunarVim version

rolling

### Neovim version (>= 0.5)

0.6 dev

### Relevant log output

_No response_

### Screenshots

_No response_

### I have

- [X] Read the readme
- [X] Read the wiki

kylo252:
# Description

- Avoid installing an overridden server. 
- call `lspconfig` directly for overridden servers.

Fixes #1972

## How Has This Been Tested?

This should now work.

```lua
lvim.lsp.automatic_servers_installation = false
vim.list_extend(lvim.lsp.override, { "clangd" })
local opts = {}
require("lvim.lsp.manager").setup("clangd", opts)
```

### DIFF
--- a/lua/lvim/lsp/manager.lua
+++ b/lua/lvim/lsp/manager.lua
@@ -54,39 +54,41 @@ function M.setup(server_name, user_config)
   if lvim_lsp_utils.is_client_active(server_name) then
     return
   end
-  local servers = require "nvim-lsp-installer.servers"
 
   local config = resolve_config(server_name, user_config)
+
+  local servers = require "nvim-lsp-installer.servers"
   local server_available, requested_server = servers.get_server(server_name)
 
-  if server_available then
-    local install_notification = false
+  local is_overridden = vim.tbl_contains(lvim.lsp.override, server_name)
 
-    if not requested_server:is_installed() then
-      if lvim.lsp.automatic_servers_installation then
-        Log:debug "Automatic server installation detected"
-        requested_server:install()
-        install_notification = true
-      else
-        Log:debug(requested_server.name .. " is not managed by the automatic installer")
-      end
-    end
-
-    requested_server:on_ready(function()
-      if install_notification then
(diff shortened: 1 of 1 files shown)


abzcoding:
what happens to `has_custom_provider, _ = pcall(require, "lvim/lsp/providers/" .. server_name)` ?

kylo252:
> what happens to `has_custom_provider, _ = pcall(require, "lvim/lsp/providers/" .. server_name)` ?

We don't really care about that anymore, according to @chase. Since it was done before the refactor in https://github.com/LunarVim/LunarVim/pull/1863. The point of this PR is to make it so that we are skipping the installer check if the server is overridden.

abzcoding (approved):

PR #1989, opened 2021-11-24
chore: bump core-plugins version

kylo252:
# Description

- chore: parsable core-plugins list
- chore: update nvim-tree's settings
- chore: add plugins' sha generator


kylo252:
@chase, I think it's better to split the PR, so we'll keep this to update the structure, and then we'll figure out a way to automate this. I also had to revert cmp's SHA since it was failing.

chase:
https://github.com/LunarVim/LunarVim/blob/24be0ef1ef36b8cc725655e920104e9676b72f25/lua/lvim/core/nvimtree.lua#L32

This is apparently deprecated now, but the filters option doesn't seem to work? Might want to pin nvimtree to a version before the bump.

kylo252:
> https://github.com/LunarVim/LunarVim/blob/24be0ef1ef36b8cc725655e920104e9676b72f25/lua/lvim/core/nvimtree.lua#L32
> 
> This is apparently deprecated now, but the filters option doesn't seem to work? Might want to pin nvimtree to a version before the bump.

I updated the setting specifically because the other one was deprecated 😄 
Make sure you've updated the binding in your `config.lua`, as it needs to be moved under `setup`: https://github.com/chase/lvim/blob/2e6c813d4dcf31cf5c1f0b4b3a1178c67aeb911b/config.lua#L190

Manndung75:
Báo

PR #1999, opened 2021-11-27
fix: correct order for cmp's setup

kylo252:
# Description

Use lazy-loading to avoid configuration errors during headless installations. 

chore: bump cmp's version
fix: lazy-load cmp's setup
fix: proper order for luasnip setup

Fixes some issues with luasnip's vscode-loader

## How Has This Been Tested?

Passes the CI

### DIFF
--- a/lua/lvim/core/cmp.lua
+++ b/lua/lvim/core/cmp.lua
@@ -146,8 +146,6 @@ end
 M.methods.jumpable = jumpable
 
 M.config = function()
-  lvim.builtin.cmp = {}
-
   local status_cmp_ok, cmp = pcall(require, "cmp")
   if not status_cmp_ok then
     return
--- a/lua/lvim/core/cmp.lua
+++ b/lua/lvim/core/cmp.lua
@@ -146,6 +146,8 @@ end
 M.methods.jumpable = jumpable
 
 M.config = function()
+  lvim.builtin.cmp = {}
+
   local status_cmp_ok, cmp = pcall(require, "cmp")
   if not status_cmp_ok then
     return
@@ -301,8 +303,7 @@ M.config = function()
   }
 end
 
-M.setup = function()
-  require("luasnip/loaders/from_vscode").lazy_load()
+function M.setup()
   require("cmp").setup(lvim.builtin.cmp)
 end
 
--- a/lua/lvim/plugins.lua
+++ b/lua/lvim/plugins.lua
@@ -15,7 +15,7 @@ local commit = {
   nlsp_settings = "1e75ac7733f6492b501a7594870cf75c4ee23e81",
   null_ls = "b07ce47f02c7b12ad65bfb4da215c6380228a959",
   nvim_autopairs = "fba2503bd8cd0d8861054523aae39c4ac0680c07",
-  nvim_cmp = "ca6386854982199a532150cf3bd711395475ebd2",
+  nvim_cmp = "092fb66b6ddb4b12b9b542d105d7af40e4fbd9f2",
   nvim_dap = "4e8bb7ca12dc8ca6f7a500cbb4ecea185665c7f1",
   nvim_lsp_installer = "52183c68baf9019c8241b1abf33ba0b6594ab3c8",
   nvim_lspconfig = "b53f89c16bcc8052aa56d3a903fcad3aaa774041",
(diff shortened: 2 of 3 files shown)


lua/lvim/core/cmp.lua
```diff
@@ -224,7 +224,9 @@ M.config = function()
     },
     snippet = {
       expand = function(args)
-        require("luasnip").lsp_expand(args.body)
+        pcall(function()
+          require("luasnip").lsp_expand(args.body)
+        end)
```
chase: This shouldn't be necessary as this should only be called when a user expands a snippet. This shouldn't be called in headless mode since there is no input.

lua/lvim/plugins.lua
```diff
@@ -76,6 +76,7 @@ return {
   -- Install nvim-cmp, and buffer source as a dependency
   {
     "hrsh7th/nvim-cmp",
+    event = "UIEnter",
```
chase: I'm a little surprised this fires for TUI since according to [the documentation](https://neovim.io/doc/user/autocmd.html#UIEnter) this is only for RPC based clients, not the default TUI.

Seems to work though :shrug:

lua/lvim/core/cmp.lua
```diff
@@ -1,3 +1,9 @@
+-- cmp's config requires cmp to be installed to run the first time
+local cmp_loaded, _ = pcall(require, "cmp")
+if not cmp_loaded then
+  return
+end
+
```
chase: This is redundant to https://github.com/LunarVim/LunarVim/blob/13f5cf436f279d882609a808654c1c5fe802d1ad/lua/lvim/core/cmp.lua#L155-L158

What errors are you encountering?

chase (changes requested):
I'm missing some context here: what are the errors you're encountering?

Some of these fixes seem redundant or unnecessary, so I feel like we're probably doing something fundamentally wrong with setting up the plugin if the config step is throwing errors.

kylo252:
> I'm missing some context here: what are the errors you're encountering?
> 
> Some of these fixes seem redundant or unnecessary, so I feel like we're probably doing something fundamentally wrong with setting up the plugin if the config step is throwing errors.

Here's what prompted this change, I can't actually bump the version because it's throwing an error in Packer during the installation. 

Even with my solution, it's still throwing an error but on startup instead of the installation.
https://github.com/LunarVim/LunarVim/runs/4340372153?check_suite_focus=true

chase:
Based on the error, it's due to `require("cmp").setup(lvim.builtin.cmp)` being called when `lvim.builtin.cmp` is nil.

This means that when either of these pcalls fail (in our case, during the installation for our tests), we have `lvim.builtin.cmp` set to nil: https://github.com/LunarVim/LunarVim/blob/rolling/lua/lvim/core/cmp.lua#L149-L156

I think we can assume the `cmp` dependency is satisfied, so that leaves the dependency on LuaSnip to be unsatisfied.

For nvim-cmp's entry, we can try adding a `requires = "L3MON4D3/LuaSnip"` to see if that resolves things:
https://github.com/LunarVim/LunarVim/blob/94c1a269cdf45a0a28280c272d67d1b26d3c5b71/lua/lvim/plugins.lua#L79

lua/lvim/plugins.lua
```diff
@@ -76,16 +76,14 @@ return {
...
     end,
-    run = function()
-      -- cmp's config requires cmp to be installed to run the first time
-      if not lvim.builtin.cmp then
-        require("lvim.core.cmp").config()
-      end
-    end,
+    requires = {
+      "L3MON4D3/LuaSnip",
+    },
```
kylo252, line 86: This combination seems to work, along with assigning an empty config regardless if cmp is available or not.
I had wanted to use "BufRead" or "InsertEnter" but I couldn't get it either to play well with Telescope which will be triggered from Dashboard.

@chase, what do you think?

chase: I guess I'm wondering why we're lazy loading nvim-cmp anyway, it just delays having it load in the cache which brings a much more significant speedup.

kylo252: I basically don't want it to try to run `setup()` while in headless mode, since it was conflicting with Packer for some reason. Otherwise, it's not really that useful.

I'll try to remove it once more to see if it's somehow sorted now.

chase (approved):

PR #2008, opened 2021-11-30
feat: null-ls code_actions interface

meijieru:
# Description

This PR adds the interface for null-ls code actions config, which has the same interface as formatters & linters.
Support for `:LvimInfo` is also added.

## How Has This Been Tested?

Having the following code in `config.lua`
```lua

  local code_actions = require "lvim.lsp.null-ls.code_actions"
  code_actions.setup {
    { name = "gitsigns" },
  }
```
Then run `:LvimInfo` or `:Telescope lsp_code_actions`.

image
image

### DIFF
--- a/lua/lvim/core/info.lua
+++ b/lua/lvim/core/info.lua
@@ -35,6 +35,23 @@ local function make_formatters_info(ft)
   return section
 end
 
+local function make_code_actions_info(ft)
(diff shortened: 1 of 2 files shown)


kylo252:
I don't really get the point with `gitsigns` but it looks useful in general, thank you!

PR #2018, opened 2021-12-01
chore: bump core-plugins' version

kylo252:


kylo252:
@abzcoding, see if this fixes the dupes' problem. I still can't reproduce it, btw.

abzcoding:
> @abzcoding, see if this fixes the dupes' problem. I still can't reproduce it, btw.

i'll test it asap, in the meantime try opening a python file in a git managed repo

abzcoding (approved):
it doesn't fix my duplicate lsp issue, but i think we should merge it anyway
tyvm!

PR #2019, opened 2021-12-01
fix: more accessible changelog

kylo252:
# Description

- Fix `lvim` version not displaying correctly on Dashboard when on rolling.
- Add `LvimVersion` that calls `get_changelog()`
- some enhancements to `get_changelog()` including copy to clipboard on `<CR>` and a better default previewer.

image

image


abzcoding (approved):
LGTM 🚀

PR #2020, opened 2021-12-01
fix: better default, ignore `.git` in `live_grep`

meijieru:
# Description

Ignore files in `.git` in `live_grep`.
I think it's a more sane default.

### DIFF
--- a/lua/lvim/core/telescope.lua
+++ b/lua/lvim/core/telescope.lua
@@ -37,6 +37,7 @@ function M.config()
         "--column",
         "--smart-case",
         "--hidden",
+        "--glob=!.git/",
       },
       mappings = {
         i = {


lua/lvim/core/telescope.lua
```diff
@@ -37,6 +37,8 @@ function M.config()
         "--column",
         "--smart-case",
         "--hidden",
+        "-g",
+        "!.git/"
```
kylo252, line 41:
```suggestion
        "--glob=!.git/"
```

kylo252 (changes requested):
I figured that most users have a `ripgreprc`, but I guess it doesn't hurt to have it here as well.

lua/lvim/core/telescope.lua
meijieru: Updated

kylo252:
Thank you!

PR #2021, opened 2021-12-02
feat: add support for fsharp

kylo252:
# Description

This add's support for the language F#. The LSP is called `fsautocomplete` and is available in [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#fsautocomplete) and in [nvm-lsp-install](https://github.com/williamboman/nvim-lsp-installer/blob/main/lua/nvim-lsp-installer/servers/init.lua#L56)
This should be available with `LspInstall`. 

## How Has This Been Tested?

Added the auto command to a clean lvim installation and opened a fsharp project work with. The language type was correctly identified.

### DIFF
--- a/ftdetect/fsautocomplete.lua
+++ b/ftdetect/fsautocomplete.lua
@@ -1,0 +1,3 @@
+vim.cmd [[
+  au BufNewFile,BufRead *.fs,*.fsx,*.fsi set filetype=fsharp
+]]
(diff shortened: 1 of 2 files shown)


kylo252:
Originally added in #1977.

Thank you, @grenaad!

PR #2027, opened 2021-12-03
fix(nvimtree): handle paths containing spaces

xeluxee:
# Description
Fixes #2007
It caused issues if telescope was started in folders containing spaces or folders that are symlinks.

## How Has This Been Tested?

- Put the cursor on folders containing spaces or folders that are symlinks
- Press `gtf` or `gtg`
- Enjoy

### DIFF
--- a/lua/lvim/core/nvimtree.lua
+++ b/lua/lvim/core/nvimtree.lua
@@ -170,9 +170,9 @@ end
 function M.start_telescope(telescope_mode)
   local node = require("nvim-tree.lib").get_node_at_cursor()
   local abspath = node.link_to or node.absolute_path
-  local is_folder = node.has_children and true
+  local is_folder = node.open ~= nil
   local basedir = is_folder and abspath or vim.fn.fnamemodify(abspath, ":h")
-  vim.api.nvim_command("Telescope " .. telescope_mode .. " cwd=" .. basedir)
+  vim.api.nvim_command("Telescope " .. telescope_mode .. " cwd=" .. string.gsub(basedir, " ", "\\ "))
 end
 
 return M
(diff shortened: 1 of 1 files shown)


lua/lvim/core/nvimtree.lua
```diff
@@ -170,9 +170,9 @@ end
 function M.start_telescope(telescope_mode)
   local node = require("nvim-tree.lib").get_node_at_cursor()
   local abspath = node.link_to or node.absolute_path
-  local is_folder = node.has_children and true
+  local is_folder = node.open ~= nil
   local basedir = is_folder and abspath or vim.fn.fnamemodify(abspath, ":h")
-  vim.api.nvim_command("Telescope " .. telescope_mode .. " cwd=" .. basedir)
+  vim.api.nvim_command("Telescope " .. telescope_mode .. " cwd=" .. string.gsub(basedir, " ", "\\ "))
```
kylo252, line 175:
```suggestion
  require("telescope.builtin")[telescope_mode] {
    cwd = vim.fn.fnameescape(basedir),
  }
```

kylo252 (changes requested):
Let's use the buitlin function for this.

lua/lvim/core/nvimtree.lua
xeluxee: There is no need to escape basedir since this is passed as an argument directly to telescope, and being an absolute path given by nvimtree it's already ready to use

kylo252: So it probably was just a problem translating vimL -> lua.

kylo252 (approved):

kylo252:
Thanks!

PR #2035, opened 2021-12-05
fix(zsh): don't set filetype to sh

meijieru:
# Description

Should not use `sh` filetype for `zsh`.
They have much different syntax.

### DIFF
--- a/lua/lvim/core/autocmds.lua
+++ b/lua/lvim/core/autocmds.lua
@@ -34,11 +34,7 @@ function M.load_augroups()
         "setlocal formatoptions-=c formatoptions-=r formatoptions-=o",
       },
     },
-    _filetypechanges = {
-      { "BufWinEnter", ".tf", "setlocal filetype=terraform" },
-      { "BufRead", "*.tf", "setlocal filetype=terraform" },
-      { "BufNewFile", "*.tf", "setlocal filetype=terraform" },
-    },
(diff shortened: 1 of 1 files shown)


kylo252:
It's mostly there to be able to use treesitter, although it's broken on ones that have a lot of zsh specific syntax.

It's also done to easily hook both `shfmt` and `shellcheck` where they work for most things.

meijieru:
This behavior has a high risk as it makes the highlight unusable once has incompatibility.
I think if the users want that behavior, they could manually `set ft=sh`.
Most of the scripts with compatibility may have the suffix `.sh`

kylo252:
Sounds good to me. 

Could you also remove the `*.tf` one as well since it's handled internally by treesitter, see: 
https://github.com/nvim-treesitter/nvim-treesitter/blob/41a097d9c543851625626618b08fc2bdf37b2894/ftdetect/hcl.vim#L2-L2

meijieru:
> Sounds good to me.
> 
> Could you also remove the `*.tf` one as well since it's handled internally by treesitter, see: https://github.com/nvim-treesitter/nvim-treesitter/blob/41a097d9c543851625626618b08fc2bdf37b2894/ftdetect/hcl.vim#L2-L2

Done.

kylo252 (approved):

kylo252:
Thanks!

PR #2037, opened 2021-12-05
feat: full compatibility with neovim v0.6

kylo252:
# Description

- Add some missing handlers configuration for full compatbility with Neovim v0.6
- Remove code that specific to v0.5.0 considering that most plugins don't support it anymore
- Add a check in the installer to verify that v0.5.1 or higher is available
- bump treesitter version with `vim.cmd` parsing support

## How Has This Been Tested?

Using `gl` now uses the new `vim.diagnostic.open_float()` call.

### DIFF
--- a/lua/lvim/lsp/config.lua
+++ b/lua/lvim/lsp/config.lua
@@ -14,6 +14,17 @@ return {
     update_in_insert = false,
     underline = true,
     severity_sort = true,
+    float = {
+      source = "always",
+      show_header = false,
+      format = function(d)
+        local t = vim.deepcopy(d)
+        if d.code then
+          t.message = string.format("%s [%s]", t.message, t.code):gsub("1. ", "")
+        end
+        return t.message
+      end,
+    },
   },
   document_highlight = true,
   code_lens_refresh = true,
--- a/lua/lvim/lsp/config.lua
+++ b/lua/lvim/lsp/config.lua
@@ -15,6 +15,9 @@ return {
     underline = true,
     severity_sort = true,
     float = {
+      focusable = false,
+      style = "minimal",
+      border = "rounded",
       source = "always",
       header = "",
       prefix = "",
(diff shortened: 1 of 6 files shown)


abzcoding:
not sure why, but i'm getting this error, even after `:PackerSync` and `:LvimCacheReset`
Screen Shot 2021-12-05 at 4 32 33 PM

kylo252:
> not sure why, but i'm getting this error, even after `:PackerSync` and `:LvimCacheReset` Screen Shot 2021-12-05 at 4 32 33 PM

Yeah, it's normal in my experience, just need to do `:PackerCompile` at least once or twice, and make sure to do :TSUpdate` which I've already added in post-update hook.

abzcoding:
ok everything is falling apart now 😢 , let me see what is going on and get back to you
Screen Shot 2021-12-05 at 4 39 14 PM


just to be sure, i'm on mac and this is my neovim version
Screen Shot 2021-12-05 at 4 42 11 PM

abzcoding:
@kylo252 yeah the issue was from my side, its fixed now
let me make sure there are not any other issues as well

lua/lvim/lsp/config.lua
```diff
@@ -4,16 +4,28 @@ return {
...
+        { name = "DiagnosticSignHint", text = "" },
+        { name = "DiagnosticSignInfo", text = "" },
       },
     },
     virtual_text = true,
     update_in_insert = false,
     underline = true,
     severity_sort = true,
+    float = {
+      source = "always",
```
abzcoding, line 18: should we add
```lua
      border = "rounded",
      focusable = false,
```
here as well or you prefer the current one?

kylo252: ~~do we have a border setting somewhere that we can re-use?~~
I'll switch it to use `lvim.lsp.popup_border`

abzcoding (approved):
tyvm <3

kylo252:
I'll leave it until tomorrow or until I get a second review. Just in case we missed something.

PR #2039, opened 2021-12-05
feat(terminal): lazygit can now be toggled

kylo252:
# Description

- remove `hidden` parameter, so it can be persistent
- map keybinding for both normal and term modes
- a default shell is now always reachable with `<C-t>` according to `lvim.builtin.terminal.direction`
- add an addition keybinding `<C-\>` for `lazygit`
since it's hard to hit `<leader>gg` within the `timeoutlen` < 300ms

## How Has This Been Tested?

Try both `<C-t>` and `<C-\>` and see how `lazygit` is persistent.

https://user-images.githubusercontent.com/59826753/144755634-5a398719-403f-4721-b29b-f0e222d80d03.mp4

### DIFF
--- a/lua/lvim/core/terminal.lua
+++ b/lua/lvim/core/terminal.lua
@@ -16,7 +16,7 @@ M.config = function()
     insert_mappings = true, -- whether or not the open mapping applies in insert mode
     persist_size = false,
     -- direction = 'vertical' | 'horizontal' | 'window' | 'float',
-    direction = "horizontal",
+    direction = "float",
     close_on_exit = true, -- close the terminal window when the process exits
     shell = vim.o.shell, -- change the default shell
     -- This field is only relevant if direction is set to 'float'
--- a/lua/lvim/core/terminal.lua
+++ b/lua/lvim/core/terminal.lua
@@ -16,7 +16,7 @@ M.config = function()
     insert_mappings = true, -- whether or not the open mapping applies in insert mode
     persist_size = false,
     -- direction = 'vertical' | 'horizontal' | 'window' | 'float',
-    direction = "float",
+    direction = "horizontal",
     close_on_exit = true, -- close the terminal window when the process exits
     shell = vim.o.shell, -- change the default shell
     -- This field is only relevant if direction is set to 'float'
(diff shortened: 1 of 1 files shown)


abzcoding:
i kind of prefer float instead of horizontal and i use the current `<leader>gg` on a daily basis with sub 300 😅 
so i'm probably not the right person to review this

@ChristianChiarulli might be a better candidate,

kylo252:
> i kind of prefer float instead of horizontal and i use the current `<leader>gg` on a daily basis with sub 300 😅 so i'm probably not the right person to review this

- I kept `<leader>gg` but also added the other one 
> since it's hard to hit `<leader>gg` within the `timeoutlen` < 300ms

- you can change the default terminal direction, I didn't hard-code that value, I just figured it's better to only use float for `lazygit`.

lua/lvim/core/terminal.lua
```diff
@@ -16,7 +16,7 @@ M.config = function()
     insert_mappings = true, -- whether or not the open mapping applies in insert mode
     persist_size = false,
     -- direction = 'vertical' | 'horizontal' | 'window' | 'float',
-    direction = "float",
+    direction = "horizontal",
```
abzcoding, line 19: I meant this one, i think @ChristianChiarulli preferred it to be float as well

kylo252: Yeah, I figured that it's what you meant. You guys are crazy, I'll put it back to float by default 😄

abzcoding: tyvm <3

abzcoding (approved):
LGTM 🚀

lua/lvim/core/terminal.lua
kylo252: but can you actually hit `<leader>gg` without `lazygit` reacting to the press? e.g. it would try to stage files for example when you press `space`. am I really that slow? :(

abzcoding: no i cannot, i usually use `q` to toggle off and `<leader>gg` to toggle on

soerenmartius:
This breaks the usage of multiple tmux windows unfortunately, lazygit will only work in the first window while it doesn't when working in multiple lvim instances at the same time

kylo252:
> This breaks the usage of multiple tmux windows unfortunately, lazygit will only work in the first window while it doesn't when working in multiple lvim instances at the same time

Why rely on `toggle-term` if you're using `tmux`? 

```tmux
bind g popup -E -yC -xC -w 80% -h 80% "lazygit"
# I even have this for my dotfiles
bind C-g popup -E -yC -xC -w 80% -h 80% "zsh -c \"lazygit --git-dir=<some-dir>. --work-tree=$HOME\""
```

PR #2040, opened 2021-12-05
[Feature]: GitHub support [WIP]

waltermaldonado:
# Description

Added octo.nvim as a builtin dependency and started the implementation of a section in the main menu (leader) for handling GitHub interaction.

As expected there is still some key collision to be solved. Also, I think we could try to find a way to install these dependencies such as lazygit, and now gh (github client) during the installation process. For the latter, the recommended ways of installing are very different for the various OSs, and I could not think about a better way than conda to do that.
This idea bugs me a little, even if we are doing it with miniconda.

Since it relies on the github client to perform the requests, you will need to perform `gh login auth` manually for it to start working. The downside is that you have to do this, but the bright side is that it works for github.com and for GHE (GitHub Enterprise), even if you have repos from GHE and github.com at the same machine.

It is also working good under a corporate proxy, with the `http_proxy` and `https_proxy` environment variables set up.

Additional Dependencies:
- octo.nvim
- GitHub client

image

image

image


chase:
Although I find this useful, I don't think this should be a core plugin.

This seems to essentially be a user configuration for Octo, which more than likely would fit better in our documentation: https://www.lunarvim.org/plugins/03-extra-plugins.html#octo

This does make me wonder if maybe we could offer an API for a plugin to act as basically an extension pack for LunarVim, so that it isn't merely copy and paste.

kylo252 (commented):
I like the general principle of what octo.nvim is trying to do, but I always found it clunky for my use-case. In a way, the interface is still limited and even unintuitive at points, e.g. what happens when you press `<cr>` in PR list?

I also agree with @chase that you need to tune-in your keybindings to make full use of it.

I would suggest moving the bulk of this PR into https://www.lunarvim.org/plugins/03-extra-plugins.html#octo, and then adding an entry for it in ` config.example.lua` with maybe a link to the docs for the full settings.

waltermaldonado:
I can see @chase's point of not recommending it as a core plugin. But one of the biggest challenges of vim that LunarVim got my attention is the key bindings preset. It takes a lot of time setting it up and can become a nightmare of change after change trying to get the perfect combination.

My idea here was to establish a working relation among all the core plugins and get this burden off of the user. If not being a core plugin, the idea of an extension API looks interesting, but I think we need to offer a way to facilitate that for who is installing a new plugin. For real, it's not only two lines of code to install a plugin and use it, you end up spending a lot of time on the key bindings and most certainly will fall into other compatibility issues leading to failure and giving up on the plugin.

jameswalmsley:
@waltermaldonado @chase @kylo252 @ChristianChiarulli  @Tastyep @abzcoding 

This is something that comes up quite regularly. I agree that this shouldn't be part of the default config. 

However using Octo recently it seems to be getting a lot better. I've also had to spend a long time curating my keybindings..

I like the website and having these user configs there.. but could we atleast commit them into the main repo under a contrib folder? Maybe even allo user configs to do a kind of.. include/contrib/octo?

J

Uzaaft:
> @waltermaldonado @chase @kylo252 @ChristianChiarulli @Tastyep @abzcoding
> 
> This is something that comes up quite regularly. I agree that this shouldn't be part of the default config.
> 
> However using Octo recently it seems to be getting a lot better. I've also had to spend a long time curating my keybindings..
> 
> I like the website and having these user configs there.. but could we atleast commit them into the main repo under a contrib folder? Maybe even allo user configs to do a kind of.. include/contrib/octo?
> 
> J

I got an idea. How about we add this under Lunarvim.org? @kylo252  @abzcoding @jameswalmsley 
For.example here: https://www.lunarvim.org/plugins/02-extra-plugins.html#octo
We could for example create a gist, and link to that?

abzcoding:
> How about we add this under Lunarvim.org? @kylo252 @abzcoding @jameswalmsley For example here: > https://www.lunarvim.org/plugins/02-extra-plugins.html#octo We could for example create a gist, and link to that, or write it directly in? 

sounds good to me 👍

Uzaaft:
> > How about we add this under Lunarvim.org? @kylo252 @abzcoding @jameswalmsley For example here: > [lunarvim.org/plugins/02-extra-plugins.html#octo](https://www.lunarvim.org/plugins/02-extra-plugins.html#octo) We could for example create a gist, and link to that, or write it directly in?
> 
> sounds good to me 👍

Now the question is: A long ass config written directly to the site, or a gist?

abzcoding:
> Now the question is: A long ass config written directly to the site, or a gist?
it would be nice to have all of them in one place, maybe [here](https://www.lunarvim.org/plugins/02-extra-plugins.html#octo)

Uzaaft:
@abzcoding  A block code 169 lines long seems kind of long....But I opened a PR in lunarvim.org, where we can take the discussion

Issue #2012, opened 2021-11-30
[Bug]: Multiple LSP servers spawned


youdaniel:
### Problem description

`lvim a.py`
`:LspInfo`
![image](https://user-images.githubusercontent.com/39450656/144133701-1b60f071-0732-4f64-aa25-135228f450bf.png)
Also happens in some other languages too. Started happening after #1989

### Check the following

- [X] I am in a git managed directory
- [X] I restarted Neovim after running :LspInstall
- [X] I have a valid language configuration (~/.config/lvim/ftplugin/)
- [X] I checked the options (~/.config/lvim/config.lua)
- [X] The programs specified by the config are installed

### LunarVim version

rolling

### Neovim version (>= 0.5)

0.6

### I have

- [X] Read the readme
- [X] Read the wiki
- [X] Searched for similar issues

mjlbach:
This is a feature of lspconfig. If you open a bare python file, with no way for us to identify the project structure, we spawn a single language server per directory which is shared.

youdaniel:
> per directory which is shared

What do you mean by this? Pyright always seems to spawn three times for me, except in a directory where the root directory pattern matches.

Is there a way for it to only spawn once for bare files? The duplicate error messages can be weird to work with.

kylo252:
I'm unable to replicate this issue. I only get a single `pyright` instance with these steps:

```shell
mkdir -p /tmp/test
touch /tmp/test/test.py
lvim  /tmp/test/test.py
```

Where is that file that you're testing is located?

abzcoding:
@kylo252 i'm not sure why but it seems like this part is called more than once
```lua
  requested_server:on_ready(function()
    if install_notification then
      vim.notify(string.format("Installation complete for [%s] server", requested_server.name), vim.log.levels.INFO)
    end
    install_notification = false
    requested_server:setup(config)
  end)
```

kylo252:
@abzcoding, unless something has changed in implementation in the installer, then we're only setting up once. I'm not sure which instance you're referring to as being called twice.

kylo252:
> Hey, I seem to have the same issue. Here this is regarding "sumneko_lua". When I open `config.lua` 2 instances of lua-language server is spawned 

@ashincoder, please try using [tests/minimal_lsp.lua](https://github.com/LunarVim/LunarVim/blob/rolling/tests/minimal_lsp.lua), just make sure to change the server name to whichever server you're having trouble with, and then start neovim like this:

```shell
[...]

PR #2042, opened 2021-12-06
fix(lsp): avoid duplicates server

meijieru:
# Description

Avoid launch duplicated lsp server.
Related https://github.com/LunarVim/LunarVim/issues/2012

### DIFF
--- a/lua/lvim/lsp/manager.lua
+++ b/lua/lvim/lsp/manager.lua
@@ -1,4 +1,6 @@
-local M = {}
+local M = {
+  _client_infos = {},
+}
 
 local Log = require "lvim.core.log"
 local lvim_lsp_utils = require "lvim.lsp.utils"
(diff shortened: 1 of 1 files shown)


kylo252 (changes requested):
As mentioned, I'm still not able to reproduce #2042, so I don't know what is it exactly that we're trying to solve.

Can you please expand on the intended logic here? I feel like there might be a cleaner way to do it.

abzcoding:
> As mentioned, I'm still not able to reproduce #2042, so I don't know what is it exactly that we're trying to solve.
> 
> Can you please expand on the intended logic here? I feel like there might be a cleaner way to do it.

@kylo252 basically this
Screen Shot 2021-12-07 at 4 23 45 PM

meijieru:
In short, `lvim_lsp_utils.is_client_active` is not reliable here as the server takes time to start.
So I set `_client_infos.status` to `loading` to avoid that.

kylo252:
Thanks for the explanation. I'm pretty sure that there's no need to introduce another variable to track this. I'll test a few different approaches and report back.

kylo252:
Thanks for pointing me in the right direction! 

Superseded by #2048

PR #2048, opened 2021-12-07
fix(lsp): prevent repeated setup call

### LINKED ISSUE #2012: [Bug]: Multiple LSP servers spawned

kylo252:
# Description

Check if the manager `autocomd` has already been configured, since some servers can take a while to initialize.
This helps guarding against a data-race condition where a server can get configured twice, which seems to occur only when attaching to single-files.

Fixes #2012

## How Has This Been Tested?

- Add a duplicate setup call to `pyright`

```lua
-- add this to ~/.config/lvim/after/ftplugin/python.lua
require("lvim.lsp.manager").setup "pyright"
require("lvim.lsp.manager").setup "pyright"
```

- Make sure to start in _single-file-mode_

```shell
touch test.py
lvim test.py
```

- For good-measure, try adding a filetype plugin to trigger packer as well

```lua
{ "Vimjas/vim-python-pep8-indent", ft = "python" }
```

### DIFF
--- a/lua/lvim/lsp/manager.lua
+++ b/lua/lvim/lsp/manager.lua
@@ -45,13 +45,28 @@ local function buf_try_add(server_name, bufnr)
   require("lspconfig")[server_name].manager.try_add_wrapper(bufnr)
 end
 
+-- check if the manager autocomd has already been configured since some servers can take a while to initialize
+-- this helps guarding against a data-race condition where a server can get configured twice
+-- which seems to occur only when attaching to single-files
+local function client_is_configured(server_name, ft)
+  ft = ft or vim.bo.filetype
+  local active_autocmds = vim.split(vim.fn.execute("autocmd FileType " .. ft), "\n")
+  for _, result in ipairs(active_autocmds) do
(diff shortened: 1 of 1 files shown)


abzcoding (approved):
It seems to be working fine. We might want to ask some other folks to test it as well

youdaniel:
I was having issues with texlab spawning multiple times. I tested this PR (very) briefly and it seems to have fixed the issue

PR #2054, opened 2021-12-08
feat: multiple enhancements to lvim-reload

### LINKED ISSUE #2052: [Feature]: Use Packer to load core plugins, not download/install
### Feature motivation

[It's been addressed before](https://github.com/LunarVim/LunarVim/issues/1399), but our core code updates with Packer and often becomes incredibly fragile, displaying errors that make LunarVim appear to be entirely broken and makes unwilling to continue.

Also, Packer on LvimUpdate with the new setup took around 63 seconds for me on a 200 Mbps connection.
This seems to be because the downloads themselves are now done in serial instead of parallel. This could be fixed directly, but we can optimize things things a little better on our end without having to worry about Packer messing up.

The current method of handling the initial install is still fragile because Packer still handles it as a remote git repository.

Assumptions:

1. Core plugins are pinned to a SHA
2. Core plugins reside on GitHub
3. The host system has `curl` and `unzip` (even Git for Windows has these)

We can entirely avoid the mess caused by Packer trying to manage a git repository and cloning unnecessary objects by just download the commit's zipped snapshot and referencing it as an absolute path.

Let's use `nvim-telescope/telescope.nvim` pinned at commit `27294d73e4562cbc83c9c488b4618f012deade7e` as an example:

- Check if `telescope.nvim-27294d73e4562cbc83c9c488b4618f012deade7e` exists in a central location `/.local/share/lunarvim/core_plugins`, if not then continue
- Download: `curl -LO https://github.com/nvim-telescope/telescope.nvim/archive/27294d73e4562cbc83c9c488b4618f012deade7e.zip`
- Unzip: `unzip 27294d73e4562cbc83c9c488b4618f012deade7e.zip`
- Relocate to the central location: `mv telescope.nvim-27294d73e4562cbc83c9c488b4618f012deade7e ~/.local/share/lunarvim/core_plugins/`
- In our plugin loader, only for these core plugins, rewrite the path to be the absolute path like this:
```
  {
    "/home/chase/.local/share/lunarvim/core_plugins/telescope.nvim-27294d73e4562cbc83c9c488b4618f012deade7e",
    commit = commit.telescope,
    config = function()
      require("lvim.core.telescope").setup()
    end,
    disable = not lvim.builtin.telescope.active,
  },
```

---

Installation and updates should _not_ use the usual `init.lua` because it loads the plugins and should _not_ be run directly as a Vim command in a running instance, this leads to the errors we see. **Why?** A lot of it is caused by use-before-load or having an old version loaded of a plugin loaded with a configuration expecting the updated version.

Updates should instead be run from the command line with a mostly clean state, which we could add as an option in the shim executable we have installed as `lvim`:
`nvim -u install_update.lua --noplugin -n -i NONE --headless`

An unfinished, minimalist example for `install_update.lua` init file:
```
local init_path = debug.getinfo(1, "S").source:sub(2)
local base_dir = init_path:match("(.*[/\\])"):sub(1, -2)

if not vim.tbl_contains(vim.opt.rtp:get(), base_dir) then
  vim.opt.rtp:append(base_dir)
end

require("lvim.bootstrap"):init(base_dir)

local plugins = require "lvim.plugins"

-- TODO: download plugins and unzip using vim.loop.spawn in parallel, using a counter when the last unzip finishes, finish the steps above
```

_No response_

### Additional context

_No response_

kylo252:
# Description

- feat: invoke template re-generation on `LvimReload`
- fix: use the `on_complete` callback which ensures a more consistent experience especially for `nvim-webdev-icons`
- feat: faster git fetch command for packer. Partially fixes #2052

## How Has This Been Tested?

Snappier `LvimReload` including correct loading of dev-icons!

### DIFF
--- a/lua/lvim/config/init.lua
+++ b/lua/lvim/config/init.lua
@@ -98,28 +98,20 @@ end
 --- Override the configuration with a user provided one
 -- @param config_path The path to the configuration overrides
 function M:reload()
-  local lvim_modules = {}
-  for module, _ in pairs(package.loaded) do
-    if module:match "lvim.core" then
-      package.loaded[module] = nil
-      table.insert(lvim_modules, module)
-    end
-  end
+  package.loaded["lvim.utils.hooks"] = nil
+  local _, hooks = pcall(require, "lvim.utils.hooks")
+  hooks.run_pre_reload()
 
   M:init()
   M:load()
 
+  require("lvim.core.autocmds").configure_format_on_save()
+
   local plugins = require "lvim.plugins"
-  local autocmds = require "lvim.core.autocmds"
-  autocmds.configure_format_on_save()
   local plugin_loader = require "lvim.plugin-loader"
-  plugin_loader.cache_clear()
+
   plugin_loader.load { plugins, lvim.plugins }
-  vim.cmd ":PackerInstall"
-  vim.cmd ":PackerCompile"
-  -- vim.cmd ":PackerClean"
-  require("lvim.lsp").setup()
-  Log:info "Reloaded configuration"
+  hooks.run_post_reload()
 end
 
 return M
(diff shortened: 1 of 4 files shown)


abzcoding (approved):
LGTM 🚀

PR #2064, opened 2021-12-10
feat: bump plugin versions

abzcoding:
# Description

Bump plugin versions

note: please test locally before merging ( my setup is very customized )


lua/lvim/core/lualine/components.lua
```diff
@@ -1,6 +1,12 @@
 local conditions = require "lvim.core.lualine.conditions"
 local colors = require "lvim.core.lualine.colors"
 
+local diag_source = "nvim_lsp"
+local ok, _ = pcall(require, "vim.diagnostic")
+if ok then
+  diag_source = "nvim_diagnostic"
```
abzcoding, line 7: should we make this the default one and drop the support for older versions ?

lua/lvim/plugins.lua
```diff
@@ -1,38 +1,38 @@
...
-  nvim_web_devicons = "8df4988ecf8599fc1f8f387bbf2eae790e4c5ffb",
+  nvim_web_devicons = "344331467509802e1af200f08ec3da278be5cbba",
   packer = "851c62c5ecd3b5adc91665feda8f977e104162a5",
-  plenary = "1c31adb35fcebe921f65e5c6ff6d5481fa5fa5ac",
+  plenary = "e6267f79481064eee53950571f53cbaafb08417d",
   popup = "b7404d35d5d3548a82149238289fa71f7f6de4ac",
   project = "71d0e23dcfc43cfd6bb2a97dc5a7de1ab47a6538",
   structlog = "6f1403a192791ff1fa7ac845a73de9e860f781f1",
-  telescope = "27294d73e4562cbc83c9c488b4618f012deade7e",
+  telescope = "5f37fbfa837dfee7ecd30f388b271f4a71c0a9e0",
```
kylo252, line 32:
```suggestion
  telescope = "80cdb00b221f69348afc4fb4b701f51eb8dd3120", -- see telescope.nvim#1549
```

We need to make a decision about telescope since they have made 0.6 a requirement, but that should probably be in a different PR since it's going to require some extra work updating all the bindings.

image

kylo252 (changes requested):
Otherwise, everything seems to work 👍

PR #2065, opened 2021-12-11
[Bugfix] Allow disabling the open binding for ToggleTerm

chase:
# Description

Instead of throwing an error if `lvim.builtin.terminal.open_mapping` is set to `nil`, this simply doesn't add the default binding

## How Has This Been Tested?

Set `lvim.builtin.terminal.open_mapping = nil` in `config.lua`. Note that it doesn't throw errors on startup.

### DIFF
--- a/lua/lvim/core/terminal.lua
+++ b/lua/lvim/core/terminal.lua
@@ -60,7 +60,9 @@ M.setup = function()
     direction = lvim.builtin.terminal.direction,
     size = lvim.builtin.terminal.size,
   }
-  M.add_exec(default_term_opts)
+  if lvim.builtin.terminal.open_mapping then
+    M.add_exec(default_term_opts)
+  end
 
   for i, exec in pairs(lvim.builtin.terminal.execs) do
     local opts = {


kylo252 (commented):
Not sure why you'd want to set that one specifically to nil, but otherwise I don't mind this.

And the good old macOS error means you'd have to make sure the entry exists before evaluating it.

chase:
> Not sure why you'd want to set that one specifically to nil, but otherwise I don't mind this.

If you don't want the mapping, there's no way to unset it. That would make it an exception to the rest of LunarVim, as far as I know.

An example, in my case, is that I don't use the builtin ToggleTerm with my default shell _and_ I don't want Ctrl-T to be mapped since it is mapped in my terminal emulator (Kitty) to do something else.

> And the good old macOS error means you'd have to make sure the entry exists before evaluating it.

I don't think so. The error looks completely unrelated:

```
...packer/start/plenary.nvim/lua/plenary/async_lib/util.lua:320: Blocking on future failed ...acker/start/plenary.nvim/lua/plenary/async_lib/async.lua:14: The coroutine failed with this message: Vim:E739: Cannot create directory /Users/runner/.local/share/lunarvim/site/after: file already exists
            stack traceback:
            	[C]: in function 'error'
            	...packer/start/plenary.nvim/lua/plenary/async_lib/util.lua:320: in function <...packer/start/plenary.nvim/lua/plenary/async_lib/util.lua:318>
            	[C]: in function 'wait'
            	...packer/start/plenary.nvim/lua/plenary/async_lib/util.lua:325: in function <...packer/start/plenary.nvim/lua/plenary/async_lib/util.lua:309>
            	[C]: in function 'xpcall'
            	...te/pack/packer/start/plenary.nvim/lua/plenary/busted.lua:73: in function 'call_inner'
            	...te/pack/packer/start/plenary.nvim/lua/plenary/busted.lua:174: in function 'it'
            	...acker/start/plenary.nvim/lua/plenary/async_lib/tests.lua:11: in function 'it'
            	...unner/work/LunarVim/LunarVim/tests/plugins_load_spec.lua:20: in function <...unner/work/LunarVim/LunarVim/tests/plugins_load_spec.lua:3>
            
            stack traceback:
            	...packer/start/plenary.nvim/lua/plenary/async_lib/util.lua:325: in function <...packer/start/plenary.nvim/lua/plenary/async_lib/util.lua:309>
```

chase:
In fact, the error is exactly the same as this, so I probably just need to rebase: https://github.com/LunarVim/LunarVim/runs/4411363465?check_suite_focus=true#step:5:14

chase:
Yup, that was it: the symptom aside from that error was that the commit action's run itself was okay, but the PR run wasn't

abzcoding (approved):

kylo252:
> If you don't want the mapping, there's no way to unset it. That would make it an exception to the rest of LunarVim, as far as I know.

I just figured that you'd either want to map it to something else, or you don't need the plugin in the first-place. But maybe you're like me, and you only use it for lazygit and lnav 🥲 

> I don't think so. The error looks completely unrelated

Sorry, I should've read the error message before. I had just assumed so because the CI was passing fine.

PR #2072, opened 2021-12-13
feat: update lsp-installer and lspconfig hashes to enable solidity_ls language server

edag94:
# Description

Updated package commit hashes to enable solidity_ls language server among other commits

Fixes #(issue)

## How Has This Been Tested?

- Run command `:mycommand`
- Check logs
- ...


abzcoding:
so which lang server is preferred for solidity?
- solang
- solidity_ls

so we can add one of them to the override list ( to make sure we are not invoking two lang servers at the same time )
https://github.com/LunarVim/LunarVim/blob/8ffbe596c0c1d3a03bf2913538c156210211a37c/lua/lvim/lsp/config.lua#L60-L80

edag94:
@kylo252 @abzcoding the npm package is published by me, which is exposing the solidity language server used by the solidity vscode extension [here](https://github.com/juanfranblanco/vscode-solidity). This language server supports auto completion, so I’d say is better than the current offerings from the Solang server, and making solidity_ls the primary server is the motivation for this PR. @kylo252 thanks for updating the override list.

PR #2083, opened 2021-12-14
fix: remove autopairs cmp completion

### LINKED ISSUE #2082: [Bug]: Parens added erronously when importing functions using nvim-cmp
### Problem description

I've run across a few problems with the current auto pairs autocompletion behavior for nvim-cmp in JavaScript. 

#### Example 1

Let's say we have a function in one file like the following:

``` javascript
export function example() {
  return "example";
}
```

If in another file I type out `example` and wait, the nvim-cmp auto-completion window pops up offering to import the `example` function. If I type `(` after then, then I end up with `example(())` in my code along with the import.

#### Example 2

Let's say I have the following React component:

``` jsx
export function Exmaple() {
  return <h1>Example</h1>;
}
```

If I type `<Example` and hit enter, I get `<Example()`. In this situation, the auto-completion recognizes the imported value as a function, but erroneously adds parentheses after it.

#### My Take

There are a ton of contexts where you need to import a function, but you don't actually want to *call* it. I don't think LunarVim can differentiate between these cases, so I think this behavior should be removed.

### LunarVim version

Latest

### Neovim version (>= 0.5)

v0.6.0

### Relevant log output

_No response_

### Screenshots

_No response_

### I have

- [X] Read the readme
- [X] Read the wiki

LandonSchropp:
# Description

This removes nvim-cmp/nvim-autopairs auto-completion behavior that's problematic. I give specific examples in #2082 showcasing the problem.

It's still possible to auto-complete with parens. If the auto-completion menu is open, typing `(` will add the opening and closing parens in addition to importing the function. I think this is simpler and clearer.

Fixes #2082

## How Has This Been Tested?

I ran this locally and it seems to work fine. The problematic examples I gave in #2082 have been fixed.

### DIFF
--- a/lua/lvim/core/autopairs.lua
+++ b/lua/lvim/core/autopairs.lua
@@ -85,14 +85,6 @@ M.setup = function()
     end),
   }
 
-  local cmp_status_ok, cmp = pcall(require, "cmp")
-  if cmp_status_ok then
-    -- If you want insert `(` after select function or method item
-    local cmp_autopairs = require "nvim-autopairs.completion.cmp"
-    local map_char = lvim.builtin.autopairs.map_char
-    cmp.event:on("confirm_done", cmp_autopairs.on_confirm_done { map_char = map_char })
-  end
-
   require("nvim-treesitter.configs").setup { autopairs = { enable = true } }
 
   local ts_conds = require "nvim-autopairs.ts-conds"


kylo252 (commented):
Personally, I don't mind removing this, but I don't use it enough to tell.

abzcoding (approved):
seems fine to me

LandonSchropp:
Thanks!

Issue #1397, opened 2021-08-26
[Feature]: Standalone cli entry


ahmedkhalf:
### Feature motivation

Right now, lvim is just an alias for `nvim -u ~/.local/share/lunarvim/lvim/init.lua --cmd "set runtimepath+=~/.local/share/lunarvim/lvim" "$@"`. But I believe it can be much more that that. 

### Describe the expected solution

Allowing lunarvim to be wrapped in an actual cli will enable endless possibilities, and will aid with the transition of lunarvim going from "just a neovim config", to a standalone IDE.

Consider the following cli entries:
```
lvim update
  update lunarvim without having to cd and git pull. (Perhaps packer sync too?)

lvim install {branch: ["master"], "rolling" }
  install lunarvim, or switch branch

lvim clean
  clean packer, cache, log.. etc. you name it.

lvim doctor
  diagnose common issues and checks for errors

lvim settings
  dump all "lvim." settings to terminal

...

lvim
  launch lunarvim, and enable restart functionality**
```

***restart functionality:*
Let's say the user sets `number` to be false, the problem is that this is a window option, so with the current hot reload implementation this will only change that specific `config.lua` window and not a buffer in a split. Furthermore, our current implementation of hot reload bypasses plugin keymappings. The solution to this is to restart neovim, currently, a user can manually restart neovim by closing and opening neovim, but then the splits aren't recovered so he will have to create a session too if he wants his splits recovered. This can be remedied/automated though, using the restart functionality.

### Describe the alternatives you've considered

There are many languages in which this app can be developed, but it seems golang is the clear winner. I tried using lua with `nvim --headless -u NONE`, but this would prove to be too slow with a whopping 20ms startup.

### Additional context

This is a small proof of concept of the restart functionality:

https://user-images.githubusercontent.com/36672196/130928696-d4360bf7-8610-4de2-9777-0a59881f5781.mp4

chase:
I think having additional functionality to update, clean, install, etc. would be very handy!

I however, do not agree about using Go. Although it does allow cross-platform building, I think it would introduce unnecessary complexity to distribute LunarVim.

[20ms is barely over the threshold of human perception of 18ms](https://news.mit.edu/2014/in-the-blink-of-an-eye-0116), so I don't think for one-off commands it is too terrible. Especially considering the majority of these operations will be doing relatively much slower, asynchronous IO.

[...]

PR #2085, opened 2021-12-15
[WIP]: initial core installer/updater rewrite

### LINKED ISSUE #2052: [Feature]: Use Packer to load core plugins, not download/install

chase:
# Description

## WORK IN PROGRESS

Make lvim update outside of a currently running instance, `lvim --update` is now the way to update.

`lvim --update` can take a remote branch to update on like `lvim --update rolling` so a user can easily switch between branches to test/try things, or even switch to a specific commit if they want.

This requires the user to uninstall and re-install to get the new `lvim` shim currently, but I would like to make updating the shim easier and the primary way to update from old versions.

`:PackerSync` will no longer affect the core plugins, so it won't break the user's currently running instance.

Closes #2052

## How Has This Been Tested?

The core of `lvim --update` is the following: `nvim -u "$LUNARVIM_RUNTIME_DIR/lvim/init_update.lua" --noplugin -n -i NONE --headless`

Inside of this branch's lvim dir run `rm -rf ../core ../site`, run `nvim -u "$LUNARVIM_RUNTIME_DIR/lvim/init_update.lua" --noplugin -n -i NONE --headless`.  
It installs all the core plugins, users plugins, compiles, then exits, but not treesitter yet.


chase:
Unit tests aren't running on GH actions for some reason, but they work locally with `make test`
Need to resolve merge conflict

init_update.lua
```diff
@@ -0,0 +1,183 @@
...
+    elseif packer_stage == 3 then
+      print("Installed user plugins in:", timer:stop(), "ms")
+      vim.cmd [[qall!]]
+    end
+  end
+
+  timer:start()
+  print "Installing core plugins..."
+  packer.sync()
+end)()
```
kylo252, line 183: This should all go into `bootstrap.lua` and `plugin-loader` because it's way too complex to manage otherwise.

init_update.lua
```diff
@@ -0,0 +1,183 @@
...
+local await = a.wait
+local wrap = a.wrap
+local jobs = require "packer.jobs"
+local result = require "packer.result"
+local async_stat = wrap(uv.fs_stat)
+
+---downloads and installs from a core plugin entry
+---@param plug table plugin entry
+---@return function
+local function download_and_install(plug)
```
kylo252, line 79: Why can't we write this in bash? I thought the whole point is to do this without relying on packer.

lua/lvim/bootstrap.lua
```diff
@@ -90,6 +93,22 @@ function M:init(base_dir)
     install_path = self.packer_install_dir,
   }
 
+  -- patch Packer's guess_dir_type to support custom installs using symlinks
+  local guess_dir_type = require("packer.plugin_utils").guess_dir_type
+  require("packer.plugin_utils").guess_dir_type = function(dir)
```
kylo252, line 98: This should go into the plugin-loader, since it's the actual wrapper for packer.

lua/lvim/core_plugin_type.lua
```diff
@@ -0,0 +1,55 @@
...
+      return result.ok()
+    end)
+  end
+  plugin.revert_last = function(_)
+    log.warn "Can't revert a core plugin!"
+    return result.ok()
+  end
+end
+
+return { setup = setup_local }
```
kylo252, line 55: This should go into the plugin-loader, since it's the actual wrapper for packer.

lua/lvim/plugins.lua
```diff
@@ -52,6 +55,9 @@ return {
   {
     "rcarriga/nvim-notify",
     commit = commit.nvim_notify,
+    config = function()
+      require("lvim.core.notify").setup()
+    end,
```
kylo252, line 60: Let's make this a separate PR.

utils/bin/lvim
```diff
@@ -3,4 +3,25 @@
...
+      target_branch=${1:"$(git rev-parse --abbrev-ref HEAD)"}
+      git stash --include-untracked -m "Changes before LunarVim update" && echo "Uncommitted changes saved to stash" || true
+      git fetch --force --depth=1 origin $target_branch:$target_branch
+      git checkout $target_branch
+      cd "$old"
+      nvim -u "$LUNARVIM_RUNTIME_DIR/lvim/init_update.lua" --noplugin -n -i NONE --headless
+
+      exit
+      ;;
+    *) set -- "$@" "$arg" ;;
```
kylo252, line 23: There shouldn't be a need for this when we can just call `:LvimUpdate` as long as the entry point is still `bootstrap.lua`, and we can add more handling for headless mode in the actual hook instead.

utils/installer/install.sh
```diff
@@ -339,6 +339,26 @@ function setup_shim() {
...
+      git fetch --force --depth=1 origin \$target_branch:\$target_branch
+      git checkout \$target_branch
+      cd "\$old"
+      nvim -u "\$LUNARVIM_RUNTIME_DIR/lvim/init_update.lua" --noplugin -n -i NONE --headless
+
+      exit
+      ;;
+    (*) set -- "\$@" "\$arg" ;;
+  esac
+done
```
kylo252, line 360: Same point as the one for `lvim/bin`.

kylo252 (changes requested):
I know there are improvements to be made as long as we keep using `bootstrap.lua` which should be able to handle this.

Overall, I'm not sure if this is providing enough benefit to justify all this added complexity. I would rather implement [`paq.nvim](https://github.com/savq/paq-nvim) in core instead and tweak it as necessary.

init_update.lua
chase: I'm not sure what you mean, this is only for init_update and nothing else. What parts of this do you think should go into bootstrap and plugin-loader respectively?

lua/lvim/plugins.lua
chase: This is necessary to prevent install from erroring out because no default config is provided _at all_. The default animation requires a background to be set for Normal.

utils/bin/lvim
chase: Not really, `:LvimUpdate` is fragile and was largely the point of this PR. `:LvimUpdate` breaking is one of the most common complaints, and there is no benefit to having it run while an editor is open.

Doing updates _while the editor is running_ means that we can't guarantee the state of _anything_.

Users are also unable to simply switch branches, including to be able to switch to a PR to test a fix, for example.

`:LvimUpdate` not functioning or displaying errors is a complaint so common that I made something to simply link to, and the current fixes are _terrible_ and usually lead to a lot of errors:
https://github.com/LunarVim/LunarVim/discussions/1752

The approach to "fix things" or switch to rolling that I see most often mentioned is "just uninstall using this script and reinstall using this really long command":
```
bash ~/.local/share/lunarvim/lvim/utils/installer/uninstall.sh
LV_BRANCH=rolling bash <(curl -s https://raw.githubusercontent.com/lunarvim/lunarvim/rolling/utils/installer/install.sh)
```

utils/installer/install.sh
chase: Speaking of which, why is this duplicated?

lua/lvim/bootstrap.lua
chase: Probably should go into plugin-loader:init, you're right.

lua/lvim/core_plugin_type.lua
chase: It's not, it is a plugin-type adapted from: https://github.com/wbthomason/packer.nvim/blob/master/lua/packer/plugin_types/local.lua

init_update.lua
chase: > Why can't we write this in bash?

I'm not sure I understand. Why should this be written in bash? What's the benefit over it being written in Lua?

> I thought the whole point is to do this without relying on packer.

I'd like to clarify, this is to fix the _fragile behavior_ of Packer when syncing the core plugins using the built-in git file type handler. This is to avoid volatile state issues as I mentioned in my other comment.
Specifically, this replaces the _download_ and _install_, not the load step of Packer.

chase:
Finally getting around to finishing this up

kylo252:
> Finally getting around to finishing this up

We should implement this fail-safe as well: https://github.com/LunarVim/LunarVim/pull/2093#issuecomment-1000167885

chase:
> > Finally getting around to finishing this up
> 
> We should implement this fail-safe as well: [#2093 (comment)](https://github.com/LunarVim/LunarVim/pull/2093#issuecomment-1000167885)

This fail-safe should probably be implemented and merged before this PR, but I do think it's a good idea in theory.

The only problem with this approach is that if the install is significantly out of date, it will still update past the fail-safe or in even more likely cases the update will just fail since `:LvimUpdate` doesn't exist in older versions.

The only real way I can think to get around this is to make `rolling` the legacy rolling branch and `rolling-0.6` the new rolling release branch for Neovim 0.6.  
When deprecating the branch for a specific Neovim version, we can just push an update to the unsupported branch that warns the user on load, saying the version of Neovim is no longer supported and to upgrade.  
When it's three minor versions behind, we can just delete the unsupported rolling branch, then updates will fail and they will probably go to the repo and update with the new instructions.

---

Since `:LvimUpdate` will be deprecated after this PR (no more updates with lvim running) with `lvim --update` taking its place, I think putting the fail-safe in place _along_ with the branch strategy above would be a well-rounded solution.

kylo252:
> The only real way I can think to get around this is to make `rolling` the legacy rolling branch and `rolling-0.6` the new rolling release branch for Neovim 0.6.

Oh, nice idea with the branches, but let's do it the other way around. If neovim-0.5 is detected, then we check out the "compat-0.5x" branch, which is basically just frozen. We would need to execute `git fetch --unshallow` first, but I think it should work otherwise.

> Since `:LvimUpdate` will be deprecated after this PR

Did we rule out the idea of running a headless instance to do the update? Other idea is that maybe we'll replace `LvimSyncCorePlugins` or `PackerSync` with `LvimUpdate` or we could also go for `--upgrade`.

chase:
> Oh, nice idea with the branches, but let's do it the other way around. If neovim-0.5 is detected, then we check out the "compat-0.5x" branch, which is basically just frozen. We would need to execute `git fetch --unshallow` first, but I think it should work otherwise.

I would agree, but we cannot do this for old versions that expect `rolling` or `master` to be the branch. That's the primary reason I suggested it in the first place.

> > Since `:LvimUpdate` will be deprecated after this PR
> 
> Did we rule out the idea of running a headless instance to do the update? Other idea is that maybe we'll replace `LvimSyncCorePlugins` or `PackerSync` with `LvimUpdate` or we could also go for `--upgrade`.

Yes, because inherently a Vim command occurs after the configuration runs. The user config and the core plugins cannot be loaded during an update, otherwise they can and often do break things.

Syncing user plugins using PackerSync will still function as expected. It will not update the core plugins, so no problems should occur.

I don't mind if it is `lvim --upgrade` or `lvim --update` either is okay, it communicates the idea regardless.

Under the hood it basically will just add `--cmd 'lua _G.__lvim_updating=true'` so that the init.lua script will run the update code and not the usual init after bootstrap (like we discussed)

kylo252:
> I would agree, but we cannot do this for old versions that expect `rolling` or `master` to be the branch. That's the primary reason I suggested it in the first place.

Why wouldn't it work? the version validator will be available on both rolling and master, and will then switch to the compat branch if neovim version is 0.5. There shouldn't be any other logic that would break if you're on a specific branch. I'll fix that PR separately either way.

> Yes, because inherently a Vim command occurs after the configuration runs. The user config and the core plugins cannot be loaded during an update, otherwise they can and often do break things.

Sounds good, we can always bring it back if it serves a purpose, it might even be useful to wrap `PackerSync` just in case.

chase:
> > I would agree, but we cannot do this for old versions that expect `rolling` or `master` to be the branch. That's the primary reason I suggested it in the first place.
> 
> Why wouldn't it work? the version validator will be available on both rolling and master, and will then switch to the compat branch if neovim version is 0.5. There shouldn't be any other logic that would break if you're on a specific branch. I'll fix that PR separately either way.

You're assuming that all users will be up to date with the recent changes, this is often not the case.
Consider the case where the user is using a version that is new enough to have LvimUpdate, but they have not updated to get any recent changes including the fail-safe. Suppose they decide to update a month after the fail-safe is introduced but they are using version 0.5.
`:LvimUpdate` will simply pull from `rolling` without checking and without the fail-safe, because the code for the fail-safe is not there. As a result the user will have a version that is not compatible with 0.5.

kylo252:
> Suppose they decide to update a month after the fail-safe is introduced but they are using version 0.5. `:LvimUpdate` will simply pull from `rolling` without checking and without the fail-safe, because the code for the fail-safe is not there.

Okay, I see what you mean, but that should hopefully(?) be fine as long as the validation is done in the `post_update` hook.  I'll switch their orders around in https://github.com/LunarVim/LunarVim/blob/7c1771aa4c6c0820454d5df4ec909ca6a8a74279/lua/lvim/utils/hooks.lua#L25-L31

> As a result the user will have a version that is not compatible with 0.5

The hook should still be there and available to be used by whatever function is pushing the update. So it should _theoretically_ be fine to trigger the update again in order to switch to the correct branch, assuming the bootstrap didn't fail miserably 😅

chase:
> Okay, I see what you mean, but that should hopefully(?) be fine as long as the validation is done in the `post_update` hook.

I considered something like this, but realistically it would have to be run before anything else in init.lua, not in the update hook. This is because the additions you make to the hook won't necessarily be there to run, and they may also fail running. But if it is init.lua, it will run, yet it will not run immediately after the update.

We also can't reliably switch to a 0.5 compat branch from init.lua or the hook and then just keep Neovim running.

If anything it would have to do the version check, switch the branch to the compat branch, clear the cache, then quit with a message saying that `lvim +LvimUpdate` needs to be run and that upgrading to 0.6 is recommended (we aren't backporting `lvim --upgrade`)

> So it should _theoretically_ be fine to trigger the update again in order to switch to the correct branch, assuming the bootstrap didn't fail miserably 😅

I think it's safe to say that for old code and users who rarely update, failing miserably is the common case haha

---

I feel like engineering for a deprecated system to this extent, in any manner, is far more work than doing the `rolling-0.6` branch approach. This also means that once we deprecate 0.6 for some other new version, the branch is already the "compat" branch.

kylo252 (commented):
This just nuked my site folder without asking me, I guess I should've read the code more carefully 😢 

We generally have a few problems with this (download) approach:
- this snapshot that gets downloaded is not actually a valid repo, since it doesn't have `.git`.
- this makes it harder to quickly switch and test plugins' branches and/or actually using `lvim` for plugin development because it will nuke the entire folder.
- we need to remove the original folder (fixable by making sure it's backing up first), but this would have been easier and more reliable if it was an actual repo.
- this is highly unlikely to work on windows, since anything with symlinks is barely supported with libuv and is basically asking for trouble.
- then we have the issue with Packer where it either knows about the core plugins or it ignores them completely, meaning that `PackerStatus` doesn't actually the correct info.

We can get rid of the symlinks by simply using a different `rtp`, but that means that packer doesn't know about them, which makes the whole process too much effort to be practical. 

I think it's easier if we manage the cloning/fetching part manually (headless or not), but still keep Packer responsible for loading. I'm even thinking it might be wisest to add the plugins as git submodules, which might have some overhead with the folder structure, but will shift a lot of responsibilities towards native git commands. Both methods would use shallow clones btw.

What do you think?

chase:
> This just nuked my site folder without asking me, I guess I should've read the code more carefully 😢

That's unfortunate, by "this" do you mean "lvim --update" or just running the init as normal? This code is marked as draft for a reason, I've not pushed any of my recent changes.

Did you store things in there that weren't managed by Packer?

> We generally have a few problems with this (download) approach:
> 
>     * this snapshot that gets downloaded is not actually a valid repo, since it doesn't have `.git`.

For core plugins that's the point. Please read the original issue that describes the implementation details.

For our repo, it is still a git repo.

>     * this makes it harder to quickly switch and test plugins' branches and/or actually using `lvim` for plugin development because it will nuke the entire folder.

By default it does delete core plugins, but it shouldn't delete anything else. So when you say "nuke", I'd like you to explain what was explicitly removed.

> 
>     * we need to remove the original folder (fixable by making sure it's backing up first), but this would have been easier and more reliable if it was an actual repo.

I'm not sure what you mean "remove the original folder", could you specify which folder?
 
>     * this is highly unlikely to work on windows, since anything with symlinks is barely supported with libuv and is basically asking for trouble.

Incorrect, symlinks are explicitly supported by Packer and work for all local plugins already.


>     * then we have the issue with Packer where it either knows about the core plugins or it ignores them completely, meaning that `PackerStatus` doesn't actually the correct info.

PackerStatus only works for git, these are not git. They are local, this is intended behavior. Core plugins should only be updated using `lvim --update`

> We can get rid of the symlinks by simply using a different `rtp`, but that means that packer doesn't know about them, which makes the whole process too much effort to be practical.

Again, symlinks are supported. This is a non-issue. However, Windows support does need to be tested explicitly as I'm not sure if it works out of the box. Are there official Windows setup instructions for LunarVim anywhere?

> I think it's easier if we manage the cloning/fetching part manually (headless or not), but still keep Packer responsible for loading. I'm even thinking it might be wisest to add the plugins as git submodules, which might have some overhead with the folder structure, but will shift a lot of responsibilities towards native git commands. Both methods would use shallow clones btw.

We do manage the clone/fetch ourselves. We do use Packer for loading. Please re-read the original issue for implementation details.

I do not think introducing git submodules solves our problems.
My approach works, but unfortunately you ran it in an incomplete or draft state.

---

You did identify a case where you wanted to override a core plugin and you wanted the full git repo. I'll have to think of a better solution, since like you, I occasionally submit upstream PRs.

kylo252:
> That's unfortunate, by "this" do you mean "lvim --update" or just running the init as normal? This code is marked as draft for a reason, I've not pushed any of my recent changes.
> By default it does delete core plugins, but it shouldn't delete anything else. So when you say "nuke", I'd like you to explain what was explicitly removed.

Sorry, I didn't clarify it better. I just meant the actual plugin folder. So one would basically lose the ability to have multiple branches, or really anything git-related with those plugins.

> I'm not sure what you mean "remove the original folder", could you specify which folder?

The plugin folder itself, same as the previous point.

> PackerStatus only works for git, these are not git. They are local, this is intended behavior. Core plugins should only be updated using `lvim --update`

That's the point I was trying to make, that `PackerStatus` is out of the picture now for better or for worse. I'm also not sure that it's actually ignoring the core plugins since `PackerSync` kept wanting to remove and re-install them or something. But like you said, it's at a draft state, so I can't say for certain.

> Again, symlinks are supported. This is a non-issue. However, Windows support does need to be tested explicitly as I'm not sure if it works out of the box. Are there official Windows setup instructions for LunarVim anywhere?

I know that they work, but my point is specifically about windows being troublesome with symlinks.

> I do not think introducing git submodules solves our problems. My approach works, but unfortunately you ran it in an incomplete or draft state.

I didn't say it doesn't work, everything was fine, but it's more about how pros and cons. Try this as a test if you'd like: https://github.com/kylo252/LunarVim/tree/submodules

chase:
I think that I put everything in the right places, but let me know @kylo252 

A few things left that I need to do:

- Test and fix whatever Windows throws at me
- Remove the stale plugins when updating
- Add the "developer mode" that uses shallow git repos

This also introduces `lvim --update-core` which only updates the core plugins based on the local branch, without trying to fetch and reset to a remote branch. Essentially, it's what test-runner.sh uses.

utils/bin/lvim
```diff
@@ -3,4 +3,36 @@
 export LUNARVIM_RUNTIME_DIR="${LUNARVIM_RUNTIME_DIR:-$HOME/.local/share/lunarvim}"
 export LUNARVIM_CONFIG_DIR="${LUNARVIM_CONFIG_DIR:-$HOME/.config/lvim}"
 
+headless_update() {
+  nvim --cmd 'lua _G.__lvim_updating=true' -u "$LUNARVIM_RUNTIME_DIR/lvim/init.lua" --noplugin -n -i NONE --headless
```
kylo252, line 7: Would this quit at the end? maybe with an explicit `+'quitall'`, or some `on_complete` event.

utils/bin/lvim
```diff
@@ -3,4 +3,36 @@
...
+
+for arg; do
+  shift
+  case $arg in
+    --update)
+      :
+      echo 'Updating LunarVim...'
+      old="$(pwd)"
+      cd "$LUNARVIM_RUNTIME_DIR/lvim"
+      target_branch="${1:-$(git rev-parse --abbrev-ref HEAD)}"
```
kylo252, line 18: If it doesn't quit, then there's a problem here because you're expecting the name of the branch as a positional argument, but also assigning a default one, so any arguments that are passed after will be passed to `git fetch` without sanitization.

```suggestion
      target_branch="${LV_BRANCH:-$(git rev-parse --abbrev-ref HEAD)}"
```

utils/bin/lvim
```diff
@@ -3,4 +3,36 @@
...
+      git stash --include-untracked -m "Changes before LunarVim update" && echo "Uncommitted changes saved to stash" || true
+      git fetch --force --depth=1 origin $target_branch:$target_branch
```
kylo252, line 20: If you're starting with a shallow clone, then you won't be able to see any remote branch other the current. You can see the workaround I added in https://github.com/LunarVim/LunarVim/pull/2127/files#diff-592059d4c0b81bb5e3509fc175b0d838215ff52d6cfd5015ab13181b29f569deR37-R51

utils/bin/lvim
```diff
@@ -3,4 +3,36 @@
...
+}
+
+for arg; do
+  shift
+  case $arg in
+    --update)
+      :
+      echo 'Updating LunarVim...'
+      old="$(pwd)"
+      cd "$LUNARVIM_RUNTIME_DIR/lvim"
```
kylo252, line 17: You don't need to change the directory if you only need it for `git`, instead you can use
```console
git -C $LVIM_BASE_DIR status
```

utils/bin/lvim
```diff
@@ -3,4 +3,36 @@
...
+      target_branch="${1:-$(git rev-parse --abbrev-ref HEAD)}"
+      git stash --include-untracked -m "Changes before LunarVim update" && echo "Uncommitted changes saved to stash" || true
```
kylo252, line 19: What's the point of stashing if we're going to ignore any errors?

kylo252 (commented):
The base-repo update logic should be moved into a specialized updater script: 
1. it would make it easier to work on and debug without worrying about updating the binary every time
2. this currently lacks nearly any error handling or logging. Once we start adding those, then it will end up getting longer and more complicated, bringing us back to point 1
3. separation of concerns, since the binary, _ideally at least_, shouldn't have too many responsibilities

utils/bin/lvim
chase: Please look at `updater.lua`

utils/bin/lvim
chase: It does quit.
Not expecting a branch, it is optional. If a branch is provided it will fetch and use it, otherwise it uses the current branch.
If it fails, the script also fails.

utils/bin/lvim
chase: That's not true

utils/bin/lvim
chase: Doing that for every command is tedious, but I guess we could? I'm not sure why cd is bad here.

utils/bin/lvim
chase: stash seems to return non-0 for normal operation every so often, do you have a better approach for shelving changes made by developers like us messing around with the internals?

utils/bin/lvim
chase: image

chase:
> The base-repo update logic should be moved into a specialized updater script:
> 
>     1. it would make it easier to work on and debug without worrying about updating the binary every time

I don't think completely splitting it off is the right choice, how about something like this?

```
#!/bin/sh

export LUNARVIM_RUNTIME_DIR="${LUNARVIM_RUNTIME_DIR:-$HOME/.local/share/lunarvim}"
export LUNARVIM_CONFIG_DIR="${LUNARVIM_CONFIG_DIR:-$HOME/.config/lvim}"

# flags
UPDATE=
UPDATE_CORE=

for arg; do
  shift
  case $arg in
    --update) UPDATE=1 ;;
    --update-core) UPDATE_CORE=1 ;;
    *) set -- "$@" "$arg" ;;
  esac
done

if [ $UPDATE ]; then
  $LUNARVIM_RUNTIME_DIR/utils/bin/update.sh
elif [ $UPDATE_CORE ]; then
  echo 'Updating core plugins from local...'
  $LUNARVIM_RUNTIME_DIR/utils/bin/update_core.sh
fi

exec nvim -u "$LUNARVIM_RUNTIME_DIR/lvim/init.lua" "$@"
```

>     2. this currently lacks nearly any error handling or logging. Once we start adding those, then it will end up getting longer and more complicated, bringing us back to point 1

I'm not sure what more logging you want. It prints to the console for just about everything it does? Are you saying logging to a file as well?
As for error handling, there's plenty of it in `updater.lua`. Do you explicitly mean in the bash script?

> 
>     3. separation of concerns, since the binary, _ideally at least_, shouldn't have too many responsibilities

Right now, the shim has no responsibilities, it's little more than an alias that isn't terribly helpful.

Where the idea started was here: https://github.com/LunarVim/LunarVim/issues/1397

I think _most_ of those things can be done using user commands, as you've largely implemented. Update just happens to be one that cannot reliably be done as a user command.

utils/bin/lvim
chase: Can drop the `|| true`, since for regular users this shouldn't ever matter

kylo252:
> I don't think completely splitting it off is the right choice, how about something like this?

Yeah, that was my intent as well, just call another script to handle any nitty-gritty details.

> I'm not sure what more logging you want.[...] Do you explicitly mean in the bash script?

I was just referring to the bash-script section.

> Right now, the shim has no responsibilities, it's little more than an alias that isn't terribly helpful.

Yes and no, it's always been a shim that's used for setting some environment variables, but it also needs to maintain compatibility with any `nvim` flag. The main reason I want to stick with user-commands as much as possible is to keep the output of `lvim --help` consistent with what's available, because the next thing you know is that somebody thinks that `--help` should include an entry for `--update` and makes a PR to fix it. Now you're stuck re-inventing the wheel with some argparser hack, or maintaining a duplicate list manually.

.gitignore
```diff
@@ -21,6 +21,7 @@ nv-settings.lua
 lv-settings.lua
 lua/lv-user/
 lua/lv-user-config/
+tests/artifacts/
```
kylo252, line 24:
```suggestion
```
This is now fixed

init.lua
```diff
@@ -5,17 +5,25 @@ if not vim.tbl_contains(vim.opt.rtp:get(), base_dir) then
...
+---@diagnostic disable-next-line: undefined-field
+local updating = _G.__lvim_updating or false
 
-local plugins = require "lvim.plugins"
-require("lvim.plugin-loader").load { plugins, lvim.plugins }
+require("lvim.bootstrap"):init(base_dir, updating)
 
-local Log = require "lvim.core.log"
-Log:debug "Starting LunarVim"
+if not updating then
```
kylo252, line 16:
```suggestion
```

It shouldn't be necessary to do this, since `updater.init()` should always quit at the end. We should even go a step further and force it to quit on a timer, just so it doesn't get stuck in the CI.

init.lua
```diff
@@ -5,17 +5,25 @@ if not vim.tbl_contains(vim.opt.rtp:get(), base_dir) then
   vim.opt.rtp:append(base_dir)
 end
 
-require("lvim.bootstrap"):init(base_dir)
+require "lvim.utils.headless_fix"
 
-require("lvim.config"):load()
+-- This is only defined when updating using `lvim --update`
+---@diagnostic disable-next-line: undefined-field
+local updating = _G.__lvim_updating or false
```
kylo252, line 13: These belong in bootstrap. There shouldn't be any errors from just calling the bootstrap, and if there are, then we have a different problem and the updater shouldn't even run 😅

lua/lvim/core/log.lua
```diff
@@ -36,21 +36,28 @@ function Log:init()
...
-            "%s [%-5s] %s: %-30s",
-            { "timestamp", "level", "logger_name", "msg" }
-          ),
-        }),
       },
     },
   }
 
+  -- if the cache dir doesn't exist, the file sink causes a crash
+  if vim.loop.fs_stat(get_cache_dir()) then
```
kylo252, line 44: Can you show me how this error was getting triggered? The installer should be taking care of this 
https://github.com/LunarVim/LunarVim/blob/e1b8eb01a2e2ca092f761b25777ba60d76609c30/utils/installer/install.sh#L292-L302

lua/lvim/plugins.lua
```diff
@@ -272,3 +276,14 @@ return {
     disable = not lvim.builtin.terminal.active,
   },
 }
+
+if os.getenv "LVIM_DEV" ~= "1" then
```
kylo252, line 280: We should probably create a global variable instead, just so we have more flexibility with how to enable dev-mode.

lua/lvim/updater.lua
```diff
@@ -0,0 +1,298 @@
...
+    total = total + 1
+    print_update()
+  end
+  return disp
+end
+
+function M.init(opts)
+  local core_install_dir = opts.core_install_dir or vim.fn.stdpath "data" .. "/core"
+  assert(vim.fn.executable "curl", "curl is not installed")
+  assert(vim.fn.executable "unzip", "unzip is not installed")
```
kylo252, line 177: These should be moved into `download_and_install()` or wrapped to not execute in `LVIM_DEV`

lua/lvim/updater.lua
```diff
@@ -0,0 +1,298 @@
...
+  vim.fn.delete(path_to_remove, "rf")
+end
+
+local function cleanup_stale(core_install_dir, core_plugins)
+  return async(function()
+    local err, dir = await(async_opendir(core_install_dir))
+    assert(not err, "unable to open core plugin install directory")
+    local entries
+    err, entries = await(async_readdir(dir))
+    assert(not err, "unable to read core plugin install directory")
```
kylo252, line 117: Should we wrap this `assert` to include an error handler (which may decide to do `cquit`) and potentially dump the error string to some logfile, either the default one, or a temporary one?

kylo252 (changes requested):

.gitignore
chase: Ah, tests no longer produce these locally?

init.lua
chase: As I mentioned previously, this is incorrect, the actions are asynchronous and handled in a callback tied to an even in Packer.

init.lua
chase: I know you don't like things getting put into init.lua, but see above as to why the `if` guard is necessary.

It is being put into bootstrap.

It just doesn't arbitrarily choose to pick up on a global which may not be set if bootstrap.lua is used from nvim without the init.lua file like you described to me, instead it takes it as a parameter which changes the mode.

lua/lvim/core/log.lua
chase: You're assuming that it is installed using the installer, which may not be the case. For example, while testing the directories are changed _after_ installation. What if the environment variables are changed after the installation?

It's also possible that a user might delete it, or simply not transfer it during a backup, because it's the cache directory. There's also little overhead to this check, there's actually more overhead for doing a pcall instead.

lua/lvim/updater.lua
chase: Good catch!

lua/lvim/plugins.lua
chase: I'm hesitant to introduce more globals, but I do agree it's more flexible that way. Although I'd like there to be "one source of truth" for this, not several.

Should we add a global for test environment as well?

lua/lvim/updater.lua
chase: These are fatal errors, if they occur, they are logged to the console and the process exits.
Mind you, this is _before_ the logging facilities with structlog are available. I also don't see why logging this to a file is necessary in any way.

init.lua
chase: TLDR: if you want to use update mode elsewhere just use `bootstrap:init(basedir, true)` instead of relying on a global to be set before

init.lua
kylo252: Yup, fixed here: https://github.com/LunarVim/LunarVim/pull/2132/files#diff-8a5cb93842556000a3a4432fecd8e2e1c6d30fe458092fbb252494d5c94727a9R11

lua/lvim/core/log.lua
kylo252: It generally sounds like a problem that structlog should handle, but otherwise it's easier and cleaner to just add this line in 
```lua
function Log:init()
  local status_ok, structlog = pcall(require, "structlog")
  if not status_ok then
    return nil
  end
  vim.fn.mkdir(get_lvim_cache_dir(), "p")
```
If you want to be really strict about it, then you can add
```lua
  local utils = require "lvim.utils"
  local cache_dir = get_cache_dir()
  if not utils.is_directory(cache_dir) then
	vim.fn.mkdir(cache_dir, "p")
  end
```

lua/lvim/plugins.lua
kylo252: > Should we add a global for test environment as well?

We shouldn't need a test variable as long as there are no issues with impatient (regardless if it's enabled or not), so you can replace that test completely with new tests for the new updater mode.

lua/lvim/updater.lua
kylo252: I'm mostly thinking about the first part, how did you make them to be fatal? A normal `assert` will just raise an exception without really killing the process.

chase: It will raise an error, yes. Since there is no pcall wrapping it, and there shouldn't be, that causes it to terminate.

This is the point of an assertion, it asserts that the condition is valid and otherwise will not continue executing the code that follows.

Very simple way of testing this is to make the directory then remove permissions to read the directory.

kylo252: What I mean is that it will not terminate the actual neovim process, so the CI workflow will get stuck because neovim isn't exiting.

```lua
-- add this to your init.lua
assert(os.getenv "MY_ENV", "missing env")

vim.schedule(function()
  print "hello"
end)
```

init.lua
chase: Returning early works, but personally I think that makes it less clear. We can go with that if you think it's better.

lua/lvim/updater.lua
chase: I assumed incorrectly that like for other print statements the headless fix would prevent the short message handler. I guess ultimately even with patching some functions, [the neovim bug remains forever punted](https://github.com/neovim/neovim/issues/9358)

I'll just make an assert_or_fail function that exits with 1 after echoing the error and use that.

Thanks for pointing this out.

lua/lvim/plugins.lua
chase: I ended up using it to not display the progress indicator (since it spams lines only in the GitHub logs, but not for users due to using `\r`), so I re-added it as a global in a similar manner

chase:
After much consideration and testing, I have reverted the part of this PR that implements a custom downloader and Packer plugin type. At this point, it is hardly the most valuable portion of this PR.

**Why?**

It seems that on occasion, GitHub does extensive throttling on downloading the `.zip` archives. I've had a few installs where the worst case took almost 2x as long as the original implementation by Packer, where as the average case for a git-based install directly by Packer is only 2-3s slower on a fast connection.

The maintenance burden as Packer's API surface changes over time does not justify such negligible gains.

chase:
What remains to be done for WIndows:
- [ ] Update lvim stub
- [ ] Implement `lvim --update` in Powershell
- [ ] Update installation Powershell to use `lvim --update`

lua/lvim/bootstrap.lua
```diff
@@ -92,31 +95,36 @@ function M:init(base_dir)
...
 end
 
----Update LunarVim
----pulls the latest changes from github and, resets the startup cache
+---@deprecated
 function M:update()
-  require_clean("lvim.utils.hooks").run_pre_update()
-  require_clean("lvim.utils.git").update_base_lvim()
-  require_clean("lvim.utils.hooks").run_post_update()
+  vim.api.nvim_err_writeln "LvimUpdate has been deprecated. Exit lvim and use `lvim --update` instead"
```
kylo252, line 127:
```suggestion
  local msg =
    "Deprecation notice: LvimUpdate is no longer supported. See https://github.com/LunarVim/LunarVim#breaking-changes"

  vim.schedule(function()
    require("lvim.core.log"):warn(msg)
  end)
```

lua/lvim/core/terminal.lua
```diff
@@ -3,6 +3,8 @@ local Log = require "lvim.core.log"
 
 M.config = function()
   lvim.builtin["terminal"] = {
+    active = true,
```
kylo252, line 6:
```suggestion
    active = false,
```
We should probably leave this to false

lua/lvim/updater.lua
```diff
@@ -0,0 +1,233 @@
...
+    local user_lvim
+    packer.on_complete = function() end
+    packer.on_compile_done = function()
+      if packer_stage == 0 then
+        print("\rInstalled core plugins in:", timer:stop(), "ms")
+        packer_stage = 1
+
+        timer:start()
+        print "Loading core plugins..."
+        packer.compile()
```
kylo252, line 150: You need to be careful with this, since it's async and it might actually be slower than you think.

utils/bin/test_runner.sh
```diff
@@ -11,6 +11,9 @@ TEST_BASE_DIR="$(mktemp -d)"
 export LUNARVIM_CONFIG_DIR="$TEST_BASE_DIR"
 export LUNARVIM_CACHE_DIR="$TEST_BASE_DIR"
 
+# update
+nvim --cmd 'lua _G.__lvim_updating=true' -u "$LUNARVIM_RUNTIME_DIR/lvim/init.lua" --noplugin -n -i NONE --headless
+
```
kylo252, line 16: This doesn't need to be here, right?

utils/installer/download_unzip.ps1
```diff
@@ -0,0 +1,10 @@
+$Url = $args[0]
+$ZipFile = $env:TEMP + '\' + $(Split-Path -Path $Url -Leaf) 
+$Destination= $args[1]
+ 
+Invoke-WebRequest -Uri $Url -OutFile $ZipFile 
+ 
+$ExtractShell = New-Object -ComObject Shell.Application 
+$Files = $ExtractShell.Namespace($ZipFile).Items() 
+$ExtractShell.NameSpace($Destination).CopyHere($Files) 
+Start-Process $Destination
```
kylo252, line 10: We don't need this file anymore, right?

kylo252 (changes requested):

lua/lvim/updater.lua
```diff
@@ -0,0 +1,233 @@
...
+local function get_lvim_after_user_config()
+  local original_lvim = lvim
+  local user_lvim = vim.deepcopy(lvim)
+  local original_package_loaded = package.loaded
+  local user_package_loaded = {}
+  _G.lvim = user_lvim
+  _G.package.loaded = user_package_loaded
+  local ok, err = pcall(dofile, require("lvim.config"):get_user_config_path())
+  if not ok then
+    print(err)
```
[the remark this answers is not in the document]
chase: I believe you misunderstand the use of this function, so it the variable swaps and the rest of it doesn't make sense to you. Think of it like a sandbox for the extensive damage the user config file does by default, it is inherently unpredictable and full of side-effects.

The user's `config.lua` file (as retrieved by get_user_config_path) can run arbitrary code. As such, it can and by default does mutate globals such as `_G.lvim` and `_G.package`. For the purpose of the updater it must, not influence the state of neovim, which [`config:load()` does extensively even after loading](https://github.com/LunarVim/LunarVim/blob/rolling/lua/lvim/config/init.lua#L82-L100). The user configured values for certain parts however, are valuable.

> currently re-assigning the values without even using it

This is incorrect, I think you may have misread the code or misunderstand the effect of what `dofile` is here. There are value swaps for globals that the user configuration mutates, but are integral to the updater script without allowing it to mutate everything else. For example, the state of core plugins, the user installed plugins, etc.

Should it use `getenv` and `setenv` for sandboxing behavior and sideffects? Absolutely, but the level of complexity to get that work work with how `lvim` is setup was far greater than this short function when I wrote this.

---

TLDR: it loads the user's configuration file in such a way that it does not mutate state, it is required and is not redundant. It's not missing error handling.

lua/lvim/bootstrap.lua
chase: I don't see why this is better.

1. It doesn't state what the user should use instead, which extremely unhelpful.
2. Instead it requires the user to visit a URL which _does not_ have a permalink.
3. It unnecessarily uses logging for something which is direct feedback to a user action

lua/lvim/core/terminal.lua
chase: Why is the terminal plugin false by default when shortcuts in whichkey require it?
Honest question, I'm fine with false, I just don't recall why this was disabled.

It wasn't false before, but nil, which means that it isn't available in the user's config for autocomplete.

lua/lvim/updater.lua
chase: No, not really.

This entire section of code handles the asynchronous callbacks of Packer in stages and it is in fact not slow. I'm well aware of how the internals of Packer work.

lua/lvim/core/terminal.lua
kylo252: It feels weird to add it just for `lazygit`, but I don't really mind either way.

utils/installer/download_unzip.ps1
chase: Good catch, you're right.

lua/lvim/bootstrap.lua
kylo252: 1. It's more of a pattern we've been using: https://github.com/LunarVim/LunarVim/blob/ec1e4140fa4156df50785da5d6297145762a7daa/lua/lvim/config/init.lua#L47-L60
2. I realize it's not a permalink, but it's intentional since it can be more flexible with the instructions
3. I don't really feel it needs to be an error, but rather a warning. And this amount of code is the same it would take for `vim.notify`, which is already handled directly in the logger

That said, we can still add a note, while still pointing to the README
> use `lvim --update` instead

chase: That pattern is terribly unhelpful for a command, in contrast to user settings. It also doesn't point out _which_ change in the breaking changes. This _immediately_ provides the correct information and _only_ the information they need. It only displays when the user tries, not when the user loads `lvim`.

For a user's _config_ I see the usefulness of the pattern because a setting is deprecated, of which there can be many, and it should be logged so that a user can review any series of them.

kylo252:
Reminder: if we're doing the plugins' version uplift in headless, then we need to log all their "breaking" changes. Just use whatever packer is using, hopefully it's logging those already.

PR #1702, opened 2021-10-05
feat: better telescope integration (WIP)

kylo252:
# Description

Better telescope integration

PR #2089, opened 2021-12-16
[Feature]: Use Telescope's git_files with fallback

### LINKED ISSUE #2088: [Feature]: Use Telescope's git_files instead of find_files by default
### Feature motivation

Using Telescope's `find_files` keybinding can be a bit of a pain. By default, this doesn't display hidden files in LunarVim, and if you set `lvim.builtin.telescope.pickers = { find_files = { hidden = true } }`, then the finder displays all of the files in the `.git` directory.

I just discovered that Telescope has a built-in `git_files` function. This works exactly like `find_files`, except it displays all files in the current repo not ignored by the `.gitignore` file. 

The one downside to this picker is that it produces an error when not in a git directory. I'd propose that LunarVim add a function that displays `git_files` when in a Git repo, and falls back to `find_files` when not.

I've already set my local config to use this, but this feels like a sane default most developers would want. 

### Additional context

_No response_

LandonSchropp:
# Description

Currently, LunarVim uses Telescope's `find_files` function by default. This has a couple of drawbacks:

* It doesn't respect the `.gitignore` file.
* By default, it doesn't display hidden files.

It turns out Telescope has a `git_files` function. The only downside to this function is that it produces an error when not in a Git repo. Luckily, Telescope provides [documentation](https://github.com/nvim-telescope/telescope.nvim/wiki/Configuration-Recipes#falling-back-to-find_files-if-git_files-cant-find-a-git-directory) on how to work around this and use `find_files` as a fallback.

Fixes #2088 

## How Has This Been Tested?

I ran this locally in both a Git repo as well as in a normal directory. It seems to work fine.


kylo252 (changes requested):
I personally think this binding should just go into the personal `config.lua`, since #1702 allows you to override the default find command to whatever you want anyway. Also, `fd` actually respects `.gitignore` by default.

LandonSchropp:
Continuing discussion on this in #2088. This PR might be closed in favor of a better solution, but I'll leave it open until we come up with a resolution in that issue.

lua/lvim/core/telescope.lua
```diff
@@ -153,4 +153,15 @@ function M.setup()
...
+-- contained in a Git repo.
+function M.git_files_with_fallback()
+  local builtin = require "telescope.builtin"
+  local ok = pcall(builtin.git_files)
+
+  if not ok then
+    builtin.find_files()
+  end
+end
+
```
kylo252, line 166: Please move this to `telescope/custom-finders` and then we can merge it, since it's a good addition anyway.

kylo252 (changes requested):

lua/lvim/core/telescope.lua
kylo252: Maybe call it `find_project_files()` instead?

LandonSchropp: Sure, can do. 🙂

lua/lvim/core/telescope/custom-finders.lua
```diff
@@ -82,4 +82,15 @@ function M.view_lunarvim_changelog()
   }):find()
 end
 
+-- Smartly opens either git_files or find_files, depending on whether the working directory is
+-- contained in a Git repo.
+function M.find_project_files()
+  local builtin = require "telescope.builtin"
```
kylo252, line 88:
```suggestion
```

This is what's causing the CI error, since this variable is already defined.

LandonSchropp: Sorry, I'm just looping back to this now. I'll get it fixed.

lua/lvim/core/which-key.lua
```diff
@@ -68,7 +70,7 @@ M.config = function()
       ["q"] = { "<cmd>q!<CR>", "Quit" },
       ["/"] = { "<cmd>lua require('Comment').toggle()<CR>", "Comment" },
       ["c"] = { "<cmd>BufferClose!<CR>", "Close Buffer" },
-      ["f"] = { "<cmd>Telescope find_files<CR>", "Find File" },
+      ["f"] = { custom_finders.find_project_files, "Find File" },
```
kylo252, line 73: I had no idea that this actually works. I think it's better to not require that file too early, since it might cause errors on new installations.

```suggestion
      ["f"] = { require ("lvim.core.telescope.custom-finders").find_project_files, "Find File" },
```

lua/lvim/core/which-key.lua
```diff
@@ -1,3 +1,5 @@
+local custom_finders = require "lvim/core/telescope/custom-finders"
+
```
kylo252, line 2:
```suggestion
```

kylo252 (changes requested):

lua/lvim/core/which-key.lua
LandonSchropp: ✅

lua/lvim/core/which-key.lua
LandonSchropp: ✅

kylo252:
Thank you!

kylo252 (approved):

LandonSchropp:
Thank you!

PR #2093, opened 2021-12-18
refactor: uplift neovim's minimum version requirement to 0.6.0

abzcoding:
# Description

update plugins and require nvim > 0.6


kylo252 (commented):
We should remove the deprecated handlers as well as the unnecessary custom `show_line_diagnostics()` in: https://github.com/LunarVim/LunarVim/blob/ccb5a6f0605c9429c2a44f3a800d7c6da6ca7752/lua/lvim/lsp/handlers.lua

abzcoding:
> We should remove the deprecated handlers as well as the unnecessary custom `show_line_diagnostics()` in: https://github.com/LunarVim/LunarVim/blob/ccb5a6f0605c9429c2a44f3a800d7c6da6ca7752/lua/lvim/lsp/handlers.lua

sounds good to me 👍 i'll do it in a couple of hours

lua/lvim/lsp/null-ls/init.lua
```diff
@@ -9,14 +9,7 @@ function M:setup()
...
 
-  null_ls.config(lvim.lsp.null_ls.config)
-  local default_opts = require("lvim.lsp").get_common_opts()
-
-  if vim.tbl_isempty(lvim.lsp.null_ls.setup or {}) then
-    lvim.lsp.null_ls.setup = default_opts
-  end
-
-  require("lspconfig")["null-ls"].setup(lvim.lsp.null_ls.setup)
+  null_ls.setup(lvim.lsp.null_ls.config)
```
kylo252, line 12:
```suggestion
  local default_opts = require("lvim.lsp").get_common_opts()
  null_ls.setup(vim.tbl_deep_extend("force", default_opts, lvim.lsp.null_ls.setup))
```
And then we can remove `lvim.lsp.null_ls.config` which hopefully is seldom used anyway.

utils/installer/config.example.lua
```diff
@@ -100,11 +100,9 @@ lvim.builtin.treesitter.highlight.enabled = true
 --   buf_set_option("omnifunc", "v:lua.vim.lsp.omnifunc")
 -- end
 -- you can overwrite the null_ls setup table (useful for setting the root_dir function)
--- lvim.lsp.null_ls.setup = {
---   root_dir = require("lspconfig").util.root_pattern("Makefile", ".git", "node_modules"),
--- }
+-- lvim.lsp.null_ls.config.root_dir = require("lspconfig").util.root_pattern("Makefile", ".git", "node_modules")
```
kylo252, line 103:
```suggestion
-- lvim.lsp.null_ls.setup.root_dir = require("lspconfig").util.root_pattern("Makefile", ".git", "node_modules")
```

utils/installer/config.example.lua
```diff
@@ -100,11 +100,9 @@ lvim.builtin.treesitter.highlight.enabled = true
...
 -- or if you need something more advanced
--- lvim.lsp.null_ls.setup.root_dir = function(fname)
+-- lvim.lsp.null_ls.config.root_dir = function(fname)
```
kylo252, line 105:
```suggestion
-- lvim.lsp.null_ls.setup.root_dir = function(fname)
```

utils/installer/install.sh
```diff
@@ -172,11 +172,11 @@ function print_missing_dep_msg() {
 
 function check_neovim_min_version() {
   # TODO: consider locking the requirement to 0.6+
```
kylo252, line 174:
```suggestion
```

kylo252 (changes requested):

abzcoding:
@kylo252, can we do a release before merging this and mark that one the last compatible version with neovim 0.5 🥺

kylo252:
> @kylo252, can we do a release before merging this and mark that one the last compatible version with neovim 0.5 🥺 

Yeah, that's what I was hoping for as well, although 1.0.0-rc is actually a tag.

kylo252:
I broke just a small part of this to fix the issue with lsp-installer.

lua/lvim/lsp/handlers.lua
```diff
@@ -3,138 +3,13 @@
...
-      local diagnostics = params.diagnostics
-      vim.lsp.diagnostic.save(diagnostics, bufnr, client_id)
-      if not vim.api.nvim_buf_is_loaded(bufnr) then
-        return
-      end
-      vim.lsp.diagnostic.display(diagnostics, bufnr, client_id, config)
-    end
-
-    vim.lsp.handlers["textDocument/hover"] = vim.lsp.with(vim.lsp.handlers.hover, {
-      border = lvim.lsp.popup_border,
```
kylo252, line 33: We need to remove the corresponding option as well, see https://github.com/LunarVim/LunarVim/issues/2095

abzcoding:
> I'm getting this error on `:LvimReload` 😢
> 
> ```lua
> Error executing vim.schedule lua callback: .../site/pack/packer/start/nvim-lspconfig/lua/lspconfig.lua:58: attempt to call field 'launch' (a nil value)
> stack traceback:
>         .../site/pack/packer/start/nvim-lspconfig/lua/lspconfig.lua:58: in function 'fn'
>         vim.lua:415: in function 'cb'
>         vim.lua:285: in function <vim.lua:285>
> Press ENTER or type command to continue
> ```
> 
> I'm also thinking we should add the deprecated settings to
> 
> https://github.com/LunarVim/LunarVim/blob/5e44e6d59bb8bac477c230dd8a463ddc1dd91799/lua/lvim/config/init.lua#L46

https://github.com/neovim/nvim-lspconfig/issues/1582

abzcoding:
@kylo252 can you test it again?

kylo252:
Now I'm getting this error: 
> LSP[id=6] client has shut down after sending the message

abzcoding:
> Now I'm getting this error:
> 
> > LSP[id=6] client has shut down after sending the message
> 
> Edit: there's also something wrong with that implementation. Check `:LspInfo` before and after `:LspRestart`

yeah wanted to mention the error has been mentioned in nvim-lspconfig and null-ls, buy you already commented on those so 😅

abzcoding:
should we wait to make sure those are fixed and then merge this?

kylo252:
> should we wait to make sure those are fixed and then merge this?

Yeah, I would say so, since it's probably not worth it to add more workarounds for this. Alternatively, you can break up the null-ls changes into a separate PR if you'd like.

lua/lvim/lsp/handlers.lua
kylo252: I was actually wrong, sorry 😢 
We need to restore these along with all the lsp options.

abzcoding: 🚬

abzcoding:
@kylo252 can you test this again? i just bumped the `null_ls` and `nvim-lspconfig`, so those issues should be fixed now ( though you might need a `LvimCacheReset` probably

lua/lvim/lsp/handlers.lua
abzcoding: we can merge your PR on top of this

kylo252:
Works fine! 🎉

abzcoding:
@kylo252 can you test it one last time 🥺

kylo252 (approved):
Thanks for all the effort you put into this, appreciated as always ❤️

kylo252:
There are one or two issues that I can only notice on `:LvimReload`, probably related to the new null-ls workflow, but nothing too major. We can deal with them later on, since they require further investigations.

abzcoding:
> There are one or two issues that I can only notice on `:LvimReload`, probably related to the new null-ls workflow, but nothing too major. We can deal with them later on, since they require further investigations.

yeah, also we might need to change the which-key workflow as well after this, I didn't update it here because it acted very weird

kylo252:
Shall we give it a day before merging into master?

abzcoding:
> Shall we give it a day before merging into master?
> 
> Edit: I got an idea for dealing with this ~on master~! We can push a change after we merge this that prevents `LvimUpdate` from pulling anything after that tag/commit if you don't have v0.6. That way, we'll be certain it won't brick itself 😃
> 
> Edit2: we should probably do this for both rolling and master.

sounds good to me 👍

PR #2094, opened 2021-12-18
refactor: remove unused old language configs

kylo252:
# Description

- remove utils/julia since it's supported fully upstream
- remove hard-coded `config/supported_languages.lua` and rely
on nvim-lsp-installer to get an updated list of supported languages
- remove dependency on deprecated `close_preview_autocommand`


abzcoding (approved):
👍

PR #2107, opened 2021-12-21
refactor(install.sh): fix typo in node error message

stonepreston:
# Description

Typos existed  in the error message when installing node dependencies. You was changed to your.  "Please set you NPM_HOME correctly and try again." is now "Please set your NPM_HOME correctly and try again." The misspelling of privileges was also fixed. 

Testing:

- Run command `./install.sh` (do not use sudo, and also ensure you have not set up a prefix for the global installation directory as described [here](https://docs.npmjs.com/resolving-eacces-permissions-errors-when-installing-packages-globally). 
- Ensure the error message regarding NPM is now grammatically correct.

### DIFF
--- a/utils/installer/install.sh
+++ b/utils/installer/install.sh
@@ -226,7 +226,7 @@ function __validate_node_installation() {
   manager_home="$($pkg_manager config get prefix 2>/dev/null)"
 
   if [ ! -d "$manager_home" ] || [ ! -w "$manager_home" ]; then
-    echo "[ERROR] Unable to install without administrative privilages. Please set you NPM_HOME correctly and try again."
+    echo "[ERROR] Unable to install without administrative privileges. Please set your NPM_HOME correctly and try again."
     exit 1
   fi
 }


utils/installer/install.sh
```diff
@@ -226,7 +226,7 @@ function __validate_node_installation() {
   manager_home="$($pkg_manager config get prefix 2>/dev/null)"
 
   if [ ! -d "$manager_home" ] || [ ! -w "$manager_home" ]; then
-    echo "[ERROR] Unable to install without administrative privilages. Please set you NPM_HOME correctly and try again."
+    echo "[ERROR] Unable to install without administrative privilages. Please set your NPM_HOME correctly and try again."
```
soufianeisnotafunction, line 229: I think you can also fix _privilages_ to _privileges_

stonepreston: fixed that one as well! Thanks for the review. Let me know if you see anything else

abzcoding (approved):
tyvm

PR #2109, opened 2021-12-22
fix(lsp): set the handlers opts for v0.6 as well

### LINKED ISSUE #2095: [Bug]: lvim.lsp.popup_border ignored by handlers
### Problem description

change `lvim.lsp.popup_border` doesn't has any effect.
I have checked `vim.lsp.with`, when called, the border is "none".

### Check the following

- [X] I am in a git managed directory
- [X] I restarted Neovim after running :LspInstall
- [X] I have a valid language configuration (~/.config/lvim/ftplugin/)
- [X] I checked the options (~/.config/lvim/config.lua)
- [X] The programs specified by the config are installed

### LunarVim version

rolling

### Neovim version (>= 0.5)

NVIM v0.7.0-dev+746-g95803f0e90

### Relevant log output

_No response_

### Screenshots

_No response_

### I have

- [X] Read the readme
- [X] Read the wiki

kylo252:
# Description

- unify the style between the handlers and diagnostics float-opts
- deprecate `lvim.lsp.popup_border` in favor of `lvim.lsp.float`
to allow further customizations

Fixes #2095

## How Has This Been Tested?

Invoking `vim.lsp.buf.hover()` (you can use `<K>`) should show borders now by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,7 @@ lvim.plugins = {
 ## Breaking changes
 
 - `lvim.lang.FOO` is no longer supported. Refer to <https://www.lunarvim.org/languages> for up-to-date instructions.
+- `lvim.lsp.popup_border` has been deprecated in favor of `lvim.lsp.float.border` and `lvim.lsp.diagnostics.float.border`.
 
 ## Resources
 
--- a/lua/lvim/config/init.lua
+++ b/lua/lvim/config/init.lua
@@ -66,6 +66,11 @@ local function handle_deprecated_settings()
       deprecation_notice(string.format("lvim.lang.%s", lang))
     end
   end
+
+  -- lvim.lsp.popup_border
+  if vim.tbl_contains(vim.tbl_keys(lvim.lsp), "popup_border") then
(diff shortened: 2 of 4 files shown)


kylo252:
@abzcoding, I think I changed my mind at least 4 times about the option name, so hopefully this one is fine!

abzcoding:
can we merge this and https://github.com/LunarVim/LunarVim/pull/2093 into one PR 🥺

abzcoding:
either github is bugging or we are overriding `lvim.lsp.float`
Screen Shot 2021-12-23 at 12 09 23 PM

kylo252:
@abzcoding, I thought I could push this, and then we can rebase #2093 on it. I can fix both if you want, but if you think it's better to do it in a single PR, then we can do that instead.

abzcoding:
> @abzcoding, I thought I could push this, and then we can rebase #2093 on it. I can fix both if you want, but if you think it's better to do it in a single PR, then we can do that instead.

yeah lets do the rebase thingy

PR #2119, opened 2021-12-27
refactor: avoid running ts.setup in headless

kylo252:
# Description

Avoid running the setup for treesitter in headless mode since it's harder to debug.


abzcoding (approved):
👍

PR #2125, opened 2021-12-31
refactor(info): use new null-ls api for sources

kylo252:
# Description

- Fix missing entries for supported sources from `LvimInfo`
- Remove supported action sources since it's not unique per filetype

## How Has This Been Tested?

`LvimInfo` should now show the correct list of supported sources


abzcoding (approved):
LGTM 🚀

abzcoding:
Btw can we do a new release? 
0.6.1 is a bit outdated

kylo252:
> Btw can we do a new release? 0.6.1 is a bit outdated

I'll fix the version validator, and then we can finally have 1.0 😃

PR #2127, opened 2022-01-01
refactor(bootstrap): more robust git module

kylo252:
# Description

- refactor: add validate_nvim_version at the beginning of bootstrap
- refactor: more robust git module

### DIFF
--- a/lua/lvim/utils/git.lua
+++ b/lua/lvim/utils/git.lua
@@ -1,0 +1,158 @@
+local M = {}
+
+local Log = require "lvim.core.log"
+
+local function git_cmd(opts)
+  local plenary_loaded, Job = pcall(require, "plenary.job")
+  if not plenary_loaded then
+    vim.cmd "packadd plenary.nvim"
+  end
+
+  opts = opts or {}
+  opts.cwd = opts.cwd or get_lvim_base_dir()
+
+  local stderr = {}
+  local stdout, ret = Job
+    :new({
+      command = "git",
+      args = opts.args,
+      cwd = opts.cwd,
+      on_stderr = function(_, data)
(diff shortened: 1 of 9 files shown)


chase:
@kylo252 Are you pushing to the right branch?

Thought we were dropping the compat branch idea and things and just telling the user to upgrade.

kylo252:
> Thought we were dropping the compat branch idea and things and just telling the user to upgrade.

I'll keep the new cleanups and remove any hopeless attempts to salvage the situation. Let's just go with bailing out early instead.

abzcoding (approved):
good work on this 👍

lua/lvim/utils/git.lua
```diff
@@ -0,0 +1,158 @@
...
+function M.get_lvim_version(type)
+  type = type or ""
+  local branch = M.get_lvim_branch()
+  if branch == "master" then
+    return M.get_lvim_tag(type)
+  end
+  local ret, log_results = git_cmd { args = { "log", "--pretty=format:%h", "-1" } }
+  local abbrev_version = log_results[1] or ""
+  if ret ~= 0 or string.match(abbrev_version, "%d") == nil then
+    Log:error "Unable to retrieve current version. Check the log for further information"
```
ryaminal, line 131: I see that this has already been merged, but I think this line is causing some annoying behavior when using the install.sh script for lunarvim. This is because the install script, I believe, does a shallow checkout of `master` and there are no tags/refs that match the pattern from line [130](https://github.com/LunarVim/LunarVim/pull/2127/files#diff-592059d4c0b81bb5e3509fc175b0d838215ff52d6cfd5015ab13181b29f569deR130) above e.g. `"%d"`.

PR #2133, opened 2022-01-02
refactor: more consistent autocmds

kylo252:
# Description

Fixes the issue where `highlight_document` is still active when the server has already been stopped, when using `LspStop` or `LvimReload` for example.

## How Has This Been Tested?

You should no longer see any warnings after about `document_highlight` after calling `LspStop`.


abzcoding (approved):
👍

PR #2134, opened 2022-01-03
refactor(settings): add headless-mode settings

kylo252:
# Description

- add headless-mode settings that can help prevent truncating error messages
- move the colorscheme autocmds into their correct module


abzcoding (approved):
👍

PR #2135, opened 2022-01-03
fix(lsp): formatter now use new null-ls api function

cpea2506:
# Description

- The formatter in the `select_default_formater` function is still using its old function, that make `format_on_save` doesn't choose `null-ls` as default formatter
- Change the name of that function to new null-ls api function to make it work correctly


abzcoding:
tyvm <3

PR #2140, opened 2022-01-04
[Bugfix] Git lua annoying error tag

ryaminal:
# Description
Removing a `Log:error` line that was added but is, imo, more problematic than helpful.
Having this message causes a message to be displayed as well as a notification, if enabled. This is at best annoying and at worst could cause someone(me, for instance. :) ) to spend some time figuring out why there is a problem.

## How Has This Been Tested?
1. Uninstall lunarvim with the uninstall script.
2. Install lunarvim with the install script.
3. Open lunarvim/lvim. Before this fix I was seeing a notify and a message: "Unable to retrieve current tag. Check the log for further information"


ryaminal:
Bleh. I've mucked up my fork. Closing this while I try and fix this PR.

PR #2141, opened 2022-01-05
fix: remove "error" message from git tag

ryaminal:
# Description
Removing a `Log:error` line that was added but is, imo, more problematic than helpful.

I think the root cause is that the install script for lunarvim uses `--depth 1` when cloning which means the only ref/tag is `master` by default. So, the check for if the first tag is a number `%d` will result in the error message always being printed, causing a notification and message that is annoying/misleading and could potentially cause someone to waste time figuring the "problem". 

## How Has This Been Tested?
1. Uninstall lunarvim with the uninstall script.
2. Install lunarvim with the install script.
3. Open lunarvim/lvim. Before this fix I was seeing a notify and a message: "Unable to retrieve current tag. Check the log for further information"


kylo252:
Thank you!

PR #2154, opened 2022-01-08
fix(null-ls): avoid sending invalid opts.args

kylo252:
# Description

- fix(null-ls): avoid sending invalid opts.args
- fix(null-ls): query registered sources correctly
- chore: bump plugins version

## How Has This Been Tested?

This still works
```lua
local formatters = require "lvim.lsp.null-ls.formatters"
formatters.setup { { exe = "shfmt", args = { "-i", "2", "-ci", "-bn" } } }
```

**note** it's recommended to use `command`, `extra_args` instead to ensure consistency


abzcoding (approved):
That seems good to me 👍

abzcoding:
btw @kylo252 can we add this as well?
```
--- a/lua/lvim/interface/popup.lua
+++ b/lua/lvim/interface/popup.lua
@@ -48,7 +48,7 @@ function Popup:display(content_provider)
   )

   local lines = content_provider(self.layout)
-  vim.api.nvim_buf_set_lines(self.bufnr, 0, -1, false, lines)
+  vim.api.nvim_buf_set_lines(self.bufnr or 0, 0, -1, false, lines)
```

kylo252:
> btw @kylo252 can we add this as well?
> to be compatible with the neovim head

I'll add it in the next PR which is more appropriate!

PR #2156, opened 2022-01-08
refactor(term): leave the first few ids unassigned

### LINKED ISSUE #2146: Cannot open multiple terminals
https://github.com/LunarVim/LunarVim/blob/8233522e0b9b026c5ece9e420c4d9642f5a0c889/lua/lvim/core/terminal.lua#L53-L64

The above lines break the ability to open multiple terminals using the keymap (e.g. `1<C-t>`, `2<C-t>`, `3<C-t>`, etc). Using `:1ToggleTerm`, `:2ToggleTerm`  command works, but hiding the terminal with `<C-t>` not. Commenting out those lines fixes everything for me. Why do we need those lines?

kylo252:
# Description

This is due to inconsistencies of `:2ToggleTerm`

Fixes #2146

---

Thank you @00sapo for looking into this!

### DIFF
--- a/lua/lvim/core/terminal.lua
+++ b/lua/lvim/core/terminal.lua
@@ -50,25 +50,13 @@ M.setup = function()
   local terminal = require "toggleterm"
   terminal.setup(lvim.builtin.terminal)
 
-  -- setup the default terminal so it's always reachable
-  local default_term_opts = {
-    cmd = lvim.builtin.terminal.shell,
-    keymap = lvim.builtin.terminal.open_mapping,
-    label = "Toggle terminal",
-    count = 1,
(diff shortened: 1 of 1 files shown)


kylo252:
@00sapo, was this change enough?

00sapo:
At now, in my config I have also removed lines 53-64 since they are not needed anymore. https://github.com/LunarVim/LunarVim/blob/0e9ea2f7b3031046647c94897676dcc02ed0c0b2/lua/lvim/core/terminal.lua#L53-L64

I would also add a note to the documentation website about this

00sapo:
Thank you so much!

PR #2157, opened 2022-01-09
fix(install): avoid data-races for `on_packer_complete`

kylo252:
# Description

Fixes an issue where the hook is deleting `packer-compiled.lua` during installation.

### DIFF
--- a/lua/lvim/lsp/templates.lua
+++ b/lua/lvim/lsp/templates.lua
@@ -43,7 +43,7 @@ end
 ---@param servers_names table list of servers to be enabled. Will add all by default
 function M.generate_templates(servers_names)
   if package.loaded["lspconfig"] then
-    vim.cmd [[ LspStop ]]
+    pcall(vim.cmd, "LspStop")
   end
 
   servers_names = servers_names or {}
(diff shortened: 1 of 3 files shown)


lua/lvim/lsp/templates.lua
```diff
@@ -42,6 +42,10 @@ end
 ---The files are generated to a runtimepath: "$LUNARVIM_RUNTIME_DIR/site/after/ftplugin/template.lua"
 ---@param servers_names table list of servers to be enabled. Will add all by default
 function M.generate_templates(servers_names)
+  if package.loaded["lspconfig"] then
+    pcall(vim.cmd, "LspStop")
```
chase, line 46: I'm not sure if pcall with a vim.cmd will work or not.

If you ever want to look under the hood and see what the lua is for a command you can just use `:command LspStop` etc. to list what it is.
For example LspStop is: `require'lspconfig'["_root"].commands["LspStop"][1](<f-args>)` where `<f-args>` are the arguments passed to LspStop in the vim command. For this case, there are none so: `require'lspconfig'["_root"].commands["LspStop"][1]()`

chase (approved):

lua/lvim/lsp/templates.lua
kylo252: In this case I know what `LspStop` does, it's supposed to stop all active clients, which is actually easier to write directly compared to `LspRestart`.

Since I do recall seeing it, I did a quick search and out of all the plugins I use, these are all occurrences

image

Maybe it doesn't actually work 😅

chase: Ah, to explain, I meant using pcall for handling an error from the vimscript runtime might not work. I'm not certain about this.

kylo252: Yeah, I got what you meant. I've switched it to use the lua API directly instead.

PR #2171, opened 2022-01-13
Revert "fix(which-key): temporary solution for which-key (#2150)"

hulufei:
This reverts commit 6740afd743a05028cc48e8f1e203e7f81345aced.

# Description

Since this fix folke/which-key.nvim#227 merged, and the patch branch have been deleted, the temporary solution should revert to prevent lvim broken. @meijieru

### DIFF
--- a/lua/lvim/plugins.lua
+++ b/lua/lvim/plugins.lua
@@ -177,10 +177,8 @@ return {
 
   -- Whichkey
   {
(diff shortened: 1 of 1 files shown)


kylo252:
Sorry that I didn't notice that this was posted before #2172, thanks however!

PR #2180, opened 2022-01-15
fix(diag): show lsp-diag code in open_float

kylo252:
# Description

Show the code of lsp diagnostics when available. 

This was working fine already for null-ls, now it should work fine as well for lsp servers.

## How Has This Been Tested?

**old**
image

**new**
image


abzcoding (approved):
👍

PR #2182, opened 2022-01-15
refactor(nvim-tree): cleanup and update settings

kylo252:
# Description

- override nvim_tree.view.close() directly instead of
defining an autocmd
- convert the callbacks to local since they're not used globally
- disable git ignore settings by default
- add new available settings


abzcoding (approved):
👍

PR #2184, opened 2022-01-15
refactor: remove unused outdated files

kylo252:
# Description

Remove some outdated files that are currently either deprecated or unused.


kylo252:
@abzcoding, wdyt?

abzcoding (approved):
I am not using the vscode integration, but aren’t those necessary?

kylo252:
> I am not using the vscode integration, but aren’t those necessary?

We're not using lush, and I'm pretty sure that half the vscode settings aren't even working anymore 🤔 
I guess there's always the `freeze` branch to fallback to.

orsenkucher:
@kylo252 What is the way to integrate with vscode now?
(and `VSCode Neovim` extension)

kylo252:
@orsenkucher, if you read through this section: https://github.com/vscode-neovim/vscode-neovim#performance-problems they basically recommend to load only a few select plugins that offer functionality _not available_ in vscode. 

I don't really see how lunarvim can help in that regard, since the files I removed had practically nothing in common with the rest of the codebase. Combine that with the fact that they weren't maintained and you hopefully you can see where I'm coming from.

PR #2216, opened 2022-01-26
fix(lsp): avoid accessing undefined user_data

svirschevskiy:
# 

Do not throw exception in case of incorrect eslint configuration

## How Has This Been Tested?
1. Open a project with eslint
1. Add a syntax error to `.eslintrc`
1. Open any js file in the project and hover a first line

**Before change**
Lunarvim throws the following error:
```
E5108: Error executing /.local/share/lunarvim/lvim/lua/lvim/lsp/config.lua:26: attempt to index field 'user_data' (a nil value)
```

**After change**
There is a float window with a correct message
```
eslint_d: SyntaxError: Cannot read config file: .eslintrc.json
Error: Unexpected token { in JSON at position 1
```

### DIFF
--- a/lua/lvim/lsp/config.lua
+++ b/lua/lvim/lsp/config.lua
@@ -23,7 +23,7 @@ return {
       prefix = "",
       format = function(d)
         local t = vim.deepcopy(d)
-        local code = d.code or d.user_data.lsp.code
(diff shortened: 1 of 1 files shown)


abzcoding (approved):
tyvm <3

PR #2239, opened 2022-02-02
feat: use schemastore.nvim to provide extended json schema support

### LINKED ISSUE #1957: [Feature]: (Minor) use SchemaStore.nvim for jsonls
### Feature motivation

I think one of the features I enjoy in IDE such as vs-code is the whole compatibility with schemas file (package.json, angular.config.json, etc...).

As of right now, I am seeing that LunarVim does not support all the those

My solution to this problem is using the [SchemaStore.nvim](https://github.com/b0o/schemastore.nvim) plugin which let jsonls access the whole [schemastore](http://schemastore.org) catalog. 

I don't think there's other applicable solutions except taking all the schemas by hand.

### Additional context

Nothing to note.

abzcoding:
# Description

Use [SchemaStore.nvim](https://github.com/b0o/SchemaStore.nvim) for extended schema support

Fixes #1957


lua/lvim/lsp/providers/jsonls.lua
```diff
@@ -1,186 +1,12 @@
...
-  end
-  return tab1
-end
-
-local extended_schemas = extend(schemas, default_schemas)
-
+local full_schemas = vim.tbl_deep_extend(
+  "force",
+  require("schemastore").json.schemas(),
+  require("nlspsettings.jsonls").get_default_schemas()
```
kylo252, line 4: We don't need `pcall` anymore?

lua/lvim/plugins.lua
```diff
@@ -272,4 +273,10 @@ return {
     end,
     disable = not lvim.builtin.terminal.active,
   },
+
+  -- SchemaStore
+  {
+    "b0o/schemastore.nvim",
+    commit = commit.schemastore,
```
kylo252, line 280: should this be loaded on `ft = 'json'`? or maybe it's not that heavy anyway?

abzcoding: sadly lazy loading causes a lot of weird issues on this one

lua/lvim/lsp/providers/jsonls.lua
abzcoding: we can use it 🤔 but if it fails, we better learn about it as verbose as possible 😅 , so i think this should be fine as long as we are not lazy loading anything about it

abzcoding: do you prefer sth like this?
```lua
local has_schemastore, schemastore = pcall(require, "schemastore")
if not has_schemastore then
  return {}
end

local has_nlspsettings, nlspsettings = pcall(require, "nlspsettings.jsonls")
if not has_nlspsettings then
  return {}
end

local full_schemas = vim.tbl_deep_extend("force", schemastore.json.schemas(), nlspsettings.get_default_schemas())
local opts = {
  settings = {
    json = {
      schemas = full_schemas,
    },
  },
  setup = {
    commands = {
      Format = {
        function()
          vim.lsp.buf.range_formatting({}, { 0, 0 }, { vim.fn.line "$", 0 })
        end,
      },
    },
  },
}

return opts
```

kylo252 (approved):

PR #2243, opened 2022-02-03
refactor(lsp): cleanup servers' override configuration

kylo252:
# Description

- rename `lvim.lsp.override` to ~~`lvim.lsp.automatic_configuration.ignored_servers`~~ `lvim.lsp.automatic_configuration.skipped_servers`
- add ~~`lvim.lsp.automatic_configuration.ignored_filetypes`~~ `lvim.lsp.automatic_configuration.skipped_filetypes`
- refactor(lsp): allow installing overridden servers during setup

Fixes issue with keeping up with new markdown servers that keeps getting added..

---

**TODOs**

- [ ] decide on variable names
- [ ] update examples in `config.example.lua`
- [x] update info-panel
- [x] handle deprecation
  - [x] add deprecation message
- [ ] add "dynamic" installation alternative (probably needs a separate PR)
- [ ] do not attempt to install deprecated servers, see #2402


lua/lvim/config/init.lua
```diff
@@ -67,6 +68,11 @@ local function handle_deprecated_settings()
     end
   end
 
+  -- lvim.lsp.override
+  if vim.tbl_contains(vim.tbl_keys(lvim.lsp), "override") then
```
abzcoding, line 72: because we have `override = {}`  down there, this warning always shows up no matter what,
shouldn't we check if exists and the length is bigger than one or sth?

kylo252: Oops 😅

academo:
This renaming is a great initiative. When I just started using lunarvim (not that long ago), the whole "overwrites" was confusing because in fat, it was not overwriting anything, it was disabling.

I also like the idea of putting this inside "automatic_configuration". In general in natural language it should sound like "ignored servers by the automatic configuration".

My only concern in the change, and this is of course something that needs to be changed in the website documentation, is how you document it.

A new user to the editor could easily think that because you are "ignoring" a server, it means it will never be used or installed when in fact is merely ignored for automatic configuration.

abzcoding:
```
"require(\"lvim.lsp.manager\").setup(\"crystalline\"), require(\"lvim.lsp.manager\").setup(\"scry\")" 
[38](https://github.com/LunarVim/LunarVim/runs/6005277771?check_suite_focus=true#step:5:38)
```

just fix this and we are good to go 👍

abzcoding:
also, can you fix the windows example config?
see this -> https://github.com/LunarVim/LunarVim/pull/2452

kylo252:
> just fix this and we are good to go 👍

I must've messed something up in one of the rebases, since I need to add all those servers manually again 🥲 

also, wdyt about the variable names?

abzcoding:
> > just fix this and we are good to go 👍
> 
> I must've messed something up in one of the rebases, since I need to add all those servers manually again 🥲
> 
> also, wdyt about the variable names?

variable name looks fine to me 👍

abzcoding (approved):
LGTM 🚀

PR #2246, opened 2022-02-04
fix(lualine): use 1-char width symbol for changed

meijieru:
The 2-char width characters cause problems in windows terminal.
Related to #1897


abzcoding:
can you fix the conflict, and we can merge it after

abzcoding (approved):
tyvm👍

PR #2248, opened 2022-02-04
chore: lazy load dap_install

meijieru:
As title.


abzcoding:
since there are no other modules in that plugin and the fact that we are caching it anyways, this doesn't have any impacts

thank you for your effort, but sadly I think I would reject this PR

cc: @kylo252

kylo252:
To add to that, it's confusing for some users when `:DIInstall` doesn't have autocomplete on the first call.

PR #2249, opened 2022-02-04
chore(lualine): no bold for lsp component

meijieru:
image
The bold is inconsistent here.


abzcoding:
it was supposed to be Bold to make sure it is different from the other components

tyvm for your effort, but i think we don't want to merge this PR, i'd suggest creating your own style according to [our documentation](https://www.lunarvim.org/configuration/06-statusline.html#style)

PR #2265, opened 2022-02-11
[Bugfix]: keep splits when closing buffer

bob3000:
# Description

After the switch to `akinsho/bufferline.nvim` when a buffer was closed with `<leader>c` the split window
displaying the buffer was also closed which differs from the former behaviour.

This PR installs `nvim-bufdel` and changes to key binding accordingly to return to the former behaviour
where the window layout is maintained unaltered after a buffer was closed.

Fixes #(issue)

## How Has This Been Tested?

The fix was tested manually. To test open two buffers in a split window layout and close one of them by pressing `<leader>c`.


kylo252:
I have my own implementation, because I found the one from nvim-bufdel to be kinda confusing because it's cycling to the _next_ buffer instead of previous one. It's also barely using any API functions. So, I ended up re-writing my own "minimal" version. 

I had considered adding it to the bufferline PR, but I wasn't sure there's enough demand for it, but I've pulled it from my own config into a new draft, please take a look https://github.com/LunarVim/LunarVim/pull/2267

bob3000:
Feel free to close the PR if there is a better solution already on the way. I just needed a quick fix and I thought I might share it here.

PR #2267, opened 2022-02-11
feat: add a minimal implementation of bbye

kylo252:
# Description

Add a minimal implementation for bbye to use with bufferline

Related #2265


kylo252:
@abzcoding, we should probably include this as well in the release.

abzcoding (approved):
LGTM 🚀

PR #2269, opened 2022-02-12
chore(plugins): bump version

abzcoding:
# Description

Bump plugin version


kylo252:
should we merge into master and do a release before this?

abzcoding:
> should we merge into master and do a release before this?

agreed 👍

kylo252:
> agreed 👍

What version number should this be? I feel like semver doesn't have to be linear, is this 1.2.0 or 1.1.0 or something else?

abzcoding:
> > agreed 👍
> 
> What version number should this be? I feel like semver doesn't have to be linear, is this 1.2.0 or 1.1.0 or something else?

i think this should be 1.1.0

kylo252:
> i think this should be 1.1.0

I've pushed the changes and updated the changelog, you should do the release this time tho :)

PR #2287, opened 2022-02-16
fix(timeoutlen): this has caused way too many issues in the past

abzcoding:
# Description

People usually are not that fast, this has caused a lot of issues, the most recent one being https://github.com/tpope/vim-fugitive/issues/1951

also make sure that lsp floating window is focusable to fix issues like https://www.reddit.com/r/lunarvim/comments/srwss8/unable_to_scroll_inside_the_floating_popup/


codingluke:
@abzcoding Thank you for beeing so pro-active 💯

kylo252 (approved):
LGTM! 🚀

PR #2070, opened 2021-12-12
fix(lsp): allow overriding servers with custom providers

kylo252:
# Description

- fix(lsp): correct client_id parsing in lvim-info
- fix(lsp): allow overriding custom providers

## How Has This Been Tested?

- Can open `LvimInfo`
- Can override `sumneko_lua` with "lua-dev"
- Can disable `vuels` and actually remove the generated `vue.lua`
```lua
vim.list_extend(lvim.lsp.override, {"vuels" })
```

PR #2295, opened 2022-02-19
[Feature]: add lvim_provider to opt in provider

### LINKED ISSUE #2290: Allow to override LS settings from "lsp/providers"
### Feature Description

I want to get rid of the settings from `lua/lvim/lsp/providers/sumneko_lua.lua`.

It seems no possible to do so in `config.lua`, because the settings from provider always merged in config:

https://github.com/LunarVim/LunarVim/blob/0536775c8ba06fbc3044a2dc35262d429d4dce4b/lua/lvim/lsp/manager.lua#L31-L35

To  make it clear, what I want is not overriding but removal of the settings from provider

_No response_

### Support information

_No response_

younger-1:
# Description

Fixes #2290. Allow users to opt in LS settings from "lsp/providers":

Only when `user_config.lvim_provider` is set to false, providers' settings will not be used.

### DIFF
--- a/lua/lvim/core/which-key.lua
+++ b/lua/lvim/core/which-key.lua
@@ -74,7 +74,7 @@ M.config = function()
         name = "Buffers",
         j = { "<cmd>BufferLinePick<cr>", "Jump" },
         f = { "<cmd>Telescope buffers<cr>", "Find" },
-        b = { "<cmd>BufferLineCyclePrev<cr>", "Previous" },
+        b = { "<cmd>b#<cr>", "Previous" },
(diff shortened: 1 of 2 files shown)


younger-1:
@kylo252 @abzcoding Do you prefer this kind of change or something not bother with `user_config`? Is it ok to use `lvim_provider` as switching field name?

kylo252:
This is actually not needed, since the user path will always take precedence `~/.config/lvim/lua/lvim/lsp/providers/*.lua`, see the discussion earlier in https://github.com/LunarVim/LunarVim/pull/2070.

Full example in https://github.com/LunarVim/LunarVim/issues/2290#issuecomment-1046209783

PR #2300, opened 2022-02-20
[Feature]: Add possibility to override plugin loading event

epontan:
# Description

It can be useful be able to override which event a core plugin should be loaded from the user configuration.

Add an `event` attribute to the default configuration of all core plugins that have an event configured and refer to that from the plugins module.

This implements the approach suggested in #1692.

## How Has This Been Tested?

My use case was to make `gitsigns.nvim` load on startup when not starting vim with a file as argument (not only when `BufRead` event is triggered).
I use the [gitsigns](https://github.com/lewis6991/gitsigns.nvim) plugin to review all the changes of last commit by changing its base (e.g. `HEAD~1`).

In my `config.lua` I override the default configuration like this:

```lua
lvim.builtin.gitsigns.event = "VimEnter"
```

This makes the plugin available at start without having to open a file first.

### DIFF
--- a/lua/lvim/core/bufferline.lua
+++ b/lua/lvim/core/bufferline.lua
@@ -36,6 +36,7 @@ end
 M.config = function()
   lvim.builtin.bufferline = {
     active = true,
+    event = "BufWinEnter",
     on_config_done = nil,
     keymap = {
       normal_mode = {},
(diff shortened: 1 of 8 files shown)


kylo252:
I've been thinking about this for a while for the `commit` part which has now become relevant, so we still need to handle the rest of the supported attributes. I'm not sure what is a clean way to do this, so maybe we should instead resurrect #1692, or something similar?

PR #2306, opened 2022-02-23
feat: add option to automatically answer 'yes' for sh install script

khaosdoctor:
# Description

Add possibility for the user to auto install dependencies based on a flag that can be either `-y, --install-dependencies` which will not prompt the user for dependency installing, or `-n, -no-install-dependencies` which won't install deps altogether.

The default behavior of prompting is maintained.

## Why

Because while fiddling around with my dotfiles and automatic setup, I'd like to be able to run the script without any sort of interaction (heavily inspired on `npm init -y`

## How Has This Been Tested?

Ran the test locally in docker containers using the same installation configuration but with the different file:

- `bash <(curl -shttps://raw.githubusercontent.com/khaosdoctor/LunarVim/feature/auto-install-deps-sh/utils/installer/install.sh) -y`
- `bash <(curl -shttps://raw.githubusercontent.com/khaosdoctor/LunarVim/feature/auto-install-deps-sh/utils/installer/install.sh) -n`
- `bash <(curl -shttps://raw.githubusercontent.com/khaosdoctor/LunarVim/feature/auto-install-deps-sh/utils/installer/install.sh) --install-dependencies`
- `bash <(curl -shttps://raw.githubusercontent.com/khaosdoctor/LunarVim/feature/auto-install-deps-sh/utils/installer/install.sh) -no-install-dependencies`

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,14 +28,18 @@ You can find all the documentation for LunarVim at [lunarvim.org](https://www.lu
 
 Make sure you have the release version of Neovim (0.6.1+).
 
-Linux:
+### Linux:
+
 ```bash
 bash <(curl -s https://raw.githubusercontent.com/lunarvim/lunarvim/master/utils/installer/install.sh)
 ```
 
-> If you want to run it without any interaction you can pass the `-y` or `--install-dependencies` flag to automatically install all dependencies and have no prompts. The same way, you can use `-n` or `--no-install-dependencies` to skip the dependency installation.
+To run the install script without any interaction you can pass the `-y` flag to automatically install all dependencies and have no prompts. This is particularly useful in automated installations.
 
-Windows (Powershell):
+The same way, you can use `--no-install-dependencies` to skip the dependency installation.
+
+### Windows (Powershell):
+
 ```powershell
 Invoke-WebRequest https://raw.githubusercontent.com/LunarVim/LunarVim/master/utils/installer/install.ps1 -UseBasicParsing | Invoke-Expression
 ```
--- a/README.md
+++ b/README.md
@@ -33,6 +33,8 @@ Linux:
 bash <(curl -s https://raw.githubusercontent.com/lunarvim/lunarvim/master/utils/installer/install.sh)
 ```
 
+> If you want to run it without any interaction you can pass the `-y` or `--install-dependencies` flag to automatically install all dependencies and have no prompts. The same way, you can use `-n` or `--no-install-dependencies` to skip the dependency installation.
+
 Windows (Powershell):
 ```powershell
 Invoke-WebRequest https://raw.githubusercontent.com/LunarVim/LunarVim/master/utils/installer/install.ps1 -UseBasicParsing | Invoke-Expression
@@ -159,7 +161,7 @@ lvim.plugins = {
 > - @mvllow, Potential LunarVim user.
 
 <div align="center" id="madewithlua">
-	
+
 [![Lua](https://img.shields.io/badge/Made%20with%20Lua-blue.svg?style=for-the-badge&logo=lua)](#madewithlua)
-	
+
 </div>
--- a/utils/installer/install.sh
+++ b/utils/installer/install.sh
@@ -24,7 +24,8 @@ readonly BASEDIR
 
 declare ARGS_LOCAL=0
 declare ARGS_OVERWRITE=0
-declare ARGS_INSTALL_DEPENDENCIES=2
+declare ARGS_INSTALL_DEPENDENCIES=1
+declare INTERACTIVE_MODE=1
 
 declare -a __lvim_dirs=(
   "$LUNARVIM_CONFIG_DIR"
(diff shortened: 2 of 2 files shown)


kylo252 (commented):
Wouldn't it be better to include a flag to answer "yes" to all interactive prompts instead? I feel like that would be easier to document and explain.

khaosdoctor:
> Wouldn't it be better to include a flag to answer "yes" to all interactive prompts instead? I feel like that would be easier to document and explain.

Can you please elaborate a bit more on that? Because I think the end result, in terms of code, would be much the same.

kylo252:
> Can you please elaborate a bit more on that? Because I think the end result, in terms of code, would be much the same.

I just think it's a common pattern to disable _all_ prompts when you pass the `-y` flag, so it's not just the dependencies. Currently, it only applies to these dependencies, but might as well use a standardized method to deal with it.

image

utils/installer/install.sh
```diff
@@ -60,10 +60,10 @@ function parse_arguments() {
       --overwrite)
         ARGS_OVERWRITE=1
         ;;
-      --install-dependencies)
+      -y | --install-dependencies)
```
kylo252, line 63:
```suggestion
      -y )
        INTERACTIVE_MODE=0
        ;;
      --install-dependencies)
```

utils/installer/install.sh
```diff
@@ -60,10 +60,10 @@ function parse_arguments() {
...
         ARGS_INSTALL_DEPENDENCIES=1
         ;;
-      --no-install-dependencies)
+      -n | --no-install-dependencies)
```
kylo252, line 66:
```suggestion
      --no-install-dependencies)
```

utils/installer/install.sh
```diff
@@ -24,7 +24,7 @@ readonly BASEDIR
 
 declare ARGS_LOCAL=0
 declare ARGS_OVERWRITE=0
-declare ARGS_INSTALL_DEPENDENCIES=1
+declare ARGS_INSTALL_DEPENDENCIES=2
```
kylo252, line 27:
```suggestion
declare ARGS_INSTALL_DEPENDENCIES=1
declare INTERACTIVE_MODE=1
```

utils/installer/install.sh
```diff
@@ -45,10 +45,10 @@ function usage() {
...
   echo ""
   echo "Options:"
-  echo "    -h, --help                       Print this help message"
-  echo "    -l, --local                      Install local copy of LunarVim"
-  echo "    --overwrite                      Overwrite previous LunarVim configuration (a backup is always performed first)"
-  echo "    --[no]-install-dependencies      Whether to prompt to install external dependencies (will prompt by default)"
+  echo "    -h, --help                               Print this help message"
+  echo "    -l, --local                              Install local copy of LunarVim"
+  echo "    --overwrite                              Overwrite previous LunarVim configuration (a backup is always performed first)"
+  echo "    -y, -n, --[no]-install-dependencies      Whether to automatically install external dependencies (will prompt by default)"
```
kylo252, line 51:
```suggestion
  echo "    -h, --help                       Print this help message"
  echo "    -l, --local                      Install local copy of LunarVim"
  echo "    -y                               Disble confirmation prompt (answer yes to all questions)"
  echo "    --overwrite                      Overwrite previous LunarVim configuration (a backup is always performed first)"
  echo "    --[no]-install-dependencies      Whether to install external dependencies (will prompt by default)"

```

README.md
```diff
@@ -33,6 +33,8 @@ Linux:
 bash <(curl -s https://raw.githubusercontent.com/lunarvim/lunarvim/master/utils/installer/install.sh)
 ```
 
+> If you want to run it without any interaction you can pass the `-y` or `--install-dependencies` flag to automatically install all dependencies and have no prompts. The same way, you can use `-n` or `--no-install-dependencies` to skip the dependency installation.
```
kylo252, line 36:
```suggestion
> If you want to run it without any interaction you can pass the `-y` flag to automatically install all dependencies and have no prompts. The same way, you can use `--no-install-dependencies` to skip the dependency installation.
```

utils/installer/install.sh
```diff
@@ -92,7 +92,7 @@ function main() {
 
   check_system_deps
 
-  if [ "$ARGS_INSTALL_DEPENDENCIES" -eq 1 ]; then
+  if [ "$ARGS_INSTALL_DEPENDENCIES" -eq 2 ]; then
```
kylo252, line 95:
```suggestion
  if [ "$ARGS_INSTALL_DEPENDENCIES" -eq 1 ]; then
  	if [ "$INTERACTIVE_MODE" -eq 1 ]; then
```

utils/installer/install.sh
```diff
@@ -104,6 +104,10 @@ function main() {
     msg "Would you like to install LunarVim's Rust dependencies?"
     read -p "[y]es or [n]o (default: no) : " -r answer
     [ "$answer" != "${answer#[Yy]}" ] && install_rust_deps
+  elif [ "$ARGS_INSTALL_DEPENDENCIES" -eq 1 ]; then
+    install_nodejs_deps
+    install_python_deps
+    install_rust_deps
```
kylo252, line 110:
```suggestion
  else
    install_nodejs_deps
    install_python_deps
    install_rust_deps
   fi
```

kylo252 (commented):
Here's a suggestion of how that might look. Sorry in advance about the weird formatting (it's hard when I can't suggest over non-edits)

khaosdoctor:
Oh, ok I see! Thanks for the suggestion, I think it looks better with it too, let me make the changes over here and push it again. I'll probably need to rebase it as well since the commit lint complained about the message size

kylo252 (approved):
LGTM! 🚀

abzcoding (approved):
LGTM 🚀

PR #2315, opened 2022-02-26
chore(plugins): bump version

abzcoding:
# Description

Update plugins


kylo252 (dismissed):
LGTM! 🚀

lua/lvim/core/telescope.lua
```diff
@@ -25,7 +25,16 @@ function M.config()
       layout_config = {
         width = 0.75,
         preview_cutoff = 120,
-        horizontal = { mirror = false },
+        prompt_position = "bottom",
```
kylo252, line 28: What was the reason behind this? It breaks other layouts, e.g. `<leader>Lf`

```suggestion
```

abzcoding: shit this was not supposed to be in here 😅

lua/lvim/core/telescope/custom-finders.lua
```diff
@@ -14,6 +14,7 @@ local utils = require "lvim.utils"
 function M.find_lunarvim_files(opts)
   opts = opts or {}
```
kylo252, line 15: The problem is that we're merging over the theme.

abzcoding: ah i see, i wanted to make it like this
Screen Shot 2022-02-27 at 3 39 52 PM
Screen Shot 2022-02-27 at 3 40 04 PM

i'll revert the changes and do a separate one for custom styles

kylo252: Yeah, creating a custom style/theme should do it. It's tricky when modifying the global default.

kylo252 (approved):
LGTM! 🚀

PR #2322, opened 2022-03-03
feat: enable nlsp-settings schemas

youdaniel:
# Description

Allows for nicer autocompletion for `:NlspConfig` by default.

## How Has This Been Tested?

`:NlspConfig <language server>`

### DIFF
--- a/lua/lvim/lsp/init.lua
+++ b/lua/lvim/lsp/init.lua
@@ -144,7 +144,10 @@ function M.setup()
     require("lvim.lsp.templates").generate_templates()
   end
 
-  bootstrap_nlsp { config_home = utils.join_paths(get_config_dir(), "lsp-settings") }
+  bootstrap_nlsp {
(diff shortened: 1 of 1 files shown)


abzcoding:
this should already be working?
Screen Shot 2022-03-04 at 3 26 19 PM



because we have this
https://github.com/LunarVim/LunarVim/blob/5fd8b4b72642ab94c33bb6e706617a554c0d2f19/lua/lvim/lsp/providers/jsonls.lua#L1-L20


cc: @kylo252

kylo252:
I'm not quite sure that I understand the implementation, because it feels like it should have be on by default. However, it seems that the autocomplete inside `~/.config/lvim/lsp-settings/pyright.json` is now broken without enabling this flag.

@abzcoding, can you please verify that part?

abzcoding:
w8 you are right, maybe it is the new update from them
@youdaniel thank you

Issue #850, opened 2021-07-10
Add sourcery to config or LunarModules?


ChristianChiarulli:
https://sourcery.ai/blog/sourcery-vim/

Nikhil0504:
A bit more detailed explanation:
This module should give us code refactoring based the level of complexity of our code and a bit more factors that we can change [see](https://github.com/sourcery-ai/sourcery/wiki/Sourcery-configuration)

also this should be added a separate module as we might add a lot flexibility of all of its features over here and we don’t want the config to be clutter over all this.

Here is the guide for installation in vim https://github.com/sourcery-ai/sourcery/wiki/Vim-Usage but they Coc as the LSP but we can figure out another way to use the default LSP for this.

Nikhil0504:
P.S also assign me to this.

mjlbach:
It should be easy to add this to lspconfig, so file a PR there, I'll approve it, then y'all get it for free.

Nikhil0504:
> It should be easy to add this to lspconfig, so file a PR there, I'll approve it, then y'all get it for free.

I am new to lua so I don't know how I would proceed for this PR

mjlbach:
Go to lspconfig, read CONTRIBUTING.md, copy a language server configuration (pyright is a solid one in this case) and document the options that need to be sent. Test if it works. Then you're done ✨

Uzaaft:
Just to make sure: 
Will this feature be optional?
I thought the whole idea behind LunarVim is to use open source software. 
Correct me if I am wrong, but isn't sourcery closed? I think users would prefer this to be optional.

ChristianChiarulli:
@Uzaaft the lspconfig will be there, you will not have to use it, it will not be enabled by default as with all external lang servers.

PR #2347, opened 2022-03-12
chore(plugins): last manual plugin update

abzcoding:
# Description

hopefully this is the last manual plugin update


kylo252 (approved):
LGTM! 🚀 

Also fixes #850 since it's should be available in lspconfig 💪

Issue #2164, opened 2022-01-11
Weird Lunarvim theme behavior


fabioluciano:
### Problem description

I added the dracula theme in Lunarvim and the colorscheme configuration dont persist.

### LunarVim version

1.0.0

### Neovim version (>= 0.6)

NVIM v0.6.1

### Operating system/version

Linux iorek 5.15.13-arch1-1 #1 SMP PREEMPT Wed, 05 Jan 2022 16:20:59 +0000 x86_64 GNU/Linux

### Relevant log output

```shell
NA
```

### Screenshots

![Dracula theme Issue](https://i.imgur.com/t9G2JYt.gif)

abzcoding:
please make sure you have this in your config
```lua
lvim.colorscheme = "dracula"
```

also you can try doing a `:LvimCacheReset` if that didn't help

fabioluciano:
this is my config.lua


``` lua

-- general
lvim.log.level = "warn"
lvim.format_on_save = true
lvim.colorscheme = "dracula"

-- keymappings [view all the defaults by pressing <leader>Lk]
lvim.leader = "space"
-- add your own keymapping
lvim.keys.normal_mode["<C-s>"] = ":w<cr>"

lvim.builtin.dashboard.active = true
lvim.builtin.notify.active = true
lvim.builtin.terminal.active = true
lvim.builtin.nvimtree.setup.view.side = "left"
lvim.builtin.nvimtree.show_icons.git = 0

-- if you don't want all the parsers change this to a table of the ones you want
lvim.builtin.treesitter.ensure_installed = {
  "bash",
  "c",
  "javascript",
  "json",
  "lua",
  "python",
  "typescript",
  "css",
  "rust",
  "java",
  "yaml",
}

lvim.builtin.treesitter.ignore_install = { "haskell" }
lvim.builtin.treesitter.highlight.enabled = true

-- Additional Plugins
lvim.plugins = {
    {
      "dracula/vim",
      config = function()
        vim.g.dracula_colorterm = 0
      end
    },
}


Also, i tried with `:LvimCacheReset`, ans keeps the same behavior.

```

abzcoding:
cc: @kylo252

kylo252:
Can you try commenting out this line: 
https://github.com/LunarVim/LunarVim/blob/e9caf5a69a0c31adcfeeeee928ffe76ad6ba4ceb/lua/lvim/utils/hooks.lua#L16

[...]

PR #2351, opened 2022-03-14
feat(plugins): add support for packer snapshots

### LINKED ISSUE #2343: Packer snapshot integration
### Feature Description

Replace/integrate the current workflow for pinning plugins with the new snapshot implementation `:h packer.snapshot()`. 

The snapshot file is in json (even though the documentation states that they're in lua).

kylo252:
# Description

Replace/integrate the current workflow for pinning plugins with the new snapshot implementation, `:h packer.snapshot()`. 

_Note: The snapshot file is in `json` (even though the documentation states that they're in `lua`)._

Fixes #2343
Related #2085

## Features

1. add a default snapshot, `snapshots/default.json`, to hold the the commits of all the core plugins
2. `LvimSyncCorePlugins` now uses `:PackerSnapshotRollback`
3. reduce errors caused by breaking changes in user plugins since they're not handled anymore with `packer.sync()`
4. `PackerSync` will now ignore all the core plugins which are already handled with snapshots, this makes it signficantly faster to finish.
5. allow "unlocking" all the core plugins, by setting an environment variable `$LVIM_DEV_MODE`, e.g. can be defined in `~/.local/bin/lvim`
6. `$LUNARVIM_CACHE_DIR/snapshots/` can be used to store complete snapshots of _all_ the installed plugins, `:h packer.snapshot()`.
7. add a new handler to allow updating `snapshots/default.json` that does not rely on packer or lvim's runtime. It could be completely re-written in any other language.

## How Has This Been Tested?

- Check that `:PackerSync` does not include any core plugins
- Added a new test case that would revert one of the plugins to an earlier commit, and proceed to verify that `:LvimSyncCorePlugins` can restore the previous state

https://github.com/LunarVim/LunarVim/blob/0d761db45777b0d86764bf3a545e47ce5ff9986f/tests/specs/plugins_load_spec.lua#L35

### DIFF
--- a/lua/lvim/plugin-loader.lua
+++ b/lua/lvim/plugin-loader.lua
@@ -1,33 +1,29 @@
 local plugin_loader = {}
 
-local in_headless = #vim.api.nvim_list_uis() == 0
-
 local utils = require "lvim.utils"
 local Log = require "lvim.core.log"
+local join_paths = utils.join_paths
+
 -- we need to reuse this outside of init()
-local compile_path = get_config_dir() .. "/plugin/packer_compiled.lua"
+local compile_path = join_paths(get_config_dir(), "plugin", "packer_compiled.lua")
+local snapshot_path = join_paths(get_lvim_base_dir(), "snapshots")
 
 function plugin_loader.init(opts)
   opts = opts or {}
 
-  local install_path = opts.install_path or vim.fn.stdpath "data" .. "/site/pack/packer/start/packer.nvim"
-  local package_root = opts.package_root or vim.fn.stdpath "data" .. "/site/pack"
+  local install_path = opts.install_path
+    or join_paths(vim.fn.stdpath "data", "site", "pack", "packer", "start", "packer.nvim")
 
   if vim.fn.empty(vim.fn.glob(install_path)) > 0 then
     vim.fn.system { "git", "clone", "--depth", "1", "https://github.com/wbthomason/packer.nvim", install_path }
     vim.cmd "packadd packer.nvim"
   end
 
-  local log_level = in_headless and "debug" or "warn"
-  if lvim.log and lvim.log.level then
-    log_level = lvim.log.level
-  end
-
-  local _, packer = pcall(require, "packer")
-  packer.init {
-    package_root = package_root,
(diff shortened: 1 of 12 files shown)


abzcoding:
are you sure the `make generate_new_lockfile` is actually doing sth?
Screen Shot 2022-03-15 at 4 17 04 PM

kylo252:
> are you sure the `make generate_new_lockfile` is actually doing sth? 

Try altering the values in `snapshots/defaults.json`, maybe something like this:
```json
  "schemastore.nvim": {
    "commit": "DUMMY"
  },
```
Now run the script again.

abzcoding:
@kylo252 you might need to rebase this, since we've added a lot of new changes to the `rolling` branch

kylo252:
@abzcoding, I've updated the PR description, let me know if you think I missed something 😅

abzcoding (approved):
LGTM 🚀 

we should probably do one last minor release before merging this,

kylo252:
> we should probably do one last minor release before merging this,

@abzcoding, maybe update the screenshot to show alpha.nvim instead as well? 😄

abzcoding:
@kylo252 now we can merge this 🎉

kylo252:
> now we can merge this 🎉

I'm also pretty hopeful it would solve https://github.com/LunarVim/LunarVim/issues/2164

lua/lvim/plugin-loader.lua
```diff
@@ -81,16 +91,14 @@ function plugin_loader.load(configurations)
...
     packer.startup(function(use)
       for _, plugins in ipairs(configurations) do
         for _, plugin in ipairs(plugins) do
           use(plugin)
         end
       end
     end)
-    -- colorscheme must get called after plugins are loaded or it will break new installs.
-    vim.g.colors_name = lvim.colorscheme
-    vim.cmd("colorscheme " .. lvim.colorscheme)
```
MaxVerevkin, line 93: This breaks the colorscheme...

abzcoding: https://github.com/LunarVim/LunarVim/pull/2392

PR #2357, opened 2022-03-16
fix: nvim-tree taking half the window on open

abzcoding:
# Description

Fixes the issue with nvim-tree when you do a `lvim .`
instead of taking half of the window, it will resize to`30`

### DIFF
--- a/lua/lvim/core/nvimtree.lua
+++ b/lua/lvim/core/nvimtree.lua
@@ -79,6 +79,7 @@ function M.config()
           global = false,
         },
         open_file = {
+          resize_window = true,
           quit_on_open = false,
         },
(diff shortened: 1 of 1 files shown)


kylo252 (commented):
I'm not sure what the problem is originally, but I don't think I see a difference with/without this change :thinking:

abzcoding:
run `lvim .`
open a file from nvim-tree using `<CR>`


without this:
Screen Shot 2022-03-18 at 12 18 05 PM

with this
Screen Shot 2022-03-18 at 12 18 25 PM

kylo252 (approved):
LGTM! :rocket:

PR #2359, opened 2022-03-18
refactor: load onedarker theme externally

### LINKED ISSUE #2310: We could probably remove the colorscheme from the repo and manage it elsewhere
### Problem description

we set this here, we can just wrap it in a pcall and use default if colorscheme isn't found.

https://github.com/LunarVim/LunarVim/blob/ec1e4140fa4156df50785da5d6297145762a7daa/lua/lvim/plugin-loader.lua#L99

We may also consider updating the colorscheme, and maybe adding some new images to the readme.

### LunarVim version

latest

### Neovim version (>= 0.6.1)

latest

### Operating system/version

I use arch btw

### Relevant log output

```shell
...
```

### Screenshots

_No response_

kylo252:
# Description

Load onedarker as a plugin.

- define a setup function to be used as an entry point
- remove C and Config from the global scope
- the plugin can be disabled with `lvim.colorscheme`

Fixes #2310

## How Has This Been Tested?

1. set this variable to something else

```lua
lvim.colorscheme = "blah"
```

2. run `:PackerSync` in case it didn't get triggered already


abzcoding (approved):
it seems to be fine

ChristianChiarulli (approved):
👍🏻 lgtm

PR #2362, opened 2022-03-18
chore(plugins): let users enjoy the new global statusline

abzcoding:
# Description

you need neovim nightly and adding this to your config.lua
```lua
lvim.builtin.lualine.options.globalstatus = true
```
Screen Shot 2022-03-18 at 10 39 20 PM


kylo252 (approved):
LGTM! 🚀

PR #2363, opened 2022-03-19
fix(lualine): conditional theme loading

kylo252:
# Description

Only assign lualine's theme to onedarker after the plugin has been enabled and installed.

```lua
lvim.builtin.lualine.options.theme = "onedarker"
```

## How Has This Been Tested?

Should pass the CI more consistently now.


abzcoding (approved):
👍

kylo252:
Okay, I think this finally fixes it!

PR #2367, opened 2022-03-19
fix(bufferline): add an additional space before diagnostics

MaxVerevkin:
# Description

I think there should be a space between icon and the diagnostics count. See below for screenshots. Let me know if this issue is specific to my terminal emulator.

## How Has This Been Tested?

- Use `alacritty`
- Open a project with LSP
- Open a second buffer to show the bufferline
- Make a mistake in your code

## Currently

image

## With this change

image

### DIFF
--- a/lua/lvim/core/bufferline.lua
+++ b/lua/lvim/core/bufferline.lua
@@ -9,7 +9,7 @@ local function diagnostics_indicator(_, _, diagnostics)
   local symbols = { error = "", warning = "", info = "" }
   for name, count in pairs(diagnostics) do
     if symbols[name] and count > 0 then
-      table.insert(result, symbols[name] .. count)
+      table.insert(result, symbols[name] .. " " .. count)
     end
   end
   result = table.concat(result, " ")


abzcoding:
before:
Screen Shot 2022-03-20 at 2 20 22 PM

after:
Screen Shot 2022-03-20 at 2 21 19 PM

on wezterm

@kylo252 can you check on your end?

MaxVerevkin:
By the way, I think it would be nice to have a config option to disable those diagnostics as the status line already displays them. Any thoughts?

abzcoding:
@MaxVerevkin 
```lua
lvim.builtin.bufferline.options.diagnostics = false
```

MaxVerevkin:
Oh, thanks, I couldn't find it somehow :/

kylo252:
It seems to be a font issue, I'm getting slightly different results based on which font I'm using.

@abzcoding, you're also using different symbols 👀

abzcoding:
> It seems to be a font issue, I'm getting slightly different results based on which font I'm using.
> 
> @abzcoding, you're also using different symbols 👀

yep, fill free to merge this, it won't hurt anyways and uses can always make their own `diagnostics_indicator` function

kylo252 (approved):

PR #2380, opened 2022-03-22
fix(terminal): whichkey -> which-key

MaxVerevkin:
# Description

There's a typo in `terminal.lua` which results in the keymap label not being registered.


abzcoding (approved):
👍

PR #2382, opened 2022-03-22
fix(terminal): weird lazygit commit message bug

abzcoding:
# Description

Basically if you use `<space>` as your leader and keep the removed line in this commit
you would face some weird behavior if you try to use lazygit commit message ( hitting `<space>c` and then using space in your commit message)

### DIFF
--- a/lua/lvim/core/terminal.lua
+++ b/lua/lvim/core/terminal.lua
@@ -93,7 +93,6 @@ M.add_exec = function(opts)
     return
   end
   wk.register({ [opts.keymap] = { opts.label } }, { mode = "n" })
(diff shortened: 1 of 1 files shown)


MaxVerevkin:
There is also a which-key mapping a few lines below. I think it should be removed as well.

MaxVerevkin:
> term_mode = { [opts.keymap] = exec_func },

I think this like should be removed too, as it causes (unwanted) terminal switching.

abzcoding:
> > term_mode = { [opts.keymap] = exec_func },
> 
> I think this like should be removed too, as it causes (unwanted) terminal switching.

yep

PR #2383, opened 2022-03-22
fix(terminal): remove term_mode keybinding for exec_funcs

abzcoding:
# Description

I'm not sure why we had this in the first place

### DIFF
--- a/lua/lvim/core/terminal.lua
+++ b/lua/lvim/core/terminal.lua
@@ -85,7 +85,6 @@ M.add_exec = function(opts)
 
   require("lvim.keymappings").load {
     normal_mode = { [opts.keymap] = exec_func },
-    term_mode = { [opts.keymap] = exec_func },
   }
 
   local wk_status_ok, wk = pcall(require, "which-key")


kylo252 (commented):
It's there so that the same binding will work in both modes, if you always dismiss lazygit with `q`, then you wouldn't notice it in the first place 😄 

try something like this to see the "toggle" feature
```lua
-- or use any other keymap other than `<c-t>`
lvim.builtin.terminal.execs[#lvim.builtin.terminal.execs + 1] = { "zsh", "<c-t>", "term", "float" }
```

MaxVerevkin:
`C-t` will work even without this line.

MaxVerevkin:
But this line makes some floating terminals basically unusable in case the keybinding starts with a space.

kylo252:
> `C-t` will work even without this line.

try something other than `<c-t>`, here are more examples:
```lua
lvim.builtin.terminal.execs[#lvim.builtin.terminal.execs + 1] = { "lf", "<M-e>", "lf file manager", "float" }
lvim.builtin.terminal.execs[#lvim.builtin.terminal.execs + 1] = { "gitui", "<M-g>", "gitui", "float" }
lvim.builtin.terminal.execs[#lvim.builtin.terminal.execs + 1] = { "glow", "<M-p>", "glow viewer", "float" }
```

abzcoding:
@kylo252 it is up to you, I'm fine either way 🤷

kylo252:
The way I see it is that it's much easier to not use this API or even undo its changes if necessary, simply do this
```vim
:tunmap <M-E>
```
compared to what is actually required to set it if this API didn't exist
```vim
:tmap <M-E>
t  <M-E>       * <Cmd>lua require('lvim.core.terminal')._exec_toggle({ cmd = 'lf', count = 103, direction = 'float'})<CR>
```
_Note: that `count` value needs to match the one used in normal mode._

---

If you want, you can leave this open for a while and see if others would prefer to have it, there's also another method discussed in https://github.com/LunarVim/LunarVim/issues/2146

MaxVerevkin:
> simply do this

It's a bit more complicated

```lua
lvim.builtin.terminal.on_config_done = function ()
  vim.api.nvim_del_keymap("t", "<leader>tp")
end
```

> compared to what is actually required to set it if this API didn't exist

I think the same effect can be achieved with 

```lua
vim.api.nvim_set_keymap("t", "<leader>tp", "<C-t>", {})
```

PR #2386, opened 2022-03-24
docs: update demo images on the main readme

kylo252:
- updated the demo image to show alpha.nvim
- added a demo of `LvimInfo`
- updated the LSP section
- removed redundant documentation section (it's already available at the references section)


kylo252:
> instead of uploading the images inside the repo, can we use the github upload image link to reduce the repo size? or is that unnecessary, for example this is [my own repo](https://github.com/abzcoding/lvim)

I guess the only downside is not being able to view them offline, but I don't think we're quite there yet in supporting that kind of setup 😄

abzcoding:
fantastic, just one thing, i think you forgot the image with startify dashboard

kylo252:
> fantastic, just one thing, i think you forgot the image with startify dashboard

I had decided to add it to that media folder but not the main page, since I got worried it's too many images. maybe there's a markdown trick to create clickable thumbnails.
also, feel free to set replacements for other 2 images (maybe something with lsp actions and/or diagnostics)

abzcoding:
Yeah, good point. I'll add some fancier images later on. For now, this makes do :)

abzcoding:
@cizordj follow this readme -> https://github.com/nvim-telescope/telescope-media-files.nvim

cizordj:
@abzcoding Thanks

abzcoding:
> @abzcoding Thanks

something like this -> 
```lua
lvim.plugins = {
  {
    "nvim-telescope/telescope-media-files.nvim",
    requires = "nvim-telescope/telescope.nvim",
  } 
}
lvim.builtin.telescope.extensions.media_files = {
      -- filetypes whitelist
      -- defaults to {"png", "jpg", "mp4", "webm", "pdf"}
      filetypes = {"png", "webp", "jpg", "jpeg"},
      find_cmd = "rg" -- find command (defaults to `fd`)
}
```

cizordj:
Thanks for sharing your config file, it'll be extremely helpful for my setup and probably for others as well.

PR #2391, opened 2022-03-25
fix: colorscheme forget to set on start

hulufei:


abzcoding:
it's there already
https://github.com/LunarVim/LunarVim/blob/ea0e1f99e7df476638569a10ddcfdff1dbce52dd/lua/lvim/plugin-loader.lua#L102-L104

PR #2392, opened 2022-03-25
fix(theme): make sure the new theme is fully loaded

abzcoding:
# Description

Make sure the theme is loaded


kylo252 (commented):
I still think it's a race-condition with packer, let's get some testers to verify that this fixes the problem for everyone.

abzcoding:
@angryheng can you check if this fixes your issue or not?

reisnobre:
Just tested locally and it seems to fixed my issue

kylo252 (approved):

PR #2394, opened 2022-03-25
ci: test automated gh action

kylo252:
# Description

Creates a pull request with an updated `snapshots/default.json` on a fixed schedule.

## How Has This Been Tested?

I think we need this to be on rolling first, 
image

### DIFF
--- a/.github/workflows/plugins.yml
+++ b/.github/workflows/plugins.yml
@@ -1,0 +1,61 @@
+name: plugins-version-bump
+
+on:
+  workflow_dispatch:
+    inputs:
+      logLevel:
+        description: 'Log level'
+        required: false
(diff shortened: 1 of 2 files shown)


abzcoding (approved):
LGTM 🚀

kylo252:
@abzcoding, should we just merge it and see what happens? we can always switch back to https://github.com/stefanzweifel/git-auto-commit-action if this doesn't work out.

kylo252:
@abzcoding, this wasn't there before, so I think it actually works! 🎉 

image

Edit: I may have forgotten to mention that it supports manual trigger mode, https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

PR #2417, opened 2022-04-02
ci(test): add write-permission for prs

kylo252:


kylo252:
@abzcoding, I'll merge this for now, since we need to be on rolling for the ci tests anyway, and to also push the stylua fix.

PR #2446, opened 2022-04-10
feat(luasnip):  lazy-load all user snippets by default

danielo515:
just add snipmate loading, so users don't have to add it to their own config.lua

# Description

Just call the function to load snipmate snippets

## How Has This Been Tested?

I use it in my own config

### DIFF
--- a/lua/lvim/plugins.lua
+++ b/lua/lvim/plugins.lua
@@ -70,6 +70,8 @@ local core_plugins = {
     "L3MON4D3/LuaSnip",
     config = function()
       require("luasnip/loaders/from_vscode").lazy_load()
+      require("luasnip/loaders/from_snipmate").lazy_load()
+
     end,
   },
   {
--- a/lua/lvim/plugins.lua
+++ b/lua/lvim/plugins.lua
@@ -69,9 +69,15 @@ local core_plugins = {
   {
     "L3MON4D3/LuaSnip",
     config = function()
-      require("luasnip/loaders/from_vscode").lazy_load()
-      require("luasnip/loaders/from_snipmate").lazy_load()
-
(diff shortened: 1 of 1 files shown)


lua/lvim/plugins.lua
```diff
@@ -70,6 +70,8 @@ local core_plugins = {
     "L3MON4D3/LuaSnip",
     config = function()
       require("luasnip/loaders/from_vscode").lazy_load()
+      require("luasnip/loaders/from_snipmate").lazy_load()
+
```
kylo252, line 74:
```suggestionf
      local utils = require "lvim.utils"
      require("luasnip.loaders.from_lua").lazy_load()
      require("luasnip.loaders.from_vscode").lazy_load {
        paths = {
          utils.join_paths(get_config_dir(), "snippets"),
          utils.join_paths(get_runtime_dir(), "site", "pack", "packer", "start", "friendly-snippets"),
        },
      }
      require("luasnip.loaders.from_snipmate").lazy_load()
```

kylo252 (changes requested):
I had done some tests on the json snippets and found out that we need to define _all_ the expected paths beforehand, otherwise it would only pick the friendly-snippets one.

But now I'm not sure if this will interfere with snipmate, so please test it out (IMPORTANT: you need to run `:PackerSync` because a recompile is required)

We should also load lua-based snippets, as long as they're in `luasnipt/lua` folder somewhere on the rtp.

Lastly, please add more details how this was tested.

lua/lvim/plugins.lua
abzcoding: i like this version 👍

abzcoding:
@danielo515 it should be merged into `rolling` and not `master`

PR #2465, opened 2022-04-14
fix(lualine): color theme gaps in some components

### LINKED ISSUE #2403: Lualine color theme gaps
### Problem description

Since I last updated lvim there are some gaps in the lualine where the color doesn't match the rest of the line. The background seems to transparent. I tried different color schemes but I always get the same behaviour. Does anybody have the same issue?

### LunarVim version

7807f147c192fdb6aec7659a3098b4bf31852fcd

### Neovim version (>= 0.6.1)

0.6.1

### Operating system/version

Arch Linux

### Relevant log output

```shell
nil
```

### Screenshots

Screenshot from 2022-03-28 13-57-33
Screenshot from 2022-03-28 13-45-29

chsjr1996:
# Description
Fix some components on lualine with color theme gaps (diff, filetype, diagnostics)

Fixes #2403

## How Has This Been Tested?
Check if the "diff", "filetype" and "diagnostics" components on lualine respect the current theme.

### DIFF
--- a/lua/lvim/core/lualine/components.lua
+++ b/lua/lvim/core/lualine/components.lua
@@ -29,7 +29,6 @@ return {
   },
   filename = {
     "filename",
-    color = {},
     cond = nil,
   },
   diff = {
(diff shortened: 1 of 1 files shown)


abzcoding (approved):
Thank you for your contribution 👍

PR #2475, opened 2022-04-15
fix(luasnip-cmp): <CR> removes default snip

FT-Labs:
# Description

- Tab mapping in luasnip skippes default value if nothing is typed.
- CR mapping does not behave same way. It deletes default value of snippet if nothing is entered. Make CR to behave same way as Tab.

### DIFF
--- a/lua/lvim/core/cmp.lua
+++ b/lua/lvim/core/cmp.lua
@@ -302,7 +302,10 @@ M.config = function()
         else
           fallback()
         end
-      end),
+      end, {
+        "i",
+        "s",
+      }),
     },
   }
 end


kylo252 (changes requested):
Thank you! 

I personally like this change, but it will add a regression, see https://github.com/LunarVim/LunarVim/issues/1645

The solution needs to take https://github.com/LunarVim/LunarVim/pull/1708 into the consideration, e.g. probably make use of `jumpable()` or `require('luasnip').jumpable()` instead.

danielo515:
this was already fixed in #2980

FT-Labs:
Closing!

PR #2484, opened 2022-04-19
fix(installer): latest and specified release version for neovim have different urls

lmckiwo:
# Description

the URL used to download neovim are different if a version is specified vs if we leave it to default to the latest build.

Fixes #(issue)

## How Has This Been Tested?

This script has been tested as follows:

RELEASE_VER=v0.6.1 bash /scripts/install-neovim-from-release
bash /scripts/install-neovim-from-release

and the nvim version verified after the installation.

### DIFF
--- a/utils/installer/install-neovim-from-release
+++ b/utils/installer/install-neovim-from-release
@@ -23,7 +23,11 @@ else
   exit 1
 fi
 
-declare -r RELEASE_URL="https://github.com/neovim/neovim/releases/$RELEASE_VER/download/$ARCHIVE_NAME.tar.gz"
+if [[ "${RELEASE_VER}" == "latest" ]]; then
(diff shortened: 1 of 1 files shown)


kylo252:
Thank you for your contribution! some minor fixes are needed to pass the CI tests


1. commit-lint: "subject may not be empty [subject-empty]" 
```sh
git commit --amend --no-edit --message "fix(installer): set correct url for nightly"
git push --force -u
```

2. run `shfmt`: see https://github.com/LunarVim/LunarVim/blob/rolling/CONTRIBUTING.md#code-conventions

Check the contributing guidelines in https://github.com/LunarVim/LunarVim/blob/rolling/CONTRIBUTING.md.

kylo252 (approved):
Thanks!

Issue #2497, opened 2022-04-21
server not found in `lsp_servers` folder


alealv:
### Problem description

I've copied [the documentation](https://www.lunarvim.org/languages/rust.html#extra-plugins) on how to setup `rust-tools` but the following line is causing errors.

```lua
cmd = { vim.fn.stdpath("data") .. "/lsp_servers/rust/rust-analyzer" },
```
I replaced it with an absolute path `~/.local/share/nvim/lsp_servers/rust/rust-analyzer` and now it's working.

So the runtime path where `lsp-installer` should put the binaries should be `~/.local/share/lunarvim` is not `~/.local/share/nvim`, right?

### LunarVim version

e81d24e  chore: update changelog

### Neovim version (>= 0.6.1)

NVIM v0.8.0-dev

### Operating system/version

Ubuntu 21.10

### Affected language servers

rust-analyzer

### Actual behavior

I'm getting the message

`Spawning language server with cmd: `/home/aalvarez/.local/share/lunarvim/lsp_servers/rust/rust-analyzer` failed. The language server is either not installed, missing from PATH, or not executable.`

### log and support info

```shell
no need
```

abzcoding:
Nah we should just update the documentation

abzcoding:
cc: @kylo252

abzcoding:
> There's an API from the installer that should be leveraged, instead of hard-coding any path, see: https://github.com/williamboman/nvim-lsp-installer/wiki/Rust#setting-up
> 
> I wonder if we should just undo the override to `vim.fn.stdpath "data"`, if it's starting to become a burden to maintain. We'd only keep the one for `cache`, which should be a lot less problematic.
> 
> I don't use that many plugins, so could you guys run a search for `vim.fn.stdpath "data"`/`vim.fn.stdpath("data")` in `~/.local/share/lunarvim/site` to see how many plugins have some implicit dependency on it.

we should use this in our documentation instead
````lua
      cmd_env = requested_server._default_options.cmd_env,
```

abzcoding:
Fixed via https://github.com/LunarVim/lunarvim.org/commit/b817fc1dcfe33d7cf131e747afb22c8036a076b9

alealv:
Thanks for the fix!

I have one doubt, though. With these changes, will `nlsp-settings` still be able to inject custom config to `nlsp-config`?

abzcoding:
> Thanks for the fix!
> 
[...]

PR #2502, opened 2022-04-22
fix(lsp): make sure servers have cmd_env

abzcoding:
# Description

make sure `jsonls`, `yamlls` and other custom servers work properly

### DIFF
--- a/lua/lvim/lsp/manager.lua
+++ b/lua/lvim/lsp/manager.lua
@@ -31,6 +31,13 @@ local function resolve_config(name, user_config)
   local has_custom_provider, custom_config = pcall(require, "lvim/lsp/providers/" .. name)
   if has_custom_provider then
     Log:debug("Using custom configuration for requested server: " .. name)
+    local lsp_installer_servers = require "nvim-lsp-installer.servers"
+    local server_available, requested_server = lsp_installer_servers.get_server(name)
+    if server_available then
+      config.server = {
+        cmd_env = requested_server._default_options.cmd_env,
+      }
+    end
     config = vim.tbl_deep_extend("force", config, custom_config)
   end
 


lua/lvim/lsp/manager.lua
```diff
@@ -31,6 +31,13 @@ local function resolve_config(name, user_config)
   local has_custom_provider, custom_config = pcall(require, "lvim/lsp/providers/" .. name)
   if has_custom_provider then
     Log:debug("Using custom configuration for requested server: " .. name)
+    local lsp_installer_servers = require "nvim-lsp-installer.servers"
+    local server_available, requested_server = lsp_installer_servers.get_server(name)
+    if server_available then
+      config.server = {
```
kylo252, line 37: I'm not really sure what sure what the use case is here? because if a server is available then all the options are passed to it anyway: https://github.com/LunarVim/LunarVim/blob/0481ec8dddf4bd8ed81c10ae98807fec97f6f872/lua/lvim/lsp/manager.lua#L105

what is `config.server`? this config table is the same as the one passed to the `setup()` function in lspconfig. I just tried with the latest commit on main branch for nvim-lsp-installer, and jsonls is still working as expected (at least with nlsp-settings).

abzcoding: wait i think i fucked up 😅 

this is not needed
i was testing two branches at the same time and i though nlspsettings is not working

my bad 🤦

kylo252: I was just reading the comment from https://github.com/LunarVim/LunarVim/issues/2497#issuecomment-1106188017, and I realized that I haven't really tested if `:LspSettings rust-analyzer` would still work with rust-tools, so please test that 🥲 

_hopefully it's fine_

abzcoding: Screen Shot 2022-04-22 at 6 25 35 PM
😄

PR #2506, opened 2022-04-22
feat(installer): ensure correct responses when prompting user

williamboman:
# Description

This makes sure users provide an acceptable answer when being prompted to confirm an action. I was installing LunarVim and accidentally answered `u` instead of `y` to a question, and the script interpreted it as a no and skipped the step.

Figured I'd just create a PR immediately, wdyt? Didn't touch the powershell script because that's scary.

## How Has This Been Tested?

I've ran the install script from my branch directly:

```
$ LV_BRANCH=rolling bash <(curl -s https://raw.githubusercontent.com/williamboman/LunarVim/install-script-ensure-proper-prompt-answer/utils/installer/install.sh)
```

### DIFF
--- a/utils/installer/install.sh
+++ b/utils/installer/install.sh
@@ -85,6 +85,25 @@ function msg() {
   printf "%s\n" "$text"
 }
 
+function confirm() {
+  local question="$1"
+  while true; do
+    msg "$question"
+    read -p "[y]es or [n]o (default: no) : " -r answer
+    case "$answer" in
+      y | Y | yes | YES | Yes)
+        return 0
+        ;;
+      n | N | no | NO | No | *[[:blank:]]* | "")
+        return 1
+        ;;
+      *)
+        msg "Please answer [y]es or [n]o."
+        ;;
+    esac
+  done
+}
+
 function main() {
   parse_arguments "$@"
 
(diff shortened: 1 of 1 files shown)


kylo252:
I like the idea in principle, but the default value doesn't do anything anymore 😅 

We need it to use "no", i.e. the default value, when you just press `ENTER` without anything else.

utils/installer/install.sh
```diff
@@ -85,6 +85,25 @@ function msg() {
...
+function confirm() {
+  local question="$1"
+  while true ; do
+    msg "$question"
+    read -p "[y]es or [n]o (default: no) : " -r answer
+    case "$answer" in
+      y | Y | yes | YES | Yes)
+        return 0
+        ;;
+      n | N | no | NO | No)
```
kylo252, line 97:
```suggestion
      n | N | no | NO | No | *[[:blank:]]* | "")
```
this seems to work for me, wdyt? but we still need to test it on macOS, just in case, although I believe this is POSIX compliant.

williamboman: Works fine!

utils/installer/install.sh
```diff
@@ -85,6 +85,25 @@ function msg() {
   printf "%s\n" "$text"
 }
 
+function confirm() {
+  local question="$1"
+  while true ; do
```
williamboman, line 90:
```suggestion
  while true; do
```

kylo252 (approved):
LGTM! 🚀

PR #2509, opened 2022-04-22
fix(nvimtree): make sure on_config_done is using the correct require

### LINKED ISSUE #2508: on_config_done callback of nvimtree is getting a table instead of the setup function
### Problem description

[On a previous discussion](https://github.com/LunarVim/LunarVim/issues/2426#issuecomment-1093364338) with @abzcoding about how on_config_done callbacks are supposed to behave, this is the conclusions we got:

>> Also, I still don't know what the thing you get is exactly. Is the instance of the plugin? Is the configuration object? Is it equivalent to do require"plugin-name"? Does it change from plugin to plugin?
> it is equivalent to do require"plugin-name"

However, looking at the code of nvim tree, it is not passing down the equivalent of requiring the module, it is passing down the config submodule, which I guess contains the relevant configuration, but it is not what the expected API is. Take a look here:

https://github.com/LunarVim/LunarVim/blob/f42008ddde7ce9a7a03081a64155558eae99d25a/lua/lvim/core/nvimtree.lua#L124-L129

Then here:

https://github.com/LunarVim/LunarVim/blob/f42008ddde7ce9a7a03081a64155558eae99d25a/lua/lvim/core/nvimtree.lua#L161-L163

### LunarVim version

f41edc6

### Neovim version (>= 0.6.1)

NVIM 0.6.1

### Operating system/version

macOS 12.1

### Relevant log output

```shell
all is good
```

### Screenshots

_No response_

abzcoding:
# Description

make sure nvim_tree.on_config_done is consistent with the rest

Fixes #2508

### DIFF
--- a/lua/lvim/core/nvimtree.lua
+++ b/lua/lvim/core/nvimtree.lua
@@ -122,9 +122,9 @@ function M.config()
 end
 
 function M.setup()
-  local status_ok, nvim_tree_config = pcall(require, "nvim-tree.config")
+  local status_ok, nvim_tree = pcall(require, "nvim-tree")
   if not status_ok then
-    Log:error "Failed to load nvim-tree.config"
+    Log:error "Failed to load nvim-tree"
     return
   end
 
@@ -158,11 +158,11 @@ function M.setup()
     }
   end
 
-  if lvim.builtin.nvimtree.on_config_done then
-    lvim.builtin.nvimtree.on_config_done(nvim_tree_config)
-  end
+  nvim_tree.setup(lvim.builtin.nvimtree.setup)
 
-  require("nvim-tree").setup(lvim.builtin.nvimtree.setup)
+  if lvim.builtin.nvimtree.on_config_done then
+    lvim.builtin.nvimtree.on_config_done(nvim_tree)
+  end
 end
 
 function M.start_telescope(telescope_mode)


lua/lvim/core/nvimtree.lua
```diff
@@ -159,7 +159,7 @@ function M.setup()
   end
 
   if lvim.builtin.nvimtree.on_config_done then
-    lvim.builtin.nvimtree.on_config_done(nvim_tree_config)
+    lvim.builtin.nvimtree.on_config_done(nvim_tree)
```
danielo515, line 162: can we make sure this happens after line 165? After all, this is on_config done, not on_pre_config done...

abzcoding: thanks for pointing that out,fixed it

lua/lvim/core/nvimtree.lua
```diff
@@ -122,7 +122,7 @@ function M.config()
 end
 
 function M.setup()
-  local status_ok, nvim_tree_config = pcall(require, "nvim-tree.config")
+  local status_ok, nvim_tree = pcall(require, "nvim-tree")
   if not status_ok then
     Log:error "Failed to load nvim-tree.config"
```
kylo252, line 127:
```suggestion
    Log:error "Failed to load nvim-tree"
```

lua/lvim/core/nvimtree.lua
```diff
@@ -158,11 +158,11 @@ function M.setup()
     }
   end
 
+  require("nvim-tree").setup(lvim.builtin.nvimtree.setup)
```
kylo252, line 161:
```suggestion
  nvim_tree.setup(lvim.builtin.nvimtree.setup)
```

lua/lvim/core/nvimtree.lua
kylo252: I think it might've had something to do with global variables at the time, since they needed to be called before the setup function itself.

kylo252 (approved):

Issue #2507, opened 2022-04-22
Unable to install language servers through nvim-lsp-installer if using lvim.lsp.manager


Ninja3047:
### Problem description

1. Add this to your config.lua. ccls is an example, using other things also breaks it.
```
require("lvim.lsp.manager").setup("ccls", {})
```
2. Use `:LspInstall` to install an lsp you don't have

See also https://github.com/williamboman/nvim-lsp-installer/issues/621

### LunarVim version

rolling-8fdd4b4

### Neovim version (>= 0.6.1)

NVIM v0.8.0-dev+31-g6c8a3013ac

### Operating system/version

Arch Linux

### Relevant log output

```shell
Installed servers (0) 
  
  Pending servers (2) 
    ◍ ccls (failed)
        ...rt/nvim-lsp-installer/lua/nvim-lsp-installer/core/fs.lua:36: Refusing to rmrf "/home/user/.local/share/lunarvim/lsp_servers/ccls.tmp" which is outside of the allowed boundary "/home/user/.local/share/nvim/lsp_servers". Please report this error at https://github.com/williamboman/nvim-lsp-installer/issues/new
        
    ◍ sumneko_lua (failed)
        ...taller/lua/nvim-lsp-installer/core/installer/context.lua:113: "/home/user/.local/share/lunarvim/lsp_servers/sumneko_lua.tmp" is not a subdirectory of "/home/user/.local/share/nvim/lsp_servers"
```

```
[INFO  Thu 21 Apr 2022 11:07:24 PM EDT] ...-installer/lua/nvim-lsp-installer/ui/status-win/init.lua:682: Starting install server_name="ccls", requested_version=""
[INFO  Thu 21 Apr 2022 11:07:32 PM EDT] ...-installer/lua/nvim-lsp-installer/ui/status-win/init.lua:682: Starting install server_name="ccls", requested_version=""
[ERROR Thu 21 Apr 2022 11:07:32 PM EDT] ...installer/lua/nvim-lsp-installer/core/installer/init.lua:79: Installation failed, name="ccls", error='...rt/nvim-lsp-installer/lua/nvim-lsp-installer/core/fs.lua:36: Refusing to rmrf "/home/user/.local/share/lunarvim/lsp_servers/ccls.tmp" which is outside of the allowed boundary "/home/user/.local/share/nvim/lsp_servers". Please report this error at https://github.com/williamboman/nvim-lsp-installer/issues/new'
[...]

PR #2536, opened 2022-04-28
refactor(lsp): decouple the installer setup-hook

kylo252:
# Description

remove the usage of the deprecated setup-hook from `nvim-lsp-installer`, see https://github.com/williamboman/nvim-lsp-installer/pull/631

## How Has This Been Tested?

custom setup works fine

```lua
vim.list_extend(lvim.lsp.automatic_configuration.skipped_servers, { "clangd", "pyright" })

local opts = {
  cmd = { clangd_bin, unpack(clangd_flags) },
  on_attach = custom_on_attach,
}

require("lvim.lsp.manager").setup("clangd", opts)
-- require("lvim.lsp.manager").setup("clangd") --- this also works now, since opts has been made actually optional :)
```

### DIFF
--- a/lua/lvim/lsp/manager.lua
+++ b/lua/lvim/lsp/manager.lua
@@ -16,29 +16,27 @@ function M.init_defaults(languages)
   end
 end
 
----Resolve the configuration for a server based on both common and user configuration
----@param name string
----@param user_config table [optional]
+---Resolve the configuration for a server by merging with the default config
+---@param server_name string
+---@vararg any config table [optional]
 ---@return table
-local function resolve_config(name, user_config)
-  local config = {
+local function resolve_config(server_name, ...)
+  local defaults = {
     on_attach = require("lvim.lsp").common_on_attach,
     on_init = require("lvim.lsp").common_on_init,
     on_exit = require("lvim.lsp").common_on_exit,
     capabilities = require("lvim.lsp").common_capabilities(),
   }
 
-  local has_custom_provider, custom_config = pcall(require, "lvim/lsp/providers/" .. name)
+  local has_custom_provider, custom_config = pcall(require, "lvim/lsp/providers/" .. server_name)
   if has_custom_provider then
-    Log:debug("Using custom configuration for requested server: " .. name)
-    config = vim.tbl_deep_extend("force", config, custom_config)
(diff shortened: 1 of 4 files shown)


abzcoding:
@kylo252 should this happen?
Screen Shot 2022-04-28 at 5 10 28 PM

kylo252:
You might've missed that I've updated the "lsp" plugins as well [`6b8b94b` (#2536)](https://github.com/LunarVim/LunarVim/pull/2536/commits/6b8b94beb5d40acb0baf6026656075c02dbc7cf3).

it should be fine after a single `:LvimSyncCorePlugins`

williamboman:
Just a FYI: Note that `server:on_ready()` was deprecated as well (no worries though, these methods will remain for the foreseeable future).

I think the remaining use cases for that method can simply be replaced by accepting a callback to the install method like:

```lua
server:install(function (success) end)
```

I wanted to avoid doing this before because of reasons, but with the new lsp-installer setup I'm fine with adding that possibility.

lua/lvim/lsp/manager.lua
```diff
@@ -62,47 +60,52 @@ local function client_is_configured(server_name, ft)
...
-    if install_notification then
-      vim.notify(string.format("Installation complete for [%s] server", requested_server.name), vim.log.levels.INFO)
+  server:on_ready(function()
+    if install_in_progress then
+      vim.notify(string.format("Installation complete for [%s] server", server.name), vim.log.levels.INFO)
     end
-    install_notification = false
-    requested_server:setup(config)
+    install_in_progress = false
+    local config = resolve_config(server_name, server:get_default_options(), user_config)
```
williamboman, line 107: I don't believe you need to retrieve nvim-lsp-installer's options here. This will all be applied automatically by nvim-lsp-installer behind the scenes, or am I missing something?

```diff
-    local config = resolve_config(server_name, server:get_default_options(), user_config)
+    local config = resolve_config(server_name, user_config)
```

kylo252: @williamboman, it's the same problem as https://github.com/LunarVim/LunarVim/issues/2507, I'm worried that the setup is called too late, since `lvim.lsp.manager` is usually called from an ftplugin file.

Instead I want to do the setup more imperatively, to avoid relying on load order.

PR #2538, opened 2022-04-28
feat(lua-dev): use the newer lua-dev branch till folke comes back

abzcoding:
# Description

Since the original one has not been updated in a while, use this branch temporarily


kylo252 (approved):

danielo515:
Why was that for paricularly chosen?

folke:
fyi, lua-dev now has a github action that trigger every night to update the docs. So even when I'm away, it gets updated ;)

PR #2542, opened 2022-04-29
feat: add cmp-tmux to the list of sources

pschmitt:
# Description

This adds tmux to the cmp sources.

https://github.com/andersevenrud/cmp-tmux

## How Has This Been Tested?

Tested locally, cmp-tmux works as expected.

### DIFF
--- a/lua/lvim/core/cmp.lua
+++ b/lua/lvim/core/cmp.lua
@@ -207,6 +207,7 @@ M.config = function()
         vsnip = "(Snippet)",
         luasnip = "(Snippet)",
         buffer = "(Buffer)",
+        tmux = "(TMUX)",
       },
       duplicates = {
         buffer = 1,
@@ -247,6 +248,7 @@ M.config = function()
       { name = "emoji" },
       { name = "treesitter" },
       { name = "crates" },
+      { name = "tmux" },
     },
     mapping = cmp.mapping.preset.insert {
       ["<C-k>"] = cmp.mapping.select_prev_item(),


kylo252 (approved):
Thanks!

danielo515:
What is the intention behind this? As far as I know ,that cmp extension is not shipped by default

kylo252:
> What is the intention behind this? As far as I know ,that cmp extension is not shipped by default

it makes it so that you don't have to manually add tmux to the list of recognizable sources, so just installing the plugin would make it work fine with cmp, tabnine is another example.

danielo515:
While that is understandable, I think it serves no purpose. First because
it is not documented , so unless tu take the time to read the entire
lunarvim code base, you are the original implementers of this,or you are
being told about it in discord, you will have no idea. Second, because if
you follow the extension instructions ,you will end with this config
duplicated, which I think it is undefined behavior .
And last, if I get it correctly, there is no way to configure from users
space? Please let me know if I'm wrong.
So, probably the best solution is to document it

thyseus:
Another solution would be to simply ship

https://github.com/andersevenrud/cmp-tmux

with the default plugins of lunarvim. This plugin is lightweight and tmux is very common imho.

kylo252:
> Another solution would be to simply ship
> 
> [andersevenrud/cmp-tmux](https://github.com/andersevenrud/cmp-tmux)
> 
> with the default plugins of lunarvim. This plugin is lightweight and tmux is very common imho.

I think that would open the flood doors to including all of cmp's sources :thinking:

thyseus:
Thats true. Maybe a _separate_ configuration file for cmp that has all sources listed at https://github.com/hrsh7th/nvim-cmp/wiki/List-of-sources but _COMMENTED_ by default could be shipped with lunarvim. One just needs to _uncomment_ the cmp plugins he is interested in ?

kylo252:
> One just needs to _uncomment_ the cmp plugins he is interested in ?

@thyseus, please create a PR to update this list with some of the missing useful/common sources

https://github.com/LunarVim/LunarVim/blob/9f1f98312bfc919b695935ec80ba3501b6cfbab9/utils/installer/config.example.lua#L96-L102

and maybe another PR to add them to the [site](https://github.com/LunarVim/lunarvim.org/blob/master/versioned_docs/version-1.2/plugins/extra-plugins.md) :smile:

PR #2553, opened 2022-05-03
Use treesitter as fold method

LintaoAmons:
# Description

fold file blocks according to treesitter as default.

## How Has This Been Tested?

I'm not familiar with the test, but I use this function already. Maybe not merge this PR, but I think this fold-default should be included out of the box.


kylo252:
Thank you for your contribution!

I used to run this setup, but after a lot of hair-pulling trying to find out why formatting certain files is slow, it turned out that this is the culprit. so we'd have to wait with this until the feature is more stable, ref https://github.com/nvim-treesitter/nvim-treesitter/blob/09efcdb0498b60f698aa1e879e4ad07d84d1c1a5/doc/nvim-treesitter.txt#L465-L467

The best approach would be to enable per filetype, but that's easier controlled from outside of lunarvim's core.

Issue #2286, opened 2022-02-16
Various themes for the main UIs


premell:
Recently I tried nvchad, and while I prefer lunarvims configuration and defaults, I found their UI to be really beatiful. These are some examples from nvchad: 

This is their theme for telescope: 
![image](https://user-images.githubusercontent.com/65544203/154215161-70e3e0c6-c87b-4a4a-9301-bd22538c7fb6.png)

This is their window for showing keymappings
![image](https://user-images.githubusercontent.com/65544203/154215263-8a04bc75-ea25-465c-ad1f-601b1e65e20a.png)

I think it would be really cool to have a few different themes for some plugins, like telescope, which you could choose from your config like "lvim.builtin... telescope.themes = 'gravity'" or something like that. Right now I feel like the structure of lunarvim is great, and the defaults are awesome. However there is not that much builtin support for customizing the look

abzcoding:
I agree with you, and I like your idea 👍 I'll work on this in a week or so.

abzcoding:
testing it on [my own config](https://github.com/abzcoding/lvim)

kanagawa:
<img width="1752" alt="Screen Shot 2022-02-17 at 12 33 20 AM" src="https://user-images.githubusercontent.com/10992695/154356707-aa206af5-de28-4d8d-9c29-e2208b42b972.png">


catppuccin:
<img width="1751" alt="Screen Shot 2022-02-17 at 12 37 39 AM" src="https://user-images.githubusercontent.com/10992695/154357253-e3550ef3-ee74-4bb8-876f-0d87c8bb69bc.png">

tokyonight:
<img width="1742" alt="Screen Shot 2022-02-17 at 12 38 51 AM" src="https://user-images.githubusercontent.com/10992695/154357439-63c1721e-0c82-4515-b8cf-a41ef56c99a8.png">

rose-pine:
<img width="1711" alt="Screen Shot 2022-02-17 at 12 40 42 AM" src="https://user-images.githubusercontent.com/10992695/154357706-a4c2ae7a-cfd5-47a2-a050-7d64e245fcfb.png">


what do you think?

abzcoding:
this would be an alternative look ( prompt being a different color)

<img width="1785" alt="Screen Shot 2022-02-17 at 1 06 46 AM" src="https://user-images.githubusercontent.com/10992695/154361278-92410d66-b96d-4803-97d0-b52484bf7cc4.png">


would this be better?

abzcoding:
[...]

Issue #1821, opened 2021-10-22
[Bug]: lvim.transparent_window = true produces error


argapost:
### Problem description

When I set `lvim.transparent_window = true` I get the following error:

```
Highlight group 'Normal' has no background highlight.

Please provide an RGB hex value or highlight group with a background value for 'background_colour' option

Defaulting to #000000
```
Can anyone reproduce?

Thanks in advance

- [X] Read the readme
- [X] Read the wiki
- [X] Searched for similar issues

PR #2557, opened 2022-05-03
Add telescope ui select

gelfand:
# Description

Adds [telescope-ui-select.nvim](https://github.com/nvim-telescope/telescope-ui-select.nvim) that returns code actions UI.

---
image

### DIFF
--- a/lua/lvim/core/telescope.lua
+++ b/lua/lvim/core/telescope.lua
@@ -87,6 +87,9 @@ function M.config()
         override_file_sorter = true, -- override the file sorter
         case_mode = "smart_case", -- or "ignore_case" or "respect_case"
       },
+      ["ui-select"] = {
+        require("telescope.themes").get_dropdown {},
+      }
     },
   })
 end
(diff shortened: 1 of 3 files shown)


kylo252:
thanks for your contribution! please take a look at  https://github.com/LunarVim/LunarVim/blob/rolling/CONTRIBUTING.md for additional info on the style guide (both commit and code)


related https://github.com/LunarVim/LunarVim/issues/2286

abzcoding:
I don't think this should be a core functionality, 
i'm using [dressing.nvim](https://github.com/stevearc/dressing.nvim) to achieve the same 

@kylo252 what do you think?

kylo252:
> ~I don't think this should be a core functionality~ i'm using [dressing.nvim](https://github.com/stevearc/dressing.nvim) to achieve the same
> on another note, the default one is really ugly 😅 , so maybe we should have sth

that's exactly why I tagged you, I'm not sure if we can detect any overrides to `vim.ui.select`, but I guess it would have to be similar to how we treat `vim.notify`, e.g. offers all the configurations (within reason), and allow to completely disable it.

that said, I think we need a compelling reason first, like a nice integration with a (few) specific theme, otherwise, it's just one more thing to maintain 😢 

oh and btw, this might be a complete headache if these plugins rely on non-default highlight groups, similar to https://github.com/LunarVim/LunarVim/issues/1821

sagikazarmark:
Why was this closed? Is there an alternative solution?

gelfand:
> Why was this closed? Is there an alternative solution?

install
`    { "stevearc/dressing.nvim" }
`

Issue #2578, opened 2022-05-06
Error when moving cursor around 


purpleyrose:
### Problem description

1. Use Neovim 0.7 or higher
2. Open any file
3. Use h,j,k,l or the arrow keys
4. Get the following error:
```
                                                                                            
Error detected while processing CursorMoved Autocommands for "<buffer=1>":                  
Error executing lua callback: /usr/share/nvim/runtime/lua/vim/lsp/util.lua:1508: Expected Lu
a number                
```

### LunarVim version

aa4c993

### Neovim version (>= 0.7)

NVIM v0.7.0 Build type: Release LuaJIT 2.1.0-beta3 Compiled by runner@fv-az316-460

### Operating system/version

Gentoo/Linux x86_64 5.15.32-gentoo-dist

### Relevant log output

```shell
checkhealth:

null-ls: require("null-ls.health").check()
========================================================================
  - INFO: no sources registered

nvim: health#nvim#check
========================================================================
## Configuration
  - WARNING: Missing user config file: /home/pam/.config/nvim/init.vim
    - ADVICE:
      - :help |init.vim|

## Performance
  - OK: Build type: Release

## Remote Plugins
  - OK: Up to date

## terminal
  - INFO: key_backspace (kbs) terminfo entry: key_backspace=\177
  - INFO: key_dc (kdch1) terminfo entry: key_dc=\E[3~
  - INFO: $COLORTERM='truecolor'

nvim-lsp-installer: require("nvim-lsp-installer.health").check()
========================================================================
## nvim-lsp-installer report
  - OK: neovim version >= 0.7.0
  - WARNING: **Composer**: not available
  - WARNING: **PHP**: not available
  - WARNING: **julia**: not available
  - OK: **bash**: `GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)`
  - OK: **tar**: `tar (GNU tar) 1.34`
  - OK: **gzip**: `gzip 1.12`
  - OK: **curl**: `curl 7.79.1 (x86_64-pc-linux-gnu) libcurl/7.79.1 OpenSSL/1.1.1n zlib/1.2.11 nghttp2/1.45.1`
  - OK: **wget**: `GNU Wget 1.21.2 built on linux-gnu.`
  - OK: **python3**: `Python 3.9.12`
  - OK: **node**: `v14.19.0`
  - OK: **Ruby**: `ruby 2.6.10p210 (2022-04-12 revision 67958) [x86_64-linux]`
[...]

Issue #2434, opened 2022-04-07
Telescope previewer times out in big projects


expoure:
### Problem description

1 - open a big project
2 - try navigate in telescope

below a compartion between nvim and lvim

my nvim configuration for telescope:
`if !exists('g:loaded_telescope') | finish | endif

nnoremap  <silent> ;f <cmd>lua require('telescope.builtin').find_files()<cr>
nnoremap  <silent> ;r <cmd>lua require('telescope.builtin').live_grep()<cr>
nnoremap  <silent> ;b <cmd>lua require('telescope.builtin').file_browser()<cr>
nnoremap <silent> \\ <cmd>Telescope buffers<cr>
nnoremap <silent> ;; <cmd>Telescope help_tags<cr>

lua << EOF
function telescope_buffer_dir()
  return vim.fn.expand('%:p:h')
end

local telescope = require('telescope')
local actions = require('telescope.actions')

telescope.setup{
  defaults = {
    mappings = {
      n = {
        ["q"] = actions.close
      },
    },
  }
}
EOF`

https://user-images.githubusercontent.com/12136268/162202466-7d695897-664f-4020-ad0a-e3ea5580bfab.mp4

### LunarVim version

master-23feb96

### Neovim version (>= 0.6.1)

0.6.1

### Operating system/version

Archlinux

### Relevant log output

```shell
No output
```

abzcoding:
hmm, @kylo252 any ideas what might cause this?

kylo252:
please share the output of `:checkhealth telescope`

expoure:
`telescope: require("telescope.health").check()
========================================================================
## Checking for required plugins
  - OK: plenary installed.
  - OK: nvim-treesitter installed.

## Checking external dependencies
  - OK: rg: found ripgrep 13.0.0
  - OK: fd: found fd 8.3.1

## ===== Installed extensions =====

## Telescope Extension: `fzf`
  - INFO: No healthcheck provided

## Telescope Extension: `projects`
  - INFO: No healthcheck provided
`

kylo252:
Looking at that demo again, the project isn't that big, and it seems that only the previewer is slow. 
- compare the performance with the previewer disabled
```vim
:Telescope find_files previewer=false
```
- run `:TSUpdate`, restart `lvim` and see if there's a difference (in case there was an update to the typescript parser)

expoure:
[...]

PR #2581, opened 2022-05-06
fix(autocmds): disable commentstring_calc on cursor-hold

kylo252:
# Description

- (fix) the autocmd on CursorHold for commentstring is redundant since`comment.nvim` already has a `pre_hook` doing the same thing.
- (fix) run the format function once: we should just return if `vim.lsp.buf.format` exists :sweat_smile:

### DIFF
--- a/lua/lvim/core/treesitter.lua
+++ b/lua/lvim/core/treesitter.lua
@@ -12,11 +12,12 @@ M.config = function()
     },
     highlight = {
       enable = true, -- false will disable the whole extension
-      additional_vim_regex_highlighting = true,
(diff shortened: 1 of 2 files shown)


kylo252:
@abzcoding, potentially fixes #2578

abzcoding (approved):
LGTM 🚀 

while at it, can you add this as well?

https://github.com/LunarVim/LunarVim/issues/2434#issuecomment-1119807392

kylo252:
> while at it, can you add this as well?

addressed in https://github.com/LunarVim/LunarVim/issues/2434#issuecomment-1120146327

PR #2610, opened 2022-05-12
chore: update templates

danielo515:
# Description

This updates the general issue template so the commands to execute to get the relevant logs do not dissapear when you paste the result of the first one. This always frustrates me. Also, the placeholder is not possible to copy, so it is even more work
 
## How Has This Been Tested?

I did not tested, but I don't think it is necessary

### DIFF
--- a/.github/ISSUE_TEMPLATE/general-issue-form.yaml
+++ b/.github/ISSUE_TEMPLATE/general-issue-form.yaml
@@ -23,7 +23,7 @@ body:
     validations:
       required: true
   - type: input
-    id: lunar-vim-version
+    id: version
     attributes:
       label: LunarVim version
       placeholder: |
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/general-issue-form.yaml
```diff
@@ -51,10 +51,14 @@ body:
...
       label: Relevant log output
-      placeholder: |
+      description: |
         :checkhealth
         :messages
         :e $LUNARVIM_CACHE/DIR/lvim.log
+      placeholder: |
+      ```
+      paste here
+      ```
```
kylo252, line 61: doesn't this need to be indented?

> There is a problem with this template

>
> YAML syntax error: (): found character that cannot start any token while scanning for the next token at line 59 column 7. [Learn more](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/common-validation-errors-when-creating-issue-forms).

danielo515: true, I'll try to fix it.
What do you think about splitting this into several inputs? one for each command, so people don't need to provide their own backtics?

kylo252: good idea!

kylo252:
Thanks a lot! I've updated the other template as well as the PR one.

PR #2644, opened 2022-05-20
fix(logger): add full time-stamp to the file sink

### LINKED ISSUE #2611: Add date to the timestamp of the logs
### Feature Description

Something that always confuses me when checking lunar vim logs are the timestamps. Because they don't include the date, there were some situations were I was reading errors of 2 or 3 days ago, and thinking they were recent.
This is a very confusing user experience with very easy fix: include the date in the logs timestamps.

Kill myself 😄 , just joking.

### Support information

_No response_

danielo515:
# Description

summary of the change

This adds the year/month/day to the log entries, in addition to the existing time

fixes #2611

### DIFF
--- a/lua/lvim/core/log.lua
+++ b/lua/lvim/core/log.lua
@@ -29,7 +29,7 @@ function Log:init()
           processors = {
             structlog.processors.Namer(),
             structlog.processors.StackWriter({ "line", "file" }, { max_parents = 0, stack_level = 2 }),
-            structlog.processors.Timestamper "%H:%M:%S",
+            structlog.processors.Timestamper "%Y/%m/%d %H:%M:%S",
           },
           formatter = structlog.formatters.FormatColorizer( --
             "%s [%-5s] %s: %-30s",


danielo515:
Didn't know we follow conventional commits. What are the available subjects?

kylo252:
Thanks!

> Didn't know we follow conventional commits. What are the available subjects?

weird that the error message is a bit weird, https://github.com/LunarVim/LunarVim/runs/6527215954?check_suite_focus=true

you can see them here: https://github.com/LunarVim/LunarVim/blob/rolling/CONTRIBUTING.md#commit-messages

lua/lvim/core/log.lua
```diff
@@ -29,7 +29,7 @@ function Log:init()
           processors = {
             structlog.processors.Namer(),
             structlog.processors.StackWriter({ "line", "file" }, { max_parents = 0, stack_level = 2 }),
-            structlog.processors.Timestamper "%H:%M:%S",
+            structlog.processors.Timestamper "%Y/%m/%d %H:%M:%S",
```
kylo252, line 32: oh you actually modified the wrong one (console), it should be the one in `structlog.sinks.File` a few lines below 😄 

and I think it might be better to use  `"%F %H:%M:%S"` just it matches the one from neovim. https://github.com/neovim/neovim/blob/3d54fbf5c04dcb405e5c28ad7361273bf563ac92/runtime/lua/vim/lsp/log.lua#L17

danielo515: Whoops, my bad. What does the `%F` do ?

danielo515: If it is DDMonYY then I prefer G. I find the American order is terrible for proper sorting

PR #2659, opened 2022-05-26
fix(autocmds): toggle format-on-save properly

choplin:
# Description

`LvimToggleFormatOnSave` can disable the format-on-save feature properly, but once the feature is disabled from the enabled state, this command cannot enable it again. This is because `vim.api.nvim_get_autocmds` returns an empty list after disabling the feature.

This commit checks the size of the returned autocmd list when `M.toggle_format_on_save` is called and enables the feature if the size is 0.

## How Has This Been Tested?

Run `LvimToggleFormatOnSave` several times, and confirm that the code is automatically formatted after running `LvimToggleFormatOnSave` from the disabled state.

### DIFF
--- a/lua/lvim/core/autocmds.lua
+++ b/lua/lvim/core/autocmds.lua
@@ -118,11 +118,11 @@ function M.configure_format_on_save()
 end
 
 function M.toggle_format_on_save()
-  local exists, _ = pcall(vim.api.nvim_get_autocmds, {
+  local exists, autocmds = pcall(vim.api.nvim_get_autocmds, {
     group = "lsp_format_on_save",
     event = "BufWritePre",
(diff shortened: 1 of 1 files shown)


kylo252:
Good catch! 

I think I originally wanted to delete the augroup itself, but that solution was flakey because I was getting errors about it still being in-use.

choplin:
Should I fix the commit message to pass the lint-commits?

kylo252:
> Should I fix the commit message to pass the lint-commits?

yes please

choplin:
@kylo252 Revised the commit message and force-pushed. Could you please approve to run CI?

kylo252:
Thanks!

PR #2683, opened 2022-06-02
fix(lualine): unique buf client names

emxxjnm:
# Description

unqiue buf client names in lualine lsp components


abzcoding (approved):
tyvm

PR #2697, opened 2022-06-07
chore(lsp): notify format request failed once when there is no client

cpea2506:
# Description
The notification message is too noisy when there are no matching language server each time we save with `format_on_save` is on. That's enough if it notifies only once.

<div align = "center">
 Screen Shot 2022-06-07 at 14 56 59
</div>

### DIFF
--- a/lua/lvim/lsp/utils.lua
+++ b/lua/lvim/lsp/utils.lua
@@ -176,7 +176,7 @@ function M.format(opts)
   end, clients)
 
   if #clients == 0 then
-    vim.notify "[LSP] Format request failed, no matching language servers."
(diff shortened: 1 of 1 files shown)


nikita-orca:
+

renaldofreire:
I am also looking for a solution to this.

abzcoding (approved):
tvym

oezguerbalataci:
The problem still exists. On every save I am getting the LSP format request failed notification.

cpea2506:
> The problem still exists. On every save I am getting the LSP format request failed notification.

This PR do not remove completely the message. It just makes the notify message appear only once each time you open project.

PR #2698, opened 2022-06-07
feat(lsp): add option to override default `nvim-lsp-installer` settings

cpea2506:
# Description

Allow user to change default settings of `nvim-lsp-installer`

## How Has This Been Tested?

User can use `lvim.lsp.installer` to change setting of `nvim-lsp-installer`, accepts the same values as [original default configuration](https://github.com/williamboman/nvim-lsp-installer#default-configuration)

```lua
lvim.lsp.installer = {
        ensure_installed = {
            "rust_analyzer",
            "sumneko_lua",
            "taplo",
            "tsserver",
            "cssls",
            "jsonls",
        },
        ui = {
            border = "rounded",
            icons = {
                server_installed = "✓",
                server_pending = "➜",
                server_uninstalled = "✗",
            },
            keymaps = {
                toggle_server_expand = "<CR>",
                install_server = "i",
                update_server = "u",
                uninstall_server = "d",
            },
        },
}
```

### DIFF
--- a/lua/lvim/lsp/config.lua
+++ b/lua/lvim/lsp/config.lua
@@ -119,6 +119,8 @@ return {
     insert_mode = {},
     visual_mode = {},
   },
+  ---@usage list of settings of nvim-lsp-installer
+  installer = {},
   null_ls = {
     setup = {},
     config = {},
--- a/lua/lvim/lsp/config.lua
+++ b/lua/lvim/lsp/config.lua
@@ -121,18 +121,16 @@ return {
   },
   ---@usage list of settings of nvim-lsp-installer
   installer = {
-    ensure_installed = {
-      "sumneko_lua",
-      "jsonls",
-    },
-    ui = {
-      icons = {
-        server_installed = "✓",
-        server_pending = "",
-        server_uninstalled = "✗",
+    setup = {
+      ensure_installed = {},
+      ui = {
+        icons = {
+          server_installed = "✓",
+          server_pending = "",
+          server_uninstalled = "✗",
+        },
       },
     },
-    log_level = lvim.log and lvim.log.level or vim.log.levels.INFO,
   },
   null_ls = {
     setup = {},
--- a/lua/lvim/lsp/config.lua
+++ b/lua/lvim/lsp/config.lua
@@ -120,7 +120,20 @@ return {
     visual_mode = {},
   },
   ---@usage list of settings of nvim-lsp-installer
-  installer = {},
+  installer = {
+    ensure_installed = {
+      "sumneko_lua",
+      "jsonls",
+    },
+    ui = {
+      icons = {
+        server_installed = "✓",
+        server_pending = "",
+        server_uninstalled = "✗",
+      },
+    },
+    log_level = lvim.log and lvim.log.level or vim.log.levels.INFO,
+  },
   null_ls = {
     setup = {},
     config = {},
--- a/lua/lvim/lsp/init.lua
+++ b/lua/lvim/lsp/init.lua
@@ -122,11 +122,7 @@ function M.setup()
     append_default_schemas = true,
   }
 
-  require("nvim-lsp-installer").setup(vim.tbl_extend("force", lvim.lsp.installer, {
-
-    -- use the default nvim_data_dir, since the server binaries are independent
-    install_root_dir = utils.join_paths(vim.call("stdpath", "data"), "lsp_servers"),
-  }))
+  require("nvim-lsp-installer").setup(lvim.lsp.installer)
 
   require("lvim.lsp.null-ls").setup()
 
--- a/lua/lvim/lsp/init.lua
+++ b/lua/lvim/lsp/init.lua
@@ -122,7 +122,7 @@ function M.setup()
     append_default_schemas = true,
   }
 
-  require("nvim-lsp-installer").setup(lvim.lsp.installer)
+  require("nvim-lsp-installer").setup(lvim.lsp.installer.setup)
 
   require("lvim.lsp.null-ls").setup()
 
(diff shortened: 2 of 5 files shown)


lua/lvim/lsp/config.lua
```diff
@@ -118,6 +118,8 @@ return {
     insert_mode = {},
     visual_mode = {},
   },
+  ---@usage list of settings of nvim-lsp-installer
+  installer = {},
```
kylo252, line 122:
```suggestion
  installer = {
    -- use the default nvim_data_dir, since the server binaries are independent
    install_root_dir = utils.join_paths(vim.call("stdpath", "data"), "lsp_servers"),
  },
```

lua/lvim/lsp/init.lua
```diff
@@ -122,10 +122,11 @@ function M.setup()
     append_default_schemas = true,
   }
 
-  require("nvim-lsp-installer").setup {
+  require("nvim-lsp-installer").setup(vim.tbl_extend("force", lvim.lsp.installer, {
```
kylo252, line 125:
```suggestion
  require("nvim-lsp-installer").setup(lvim.lsp.installer)
```

lua/lvim/lsp/init.lua
```diff
@@ -122,10 +122,11 @@ function M.setup()
...
+
     -- use the default nvim_data_dir, since the server binaries are independent
     install_root_dir = utils.join_paths(vim.call("stdpath", "data"), "lsp_servers"),
```
kylo252, line 128:
```suggestion
```

lua/lvim/lsp/init.lua
```diff
@@ -122,10 +122,11 @@ function M.setup()
...
-  }
+  }))
```
kylo252, line 129:
```suggestion
```

kylo252 (commented):
Sorry for the delay.

This is a really good idea, thanks!

lua/lvim/lsp/config.lua
cpea2506: As i see in the default setting of `nvim-lsp-installer`, it sets `install_root_dir` as the same as us. Can i completely remove this setting?

https://github.com/williamboman/nvim-lsp-installer/blob/5904749c2b7fb9d70967a9dc3599107e3d07c6d1/lua/nvim-lsp-installer/settings.lua#L53-L54

cpea2506: I have added some default settings.

kylo252 (commented):
what's left:

- [ ] consider creating an explicit `lvim.lsp.installer.setup[...]` 
- [ ] figure out if `lvim.lsp.installer.ensure_installed` should take priority over `lvim.lsp.automatic_configuration.skipped_servers` or not, see https://github.com/LunarVim/LunarVim/blob/8989984b781367b2744c4857e73d8943311db241/lua/lvim/lsp/templates.lua#L24-L27
- [ ] document this somewhere, maybe in [`config.example.lua`](https://github.com/LunarVim/LunarVim/blob/8989984b781367b2744c4857e73d8943311db241/utils/installer/config.example.lua#L90-L94)

cpea2506:
Some note about my new commit:

- [x] added config example
- [x] move all field to `lvim.lsp.installer.setup`
- [x] ensure_installed: i think that if i want to ensure install some servers. As its name,  it should take higher priority and remove those server from list of `skipped_servers`

lua/lvim/lsp/templates.lua
```diff
@@ -17,13 +17,22 @@ end
...
-    return
+    if not vim.tbl_contains(ensure_installed_servers, server_name) then
+      return
+    else
+      -- ensure_installed should take priority over skipped_servers
+      -- remove server from skipped_servers list if it is ensured install
+      lvim.lsp.automatic_configuration.skipped_servers = vim.tbl_filter(function(s)
+        return s ~= server_name
+      end, lvim.lsp.automatic_configuration.skipped_servers)
+    end
```
kylo252, line 35: I'm not sure what the best way to handle this is, I think this might get confusing to debug if we start modifying this list, keep in mind that this template generator doesn't get called all that often.

However I see the benefit, since `LvimInfo` also relies on this list.

If you want to go with this approach of modifying the skip list, then you need to do it in this `config:load()`, instead
https://github.com/LunarVim/LunarVim/blob/958030e728d3a853cc85051edf0326593c24148f/lua/lvim/config/init.lua#L104-L106

or create a new `should_skip()` function in the template generator, that checks all those hierarchy rules without modifying any of the original lists.

cc: @abzcoding, @rebuilt

cpea2506: Then i should go with the second approach. Which do you think is better?

```lua
function M.generate_ftplugin(server_name, dir)
  if vim.tbl_contains(skipped_servers, server_name) 
      and not vim.tbl_contains(ensure_installed_servers, server_name) then
    return
  end
-- the rest of templates generator
```

or 

```lua
local function should_skip(server_name)
  return vim.tbl_contains(skipped_servers, server_name) 
      and not vim.tbl_contains(ensure_installed_servers, server_name)
end

function M.generate_ftplugin(server_name, dir)
  if should_skip(server_name) then
    return
  end
-- the rest of templates generator
```

kylo252: I'd go with with the second version, because there's also the case of skipped filetypes 😅 

https://github.com/LunarVim/LunarVim/blob/283e38b733c29514ee602ae3c251a2d9f4a9c053/lua/lvim/lsp/templates.lua#L38-L45

it's more for readability than anything, and would make it easier to add some debug log entries

kylo252: @cpea2506, rebasing on rolling should now fix the CI issues, sorry about that!

cpea2506: Ah! Okay. Im crying with that thing! 😭

kylo252 (approved):
LGTM! 🚀 

we can always iterate on this later, tysm!

PR #2705, opened 2022-06-09
[Feature] File detect *.astro files

0ql:
# Description

Syntax highlighting for astro doesn't currently work even with astro lsp and TS syntax highlighting.
Added `ftdetect/astro.lua` file:
```lua
vim.cmd [[
  au BufRead,BufNewFile *.astro set filetype=astro
]]
```

## How Has This Been Tested?

It now correctly detects `.astro` files. Works fine across all my astro projects and on a clean install of lvim.


kylo252:
Thanks for your contribution!

There's already a tree-sitter [parser](https://github.com/nvim-treesitter/nvim-treesitter/blob/32e364ea3c99aafcce2ce735fe091618f623d889/lua/nvim-treesitter/parsers.lua#L1035-L1042) for astro, so I suggest you open an issue there, since it should work out of the box. 


As we've already removed all those ftdetect files in https://github.com/LunarVim/LunarVim/pull/2651, you can just create a `~/.config/lvim/filetype.lua` and add the following snippet using the new lua API

```lua
-- see :h new-filetype
vim.filetype.add {
  extension = {
    astro = "astro",
  },
}
```

PR #2707, opened 2022-06-11
fix: skip calling nvim-tree.setup() more than once

kylo252:
# Description

This change introduced in https://github.com/kyazdani42/nvim-tree.lua/pull/1308/commits/d369e7d063a76a2810dce473d237f0c7c831bab7 doesn't play nicely with packer

## How Has This Been Tested?

- Run `:PackerSync` without getting warnings..

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,5 @@ lua/lv-user-config/
 *.bak
 *.backup
 *.old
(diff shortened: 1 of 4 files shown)


abzcoding (approved):
👍

PR #2709, opened 2022-06-12
feat(logger): add open command

### LINKED ISSUE #2708: Provide a command to open LunarVim log
### Feature Description

Hello.
I think it will be nice if we provide the users a way to open the lunar vim log file (or just the log) directly from a command. For me typing the sequence required right now is a bit out of my comfort zone (space + shift + l + l + shift + d ) and I will be happier typing a command.

Doing it myself in my own config, but a built in will be something nice to have

### Support information

_No response_

danielo515:
# Description

This adds two new commands, as suggested in #2708. One for open the lunarvim log as a file, and another to just open lunarvim log.
They follow a similar naming as they which-key counterparts, one being named after open and the other after edit:

- `LvimOpenlog`
- ~~`LvimEditlog`~~

fixes #2708 

## How Has This Been Tested?

I tested it on my config and it works fine

- ~~`:LvimEditlog`~~
- See logs
- `:LvimOpenlog`

### DIFF
--- a/lua/lvim/core/commands.lua
+++ b/lua/lvim/core/commands.lua
@@ -65,6 +65,12 @@ M.defaults = {
       print(require("lvim.utils.git").get_lvim_version())
     end,
   },
+  {
+    name = "LvimOpenlog",
+    fn = function()
+      vim.fn.execute("edit " .. require("lvim.core.log").get_path())
+    end,
+  },
 }
 
 function M.load(collection)


lua/lvim/core/which-key.lua
```diff
@@ -244,6 +244,7 @@ M.config = function()
         c = { "<cmd>Telescope colorscheme<cr>", "Colorscheme" },
         f = { "<cmd>Telescope find_files<cr>", "Find File" },
         h = { "<cmd>Telescope help_tags<cr>", "Find Help" },
+        H = { "<cmd>Telescope highlights<cr>", "Find highlight groups" },
```
kylo252, line 247:
```suggestion
```

lua/lvim/core/commands.lua
```diff
@@ -65,6 +65,18 @@ M.defaults = {
...
+    fn = function()
+      vim.fn.execute("edit " .. require("lvim.core.log").get_path())
+    end,
+  },
+  {
+    name = "LvimOpenlog",
+    fn = function()
+      require("lvim.core.terminal").toggle_log_view(vim.lsp.get_log_path())
+    end,
+  },
```
kylo252, line 79:
```suggestion
  {
    name = "LvimOpenlog",
    fn = function()
      vim.fn.execute("edit " .. require("lvim.core.log").get_path())
    end,
  },
```

kylo252: I don't think there's a need for multiple commands here, also note that ToggleTerm is optional.

lua/lvim/core/which-key.lua
danielo515: Was this removed from master? Or is it somewhere else?

danielo515:
Thank you for your review and your fixes @kylo252 !

kylo252:
I'm actually having some second thoughts about this, should we use Telescope for this instead, so basically allow viewing/editing _all_ the known logs?

danielo515:
> I'm actually having some second thoughts about this, should we use Telescope for this instead, so basically allow viewing/editing _all_ the known logs?

While I agree that such thing will provide a nicer experience when you are not sure which log to open, I still thing there is value in having a quick command like this to open it directly when you know what you want.

danielo515:
Why do you think there is not? I am following the same semantics as the
wich keys do. I usually prefer to open the log file, because it is easier
to read, while I think others will prefer to just open the log.
Or are you proposing to use arguments instead ?

PR #2722, opened 2022-06-16
fix: suggest the correct gentoo command

crystal-trd:
Fixed the Gentoo dependency install from `emerge install -y` (invalid command) to `emerge -v` (A working command with verbose output)


kylo252 (approved):
Thanks!

just fix the commit-lint issue, otherwise LGTM!

abzcoding:
using #2750 instead

PR #2743, opened 2022-06-21
fix: Fixing the gentoo dependencies install command

crystal-trd:
hope this pull request works


abzcoding:
using #2750 instead

PR #2749, opened 2022-06-22
fix: Move `pickers` key in telescope config

wren:
# Description

The `pickers` config key was erroneously nested under `defaults`, but should instead be a sibling of it. The correct structure is documented here:

https://github.com/nvim-telescope/telescope.nvim#telescope-setup-structure

I didn't see any related issues to this in your backlog after a quick search, so no issue is linked here.

## How Has This Been Tested?

I first noticed this problem because `hidden = true` wasn't showing hidden files for me for the `find_files` picker. With this PR, I can now change the `hidden` on and off as expected.

To confirm the problem (before this PR):
- run `:Telescope find_files`
- search for a hidden file
- confirm that hidden files don't show in list, even though `hidden = true` is set by default [here](https://github.com/LunarVim/LunarVim/blob/rolling/lua/lvim/core/telescope.lua#L75)

To confirm the fix:
- run the code from this PR
- run `:Telescope find_files`
- search for a hidden file
- confirm that the hidden files now show in list

### DIFF
--- a/lua/lvim/core/telescope.lua
+++ b/lua/lvim/core/telescope.lua
@@ -70,14 +70,14 @@ function M.config()
       borderchars = { "─", "│", "─", "│", "╭", "╮", "╯", "╰" },
       color_devicons = true,
       set_env = { ["COLORTERM"] = "truecolor" }, -- default = nil,
-      pickers = {
-        find_files = {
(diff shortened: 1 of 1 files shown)


abzcoding:
using #2750 instead

PR #2754, opened 2022-06-24
fix(lualine): set icon color according to the status of treesitter

emxxjnm:
# Description

set icon color according to the status of treesitter

fixes #(issue)

## How Has This Been Tested?

L

image

image

### DIFF
--- a/lua/lvim/core/lualine/components.lua
+++ b/lua/lvim/core/lualine/components.lua
@@ -70,13 +70,13 @@ return {
   },
   treesitter = {
     function()
-      local b = vim.api.nvim_get_current_buf()
-      if next(vim.treesitter.highlighter.active[b]) then
-        return ""
-      end
-      return ""
+      return ""
     end,
-    color = { fg = colors.green },
+    color = function()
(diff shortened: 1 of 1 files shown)


rebuilt (approved):
Tested.  It works.  It's nice.  

The only thing to note is that anyone debugging with an end user must be aware that the tree will show up now even when the treesitter parser for the language is not installed.   So you can't just ask if the user sees a tree at the bottom right of their screen.  They must ask if that tree is green or red.

kylo252:
maybe we should move the icon closer to the filetype?
```diff
diff --git a/lua/lvim/core/lualine/styles.lua b/lua/lvim/core/lualine/styles.lua
index 8991d9d..da37ec8 100644
--- a/lua/lvim/core/lualine/styles.lua
+++ b/lua/lvim/core/lualine/styles.lua
@@ -88,9 +88,9 @@ styles.lvim = {
     },
     lualine_x = {
       components.diagnostics,
-      components.treesitter,
       components.lsp,
       components.filetype,
+      components.treesitter,
     },
     lualine_y = {},
     lualine_z = {
```

PR #2762, opened 2022-06-25
fix: use pcall for setting up project.nvim 

lytex:
I set `lvim.builtin.project.active = false` in my configuration and caused a lot of errors because project_nvim was being called nonetheless

### DIFF
--- a/lua/lvim/core/project.lua
+++ b/lua/lvim/core/project.lua
@@ -41,7 +41,10 @@ function M.config()
 end
 
 function M.setup()
-  local project = require "project_nvim"
+  local status_ok, project = pcall(require, "project_nvim")
+  if not status_ok then
+    return
+  end
 
   project.setup(lvim.builtin.project)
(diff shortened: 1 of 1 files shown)


rebuilt:
Setting the 'active' configuration option to false doesn't actually uninstall a plugin.  Anything in the `start` folder will automatically be loaded.  This is the normal behavior of vim and neovim.  To actually uninstall a plugin, you must run a packer command like `:PackerSync`.   This will remove the plugin files from the start folder and the plugin won't run.  

Can you list the error message you were seeing so I can understand how this change fixed those errors?

lytex:
I didn't know it was necessary to call `:PackerSync`

Feel free to close this PR, I assumed every package was meant to be called using a protected call and somehow project_nvim was missing thus causing an error.

kylo252 (approved):
I think this is a good practice in general, considering how flaky packer can be. Thanks!

PR #2768, opened 2022-06-26
fix(comment): add extended mappings

bncpr:
# Description
Comment default config was set with extra = false with a comment
suggesting this was for `g>`, `g<`, etc. mappings. However, the `extra`
argument is for the `gco`, `gcO`, and `gcA` mappings which should be enabled
by default and, in my opinion are very useful. The extended mappings are
still disabled by default.

## How Has This Been Tested?

Try the mappings.


kylo252 (approved):
Sorry for the delay!

LGTM! 🚀

abzcoding (approved):

PR #2775, opened 2022-06-30
Fixed a typo in utils/installer/install.sh

Yazeed1s:


Yazeed1s:
I think this pr should be done in the rolling branch, closed

PR #2781, opened 2022-07-02
fix: adding max_jobs = 40; to :PackerSync

luxus:
fixing :PackerSync on large config from being stuck while syncing

# Description

add max_jobs to make large configs work. sometimes `packersync` stuck 
recommendation is from @abzcoding mentioned in https://github.com/abzcoding/lvim/issues/67#issuecomment-1172866361
i don't see any downsides.. 
from the official repo:
`max_jobs = nil, -- Limit the number of simultaneous jobs. nil means no limit`

fixes #(https://github.com/abzcoding/lvim/issues/67)

## How Has This Been Tested?

- Run command `:PackerSync`

### DIFF
--- a/lua/lvim/plugin-loader.lua
+++ b/lua/lvim/plugin-loader.lua
@@ -20,6 +20,7 @@ function plugin_loader.init(opts)
     package_root = opts.package_root or join_paths(vim.fn.stdpath "data", "site", "pack"),
     compile_path = compile_path,
     snapshot_path = snapshot_path,
+    max_jobs = 40,
     log = { level = "warn" },
     git = {
       clone_timeout = 300,


kylo252 (commented):
seems like the CI doesn't like it, you should probably set packer's log level to "debug" temporarily, just to see why it's hanging

you could also try a smaller number like 4 or 8

luxus:
@kylo252 okay I got it fixed by changing the number to nil

abzcoding:
@kylo252 @luxus the number should be higher for our CI ( or even nil) because our CI has a fast connection and also it is supposed to finish under 4 minutes ( or you'll get the timeout error)

but users with slower internet connections would need that number to be sth like 25 to make sure they won't get stuck

luxus:
> even nil

nil works 👯

kylo252 (commented):
does this actually affect anything tho? Packer already defines this as `nil`, see 
https://github.com/wbthomason/packer.nvim/blob/00ec5adef58c5ff9a07f11f45903b9dbbaa1b422/lua/packer.lua#L17

maybe I'm wrong, but I still think this is still the upper limit of how many `git` operations to spawn, so even if you have fast internet connection, I don't think you should go above the number of cores that you have.

judging by  https://github.com/wbthomason/packer.nvim/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+max_jobs+, I don't think anyone has a solution either 😢

luxus:
looks like its a mac problem :D
that's why it goes under the radar because you need a mac and a lot of plugins active
if i think about i never notice the problem on my linux installation

luxus:
after some testing.. (back on my mac) @kylo252 was right. looks like nil is the default value. I got it working locally on all my machines and on the CI with `max_jobs = 40`

luxus:
@abzcoding does it looks good to you?

abzcoding (approved):
it does look good, tyvm <3

danielo515:
Oh! I just opened an issue about this, this will fix my problem #2793

kylo252:
@luxus, thanks!

PR #2830, opened 2022-07-20
perf(cmp): remove redundant check for emmet-ls

rebuilt:
…etion.

# Description

Removes emmet logic from cmp.lua   We were performing a check to see if emmet was active on every completion request.  This slows down completion regardless of the filetype you're editing.  Additionally, the emmet logic no longer works and is no longer required to get emmet running in lunarvim.  

## How Has This Been Tested?

If a user wants to run emmet, they can configure it like so:

```lua
lvim.plugins = {
  {
    "aca/emmet-ls",
    ft = {
      "html",
      "typescriptreact",
      "javascriptreact",
      "css",
      "sass",
      "scss",
      "less",
    },
  },
  {
    "jackieaskins/cmp-emmet",
    run = "npm run release",
  },
}

table.insert(lvim.builtin.cmp.sources, { name = "emmet" })
lvim.builtin.cmp.formatting.source_names["emmet"] = "(Emmet)"
```

### DIFF
--- a/lua/lvim/core/cmp.lua
+++ b/lua/lvim/core/cmp.lua
@@ -22,20 +22,6 @@ local function feedkeys(key, mode)
 end
 M.methods.feedkeys = feedkeys
 
----checks if emmet_ls is available and active in the buffer
----@return boolean true if available, false otherwise
-local is_emmet_active = function()
-  local clients = vim.lsp.buf_get_clients()
-
-  for _, client in pairs(clients) do
-    if client.name == "emmet_ls" then
-      return true
-    end
-  end
-  return false
-end
-M.methods.is_emmet_active = is_emmet_active
-
 ---when inside a snippet, seeks to the nearest luasnip field if possible, and checks if it is jumpable
 ---@param dir number 1 for forward, -1 for backward; defaults to 1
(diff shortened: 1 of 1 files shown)


rebuilt:
@kylo252 or @abzcoding  Can you help me out with the stylua errors please?  
It's complaining about files I have not touched:  
```bash
 ./lua/lvim/core/telescope/custom-finders.lua
./lua/lvim/utils/git.lua
./tests/lvim/helpers.lua
```

I've tried many different variations of formatting and NOT formatting those files with stylua but I'm still getting errors.  

Running `stylua --config-path .stylua.toml .` from the root directory: `/home/nelson/.local/share/lunarvim/lvim` only changes one file `bufferline.lua`.   
Running `stylua --config-path .stylua.toml ./lua/lvim/core/telescope/custom-finders.lua`  from the root directory makes no formatting changes to that file.
Running `stylua custom-finders.lua` from the folder it's in DOES modify the formatting for that file but not in a way that matches the github CI check. 

I'm not sure how else to format those files so it passes CI

kylo252:
@rebuilt, you're probably using an older version of stylua 😅 

it should be fine for other files that you haven't changed, if you rebase on rolling

PR #2549, opened 2022-05-01
refactor: use api-autocmds for lsp functions

kylo252:
# Description

use the new autocmds API

PR #2544, opened 2022-04-30
feat(neovim): neovim 0.8 compatibility

abzcoding:
# Description

Make sure we are compatible with neovim 0.8

https://www.reddit.com/r/neovim/comments/u5si2w/breaking_changes_inbound_next_few_weeks_for/

PR #2851, opened 2022-07-25
fix(lvim/lsp/manager): make client_is_configured more reliable

lvimuser:
# Description

A server_name can be the same as its filetype, which will match the given pattern if you set any filetype autocmd for it. This might prevent a server from attaching.

Consider the html lsp. Its server_name is simply `html`. This might prevent a user from setting up plugins like `nvim-colorizer`, `nvim-ts-autotag` or using the emmet lsp in tandem.

Solution:

Use the lua API so we can match the specific command string, which contains only the manager's setup call, instead of trying to match/parse the result of `:autocmd`. 
## How Has This Been Tested?

- Set a 'html' filetype autocommand.
- Check `:LvimInfo`. 
- Lsp should be attached.

### DIFF
--- a/lua/lvim/lsp/manager.lua
+++ b/lua/lvim/lsp/manager.lua
@@ -37,9 +37,9 @@ end
 -- which seems to occur only when attaching to single-files
 local function client_is_configured(server_name, ft)
   ft = ft or vim.bo.filetype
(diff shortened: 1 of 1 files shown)


kylo252 (approved):
Thanks! 

this is much nicer with the lua API 🚀 

related https://github.com/LunarVim/LunarVim/pull/2544 and https://github.com/LunarVim/LunarVim/pull/2549 and

PR #2853, opened 2022-07-26
Update keymappings.lua

Juniar-Rakhman:
Window vertical resizing should be inverted

# Description

summary of the change

fixes #(issue)

## How Has This Been Tested?

- Run command `:mycommand`
- Check logs
- ...

### DIFF
--- a/lua/lvim/keymappings.lua
+++ b/lua/lvim/keymappings.lua
@@ -58,8 +58,8 @@ local defaults = {
     -- Resize with arrows
     ["<C-Up>"] = ":resize -2<CR>",
     ["<C-Down>"] = ":resize +2<CR>",
-    ["<C-Left>"] = ":vertical resize -2<CR>",
-    ["<C-Right>"] = ":vertical resize +2<CR>",
+    ["<C-Left>"] = ":vertical resize +2<CR>",
+    ["<C-Right>"] = ":vertical resize -2<CR>",
 
     -- Tab switch buffer
     ["<S-l>"] = ":BufferLineCycleNext<CR>",


rebuilt:
As it's currently written, `<C-left>` means resize this window to make it smaller and `<C-right>` is resize this window to make it bigger.  But I've always mistakenly thought of it as "move the border to the left or right".  Is that how you think of this keybinding as well?

With this change, the right-side window seems to have the intuitive behavior of resizing the correct way but then when you switch focus to the opposite window, it resizes in the `wrong` direction (assuming you think of moving the borders vs resizing the window).   Ideally, a change to this behavior would take into account the focused window and move the border accordingly. 
This page has a method of accomplishing that in vim: https://stackoverflow.com/questions/12129059/vim-vertical-resize-to-a-specific-direction

That said, I'm not sure if it's 'OK' to think of this keybinding as shifting the border vs resizing a window.  

The changes in this PR do seem like improvement because the correct behavior(of moving a border) is applied to the newest open buffer.  And that's the one most likely to be resized.

kylo252:
one way to do it is to check if the window is on the inside or the outside, see https://github.com/aserowy/tmux.nvim/blob/3b6a1ee21131513695ff2b4bc22f74e68a52cea4/lua/tmux/resize.lua#L31-L40

rafamadriz:
> The changes in this PR do seem like improvement because the correct behavior(of moving a border) is applied to the newest open buffer. And that's the one most likely to be resized.

What do you mean ? Is not applying to newest open buffer, it is just inverting the resizing directions.

rebuilt:
> What do you mean ? Is not applying to newest open buffer, it is just inverting the resizing directions.

Ok.  Sure.  But I don't understand why this is a clear win over the keybinding that's already in Lunarvim.   

When someone asks why we changed the mappings what do we tell them?

rafamadriz:
> Ok. Sure. But I don't understand why this is a clear win over the keybinding that's already in Lunarvim.

Oh I agree 100%, I didn't say this should be merged. There's no point on this PR.

PR #2854, opened 2022-07-26
fix(nvimtree): remove `indent_markers` icons trailing space

cpea2506:
# Description

As commit <https://github.com/kyazdani42/nvim-tree.lua/commit/9a02dedd92fad67b04b2a3fee2de20555956b089>, `nvim-tree` has removed trailing spaces from the default `indent_markers` icons. That would cause an issue when unneeded space was rendered in tree in `LunarVim`. Add all available icons setting, so that this will make more consistent.

# How Has This Been Tested?
- Before:

  before
- After:

  after
</div>

### DIFF
--- a/lua/lvim/core/nvimtree.lua
+++ b/lua/lvim/core/nvimtree.lua
@@ -68,9 +68,10 @@ function M.config()
         indent_markers = {
           enable = false,
           icons = {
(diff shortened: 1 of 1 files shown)


kylo252 (approved):
LGTM! 🚀

PR #2862, opened 2022-07-27
update keybinds to use defined leader key

nunocf:
Hey!
I love LunarVim but had to override the default buttons to use my leader key as it was bugging me a little bit as all the bindings were incorrect.

I'm not 100% sure this will work because I'm not too familiar with Lua, but the `lvim` table is global, right?

If this works, the loading screen should display the leader you set, instead of the harcoded SPC, but perhaps we need a special case for the SPC, as it probably wouldn't display.

# Description

summary of the change

fixes #(issue)

## How Has This Been Tested?

- Run command `:mycommand`
- Check logs
- ...

### DIFF
--- a/lua/lvim/core/alpha/dashboard.lua
+++ b/lua/lvim/core/alpha/dashboard.lua
@@ -51,13 +51,13 @@ function M.get_sections()
 
   local buttons = {
     entries = {
-      { "SPC f", "  Find File", "<CMD>Telescope find_files<CR>" },
-      { "SPC n", "  New File", "<CMD>ene!<CR>" },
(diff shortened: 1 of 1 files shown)


kylo252 (commented):
this will not work if `lvim.leader = " "` so you need to handle that case and re-assign the prefix to "SPC".

Also, I wasn't able to get it working, probably because `vim.g.mapleader` is not assigned just yet when this is configured.  

Regardless, this binding has nothing to do with the mapleader, you could just overwrite to whatever you want, e.g. https://github.com/kylo252/lvim/blob/952c97f217811e9a604ad9495e45383a57fcd02f/lua/user/dashboard.lua

hawkinchina:
Can I define another leader, just like vim  localheader = "," ?

PR #2870, opened 2022-07-30
feat(lvim/lsp): enable tailwindcss by default

lvimuser:
# Description

The first commit adds a helper function to check for dependencies in a `package.json` file.
The second enables the server and adds a custom provider for it, where we hijack `root_dir` to check if the server be attached.

Logic follows the vscode extension https://marketplace.visualstudio.com/items?itemName=bradlc.vscode-tailwindcss.

> In order for the extension to activate you must have tailwindcss installed and a Tailwind config file named tailwind.config.js or tailwind.config.cjs in your workspace.

The PR also paves the way for enabling angularls by default, if something similar to https://github.com/neovim/nvim-lspconfig/pull/1462 gets merged there (afaik, it's working).

## How Has This Been Tested?

- Open a tailwind project, server should be attached.
- Open a non tailwind project, server should not be attached for the same filetypes.

Filetypes: https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/tailwindcss.lua#L14-L65.

Not tested on monorepos.

--
The code is mostly written by @PasiBergman, with minor modifications.

Reference:
https://github.com/PasiBergman/lvim/blob/main/lua/utils/init.lua#L9
https://github.com/PasiBergman/lvim/blob/main/lua/lsp/tailwindcss.lua

I'm not sure how the licensing works though.

### DIFF
--- a/lua/lvim/lsp/providers/tailwindcss.lua
+++ b/lua/lvim/lsp/providers/tailwindcss.lua
@@ -1,0 +1,8 @@
+local opts = {
+  root_dir = function(fname)
+    local util = require "lspconfig/util"
+    return util.root_pattern("tailwind.config.js", "tailwind.config.cjs", "tailwind.js", "tailwind.cjs")(fname)
+  end,
+}
+
+return opts
--- a/lua/lvim/lsp/config.lua
+++ b/lua/lvim/lsp/config.lua
@@ -30,7 +30,6 @@ local skipped_servers = {
   "sqlls",
   "sqls",
   "stylelint_lsp",
-  "tailwindcss",
   "tflint",
   "svlangserver",
   "verible",


kylo252:
it seems tailwind requires a `tailwind.config` file, AFAICT, so let's split this into two separate PRs:
- enable tailwind when that file exists
- add a helper for detecting package.json content (btw, we should probably upstream this to null-ls)

lua/lvim/lsp/providers/tailwindcss.lua
```diff
@@ -0,0 +1,20 @@
...
+local opts = {
+  root_dir = function(fname)
+    if not project_has_tailwindcss_dependency() then
+      return
+    end
+
+    local default_root_dir = require("lspconfig").tailwindcss.document_config.default_config.root_dir
+    return default_root_dir(fname)
+  end,
+}
```
kylo252, line 18:
```suggestion
local opts = {
  root_dir = function(fname)
     return util.root_pattern('tailwind.config.js', 'tailwind.config.ts')(fname)
        or util.root_pattern('postcss.config.js', 'postcss.config.ts')(fname)
  end,
}
```

kylo252: so this would enough for the PR regarding tailwind

lvimuser: I don't agree...

Does postcss implies tailwind? (don't think so, but I don't know).

Also, doesn't handle '.cjs'. For reference, https://github.com/tailwindlabs/tailwindcss-intellisense/blob/master/packages/vscode-tailwindcss/src/extension.ts#L50.

And I haven't seen any '.ts' mentioned on tailwind docs (or the vscode ext, but I haven't checked it thoroughly), even for a typescript project.

From what I've tested, I think it's ok to not check if tailwind is installed in the project.

I'm deciding between
`
util.root_pattern("tailwind.config.js", "tailwind.config.cjs", "tailwind.js", "tailwind.cjs")(fname)
`
or
`util.root_pattern "tailwind.*js" (fname)`, what do you think?

kylo252: > Does postcss implies tailwind? (don't think so, but I don't know).

I don't know either 😄 

if it's not, then maybe we should fix that in lspconfig itself, see https://github.com/neovim/nvim-lspconfig/blob/80fddd28de5fe0642da7adee955f5762b933e730/lua/lspconfig/server_configurations/tailwindcss.lua#L106-L107

Uzaaft: The usage of postcss depends upon the framework being used and how one set up the project.
The `tailwindcss` cli,  does not use add postcss in the package.json file.

kylo252: @lvimuser, apply this to fix the tests
```diff
diff --git a/tests/specs/lsp_spec.lua b/tests/specs/lsp_spec.lua
index 7d9a338..01e5e1d 100644
--- a/tests/specs/lsp_spec.lua
+++ b/tests/specs/lsp_spec.lua
@@ -53,10 +53,14 @@ a.describe("lsp workflow", function()
   a.it("should only include one server per generated template", function()
     require("lvim.lsp").setup()
 
+    local allowed_dupes = { "tailwindcss" }
     for _, file in ipairs(vim.fn.glob(lvim.lsp.templates_dir .. "/*.lua", 1, 1)) do
       local content = {}
       for entry in io.lines(file) do
-        table.insert(content, entry)
+        local server_name = entry:match [[.*setup%("(.*)"%)]]
+        if not vim.tbl_contains(allowed_dupes, server_name) then
+          table.insert(content, server_name)
+        end
       end
       local err_msg = ""
       if #content > 1 then
```

kylo252 (approved):
Seems to work as intended, LGTM! 🚀

Issue #2872, opened 2022-08-01
Add default keybinding to cycle to next buffer


benediktms:
### Feature Description

A simple feature to add a default which-key keybinding to cycle to the next buffer, similar to the existing feature to cycle to the next buffer (essentially calling `:BufferLineCycleNext`)

### Describe the alternatives you have considered

The user can simply add a custom keybinding to do this but, providing a default which-key keybinding seems to make sense since we already have one for cycling to the previous buffer

### Support information

I am happy to submit a PR for this if is something that makes sense to add to the project

rebuilt:
Yeah, it seems like an oversight not to have it in whichkey.  If you submit that PR, I'd approve that change.   

As an aside, I'll say that the existing non-whichkey mapping of `<S-l>` and `<S-h>` override some basic vim keys and we should consider removing those.

benediktms:
Cool, I can also remove those in my PR

rebuilt:
Maybe separate PR's since the first change can be merged without controversy but the second might need more input from others.

kobzar:
SPS + b + b working by default
SPS + b  + j
The next one idea is close buffer by SPS + b + c not by SPS + c by default

benediktms:
@kobzar Sorry the description was wrong I meant cycling to the next buffer. `<leader>bb` works to cycle to the previous buffer. I propose adding a similar keybinding (`<leader>bn`) to cycle to the next buffer.

kobzar:
Yep. It must be by default.

cpea2506:
> The next one idea is close buffer by SPS + b + c not by SPS + c by default

Agree! It would make sense if close buffer is in buffer section.

miwojc:
Why not?
next buffer: SPS b n
previous buffer: SPS b p

MycrofD:
@miwojc perhaps because the keys `b` and `n` are right next to each other and it is fast to switch between them. It is even faster to just do `SPS b b` and cycle through buffers.

PR #2874, opened 2022-08-01
refactor(keymaps)!: don't hijack H/L by default

benediktms:
# Description

Buffer cycling keymaps should be delegated to which-key. These keymaps fulfil the same role and are therefore redunant.

part of https://github.com/LunarVim/LunarVim/issues/2872#issuecomment-1201616659

## How Has This Been Tested?

- Buffer cycling is implemented and tested here https://github.com/LunarVim/LunarVim/pull/2873/commits/5f9f711c93c79ef63410390f8b47a0fc3cf0ea5f

### DIFF
--- a/lua/lvim/keymappings.lua
+++ b/lua/lvim/keymappings.lua
@@ -61,10 +61,6 @@ local defaults = {
     ["<C-Left>"] = ":vertical resize -2<CR>",
     ["<C-Right>"] = ":vertical resize +2<CR>",
 
-    -- Tab switch buffer
-    ["<S-l>"] = ":BufferLineCycleNext<CR>",
-    ["<S-h>"] = ":BufferLineCyclePrev<CR>",
-
     -- Move current line / block with Alt-j/k a la vscode.
     ["<A-j>"] = ":m .+1<CR>==",
     ["<A-k>"] = ":m .-2<CR>==",
--- a/utils/installer/config.example.lua
+++ b/utils/installer/config.example.lua
@@ -19,6 +19,8 @@ lvim.colorscheme = "onedarker"
 lvim.leader = "space"
 -- add your own keymapping
 lvim.keys.normal_mode["<C-s>"] = ":w<cr>"
+-- lvim.keys.normal_mode["<S-l>"] = ":BufferLineCycleNext<CR>"
+-- lvim.keys.normal_mode["<S-h>"] = ":BufferLineCyclePrev<CR>"
 -- unmap a default keymapping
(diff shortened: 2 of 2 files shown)


rebuilt (approved):

rebuilt:
Uppercase `H` and `L` are very basic movement keys.   Overriding these seems like a mistake.  

Even though this pr deletes the `H` and `L` bindings, LunarVim still provides whichkey bindings to cycle through buffers so users will still have that built-in.  

That said, I recognize that these keybindings have been in LunarVim for a while and some people may prefer to keep these bindings in.  

Anyone that still wants the `H` and `L` keybindings can still add them to their config

```
	lvim.keys.normal_mode["<S-l>"] = ":BufferLineCycleNext<CR>"
	lvim.keys.normal_mode["<S-h>"] = ":BufferLineCyclePrev<CR>"
```

Or they can specify other bindings, eg

```
	vim.api.nvim_set_keymap("n", "<A-l>", ":BufferLineCycleNext<CR>", { noremap = true, silent = true })
	vim.api.nvim_set_keymap("n", "<A-h>", ":BufferLineCyclePrev<CR>", { noremap = true, silent = true })

	lvim.keys.normal_mode["<A-Right>"] = ":BufferLineCycleNext<CR>"
	lvim.keys.normal_mode["<A-Left>"] = ":BufferLineCyclePrev<CR>"

```

kylo252 (commented):
How about adding them to `config.example.lua` instead? that way it's also easily searchable for anyone than having to refer to this PR.

kylo252 (approved):
LGTM! 🚀

danielo515:
Removing this keys that have been this long in LunarVim it's a major breaking change.
People (like me) is already used to this, and I would say that they are more convenient than what they do by default, and way way more convenient than having to use which-key for something you do as often as changing to next buffer.
You only have to give a quick read to the lunar vim chat to see that people love this bindings and that they are used to them.
This will frustrate a lot of people the day they update lunar vim to realise that their usual workflow does not work, and when something is so deep into your muscular memory it is hard to understand why. Then people will have to waste some time to find this PR, or open a new issue, or ask for help in the chat.
Overriding vim defaults is a sensitive thing, but once done you can not just remove it because it will became the new default of a lot of people, and removing it will do as much harm as it did when the decision of overriding it was taken.
LunarVim is not plain vim, so removing this just because "is not vim defaults" seems like a elitist choice. Vim has many key-bindings, if we were going to keep them all by default, then LunarVim will not be adding as much value.
Just my two cents

PR #2880, opened 2022-08-05
refactor!: migrate to mason.nvim

### LINKED ISSUE #2850: migrate to mason.nvim
### Discussed in https://github.com/LunarVim/LunarVim/discussions/2848

<div type='discussions-op-text'>

<sup>Originally posted by **Aceticia** July 24, 2022</sup>
According to the post [here](https://www.reddit.com/r/neovim/comments/w6w5ij/introducing_masonnvim/?utm_source=share&utm_medium=ios_app&utm_name=iossmf), lsp-installer will no longer be maintained. Mason.nvim seems to be the canon successor to it, supporting linters, formatters, dap and lsp. Should lunarvim consider transitioning to it?</div>

kylo252:
# Description

> `mason.nvim` is the next generation version of `nvim-lsp-installer`. It builds on
> top of the very same foundation as `nvim-lsp-installer`, but with a majority of
> internals refactored to improve extensibility and testability.

fixes #2850

### TODOs

- [x] replace `nvim-lsp-installer` completely
- [x] ensure that template generation still works
- [x] fix data-race on server install (on_server_ready has been removed..)
- [ ] make sure no deprecated/outdated commands exist (how to replace `LspInstallInfo`?)
- [ ] add null-ls integration (requires a separate PR)
- [ ] add DAP integration (requires a separate PR)

### DIFF
--- a/lua/lvim/config/init.lua
+++ b/lua/lvim/config/init.lua
@@ -99,6 +99,13 @@ local function handle_deprecated_settings()
       "Use vim.api.nvim_create_autocmd instead or check LunarVim#2592 to learn about the new syntax"
     )
   end
+
+  if lvim.lsp.automatic_servers_installation then
+    deprecation_notice(
+      "lvim.lsp.automatic_servers_installation",
+      "Use `lvim.lsp.installer.setup.automatic_installation` instead"
+    )
+  end
 end
 
 --- Override the configuration with a user provided one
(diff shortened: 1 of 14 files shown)


luxus (approved):

amoebaguy:
This completely broke my rust-tools configuration which is 1:1 from https://www.lunarvim.org/languages/rust.html#extra-plugins.

Are those random breaking changes supposed to happen in rolling? Then a more suited branch name would be  'unstable' instead.

kylo252:
> This completely broke my rust-tools configuration which is 1:1 from https://www.lunarvim.org/languages/rust.html#extra-plugins.

just remove all references to lsp-installer and it should work.

> Are those random breaking changes supposed to happen in rolling? Then a more suited branch name would be  'unstable' instead.

there's a reason why the master branch exists along with its version releases.

Issue #2900, opened 2022-08-11
LvimInfo fails to run


danielo515:
### Problem description

I just updated to the latest rolling, and when I do LvimInfo it reports an error.

### LunarVim version

rolling-9caa457

### Neovim version (>= 0.7.2)

NVIM v0.7.0-dev+882-g9304ee387

### Operating system/version

macOS 12.4

### support info

```
14:47:02 [WARN ] lvim: "Invalid configuration: ...ielo/.local/share/lunarvim/lvim/lua/lvim/lsp/manager.lua:91: module 'mason-lspconfig.mappings.server' not found:\n\tno field package.preload['mason-lspconfig.mappings.server']\n\tno file './mason-lspconfig/mappings/server.lua'\n\tno file '/usr/local/Cellar/luajit-openresty/2.1-20220411/share/luajit-2.1.0-beta3/mason-lspconfig/mappings/server.lua'\n\tno file '/usr/local/share/lua/5.1/mason-lspconfig/mappings/server.lua'\n\tno file '/usr/local/share/lua/5.1/mason-lspconfig/mappings/server/init.lua'\n\tno file '/usr/local/Cellar/luajit-openresty/2.1-20220411/share/lua/5.1/mason-lspconfig/mappings/server.lua'\n\tno file '/usr/local/Cellar/luajit-openresty/2.1-20220411/share/lua/5.1/mason-lspconfig/mappings/server/init.lua'\n\tno file './mason-lspconfig/mappings/server.so'\n\tno file '/usr/local/lib/lua/5.1/mason-lspconfig/mappings/server.so'\n\tno file '/usr/local/Cellar/luajit-openresty/2.1-20220411/lib/lua/5.1/mason-lspconfig/mappings/server.so'\n\tno file '/usr/local/lib/lua/5.1/loadall.so'\n\tno file './mason-lspconfig.so'\n\tno file '/usr/local/lib/lua/5.1/mason-lspconfig.so'\n\tno file '/usr/local/Cellar/luajit-openresty/2.1-20220411/lib/lua/5.1/mason-lspconfig.so'\n\tno file '/usr/local/lib/lua/5.1/loadall.so'" file="init.lua", line=119
"chezmoi/dot_config/lvim/config.lua" 61L, 1992B
"chezmoi/dot_config/lvim/lua/user/keymaps.lua" 70L, 3450B
E5108: Error executing lua vim/shared.lua:0: t: expected table, got string
stack traceback:
[...]

PR #2901, opened 2022-08-11
fix: LvimInfo dont working #2900

Nojipiz:
Add extra validation to table content
# Description

summary of the change

fixes issue #2900 

## How Has This Been Tested?

- Run command `:LvimInfo`

### DIFF
--- a/lua/lvim/core/info.lua
+++ b/lua/lvim/core/info.lua
@@ -112,6 +112,9 @@ local function make_auto_lsp_info(ft)
   end
 
   local supported = lsp_utils.get_supported_servers(ft)
+  if supported == nil then
+    return info_lines
+  end
   local skipped = vim.tbl_filter(function(name)
     return vim.tbl_contains(supported, name)
   end, skipped_servers)


lua/lvim/core/info.lua
```diff
@@ -112,6 +112,9 @@ local function make_auto_lsp_info(ft)
   end
 
   local supported = lsp_utils.get_supported_servers(ft)
+  if supported == nil then
+    return info_lines
+  end
```
kylo252, line 117: I'm not sure this fixes anything, since the function shouldn't return `nil` in the first place

https://github.com/LunarVim/LunarVim/blob/9caa457810a26369f848d4f90f6e0ec84e0048e0/lua/lvim/lsp/utils.lua#L67-L72

kylo252 (commented):
please provide some steps to reproduce the original issue we're trying to solve here

lua/lvim/core/info.lua
Nojipiz: Ohh sure, my bad.
I was running the stable release not the rolling and there are differences!

PR #2902, opened 2022-08-11
docs: update example config for automatic lsp installation

vonPB:
# Description

updated example configs

fixes deprecation notice on launch
image

## How Has This Been Tested?

- open LunarVim with example config -> no notice

## Another note:
The deprecation error is only thrown when 
`lvim.lsp.installer.setup.automatic_installation` is set to `true`.
Shouldn't the user be notified regardless of the value it is set to?

https://github.com/LunarVim/LunarVim/blob/eefde00ae80f91ecf88a93e869e346fdd04c7ba4/lua/lvim/config/init.lua#L103

### DIFF
--- a/utils/installer/config.example.lua
+++ b/utils/installer/config.example.lua
@@ -99,7 +99,7 @@ lvim.builtin.treesitter.highlight.enabled = true
 -- }
 
 -- ---@usage disable automatic installation of servers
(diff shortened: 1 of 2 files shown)


cpea2506:
fixes since LunarVim@7df4773b0bc293d50fe9979505db8db246a61607

PR #2910, opened 2022-08-12
refactor: smaller timeout for packer

danielo515:
I don't think anyone wants to wait 5 minutes for a git clone to fail. I think one minute is more than enough

# Description

Reduce the timeout of packer for git clone

## How Has This Been Tested?

I don't think it is required, but CI should test this already

### DIFF
--- a/lua/lvim/plugin-loader.lua
+++ b/lua/lvim/plugin-loader.lua
@@ -23,7 +23,7 @@ function plugin_loader.init(opts)
     max_jobs = 100,
     log = { level = "warn" },
     git = {
-      clone_timeout = 300,
+      clone_timeout = 120,
     },
     display = {
       open_fn = function()


danielo515:
Does anyone have any opinion on this?

danielo515:
@kylo252 anything against?

kylo252 (commented):
maybe 120s as a compromise? we need to strike a balance against too many false positives for users with bad connections

lua/lvim/plugin-loader.lua
```diff
@@ -23,7 +23,7 @@ function plugin_loader.init(opts)
     max_jobs = 40,
     log = { level = "warn" },
     git = {
-      clone_timeout = 300,
+      clone_timeout = 60,
```
danielo515, line 26:
```suggestion
      clone_timeout = 120,
```

danielo515:
> maybe 120s as a compromise? we need to strike a balance against too many false positives for users with bad connections

Done!

kylo252 (approved):
Thanks!

PR #2921, opened 2022-08-18
fix(statusline): display null-ls linters properly

choplin:
# Description

null-ls linters may have either `DIAGNOSTIC`, `DIAGNOSTIC_ON_OPEN`, or `DIAGNOSTIC_ON_SAVE` as its method parameter. e.g., https://github.com/jose-elias-alvarez/null-ls.nvim/blob/8914051a3d399e9715833ad76bbf5fe69ea7faf0/lua/null-ls/builtins/diagnostics/buf.lua#L12

Currently, only linters with `DIAGNOSTIC` are shown in the lsp component on the statusline regardless of whether the linters are registered and work properly.

This is due to `lvim.lsp.null-ls.linters.list_registers` returning only linters with `DIAGNOSTIC`.

This PR makes that function retrieve the linters with all three method types and return the merged list.

## How Has This Been Tested?

- Register `buf` null-ls linter, which is set to `DIAGNOSTIC_ON_SAVE`, open a supported file (i.e., proto file), then check if `buf` linter is shown in the statusline.

### DIFF
--- a/lua/lvim/lsp/null-ls/linters.lua
+++ b/lua/lvim/lsp/null-ls/linters.lua
@@ -6,9 +6,19 @@ local null_ls = require "null-ls"
 local services = require "lvim.lsp.null-ls.services"
 local method = null_ls.methods.DIAGNOSTICS
 
+local alternative_methods = {
+  null_ls.methods.DIAGNOSTICS,
+  null_ls.methods.DIAGNOSTICS_ON_OPEN,
+  null_ls.methods.DIAGNOSTICS_ON_SAVE,
+}
+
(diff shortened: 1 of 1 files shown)


choplin:
null-ls formatters may have an alternative method as `RANGE_FORMATTING` as well. But, it seems that no built-in formatters of null-ls have only `RANGE_FORMATTING` method (all formatters that have `RANGE_FORMATTING` also have `FORMATTING` method). So, I didn't make changes to formatters.

kylo252 (approved):

PR #2923, opened 2022-08-19
chore: remove unnecessary code

lvimuser:
# Description

Just a simple cleanup. The deepcopy can be removed, and the component logic was duplicated.

### DIFF
--- a/lua/lvim/core/lualine/components.lua
+++ b/lua/lvim/core/lualine/components.lua
@@ -47,15 +47,10 @@ return {
     function()
       local utils = require "lvim.core.lualine.utils"
       if vim.bo.filetype == "python" then
-        local venv = os.getenv "CONDA_DEFAULT_ENV"
+        local venv = os.getenv "CONDA_DEFAULT_ENV" or os.getenv "VIRTUAL_ENV"
         if venv then
           return string.format("  (%s)", utils.env_cleanup(venv))
         end
-        venv = os.getenv "VIRTUAL_ENV"
-        if venv then
-          return string.format("  (%s)", utils.env_cleanup(venv))
-        end
-        return ""
       end
       return ""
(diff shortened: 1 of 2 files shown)


lua/lvim/lsp/config.lua
```diff
@@ -64,12 +64,11 @@ return {
       header = "",
       prefix = "",
       format = function(d)
-        local t = vim.deepcopy(d)
```
kylo252, line 67: the `deepcopy()` was necessary to avoid messing up the original diagnostics, are you sure it's not needed anymore?

lvimuser: we're not mutating anymore though, instead of changing the field we just simply return.

kylo252: nice catch!

lua/lvim/core/lualine/components.lua
```diff
@@ -47,15 +47,10 @@ return {
     function()
       local utils = require "lvim.core.lualine.utils"
       if vim.bo.filetype == "python" then
-        local venv = os.getenv "CONDA_DEFAULT_ENV"
+        local venv = os.getenv "CONDA_DEFAULT_ENV" or os.getenv "VIRTUAL_ENV"
```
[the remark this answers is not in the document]
lvimuser: Does `vim.env` keeps track of changes? (e.g. you may change venv without restarting the editor...). I'd rather not change it (or leave it for another PR).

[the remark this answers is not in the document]
kylo252: nvm, `vim.env` is implemented with `vim.fn.getenv` so it's probably slower

kylo252 (approved):
LGTM! 🚀

PR #2933, opened 2022-08-22
refactor(config): tell the users which config file is exactly failing

danielo515:
# Description

I find frustrating that, when something fails, I just get the error message and a log message pointing to init.lua, which is not very helpful. This adds the path of the failing config file to be super explicit to the user

fixes #(issue)

## How Has This Been Tested?

- Run lvim with anything wrong at userland config
- Check logs and see the file that fails is being pointed out

### DIFF
--- a/lua/lvim/config/init.lua
+++ b/lua/lvim/config/init.lua
@@ -116,7 +116,7 @@ function M:load(config_path)
   local ok, err = pcall(dofile, config_path)
   if not ok then
     if utils.is_file(user_config_file) then
-      Log:warn("Invalid configuration: " .. err)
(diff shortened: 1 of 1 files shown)


kylo252 (commented):
Please provide a complete example of before and after, since I'm not sure I follow with what's missing.

Try adding some code in `config.lua` and save the file, you'll see a warning message specifying even line, e.g.

```log
2022-08-22T15:43:20 lvim  WARN "Invalid configuration: /home/hatsu/.config/lvim/config.lua:86: '=' expected near 'table'"
```

image

danielo515:
Yes, you're right. I think I was facing a corner case where it was erroring at compile time in the plugin instantiation. I will try to reproduce, but for now it's probably fine to close this PR

PR #2926, opened 2022-08-20
fix: update key bindings for comment.nvim to use new api

EdRW:
# Description

The new update to comment.nvim plugin has [breaking changes to its api](https://github.com/numToStr/Comment.nvim/pull/183). This has caused the comment ( \ ) keybinding to break.

This is a minor change to fix the keybinding that were broken by switching to the new comment.nvim api

This PR solves the issue caused by  #2895 

## How Has This Been Tested?

- toggle line comment in normal mode using `space -> \`
- toggle several line comments in visual mode using `space -> \`

PR #2947, opened 2022-08-25
fix(comment): fix current linewise comment

choplin:
# Description

Currently, toggling linewise comment on the current line, which is mapped to `<Leader>/` as default, does not work properly. This is due to the wrong name of <Plug> key name introduced in https://github.com/LunarVim/LunarVim/pull/2926.

The correct key name is as here in the original repository.
https://github.com/rafamadriz/Comment.nvim/blob/9b76787e273567c0e3027304bd16ffedc751c04c/after/plugin/Comment.lua#L20

## How Has This Been Tested?

Fixed a key name, then check if `<Leader>/` toggle the comment on the current line.


lua/lvim/core/which-key.lua
```diff
@@ -82,7 +82,7 @@ M.config = function()
       [";"] = { "<cmd>Alpha<CR>", "Dashboard" },
       ["w"] = { "<cmd>w!<CR>", "Save" },
       ["q"] = { "<cmd>lua require('lvim.utils.functions').smart_quit()<CR>", "Quit" },
-      ["/"] = { "<Plug>(comment_toggle_linewise_current)", "Comment toggle current line" },
+      ["/"] = { "<Plug>(comment_toggle_current_linewise)", "Comment toggle current line" },
```
kylo252, line 85: this is the _deprecated_ function between the two, did you run `:PackerSync` ?

choplin: Ah, my bad. I misunderstood. It works properly after `:PackerSync`. Sorry!

PR #2979, opened 2022-09-01
fix(cmp): do not mutate the original confirm_opts on CR

danielo515:
# Description

The CR mapping of cmp overrides the provided confirm_opts, making this change permanent. This PR avoids that.

## How Has This Been Tested?

- Test CMP CR to select a mapping


kylo252 (approved):
LGTM! 🚀

PR #2982, opened 2022-09-01
feat: warn user when setting un-installed colorscheme

### LINKED ISSUE #2856: Colorscheme cause an error
The following two lines are causing me a bit of headaches. Apparently, everything work, but those lines make nvim output the warning error `"problems detected while loading plugins' configurations"`. Nothing bad happens, but I have this message.

https://github.com/LunarVim/LunarVim/blob/6ab3e8a73920e3d2e02e57f7dd3a1f3d8e54ee63/lua/lvim/plugin-loader.lua#L109-L111

I think there's a problem of cache, but I don't know how to delete it.

cpea2506:
# Description

The two colorscheme lines below inside `xpcall` function couldn't take any advantage of the `debug.traceback` function.

https://github.com/LunarVim/LunarVim/blob/6ab3e8a73920e3d2e02e57f7dd3a1f3d8e54ee63/lua/lvim/plugin-loader.lua#L109-L111

So when users set colorscheme that hasn't been installed, they will be warned with

> "problems detected while loading plugins' configurations" 

without knowing what happen inside their config. We should move it outside and handle its own error instead.

~~Also clear core plugins' caches on saving `config.lua` file. So that the change will instantly load without exiting and entering `lvim` again.~~

fixes #2856

## How Has This Been Tested?

- Set `lvim.colorscheme="something"`
- Should be warned with 

Screenshot 2022-09-01 at 22 13 13

### DIFF
--- a/lua/lvim/plugin-loader.lua
+++ b/lua/lvim/plugin-loader.lua
@@ -106,13 +106,19 @@ function plugin_loader.load(configurations)
         end
       end
     end)
-    -- colorscheme must get called after plugins are loaded or it will break new installs.
-    vim.g.colors_name = lvim.colorscheme
-    vim.cmd("colorscheme " .. lvim.colorscheme)
   end, debug.traceback)
+
   if not status_ok then
     Log:warn "problems detected while loading plugins' configurations"
     Log:trace(debug.traceback())
+  end
+
+  -- colorscheme must get called after plugins are loaded or it will break new installs.
+  vim.g.colors_name = lvim.colorscheme
+  local set_colorscheme_ok, _ = pcall(vim.cmd, "colorscheme " .. lvim.colorscheme)
+
+  if not set_colorscheme_ok then
+    Log:warn(string.format("problem detected: could not find '%s' colorscheme", lvim.colorscheme))
   end
 end
 
--- a/lua/lvim/plugin-loader.lua
+++ b/lua/lvim/plugin-loader.lua
@@ -113,13 +113,14 @@ function plugin_loader.load(configurations)
     Log:trace(debug.traceback())
   end
 
-  -- colorscheme must get called after plugins are loaded or it will break new installs.
-  vim.g.colors_name = lvim.colorscheme
-  local set_colorscheme_ok, _ = pcall(vim.cmd, "colorscheme " .. lvim.colorscheme)
-
-  if not set_colorscheme_ok then
-    Log:warn(string.format("problem detected: could not find '%s' colorscheme", lvim.colorscheme))
+  -- ref: https://github.com/neovim/neovim/issues/18201#issuecomment-1104754564
+  local colors = vim.api.nvim_get_runtime_file(("colors/%s*"):format(lvim.colorscheme), false)
+  if #colors == 0 then
+    Log:warn(string.format("Could not find '%s' colorscheme", lvim.colorscheme))
+    return
   end
+  vim.g.colors_name = lvim.colorscheme
+  vim.cmd("colorscheme " .. lvim.colorscheme)
 end
 
 function plugin_loader.get_core_plugins()
--- a/lua/lvim/plugin-loader.lua
+++ b/lua/lvim/plugin-loader.lua
@@ -112,15 +112,6 @@ function plugin_loader.load(configurations)
     Log:warn "problems detected while loading plugins' configurations"
     Log:trace(debug.traceback())
   end
-
-  -- ref: https://github.com/neovim/neovim/issues/18201#issuecomment-1104754564
-  local colors = vim.api.nvim_get_runtime_file(("colors/%s.*"):format(lvim.colorscheme), false)
-  if #colors == 0 then
-    Log:warn(string.format("Could not find '%s' colorscheme", lvim.colorscheme))
-    return
-  end
-  vim.g.colors_name = lvim.colorscheme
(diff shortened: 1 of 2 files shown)


lua/lvim/plugin-loader.lua
```diff
@@ -106,14 +106,20 @@ function plugin_loader.load(configurations)
...
     Log:trace(debug.traceback())
   end
+
+  -- colorscheme must get called after plugins are loaded or it will break new installs.
+  vim.g.colors_name = lvim.colorscheme
+  local set_colorscheme_ok, _ = pcall(vim.cmd, "colorscheme " .. lvim.colorscheme)
+
+  if not set_colorscheme_ok then
+    Log:warn(string.format("problem detected: could not find '%s' colorscheme", lvim.colorscheme))
+  end
```
kylo252, line 122:
```suggestion
  -- ref: https://github.com/neovim/neovim/issues/18201#issuecomment-1104754564
  local colors = vim.api.nvim_get_runtime_file(("colors/%s*"):format(lvim.colorscheme), true)
  if #colors == 0 then
    Log:warn(string.format("problem detected: could not find '%s' colorscheme", lvim.colorscheme))
    return
  end
  vim.g.colors_name = lvim.colorscheme
  vim.cmd("colorscheme " .. lvim.colorscheme)
```

lvimuser: We might not need to pass `true`? if there's one, that's enough, we don't need to keep searching

For reference
```lua
nvim_get_runtime_file({name}, {all})                 *nvim_get_runtime_file()*
    Find files in runtime directories

    'name' can contain wildcards. For example
    nvim_get_runtime_file("colors/*.vim", true) will return all color scheme
    files.

    It is not an error to not find any files. An empty array is returned then.

    Parameters:  
        {name}  pattern of files to search for
        {all}   whether to return all matches or only the first
```

lua/lvim/plugin-loader.lua
```diff
@@ -106,14 +106,20 @@ function plugin_loader.load(configurations)
...
     Log:warn "problems detected while loading plugins' configurations"
     Log:trace(debug.traceback())
   end
+
+  -- colorscheme must get called after plugins are loaded or it will break new installs.
+  vim.g.colors_name = lvim.colorscheme
+  local set_colorscheme_ok, _ = pcall(vim.cmd, "colorscheme " .. lvim.colorscheme)
+
+  if not set_colorscheme_ok then
+    Log:warn(string.format("problem detected: could not find '%s' colorscheme", lvim.colorscheme))
```
lvimuser, line 121:
```suggestion
    Log:warn(string.format("Could not find '%s' colorscheme", lvim.colorscheme))
```

lua/lvim/plugin-loader.lua
cpea2506: I wonder why we need to set `vim.g.colors_name`. When most colorschemes will do it for us
https://github.com/LunarVim/LunarVim/blob/6ab3e8a73920e3d2e02e57f7dd3a1f3d8e54ee63/lua/lvim/plugin-loader.lua#L110

lua/lvim/plugin-loader.lua
```diff
@@ -106,14 +106,21 @@ function plugin_loader.load(configurations)
...
   end
+
+  -- ref: https://github.com/neovim/neovim/issues/18201#issuecomment-1104754564
+  local colors = vim.api.nvim_get_runtime_file(("colors/%s.*"):format(lvim.colorscheme), false)
+  if #colors == 0 then
+    Log:warn(string.format("Could not find '%s' colorscheme", lvim.colorscheme))
+    return
+  end
+  vim.g.colors_name = lvim.colorscheme
+  vim.cmd("colorscheme " .. lvim.colorscheme)
```
kylo252, line 123: we should now move this into `theme.lua` since #3133 got merged

```diff
diff --git a/lua/lvim/core/theme.lua b/lua/lvim/core/theme.lua
index 394963a0..7a4c469b 100644
--- a/lua/lvim/core/theme.lua
+++ b/lua/lvim/core/theme.lua
@@ -1,5 +1,6 @@
 local M = {}
 
+local Log = require "lvim.core.log"
 M.config = function()
   lvim.builtin.theme = {
     name = "tokyonight",
@@ -85,17 +86,24 @@ end
 M.setup = function()
   -- avoid running in headless mode since it's harder to detect failures
   if #vim.api.nvim_list_uis() == 0 then
-    local Log = require "lvim.core.log"
     Log:debug "headless mode detected, skipping running setup for lualine"
     return
   end
 
   local status_ok, theme = pcall(require, "tokyonight")
-  if not status_ok then
+  if status_ok and theme then
+    theme.setup(lvim.builtin.theme.options)
+  end
+
+  -- ref: https://github.com/neovim/neovim/issues/18201#issuecomment-1104754564
+  local colors = vim.api.nvim_get_runtime_file(("colors/%s.*"):format(lvim.colorscheme), false)
+  if #colors == 0 then
+    Log:warn(string.format("Could not find '%s' colorscheme", lvim.colorscheme))
     return
   end
 
-  theme.setup(lvim.builtin.theme.options)
+  vim.g.colors_name = lvim.colorscheme
+  vim.cmd("colorscheme " .. lvim.colorscheme)
 
   require("lvim.core.lualine").setup()
```

kylo252 (approved):

ChristianChiarulli (approved):

PR #2984, opened 2022-09-03
feat(telecope): set show_untracked by default

pschmitt:
# Description

- add the ability to display untracked files by default when invoking `find_project_files`
- can be turned off by setting the `show_untracked` to `false`.

## How Has This Been Tested?

```lua
-- show untracked files
require('lvim.core.telescope.custom-finders').find_project_files()

-- skip untracked files
require('lvim.core.telescope.custom-finders').find_project_files({show_untracked = false})
```

### DIFF
--- a/lua/lvim/core/telescope.lua
+++ b/lua/lvim/core/telescope.lua
@@ -79,6 +79,9 @@ function M.config()
         --@usage don't include the filename in the search results
         only_sort_text = true,
       },
+      git_files = {
+        show_untracked = true,
+      },
     },
     extensions = {
       fzf = {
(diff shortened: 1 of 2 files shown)


kylo252 (commented):
I'm in favor of making this a default behavior actually! in that case, we'd have it set in telescope's config table, and it could still be overridden.

cc: @abzcoding, @rebuilt

chris-ricketts:
I agree this should be the default. I'm actually pretty sure it used to be for the default `<space>-f` "find files" keybinding but it was changed at some point.

kylo252 (approved):

lvimuser (approved):

abzcoding (approved):
LGTM 🚀

rebuilt (approved):

PR #3004, opened 2022-09-10
fix(lsp/utils): do not register duplicate autocommands

lvimuser:
Checks for existence of buffer local `documentHighlight/codeLens` autocommands before registering another one.

## How Has This Been Tested?

- Run `:e` multiple times. 
- There should be only one `CursorHold` autocmd for codelens/documentHighlight for the current buffer.

### DIFF
--- a/lua/lvim/lsp/utils.lua
+++ b/lua/lvim/lsp/utils.lua
@@ -88,19 +88,27 @@ function M.setup_document_highlight(client, bufnr)
   if not status_ok or not highlight_supported then
     return
   end
-  local augroup_exist, _ = pcall(vim.api.nvim_get_autocmds, {
-    group = "lsp_document_highlight",
+  local group = "lsp_document_highlight"
+  local hl_events = { "CursorHold", "CursorHoldI" }
+
+  local ok, hl_autocmds = pcall(vim.api.nvim_get_autocmds, {
+    group = group,
+    buffer = bufnr,
+    event = hl_events,
   })
-  if not augroup_exist then
-    vim.api.nvim_create_augroup("lsp_document_highlight", {})
+
+  if ok and #hl_autocmds > 0 then
+    return
   end
-  vim.api.nvim_create_autocmd({ "CursorHold", "CursorHoldI" }, {
-    group = "lsp_document_highlight",
+
+  vim.api.nvim_create_augroup(group, { clear = false })
+  vim.api.nvim_create_autocmd(hl_events, {
+    group = group,
     buffer = bufnr,
     callback = vim.lsp.buf.document_highlight,
   })
   vim.api.nvim_create_autocmd("CursorMoved", {
-    group = "lsp_document_highlight",
+    group = group,
     buffer = bufnr,
     callback = vim.lsp.buf.clear_references,
   })
@@ -113,14 +121,20 @@ function M.setup_codelens_refresh(client, bufnr)
   if not status_ok or not codelens_supported then
     return
   end
-  local augroup_exist, _ = pcall(vim.api.nvim_get_autocmds, {
-    group = "lsp_code_lens_refresh",
+  local group = "lsp_code_lens_refresh"
+  local cl_events = { "BufEnter", "InsertLeave" }
+  local ok, cl_autocmds = pcall(vim.api.nvim_get_autocmds, {
+    group = group,
+    buffer = bufnr,
+    event = cl_events,
   })
-  if not augroup_exist then
-    vim.api.nvim_create_augroup("lsp_code_lens_refresh", {})
+
+  if ok and #cl_autocmds > 0 then
+    return
   end
-  vim.api.nvim_create_autocmd({ "BufEnter", "InsertLeave" }, {
-    group = "lsp_code_lens_refresh",
+  vim.api.nvim_create_augroup(group, { clear = false })
+  vim.api.nvim_create_autocmd(cl_events, {
+    group = group,
     buffer = bufnr,
     callback = vim.lsp.codelens.refresh,
   })


lua/lvim/lsp/utils.lua
```diff
@@ -88,14 +89,13 @@ function M.setup_document_highlight(client, bufnr)
   if not status_ok or not highlight_supported then
     return
   end
-  local augroup_exist, _ = pcall(vim.api.nvim_get_autocmds, {
-    group = "lsp_document_highlight",
-  })
-  if not augroup_exist then
-    vim.api.nvim_create_augroup("lsp_document_highlight", {})
+  local group = "lsp_document_highlight"
```
kylo252, line 92: we can use the same variable again for line 103

kylo252: and the same for the other variable as well^

lua/lvim/utils.lua
```diff
@@ -104,4 +104,16 @@ function M.write_file(path, txt, flag)
   end)
 end
 
+---Checks for existence of buffer-local autocommands
+---@param group string|integer name or id
+---@param bufnr integer
+function M.autocommands_exists(group, bufnr)
```
kylo252, line 110: late discovery, but I noticed that this is more like `augroup_exists`, since we're not passing any specific autocmd identifiers.

we have two ways of fixing this:
- pass more meta data about the autocmd (mainly `pattern` and `event`)
- store the autocmd id (maybe a buffer variable?), and see if it's still valid later

kylo252: @lvimuser, here's a simple implementation that fixes the original issue, we could always create that wrapper if it's needed

```diff
diff --git a/lua/lvim/lsp/utils.lua b/lua/lvim/lsp/utils.lua
index e62c3b7..022f1c3 100644
--- a/lua/lvim/lsp/utils.lua
+++ b/lua/lvim/lsp/utils.lua
@@ -1,7 +1,6 @@
 local M = {}
 
 local tbl = require "lvim.utils.table"
-local utils = require "lvim.utils"
 
 function M.is_client_active(name)
   local clients = vim.lsp.get_active_clients()
@@ -90,11 +89,19 @@ function M.setup_document_highlight(client, bufnr)
     return
   end
   local group = "lsp_document_highlight"
-  if utils.autocommands_exists(group, bufnr) then
+  local hl_events = { "CursorHold", "CursorHoldI" }
+
+  local ok, hl_autocmds = pcall(vim.api.nvim_get_autocmds, {
+    group = group,
+    buffer = bufnr,
+    event = hl_events,
+  })
+
+  if ok and #hl_autocmds > 0 then
     return
   end
   vim.api.nvim_create_augroup(group, { clear = false })
-  vim.api.nvim_create_autocmd({ "CursorHold", "CursorHoldI" }, {
+  vim.api.nvim_create_autocmd(hl_events, {
     group = group,
     buffer = bufnr,
     callback = vim.lsp.buf.document_highlight,
@@ -114,11 +121,18 @@ function M.setup_codelens_refresh(client, bufnr)
     return
   end
   local group = "lsp_code_lens_refresh"
-  if utils.autocommands_exists(group, bufnr) then
+  local cl_events = { "BufEnter", "InsertLeave" }
+  local ok, cl_autocmds = pcall(vim.api.nvim_get_autocmds, {
+    group = group,
+    buffer = bufnr,
+    event = cl_events,
+  })
+
+  if ok and #cl_autocmds > 0 then
     return
   end
   vim.api.nvim_create_augroup(group, { clear = false })
-  vim.api.nvim_create_autocmd({ "BufEnter", "InsertLeave" }, {
+  vim.api.nvim_create_autocmd(cl_events, {
     group = group,
     buffer = bufnr,
     callback = vim.lsp.codelens.refresh,
diff --git a/lua/lvim/utils.lua b/lua/lvim/utils.lua
index 8e1647f..71ede6e 100644
--- a/lua/lvim/utils.lua
+++ b/lua/lvim/utils.lua
@@ -104,16 +104,4 @@ function M.write_file(path, txt, flag)
   end)
 end
 
----Checks for existence of buffer-local autocommands
----@param group string|integer name or id
----@param bufnr integer
-function M.autocommands_exists(group, bufnr)
-  local ok, autocmds = pcall(vim.api.nvim_get_autocmds, {
-    group = group,
-    pattern = string.format("<buffer=%d>", bufnr),
-  })
-
-  return ok and #autocmds > 0
-end
-
 return M
```

lvimuser: So we can pass buffer, it's just not documented?

kylo252: > So we can pass buffer, it's just not documented?

it's not that clear maybe, but it's in the [nightly docs](https://github.com/neovim/neovim/blob/c54c37ee28cf380199a65c5849b3abe7aaae0afa/runtime/doc/api.txt#L3394) at least

> I'd like it to be a private function inside `lsp/utils.lua` then, is that ok?

maybe call it `safe_get_autocmds` and add it to `core/autocmds.lua`

lvimuser: > > So we can pass buffer, it's just not documented?
>
> it's not that clear maybe, but it's in the [nightly docs](https://github.com/neovim/neovim/blob/c54c37ee28cf380199a65c5849b3abe7aaae0afa/runtime/doc/api.txt#L3394) at least

That's `return`, not `parameter`. :shrug:

> > I'd like it to be a private function inside `lsp/utils.lua` then, is that ok?
>
> maybe call it `safe_get_autocmds` and add it to `core/autocmds.lua`

nvm, not worth it. We'd have to pass everything anyway.

kylo252 (approved):
LGTM! :rocket:

PR #3016, opened 2022-09-13
fix(example config): fix config for treesitter

yangbinji:
# Description

fix 'lvim.builtin.treesitter.highlight.enabled = false' not work.

## How Has This Been Tested?

Change config to take effect.


lvimuser (approved):

kylo252 (approved):
Thanks!

PR #3029, opened 2022-09-16
feat(document highlight): use illuminate rather than autocommand to avoid flashing

ChristianChiarulli:


lua/lvim/lsp/handlers.lua
```diff
@@ -12,7 +12,7 @@ function M.setup()
     float = lvim.lsp.diagnostics.float,
   }
   vim.diagnostic.config(config)
-  vim.lsp.handlers["textDocument/hover"] = vim.lsp.with(vim.lsp.handlers.hover, lvim.lsp.float)
+  -- vim.lsp.handlers["textDocument/hover"] = vim.lsp.with(vim.lsp.handlers.hover, lvim.lsp.float)
```
cpea2506: I wonder why this is commented. This make hover interface look good.

PR #3031, opened 2022-09-16
feat: add lir.nvim

kylo252:
# Description

- feat: add lir.nvim as a minimal alternative to nvimtree
- only allow setting up lir if `lvim.builtin.nvimtree.active=false`

## How Has This Been Tested?

```lua
lvim.builtin.nvimtree.active = false
lvim.builtin.lir.active = true
```

seems to work fine with `:edit .`

### DIFF
--- a/lua/lvim/core/builtins/init.lua
+++ b/lua/lvim/core/builtins/init.lua
@@ -9,6 +9,7 @@ local builtins = {
   "lvim.core.telescope",
   "lvim.core.treesitter",
   "lvim.core.nvimtree",
+  "lvim.core.lir",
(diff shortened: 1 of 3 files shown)


kylo252:
There's something broken about `lir.nvim` when it comes to installing with Packer in headless mode, probably related to this
https://github.com/ChristianChiarulli/lir.nvim/blob/7d8c6c4a8a1251ec8433b708ddf9b1491a07703f/plugin/lir.vim#L14-L23

will shouldn't merge until the CI is actually passing, see https://github.com/LunarVim/LunarVim/actions/runs/3066810078

PR #3032, opened 2022-09-16
fix(installer): small fix in help message of install.sh

vankodg:
Change-Id: Idbfcb8be24b7b17c2680b6ab0b5c9ee6e807c2fd

# Description

Move `-` inside the square brackets, so both version (with or without) is correct.

## How Has This Been Tested?

- Run `install.sh -h`


ChristianChiarulli:
do you have a picture demonstrating the fix?

kylo252 (approved):
LGTM, thanks! :rocket:

PR #3035, opened 2022-09-17
fix(lsp): enforce lvim completion for lua-server

### LINKED ISSUE #3002: lua-dev config broken
### Problem description

The recent update of https://github.com/folke/lua-dev.nvim deprecate the `lspconfig` field, which breaks our customized lsp config.

### LunarVim version

rolling-2519e074

### Neovim version (>= 0.7)

NVIM v0.7.2

### Operating system/version

archlinux

### Affected language servers

sumneko_lua

### Steps to reproduce

_No response_

### Actual behavior

For example, `vim`, `packer_plugins` as global for diagnostics don't take effect.

### Expected behavior

_No response_

### support info

```console
 Use [q] or [Esc] to quit the window
 
 Language client log: /home/jmei/.cache/nvim/lsp.log
 Detected filetype:   lua
 
 2 client(s) attached to this buffer: 
 
 Client: sumneko_lua (id: 1, pid: 8170, bufnr: [1])
 	filetypes:       lua
 	autostart:       true
 	root directory:  /home/jmei/lib/LunarVim
 	cmd:             lua-language-server
 
 Client: null-ls (id: 2, pid: 5001, bufnr: [1])
 	filetypes:       cpp, lua
 	autostart:       false
 	root directory:  /home/jmei/lib/LunarVim
 	cmd:             nvim
 
 Configured servers list: sumneko_lua
```

### logs

_No response_

### Screenshots

_No response_

kylo252:
# Description

`lua-dev` has changed the way it accepts settings, this will now hopefully accommodate those changes!

- fix `lvim.` auto-completion in `config.lua`
- add a select list of plugins by default that should always have auto-completion
  ```lua
  plugins = { "plenary.nvim", "telescope.nvim", "nvim-treesitter", "packer.nvim", "LuaSnip" }
  ```

fixes #3002

## How Has This Been Tested?

- open `config.lua` start typing `lvim.builtin.telescope.`
- open `config.lua` start typing `require'packer.`

### DIFF
--- a/lua/lvim/lsp/providers/sumneko_lua.lua
+++ b/lua/lvim/lsp/providers/sumneko_lua.lua
@@ -1,37 +1,38 @@
-local dev_opts = {
+local default_workspace = {
   library = {
-    vimruntime = true, -- runtime path
-    types = true, -- full signature, docs and completion of vim.api, vim.treesitter, vim.lsp and others
-    -- plugins = true, -- installed opt or start plugins in packpath
-    -- you can also specify the list of plugins to make available as a workspace library
-    plugins = { "plenary.nvim", "telescope.nvim", "nvim-treesitter", "packer.nvim", "LuaSnip" },
+    vim.fn.expand "$VIMRUNTIME",
+    get_lvim_base_dir(),
+    require("lua-dev.sumneko").types(),
   },
-  override = function(root_dir, library)
-    -- workaround for hard-coded lua_dev.util.is_nvim_config()
-    if root_dir == get_config_dir() or root_dir == get_lvim_base_dir() then
-      library.enabled = true
-      library.plugins = { "plenary.nvim", "telescope.nvim", "nvim-treesitter", "packer.nvim", "LuaSnip" }
-    end
-  end,
+
+  maxPreload = 1000,
+  preloadFileSize = 10000,
 }
 
-local lua_dev_loaded, lua_dev = pcall(require, "lua-dev")
-if lua_dev_loaded then
-  lua_dev.setup(dev_opts)
+local add_packages_to_workspace = function(packages, config)
+  -- config.settings.Lua = config.settings.Lua or { workspace = default_workspace }
+  local runtimedirs = vim.api.nvim__get_runtime({ "lua" }, true, { is_lua = true })
+  local workspace = config.settings.Lua.workspace
+  for _, v in pairs(runtimedirs) do
+    for _, pack in ipairs(packages) do
+      if v:match(pack) and not vim.tbl_contains(workspace.library, v) then
+        table.insert(workspace.library, v)
+      end
+    end
+  end
 end
 
 local lspconfig = require "lspconfig"
 
-local make_on_new_config = function(on_new_config)
+local make_on_new_config = function(on_new_config, _)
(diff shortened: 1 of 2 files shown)


lua/lvim/lsp/providers/sumneko_lua.lua
```diff
@@ -4,16 +4,41 @@ local dev_opts = {
...
+    -- this is otherwise getting overwritten by lua_dev.lsp.on_new_config()
+    new_config.settings.Lua = new_config.settings.Lua or { workspace = { library = {} } }
+    table.insert(new_config.settings.Lua.workspace.library, get_lvim_base_dir())
+  end)
+end
+
+lspconfig.util.default_config = vim.tbl_extend("force", lspconfig.util.default_config, {
+  on_new_config = make_on_new_config(lspconfig.util.default_config.on_new_config),
+})
+
```
lvimuser, line 41: I suppose passing `on_new_config` in `opts` doesn't work? 
Alternatively, `lua_dev.setup()` returns the settings, so we could try `return vim.tbl_deep_extend("force", luadev, opts)`

kylo252: > I suppose passing `on_new_config` in `opts` doesn't work?

that one is getting overwritten unfortunately, I nearly want to write our own `lua_dev.setup()` and skip this entire dance..

abzcoding:
@kylo252 I just removed my custom config and used this instead, and it worked just fine,
not sure about customization though

abzcoding:
Also, I'm not sure if we need all of those plugins activated or not
```
 { "plenary.nvim", "telescope.nvim", "nvim-treesitter", "packer.nvim", "LuaSnip" }
 ```
 it seems a bit much
Screenshot 2022-09-18 at 7 24 25 PM

lvimuser:
Agree with abz. I don't think plenary should be a default either...
Maybe leave it empty for now?

lua/lvim/lsp/providers/sumneko_lua.lua
lvimuser: > Alternatively, lua_dev.setup() returns the settings, so we could try return vim.tbl_deep_extend("force", luadev, opts)

It seems that `nlspsettings` overrides the configuration in that case (not sure if it's a bug / intended or something else going on). So that might not work either...

kylo252:
> it seems a bit much

I did change the `maxPreload=1000`, and it was pretty fast, but I have only tested on a single machine

> Agree with abz. I don't think plenary should be a default either...
> Maybe leave it empty for now?

the way I think about it is that you'd want these kind of completion for a nice dev environment:
- plenary: async, busted, curl, etc
- telescope: custom finders
- luasnip: snippets API
- treesitter: extended API that's not yet available in core
- packer: alternative async impl (probably not that useful)

I'll try to see if I can get the lua server to ignore `test{s}` directories, and how much would that save us.

abzcoding (approved):
thanks @kylo252, it is much faster now

LGTM 🚀

lvimuser:
> I did change the `maxPreload=1000`, and it was pretty fast, but I have only tested on a single machine
> 
I missed that, looks pretty good!

Issue #2891, opened 2022-08-09
`:LvimUpdate` timing out after 5000ms


brice-verint:
### Problem description

1. Run `:LvimUpdate`

```
Error executing Lua callback: .../site/pack/packer/start/plenary.nvim/lua/plenary/job.lua:499: 'git fetch --unshallow' was unable to complete in 5000 ms                                                                                                                                            
stack traceback:                                                                                                                                  
        [C]: in function 'error'                                                                                                                  
        .../site/pack/packer/start/plenary.nvim/lua/plenary/job.lua:499: in function 'wait'                                                       
        .../site/pack/packer/start/plenary.nvim/lua/plenary/job.lua:453: in function 'sync'                                                       
        ...n.Rice/.local/share/lunarvim/lvim/lua/lvim/utils/git.lua:23: in function 'git_cmd'                                                     
        ...n.Rice/.local/share/lunarvim/lvim/lua/lvim/utils/git.lua:44: in function 'safe_deep_fetch'                                             
        ...n.Rice/.local/share/lunarvim/lvim/lua/lvim/utils/git.lua:56: in function 'update_base_lvim'                                            
        ...n.Rice/.local/share/lunarvim/lvim/lua/lvim/bootstrap.lua:124: in function 'update'                                                     
        ...ce/.local/share/lunarvim/lvim/lua/lvim/core/commands.lua:47: in function <...ce/.local/share/lunarvim/lvim/lua/lvim/core/commands.lua:4
6>
```

### LunarVim version

master-47e4e5b

### Neovim version (>= 0.7.2)

NVIM v0.7.2

### Operating system/version

macOS 12.3 (Monterey)

### Steps to reproduce

1. Run `:LvimUpdate`

### support info

```console
:LvimInfo
Error executing Lua callback: vim/shared.lua:0: t: expected table, got nil                                                                        
stack traceback:                                                                                                                                  
        [C]: in function 'error'                                                                                                                  
        vim/shared.lua: in function 'validate'                                                                                                    
        vim/shared.lua: in function <vim/shared.lua:0>                                                                                            
        vim/shared.lua: in function 'tbl_filter'                                                                                                  
        ...n.Rice/.local/share/lunarvim/lvim/lua/lvim/core/info.lua:115: in function 'make_auto_lsp_info'                                         
        ...n.Rice/.local/share/lunarvim/lvim/lua/lvim/core/info.lua:159: in function 'toggle_popup'                                               
[...]

Issue #2794, opened 2022-07-06
LunarVim cannot completely be installed because of "Killing git due to timeout!"


dennis605:
### Problem description

If I try to install LunarVim with following command
` bash <(curl -s https://raw.githubusercontent.com/lunarvim/lunarvim/master/utils/installer/install.sh)
`
everything went fine until installer tries to prepare Packer setup.
If installer is in that phase there is a break for about 5 Minutes and then installer hangs with error:

`[packer.nvim] [WARN  23:32:51] jobs.lua:87: Killing git due to timeout!`  
and will throw more errors saying: "Killing git due to timeout!"
I will put the whole output in field "support info"

### LunarVim version

master-b04aefb

### Neovim version (>= 0.7.2)

NVIM v0.7.2

### Operating system/version

Debian 11

### Steps to reproduce

1. install LunarVim with following command: `bash <(curl -s https://raw.githubusercontent.com/lunarvim/lunarvim/master/utils/installer/install.sh)`
2. answer all following question with YES

### support info

` bash <(curl -s https://raw.githubusercontent.com/lunarvim/lunarvim/master/utils/installer/install.sh)

      88\                                                   88\
      88 |                                                  \__|
      88 |88\   88\ 888888$\   888888\   888888\ 88\    88\ 88\ 888888\8888\
      88 |88 |  88 |88  __88\  \____88\ 88  __88\\88\  88  |88 |88  _88  _88\
      88 |88 |  88 |88 |  88 | 888888$ |88 |  \__|\88\88  / 88 |88 / 88 / 88 |
      88 |88 |  88 |88 |  88 |88  __88 |88 |       \88$  /  88 |88 | 88 | 88 |
      88 |\888888  |88 |  88 |\888888$ |88 |        \$  /   88 |88 | 88 | 88 |
      \__| \______/ \__|  \__| \_______|\__|         \_/    \__|\__| \__| \__|

--------------------------------------------------------------------------------
Detecting platform for managing any additional neovim dependencies
--------------------------------------------------------------------------------
Would you like to install LunarVim's NodeJS dependencies?
[y]es or [n]o (default: no) : yes
[ERROR] Unable to install using [pnpm] without administrative privileges.
Installing node modules with yarn..
yarn global v1.22.19
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
[4/4] Building fresh packages...
success Installed "neovim@4.10.1" with binaries:
      - neovim-node-host
[...]

Issue #2793, opened 2022-07-06
Packer sync gets stuck when I reach 80 plugins installed


danielo515:
### Problem description

I think this issue very depends on the machine power and threading capabilities, but anyway, here it goes.
I was using lunarvim without problem for some time, and I was also using lvim rolling version. Recently, when I run packer sync it was getting stuck. I tried removing all plugins, then add them one by one and execute packer sync
 with ` lvim +PackerSync -c 'autocmd User PackerComplete quitall'`. When I reached the amount of 80 plugins the command stuck. I tried removing and adding various plugins to verify it was not tied to an specific plugin, and every time it was able to install up to 79 plugins, no matter which ones, but as soon as I reached the number of 80, stuck.
This happens consistently in my CI environment, but only for the OSX checker, the linux one usually finishes without problems.

This issue may be related (packer repo): https://github.com/wbthomason/packer.nvim/issues/751

### LunarVim version

rolling-e662039

### Neovim version (>= 0.7.2)

NVIM v0.7.0-dev+882-g9304ee387

### Operating system/version

macOS 11.6

### Steps to reproduce

This is the default lvim config:

```
{
    package_root = opts.package_root or join_paths(vim.fn.stdpath "data", "site", "pack"),
    compile_path = compile_path,
    snapshot_path = snapshot_path,
    log = { level = "warn" },
    git = { clone_timeout = 900 },
    display = {
      open_fn = function()
        return require("packer.util").float { border = "rounded" }
      end,
    },
  }
```

And if I change it to this then the command succeeds and everything works. This requires me hacking lunarvim source code, which will not be feasible on my CI pipeline:

```
{
    package_root = opts.package_root or join_paths(vim.fn.stdpath "data", "site", "pack"),
    compile_path = compile_path,
    snapshot_path = snapshot_path,
    log = { level = "warn" },
    max_jobs = 50,
    git = { clone_timeout = 900 },
    display = {
      open_fn = function()
        return require("packer.util").float { border = "rounded" }
      end,
    },
  }
```

As you can see, I just added max_jobs 50 and it now works.

### support info

```console
 Language client log: /Users/danielo/.cache/lvim/lsp.log
 Detected filetype:   lua
 
 1 client(s) attached to this buffer: 
 
 Client: sumneko_lua (id: 2, pid: 92465, bufnr: [127])
 	filetypes:       lua
 	autostart:       true
 	root directory:  /Users/danielo/.local/share/lunarvim/lvim
[...]

PR #3038, opened 2022-09-19
feat: add lir.nvim again

ChristianChiarulli:


kylo252 (commented):
You can see that the installer is broken with `lir`, I'm not really sure why but it's breaking packer :cry: 

As I said in the other PR, I suspect it's the [autocmd messing things up](https://github.com/ChristianChiarulli/lir.nvim/blob/7d8c6c4a8a1251ec8433b708ddf9b1491a07703f/plugin/lir.vim#L14-L23)

rebuilt:
We had lowered the number of max_jobs because users were experiencing problems 

https://github.com/LunarVim/LunarVim/issues/2891
https://github.com/LunarVim/LunarVim/issues/2794
https://github.com/LunarVim/LunarVim/issues/2793

JoshMock:
Using lvim nightly and have been getting this notification on startup:

```
packer.nvim: Error running config for lir.nvim: ...packer/start/nvim-web-devicons/lua/nvim-web-devicons.lua:1700: bad argument #1 to 'pairs' (table expected, got nil)
```

Maybe related to this change?

ChristianChiarulli:
@JoshMock try running `:PackerSync` and `:PackerCompile`

ChristianChiarulli:
also jj is gone, if people want the feature it's not hard to add

Uzaaft:
I believe that this PR broke Packer for me on Mac. max_jobs = 100 makes Packer freeze, while for.eks max_jobs= 50 does not.

JoshMock:
> @JoshMock try running `:PackerSync` and `:PackerCompile`

`:PackerCompile` doesn't work, just popped up another notification with the [same error](https://github.com/LunarVim/LunarVim/pull/3038#issuecomment-1254256675).

ChristianChiarulli:
we should change the max jobs depending on platform @kylo252

JoshMock:
I figured out that my issue was due to setting `lvim.builtin.nvimtree.active = false`. I wasn't using it so I'd disabled it. Reenabling it fixed [my error](https://github.com/LunarVim/LunarVim/pull/3038#issuecomment-1254256675). Should probably make sure core configs aren't dependent on plugins that can be disabled.

ChristianChiarulli:
good point, right now they're pretty intertwined, I'll try to handle it on a stream

PR #3040, opened 2022-09-19
fix(lir): warning only if both lir and nvimtree are active

cpea2506:
# Description

Make sure `lir` is active before checking conflicting with `nvimtree`

## How Has This Been Tested?

- Set `lvim.builtin.lir.active = false` and `lvim.builtin.nvimtree.active = true`
- See there is no warning about `nvimtree`

### DIFF
--- a/lua/lvim/core/lir.lua
+++ b/lua/lvim/core/lir.lua
@@ -4,11 +4,12 @@ local Log = require "lvim.core.log"
 
 M.config = function()
   lvim.builtin.lir = {
-    active = true,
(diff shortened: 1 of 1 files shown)


cpea2506:
@ChristianChiarulli Should we disable `lir` by default if `nvimtree` is active? I think a could be good for fresh install of `Lunarvim` with no warning message.

rebuilt (approved):

ChristianChiarulli:
there's no reason they can't both be active

PR #3043, opened 2022-09-20
feat: breadcrumbs

ChristianChiarulli:


lua/lvim/plugins.lua
```diff
@@ -202,6 +202,15 @@ local core_plugins = {
     disable = not lvim.builtin.lualine.active,
   },
 
+  -- breadcrumbs
+  {
+    "SmiteshP/nvim-navic",
+    config = function()
+      require("lvim.core.breadcrumbs").setup()
+    end,
+    -- disable = not lvim.builtin.breadcrumbs.active,
```
abzcoding, line 211: @ChristianChiarulli is there a reason that this is commented?

cc: @kylo252

lua/lvim/core/breadcrumbs.lua
```diff
@@ -0,0 +1,204 @@
+local M = {}
+
+-- local Log = require "lvim.core.log"
+
+M.config = function()
```
abzcoding, line 5: this should be `M.setup`
and we might need a `M.config`

I'll address these in a PR

PR #3049, opened 2022-09-20
revert: fix Packer instead of hard-coding config

kylo252:
# Description

- revert: suggest using PackerCompile and instead of hard-coding config
- fix(installer): remove useless shallow clone

## How Has This Been Tested?

- `rm ~/.config/lvim/plugin/packer_compiled.lua`
- `lvim +PackerSync`
- check that winbar works :rocket:


abzcoding (approved):
thank you <3

ChristianChiarulli (approved):

PR #3081, opened 2022-09-25
fix(log): correct add_entry code documentation

nieomylnieja:
# Description

Fixed:
- incorrect type hint in `add_entry` method description
- added missing `event` parameter type hint
- corrected reference from `vim.log.log_levels` to the current `vim.log.levels`


abzcoding (approved):
thanks <3

Issue #1987, opened 2021-11-24
Can we get session restore with tmux ?


Dulanjala007:
### Feature motivation

tmux can't restore lvim instances...
maybe because `tmux-ressuract` records processes as nvim not lvim..?
any thoughts?

kylo252:
The process name is still `nvim`, but maybe you meant it the other way around. Unfortunately, `tmux-resurrect` uses rudimentary methods to detect session files, which doesn't really help all that much. Also, do take a look at https://github.com/folke/persistence.nvim.

Dulanjala007:
@kylo252 nice to see you again... last time with the lsp-intaller bug..

with tmux i was used to seeing the vim opened and cursor in the last place i left it after the restart..
so i was trying make it happen with the lunarvim as well...

currently trying out various combinations from this documentation to see if i can get it working..
[https://github.com/tmux-plugins/tmux-resurrect/blob/master/docs/restoring_programs.md](url)

clementnuss:
any enhancement on this topic? that would indeed be really nice 👍🏻

daudcanugerah:
function lvim() { nvim -u ~/.local/share/lunarvim/lvim/init.lua "$@" }

abzcoding:
you can now probably use the new lvim binary to achieve this, feel free to reopen if the issue still persists

clementnuss:
thanks for that info. not sure where/how to find the `lvim` binary though.
the `lvim` file in `~/.local/bin` isn't producing the expected results
is that on another branch than rolling?

abzcoding:
the lvim binary should look like this ( your username instead of `abz` )
<img width="740" alt="Screen Shot 2022-02-08 at 2 35 00 PM" src="https://user-images.githubusercontent.com/10992695/152974928-53f61197-d90c-4845-9d8b-e88833967b0f.png">

clementnuss:
the `lvim`  POSIX shell script is an executable shell script and not a binary file, and its content is the same as the one you showed.

The issue however is that `tmux` restores the process name, which will be something looking like `nvim ...` and not `lvim`.
I looked a bit more in the `~/.tmux/resurrect` folder, and having one of the session restore files, noticed that tmux records the `lvim` sessions as follows:
```
pane    daily   0       1       :*      0       :/home/clement/currentWorkingDirectory   1       nvim    :nvim -u /home/clement/.local/share/lunarvim/lvim/init.lua
```

[...]

Issue #2745, opened 2022-06-22
[Resolved] tmux navigation skips vim windows/buffers


rollschild:
### Problem description

I'm using [vim-tmux-navigator](https://github.com/christoomey/vim-tmux-navigator) as the plugin to facilitate navigation between vim and tmux. And I use the suggested tmux.conf:
```
# Smart pane switching with awareness of Vim splits.
# See: https://github.com/christoomey/vim-tmux-navigator
is_vim="ps -o state= -o comm= -t '#{pane_tty}' \
    | grep -iqE '^[^TXZ ]+ +(\\S+\\/)?g?(view|n?vim?x?)(diff)?$'"
bind-key -n 'C-h' if-shell "$is_vim" 'send-keys C-h'  'select-pane -L'
bind-key -n 'C-j' if-shell "$is_vim" 'send-keys C-j'  'select-pane -D'
bind-key -n 'C-k' if-shell "$is_vim" 'send-keys C-k'  'select-pane -U'
bind-key -n 'C-l' if-shell "$is_vim" 'send-keys C-l'  'select-pane -R'
tmux_version='$(tmux -V | sed -En "s/^tmux ([0-9]+(.[0-9]+)?).*/\1/p")'
if-shell -b '[ "$(echo "$tmux_version < 3.0" | bc)" = 1 ]' \
    "bind-key -n 'C-\\' if-shell \"$is_vim\" 'send-keys C-\\'  'select-pane -l'"
if-shell -b '[ "$(echo "$tmux_version >= 3.0" | bc)" = 1 ]' \
    "bind-key -n 'C-\\' if-shell \"$is_vim\" 'send-keys C-\\\\'  'select-pane -l'"

bind-key -T copy-mode-vi 'C-h' select-pane -L
bind-key -T copy-mode-vi 'C-j' select-pane -D
bind-key -T copy-mode-vi 'C-k' select-pane -U
bind-key -T copy-mode-vi 'C-l' select-pane -R
bind-key -T copy-mode-vi 'C-\' select-pane -l
```

However, after installing the plugin, there is an issue. Pressing navigation keys such as `<C-l>` or `<C-h>` correctly switches between vim and tmux, _but_ it would completely skip other vim windows.

For example, if I have two tmux panes, `0` and `1`, and in pane `0` I have two vim windows `0` and `1`, like following:
```
tmux pane 0                                         | tmux pane 1
----------------------------------------------------|
|                         |                         |
[...]

PR #3099, opened 2022-09-29
Make tmux-resurrect plugin restore lvim sessions properly

shitcoding:
# Description

Join lvim shell script launcher commands into oneliner in order to make `tmux-resurrect` plugin properly restore Lunarvim sessions when using tmux.
(See issue "Can we get session restore with tmux ?" - https://github.com/LunarVim/LunarVim/issues/1987)

The solution proposed by clementnuss in the comments (https://github.com/LunarVim/LunarVim/issues/1987#issuecomment-1032563301):
> the lvim POSIX shell script is an executable shell script and not a binary file, and its content is the same as the one you showed.
> 
> The issue however is that tmux restores the process name, which will be something looking like nvim ... and not lvim.
> I looked a bit more in the ~/.tmux/resurrect folder, and having looked at one of the session restore files, noticed that tmux records the lvim sessions as follows:
> 
> pane    daily   0       1       :*      0       :/home/clement/currentWorkingDirectory   1       nvim    :nvim -u /home/clement/.local/share/lunarvim/lvim/init.lua
> 
> when restoring as is, this fails, because the LUNARVIM_... environment variables aren't set.
> 
> after setting the 3 env variables (as in the lvim shell script), restoring with tmux works! So I'll simply add those variables to my default env variables and this will solve the issue 

This solution works totally fine, but you need to export env variables in bashrc or zshrc in order to fix the problem.

Joining env variables export commands and lvim launch command into a one-liner fixes the tmux-resurrect issue without extra hassle.


utils/bin/lvim.template
```diff
@@ -1,7 +1,6 @@
 #!/bin/sh
 
-export LUNARVIM_RUNTIME_DIR="${LUNARVIM_RUNTIME_DIR:-RUNTIME_DIR_VAR}"
-export LUNARVIM_CONFIG_DIR="${LUNARVIM_CONFIG_DIR:-CONFIG_DIR_VAR}"
-export LUNARVIM_CACHE_DIR="${LUNARVIM_CACHE_DIR:-CACHE_DIR_VAR}"
-
+export LUNARVIM_RUNTIME_DIR="${LUNARVIM_RUNTIME_DIR:-RUNTIME_DIR_VAR}" &&\
+export LUNARVIM_CONFIG_DIR="${LUNARVIM_CONFIG_DIR:-CONFIG_DIR_VAR}" &&\
+export LUNARVIM_CACHE_DIR="${LUNARVIM_CACHE_DIR:-CACHE_DIR_VAR}" &&\
```
kylo252, line 5:
```suggestion
export LUNARVIM_RUNTIME_DIR="${LUNARVIM_RUNTIME_DIR:-RUNTIME_DIR_VAR}"
export LUNARVIM_CONFIG_DIR="${LUNARVIM_CONFIG_DIR:-CONFIG_DIR_VAR}"
export LUNARVIM_CACHE_DIR="${LUNARVIM_CACHE_DIR:-CACHE_DIR_VAR}"
```

utils/bin/lvim.template
```diff
@@ -1,7 +1,6 @@
...
 exec nvim -u "$LUNARVIM_RUNTIME_DIR/lvim/init.lua" "$@"
```
kylo252, line 6:
```suggestion
exec -a lvim nvim -u "$LUNARVIM_RUNTIME_DIR/lvim/init.lua" "$@"
```

kylo252 (commented):
please try this instead, but be sure to add this to your `tmux.conf`

```tmux
# see https://github.com/tmux-plugins/tmux-resurrect#docs
set -g @resurrect-processes 'vim nvim lvim'
```

shitcoding:
> please try this instead, but be sure to add this to your `tmux.conf`
> 
> ```shell
> # see https://github.com/tmux-plugins/tmux-resurrect#docs
> set -g @resurrect-processes 'vim nvim lvim'
> ```


Already did this, but the thing is, Lunarvim env variables need to be set in order for `tmux-resurrect` to restore `lvim` session properly.  
 
clementnuss provided a solution for this: exporting lvim env vars in `.bashrc` or `.zshrc` - https://github.com/LunarVim/LunarVim/issues/1987#issuecomment-1032563301
> when restoring as is, this fails, because the LUNARVIM_... environment variables aren't set.
> 
> after setting the 3 env variables (as in the lvim shell script), restoring with tmux works! So I'll simply add those variables to my default env variables and this will solve the issue

But extra hassle can be avoided just by joining 3 `export` commands with `exec` command into one-liner. In this case `tmux-resurrect` plugin will properly restore lvim session with required env vars already set up.

Not a critical issue, but I thought it can be helpful for tmux users.

Anyways, thanks for your reply!

kylo252:
> Already did this, but the thing is, Lunarvim env variables need to be set in order for `tmux-resurrect` to restore `lvim` session properly.

no they don't, not if you're restoring `lvim` and **not** `nvim`

> clementnuss provided a solution for this: exporting lvim env vars in `.bashrc` or `.zshrc` - [#1987 (comment)](https://github.com/LunarVim/LunarVim/issues/1987#issuecomment-1032563301)

again, this is not needed since calling `lvim` shim would resolve these automatically for you

> But extra hassle can be avoided just by joining 3 `export` commands with `exec` command into one-liner. In this case `tmux-resurrect` plugin will properly restore lvim session with required env vars already set up.
> Not a critical issue, but I thought it can be helpful for tmux users.

In case you haven't tried my suggestion, check this demo


https://user-images.githubusercontent.com/59826753/194313338-9d1b9827-fdfd-47d4-8a32-a3b843ba693a.mp4



> Anyways, thanks for your reply!

np :)

shitcoding:
Umm this is weird, obviously everything works fine in your demo.

I have the following in `tmux.conf`:
```
set -g @resurrect-processes 'lvim htop'
```
But `tmux-resurrect` restores lvim session by trying to run only the last command in lvim shell executable:
```
exec nvim -u "$LUNARVIM_RUNTIME_DIR/lvim/init.lua" "$@"
```
And it fails, because first 3 `export` commands that set the env variables aren't run before that.

Anyways, this is not a Lunarvim issue, and only appears in some cases when using `tmux` + `tmux-resurrect` together (and it worked in your case).
So probably it's better for me to stop bothering you :)

Thanks!

kylo252:
@shitcoding, maybe you missed my other comment about overriding `arg[0]`?

https://github.com/LunarVim/LunarVim/pull/3099#discussion_r985055354

otherwise, it wouldn't work just as you have experienced,



related:
- https://github.com/LunarVim/LunarVim/issues/1987#issuecomment-1190152047
- https://github.com/LunarVim/LunarVim/issues/2745#issuecomment-1166258669

shitcoding:
@kylo252, yeah, didn't notice this change. Thanks again!

andreportela:
This is still happening to me.

```python
$ lvim --version
NVIM v0.8.3
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by brew@Ventura-arm64.local

Features: +acl +iconv +tui
```
So my lvim script is already: 
```bash
#!/usr/bin/env bash

export LUNARVIM_RUNTIME_DIR="${LUNARVIM_RUNTIME_DIR:-"/Users/myuser/.local/share/lunarvim"}"
export LUNARVIM_CONFIG_DIR="${LUNARVIM_CONFIG_DIR:-"/Users/myuser/.config/lvim"}"
export LUNARVIM_CACHE_DIR="${LUNARVIM_CACHE_DIR:-"/Users/myuser/.cache/lvim"}"

export LUNARVIM_BASE_DIR="${LUNARVIM_BASE_DIR:-"/Users/myuser/.local/share/lunarvim/lvim"}"

exec -a lvim nvim -u "$LUNARVIM_BASE_DIR/init.lua" "$@"
```

I tried to put these in my .tmux.conf:
```bash
set -g @plugin 'tmux-plugins/tmux-resurrect'
set -g @resurrect-capture-pane-contents 'on'
set -g @resurrect-processes 'vim nvim lvim'
set -g @resurrect-strategy-nvim 'session'
```

But still no luck. 

1. I open a file with lvim, 
2. save the session
3. kill it
4. But when I restore the session still no luck, lvim is restored into the default dashboard and my previous session is lost

What am I missing?

kylo252:
> What am I missing?

have you tried restoring the session after you've launched lvim? I find the syntax for tmux-resurrect to be extremely awkward which is why I prefer doing that step manually.

andreportela:
> have you tried restoring the session after you've launched lvim? I find the syntax for tmux-resurrect to be extremely awkward which is why I prefer doing that step manually.

Yeah... manually restoring it works

PR #3110, opened 2022-09-30
feat: dynamic or fixed toggle terminal size

### LINKED ISSUE #3102: [Hacktoberfest] terminal is dynamic and takes up percentage of editor
instead of static amount of lines, terminals should be configured to open with a percentage of the editor.

GabMus:
# Description

- New preference table for terminal split sizes
- New utility function to get the current buffer's size (as in width and height)
- Toggle terminals default to 30% width/height depending if they're vertical/horizontal

fixes #3102

## How Has This Been Tested?

- Start lvim
- Press `<M-1>` or `<M-2>` to toggle the vertical and horizontal terminals
- Play around with the window size and see that the toggle terminal is always 30% of the size


kylo252 (commented):
why not make use of `size` being a callback?

```lua
  size = function(term)
    if term.direction == "horizontal" or term.direction == "vertical"  then
      return 20 -- replace this with custom logic
    end
  end,
```

GabMus:
> why not make use of `size` being a callback?

You mean in here?

```lua
   local opts = {
      cmd = exec[1],
      cmd = exec[1],
      keymap = exec[2],
      keymap = exec[2],
      label = exec[3],
      label = exec[3],
      -- NOTE: unable to consistently bind id/count <= 9, see #2146
      -- NOTE: unable to consistently bind id/count <= 9, see #2146
      count = i + 100,
      count = i + 100,
      direction = exec[4] or lvim.builtin.terminal.direction,
      direction = exec[4] or lvim.builtin.terminal.direction,
      size = exec[5] or lvim.builtin.terminal.size,
      size = function()
        -- use the custom logic here?
      end,
    }
    }
```

ChristianChiarulli:
I think we should continue passing the size in as an arg rather than make it a global config option.

GabMus:
I reworked the way this works: now instead of using a config we again use the argument passed in exec, and distinguish between float and int for checking if the value should be a percentage (now between 0 and 1) or a fixed number of cells

GabMus:
Sorry it took me a while to get back to this.

With [this](https://github.com/LunarVim/LunarVim/pull/3110#discussion_r988954577) proposed suggestion the terminal sizes will be set when you open lvim, and stay at that size until you restart. This means that if you resize your terminal or your buffer, the percentage will be wrong. So size needs to be a function that gets called in exec_toggle

GabMus:
Anything else blocking this?

PR #2491, opened 2022-04-20
feat: lock nvim <0.7 to a specific tag

kylo252:
# Description

- require v0.7+ for starting lunarvim (prevent issues with manual update)
- running `:LvimUpdate` when using a neovim <0.7 should switch to the tag `1.1.3` automatically and print a warning message 

Related #2476

## How Has This Been Tested?

- startup requirement: modify `~/.local/bin/lvim` to use a `nvim-0.6.1` appimage
- `LvimUpdate`: see https://github.com/LunarVim/LunarVim/issues/2476#issuecomment-1100859112

PR #3111, opened 2022-10-01
feat: lock new installations to nvim 0.8+

kylo252:
# Description

- require v0.8+ for starting lunarvim (prevent issues with manual update)
- running :LvimUpdate when using a neovim <0.8 should switch to the tag 1.1.4 automatically and print a warning message

see #2491 for more info on the implementation

closes #3019

## How Has This Been Tested?

should pass the CI?

### DIFF
--- a/.github/ISSUE_TEMPLATE/general-issue-form.yaml
+++ b/.github/ISSUE_TEMPLATE/general-issue-form.yaml
@@ -33,7 +33,7 @@ body:
   - type: input
     id: nvim-version
     attributes:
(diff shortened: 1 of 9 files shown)


kylo252:
we should probably wait a bit on this until the release is widely available, but I think it should be part of the next release in any case.

lvimuser:
Don't forget the [README](https://github.com/LunarVim/LunarVim/blob/rolling/README.md?plain=1#L33)

Will also close https://github.com/LunarVim/LunarVim/pull/3019

PR #3122, opened 2022-10-02
feat: type gf on require file

SingularisArt:
now, when you're hovering over a require('a.b.c'), you can type gf, and go to the 'c.lua' file

### DIFF
--- a/ftplugin/lua.lua
+++ b/ftplugin/lua.lua
@@ -1,0 +1,86 @@
+local fmt = string.format
+
+-- Iterator that splits a string o a given delimiter
+local function split(str, delim)
+  delim = delim or "%s"
+  return string.gmatch(str, fmt("[^%s]+", delim))
+end
+
+-- Find the proper directory separator depending
+-- on lua installation or OS.
+local function dir_separator()
+  -- Look at package.config for directory separator string (it's the first line)
+  if package.config then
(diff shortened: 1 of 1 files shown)


ftplugin/lua.lua
```diff
@@ -0,0 +1,86 @@
...
+    if vim.fn.filereadable(path) == 1 then
+      return path
+    end
+  end
+end
+
+-- Search for nvim lua include paths
+local function include_rtpaths(fname, ext)
+  ext = ext or "lua"
+  local sep = dir_separator()
```
kylo252, line 38: @SingularisArt, please fix the lint violations here, these two variables are not used

ftplugin/lua.lua
```diff
@@ -0,0 +1,86 @@
...
+    -- Look on runtime path for 'lua/*/init.lua' files
+    local path2 = table.concat({ path, ext, fname, initfile }, sep)
+    if vim.fn.filereadable(path2) == 1 then
+      return path2
+    end
+  end
+end
+
+-- Global function that searches the path for the required file
+function find_required_path(module)
```
kylo252, line 56: @SingularisArt, please avoid using global functions with generic names like this, consider move this function under `lvim/utils` or at very least add a prefix to the function.

PR #3141, opened 2022-10-04
fix: Limit max_jobs to avoid freezes in macOS

Uzaaft:
Limit the number of max_jobs to 50, to avoid freezes in macOS.
Basically re-implements https://github.com/LunarVim/LunarVim/pull/2781, but with 50(seems to work locally for me)

This should be temporary, until we find a better fix

- Run command `:PackerSync` before and after, and observe that the update process does not hangup.

### DIFF
--- a/lua/lvim/plugin-loader.lua
+++ b/lua/lvim/plugin-loader.lua
@@ -20,7 +20,7 @@ function plugin_loader.init(opts)
     package_root = opts.package_root or join_paths(vim.fn.stdpath "data", "site", "pack"),
     compile_path = compile_path,
     snapshot_path = snapshot_path,
-    max_jobs = 100,
+    max_jobs = 50,
     log = { level = "warn" },
     git = {
       clone_timeout = 120,


luxus:
try this in the terminal
`ulimit -n 10240`

Uzaaft:
> try this in the terminal
> `ulimit -n 10240`

I'm not sure if that should be the proposed solution for a user whenever they get this problem?

luxus:
> can you do a system check and set this to 50 if it's macOS?

i don't understand the question.
it works with macOS with 50 ...

danielo515:
> > can you do a system check and set this to 50 if it's macOS?
> 
> i don't understand the question. it works with macOS with 50 ...

He says that, in the config you check the operating system where lunar-vim is being executed, and based on the result you set one value or the other

Uzaaft:
I guess I'll leave to other regarding how they would go forward fixing that. From what I've read, detecting macOS in lua inside nvim is a mixed bag, and I wouldn't feel comfortable implement that myself

Issue #3137, opened 2022-10-03
colorscheme default not properly overridden for lualine


bob3000:
### Problem description

The lualine colors are currently inconsistent with the selected colorscheme in the fields branch and line number (see screenshots).
I couldn't exactly figure out what's going on but I made some observations that might be helpful.

1. In https://github.com/LunarVim/LunarVim/blob/rolling/lua/lvim/core/lualine/components.lua#L27 I wrote `print(lvim.colorscheme)` into the conditional block and `:messages` actually printed `tokyonight` even though in my configuration I set `lvim.colorscheme` to `gruvbox-material`. When I comment out the entire if block the problem goes away.

2. In https://github.com/LunarVim/LunarVim/blob/rolling/lua/lvim/config/defaults.lua#L3 when I set the colorscheme to `gruvbox-material` it also makes the problem go away.

In summary it looks like the for some reason the default is not properly overridden by the user config but even more interestingly it only has a negative effect on the two fields mentioned above.

### LunarVim version

rolling-35cea14:q

### Neovim version (>= 0.7.2)

NVIM v0.8.0

### Operating system/version

macOS 12.6

### support info

.

### Screenshots

wrong lualine style (background is black)
![lualine-wrong](https://user-images.githubusercontent.com/1550574/193647208-c08411b9-d4eb-47ea-a242-e6679233cd7c.png)

correct lualine style (background is dark brown)
![lualine-right](https://user-images.githubusercontent.com/1550574/193647115-9f3d1ead-2205-47be-8d7d-bc21daa8653b.png)

Issue #3106, opened 2022-09-30
lualine theme get stuck to default, can't change


qvieth:
### Problem description

lualine theme doesn't detect current theme, can't change through `lvim.builtin.lualine.options.theme` also

### LunarVim version

rolling-9930fc3

### Neovim version (>= 0.7.2)

NVIM v0.7.2

### Operating system/version

Fedora 36

### Steps to reproduce

after an PackerSync my lualine become like this, can't change back

### support info

    __                          _    ___         
   / /   __  ______  ____ _____| |  / (_)___ ___ 
  / /   / / / / __ \/ __ `/ ___/ | / / / __ `__ \
 / /___/ /_/ / / / / /_/ / /   | |/ / / / / / / /
/_____/\__,_/_/ /_/\__,_/_/    |___/_/_/ /_/ /_/ 

Buffer info
* filetype:                c
* bufnr:                   5
* treesitter status:       active

Active client(s)
* name:                      clangd
* id:                        2
* supported filetype(s):     [c, cpp, objc, objcpp, cuda, proto]
* attached buffers:          5
* root_dir pattern:          5
* capabilities:              referencesProvider      | definitionProvider       | documentRangeFormattingProvider
                             documentSymbolProvider  | astProvider              | documentFormattingProvider
                             workspaceSymbolProvider | clangdInlayHintsProvider | documentHighlightProvider
                             implementationProvider  | declarationProvider      | hoverProvider
                             callHierarchyProvider   | typeHierarchyProvider    | selectionRangeProvider
                             typeDefinitionProvider  | memoryUsageProvider

Automatic LSP info
* Skipped servers: [angularls, ansiblels, csharp_ls, cssmodules_ls, denols, ember, emmet_ls, eslint, golangci_lint_ls, graphql, jedi_language_server, ltex, phpactor, psalm, pylsp, quick_lint_js, rome, reason_ls, solang, solc, sorbet, sourcery, sqlls, sqls, stylelint_lsp, tflint, svlangserver, verible, vuels]

Formatters info
* Active: 
* Supported: [astyle, clang_format, uncrustify]

Linters info
* Active: 
* Supported: [clang_check, cppcheck, cpplint, gccdiag]

Code actions info
* Active: 

### Screenshots

![image](https://user-images.githubusercontent.com/71810726/193218117-2b9acf43-eee9-4432-a485-169b59a987c5.png)

lvimuser:
See also #3020

[...]

Issue #3066, opened 2022-09-23
Lualine theme can't be changed in configuration


carlossgv:
### Problem description

Lualine theme is not beign updated when changing colorscheme from the config file, nor it can be set it in using `lvim.builtin.lualine.options.theme = "onedark"`.

Current behavior:
- Statusline theme won't match selected colorscheme when `lvim.colorscheme = "some_theme"` is set in config file.
- Statusline theme won't change if set in config file like `lvim.builtin.lualine.options.theme = "some_theme"`.
- Statusline will update to current theme when set from command line `:colorscheme some_theme`.

For the time beign I managed to get it to match theme from config file by changing line 86 in `lua/lvim/core/theme.lua` from `lvim.builtin.lualine.options.theme = "tokyonight"` to `lvim.builtin.lualine.options.theme = "auto"`. This will make the statusline to match theme set in config but still won't work when setting the status line theme in config.

### LunarVim version

rolling-817337b

### Neovim version (>= 0.7.2)

NVIM v0.7.2 Release

### Operating system/version

Linux Ubunti 22.04

### Steps to reproduce

1. Go to config file.
2. Set lvim.colorscheme = "pablo"
3. Restart lvim.

1. Go to config file.
2. Set lvim.colorscheme = "pablo"
3. Set. lvim.builtin.lualine.options.theme = "pablo"
4. Restart lvim.

### support info

```
 Use [q] or [Esc] to quit the window
 
 Language client log: /home/carlossgv/.cache/lvim/lsp.log
 Detected filetype:   alpha
 
 1 client(s) attached to this buffer: 
 
 Client: copilot (id: 1, pid: 109644, bufnr: [1])
 	filetypes:       
 	autostart:       false
 	root directory:  Running in single file mode.
 	cmd:             node /home/carlossgv/.local/share/lunarvim/site/pack/packer/start/copilot.vim/copilot/dist/agent.js
 
 Configured servers list: 

```
```

    __                          _    ___         
   / /   __  ______  ____ _____| |  / (_)___ ___ 
  / /   / / / / __ \/ __ `/ ___/ | / / / __ `__ \
 / /___/ /_/ / / / / /_/ / /   | |/ / / / / / / /
/_____/\__,_/_/ /_/\__,_/_/    |___/_/_/ /_/ /_/ 


Buffer info
* filetype:                alpha
* bufnr:                   1
* treesitter status:       inactive

Active client(s)
* name:                      copilot
* id:                        1
[...]

PR #3142, opened 2022-10-04
fix: lualine custom theme

opalmay:
A few issues opened about this, #3106 #3066 and (maybe) #3137
LuaLine was forced to use TokyoNight with no option to change in config. Instead it will now use "auto" which will work for custom themes as well as TokyoNight.


opalmay:
Change was already covered in last merged commit

PR #3165, opened 2022-10-06
feat(wich-key): binding to save without formatting

danielo515:
# Description

summary of the change

A lot of people is concerned when they save a file and they don't want auto-format to be applied to it. This adds a nice which key mapping to do that.

## How Has This Been Tested?

This is a command I execute often, there is nothing fancy about it

### DIFF
--- a/lua/lvim/core/which-key.lua
+++ b/lua/lvim/core/which-key.lua
@@ -92,7 +92,7 @@ M.config = function()
         f = { "<cmd>Telescope buffers<cr>", "Find" },
         b = { "<cmd>BufferLineCyclePrev<cr>", "Previous" },
         n = { "<cmd>BufferLineCycleNext<cr>", "Next" },
-        W = {"<cmd>noautocmd w<cr>", "Save without formatting" },
+        W = { "<cmd>noautocmd w<cr>", "Save without formatting" },
         -- w = { "<cmd>BufferWipeout<cr>", "Wipeout" }, -- TODO: implement this for bufferline
         e = {
           "<cmd>BufferLinePickClose<cr>",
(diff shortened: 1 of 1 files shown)


danielo515:
@kylo252 anything against?

danielo515:
Sorry for bumping this, but I think it is something will help people.

danielo515:
@ChristianChiarulli would you mind having a look?

LostNeophyte:
I've added that keymapping to the default config (not yet merged) and even my own, it's quite useful, thanks :)
I'm not sure about adding it to lvim because formatting on save is disabled by default

PS: set the base branch to master

danielo515:
> I've added that keymapping to the default config (not yet merged) and even my own, it's quite useful, thanks :) I'm not sure about adding it to lvim because formatting on save is disabled by default
> 
> PS: set the base branch to master

Already changed the base to master.
The keybinding is also useful even if you don't use autoformat, because it just disables all autocomands when saving, so you may have a git auto commit autocommand, a copy file autocommand, and you just want to save without running them. What about changing the name conditionally based in the presence of autoformat or not?

LostNeophyte:
I guess you can change the keybind description to `Save without autocommands`, although beginners may not know that it'll work for formatting then, @kylo252 what do you think?

lua/lvim/core/which-key.lua
```diff
@@ -92,6 +92,7 @@ M.config = function()
         f = { "<cmd>Telescope buffers<cr>", "Find" },
         b = { "<cmd>BufferLineCyclePrev<cr>", "Previous" },
         n = { "<cmd>BufferLineCycleNext<cr>", "Next" },
+        W = { "<cmd>noautocmd w<cr>", "Save without formatting" },
```
kylo252, line 95:
```suggestion
        W = { "<cmd>noautocmd w<cr>", "Save without formatting (noautocmd)" },
```

kylo252 (approved):

PR #3189, opened 2022-10-09
fix(breadcrumbs): correctly ignore markdown files

Nefendi:
# Description

The Markdown file type was incorrectly specified in the breadcrumbs ignore list as `Markdown` with the first letter capitalized. This resulted in creating a winbar for Markdown files.

## How Has This Been Tested?

By simply using LunarVim in a Markdown file.

### DIFF
--- a/lua/lvim/core/breadcrumbs.lua
+++ b/lua/lvim/core/breadcrumbs.lua
@@ -88,7 +88,7 @@ M.winbar_filetype_exclude = {
   "dap-terminal",
   "dapui_console",
   "lab",
-  "Markdown",
+  "markdown",
   "",
 }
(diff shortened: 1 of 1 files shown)


ChristianChiarulli:
why don't we want winbar for markdown?

Nefendi:
Actually, I was thinking the same thing :smile:. I just thought that it was the intention. 

In that case, it would be nice to remove this `Markdown` entry if it does nothing. Or maybe it does, and I am simply unaware of that fact?

Would you like me to create a separate PR with this simple fix?

ChristianChiarulli:
hmm, yea I think we should just remove that

PR #3199, opened 2022-10-10
chore(whichkey): use V instead of L for LunarVim binds

ChristianChiarulli:


kylo252 (commented):
I'm not really in favor of this change, as it will lead to a lot of churn 😞 

we already have `<leader>L` mentioned in various places (docs, issues, discord,.. etc), not to mention that people have gotten used to it by now.

on the other hand, we could move around some bindings in `<leader>l`
```lua
      l = {
        name = "LSP",
        a = { "<cmd>lua vim.lsp.buf.code_action()<cr>", "Code Action" },
        d = { "<cmd>Telescope diagnostics bufnr=0 theme=get_ivy<cr>", "Buffer Diagnostics" },
        -- w = { "<cmd>Telescope diagnostics<cr>", "Diagnostics" }, -- should be moved to <leader>lD or removed
        f = { require("lvim.lsp.utils").format, "Format" },
        i = { "<cmd>LspInfo<cr>", "Info" },
        I = { "<cmd>Mason<cr>", "Mason Info" },
        -- j = { vim.diagnostic.goto_next, "Next Diagnostic" }, -- should be moved to ]d
        -- k = { vim.diagnostic.goto_prev, "Prev Diagnostic" }, -- should be moved to [d
        l = { vim.lsp.codelens.run, "CodeLens Action" },
        q = { vim.diagnostic.setloclist, "Quickfix" }, -- I'm not sure how useful vs (telescope+ctrl_q)
        r = { vim.lsp.buf.rename, "Rename" },
        s = { "<cmd>Telescope lsp_document_symbols<cr>", "Document Symbols" },
        S = { "<cmd>Telescope lsp_dynamic_workspace_symbols<cr>", "Workspace Symbols" },
        -- e = { "<cmd>Telescope quickfix<cr>", "Telescope Quickfix" }, -- what does this even do?
      },

```

ChristianChiarulli:
the reason is for a convention in `starter.lvim` where we're using `L` for language specific binds, can you think of a better bind that we could use there?

ChristianChiarulli:
We're gonna go with `C` instead, we can take a look at the lsp stuff in another issue/PR

PR #3206, opened 2022-10-11
fix: Remove J and K key mappings for move

tobdob:
# Description

Remove the _J_ and _K_ key mappings for moving lines of code since the _J_ mapping overrides the standard _J_ (join lines) action.
Not having the standard _J_ mapping is confusing for new users. _Alt-j_ and _Alt-k_ are still available for moving lines.

## How Has This Been Tested?

- Verified that _J_ (join lines) works in visual block mode.
- Verified that _Alt-j_ and _Alt-k_ can still be used to move lines.

### DIFF
--- a/lua/lvim/keymappings.lua
+++ b/lua/lvim/keymappings.lua
@@ -83,10 +83,6 @@ local defaults = {
   },
 
   visual_block_mode = {
-    -- Move selected line / block of text in visual mode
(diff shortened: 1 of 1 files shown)


ChristianChiarulli:
I agree @kylo252 what do you think?

Almo7aya:
+1 on this PR, I have these disabled in my config, I also prefer to remove  jk and any other mappings for esc as well, they are very confusing

PR #3207, opened 2022-10-12
fix(readme): Updated the update plugins command and resource section

Rishabh672003:


kylo252 (commented):
these commands are actually different, as `LvimSyncCorePlugins` does not touch `lvim.plugins`

It's internally needed to avoid errors when updating lunarvim itself.

Rishabh672003:
well this pr is useless now bye

PR #3221, opened 2022-10-13
fix(breadcrumbs): make winbar_filetype_exclude customizable

abzcoding:
i want to be able to exclude `noice`, `notify` and ... filetypes from breadcrumbs

### DIFF
--- a/lua/lvim/core/breadcrumbs.lua
+++ b/lua/lvim/core/breadcrumbs.lua
@@ -8,6 +8,32 @@ M.config = function()
   lvim.builtin.breadcrumbs = {
     active = true,
     on_config_done = nil,
+    winbar_filetype_exclude = {
+      "help",
+      "startify",
+      "dashboard",
+      "packer",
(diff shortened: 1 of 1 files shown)


vonPB:
I've never tried noice before. Checked this pr out and added `noice` to the plugins and got a bunch of errors on every command.
But should not be related to this specific problem though.
image

abzcoding:
> I've never tried noice before. Checked this pr out and added `noice` to the plugins and got a bunch of errors on every command.
> But should not be related to this specific problem though.
> image

look at my commit messages,
it does exactly have to do with this, just add `vim` to the excluded list
since noice uses `vim` filetype for it's cmdline
https://github.com/abzcoding/lvim/commit/66378a1b93e9482553027da89ba0ec2da8250138

vonPB:
Ah! Great that works.
Just had to remove the `and lvim.builtin.noice.active`
Thanks @abzcoding

PR #3223, opened 2022-10-13
fix: name treesitter source in cmp

danielo515:
# Description

Just added a name for treesitter to cmp to make sure al built-in sources are named

## How Has This Been Tested?

Just use it


abzcoding (approved):
thanks

Issue #3215, opened 2022-10-12
[spike][neotest] should we add neotest?


ChristianChiarulli:
LunarVim would like to provide a simple and standard interface for testing. Neotest seems to be blazing that trail.

- find out if neotest is a good addition to lunarvim
- implement as core plugin if so

JBSK8NC:
I can take it.

kylo252:
## Considerations

#### what are the benefits of adding this to core?

so far, this is only setting up keybinds, which are typically opinionated anyway and are probably better configured by the end-user

#### how does this play with dap?

is there any conflict? is there an integration

#### how to enable extra runners? 

we can **not** add 20 plugins just for test-runner, so this might be better added in `lvim.plugins` where one can add the test-runners they care about

kylo252:
as discussed in https://github.com/LunarVim/LunarVim/pull/3227#issuecomment-1279223228, this will be moved into starter.lvim

PR #3227, opened 2022-10-13
feat: add neotest to the core plugins

JBSK8NC:
# Description

Add neotest to LunarVim.

### DIFF
--- a/lua/lvim/core/builtins/init.lua
+++ b/lua/lvim/core/builtins/init.lua
@@ -22,6 +22,7 @@ local builtins = {
   "lvim.core.lualine",
   "lvim.core.alpha",
   "lvim.core.mason",
+  "lvim.core.neotest",
 }
 
 function M.config(config)
--- a/lua/lvim/core/neotest.lua
+++ b/lua/lvim/core/neotest.lua
@@ -30,8 +30,6 @@ M.setup = function(config)
     o = { "<cmd>lua require('neotest').output.open({ enter = true })<cr>", "Output" },
     S = { "<cmd>lua require('neotest').run.stop()<cr>", "Stop" },
     s = { "<cmd>lua require('neotest').summary.toggle()<cr>", "Summary" },
-    v = { "<cmd>TestVisit<cr>", "Visit" },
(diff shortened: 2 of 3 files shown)


JBSK8NC:
lvim.plugins = {
 "nvim-neotest/neotest-python",
 }


lvim.builtin.test.active = true
local test = require "lvim.core.neotest"
test.setup {
   adapters = {
     require("neotest-python")({
       dap = { justMyCode = false },
     }),
   },
}

JBSK8NC:
resolves #3215

ChristianChiarulli:
I'll try this on stream soon

kylo252 (changes requested):
## Considerations

#### what are the benefits of adding this to core?

so far, this is only setting up keybinds, which are typically opinionated anyway and are probably better configured by the end-user

#### how does this play with dap?

is there any conflict? is there an integration

#### how to enable extra runners? 

we can **not** add 20 plugins just for test-runner, so this might be better added in `lvim.plugins` where one can add the test-runners they care about

kylo252:
I really doubt we can do this in a reasonable and configurable way.

cc: @abzcoding

abzcoding:
> I really doubt we can do this in a reasonable and configurable way.
> 
> cc: @abzcoding

I'm using this on my own lvim, but I'm not sure what are the benefits of adding this to the core,
we added debugging stuff because it was really hard to configure them properly, but neotest you can easily integrate,
so I'm with kylo on this one

JBSK8NC:
@kylo252 Neotest integrates with nvim-dap, check here. https://github.com/nvim-neotest/neotest#strategies. If no debug adapter is provided. Notify will trigger telling the user dap is not configured.

@kylo252 , @abzcoding 
I agree with the assessment. I think neotest in core does not really provide any less setup of the plugin as the user will have to do most of the setup for the specific test adapters in their config.lua. @ChristianChiarulli I was originally thinking it might be better to add this to the starter.lvim. We could ask the individual language contributors add a testing section for the specific languages. I can also make a config there with the keymaps as an example of setting up testing.

PR #3245, opened 2022-10-15
fix(nvim-cmp-lsp): update_capabilities has been deprecated

abzcoding:
more info -> https://github.com/hrsh7th/cmp-nvim-lsp/commit/389f06d3101fb412db64cb49ca4f22a67882e469

### DIFF
--- a/lua/lvim/lsp/init.lua
+++ b/lua/lvim/lsp/init.lua
@@ -25,6 +25,11 @@ local function add_lsp_buffer_keybindings(bufnr)
 end
 
 function M.common_capabilities()
+  local status_ok, cmp_nvim_lsp = pcall(require, "cmp_nvim_lsp")
+  if status_ok then
+    return cmp_nvim_lsp.default_capabilities()
+  end
+
   local capabilities = vim.lsp.protocol.make_client_capabilities()
   capabilities.textDocument.completion.completionItem.snippetSupport = true
   capabilities.textDocument.completion.completionItem.resolveSupport = {
@@ -34,11 +39,6 @@ function M.common_capabilities()
       "additionalTextEdits",
     },
   }
-
-  local status_ok, cmp_nvim_lsp = pcall(require, "cmp_nvim_lsp")
-  if status_ok then
-    capabilities = cmp_nvim_lsp.update_capabilities(capabilities)
-  end
 
   return capabilities
 end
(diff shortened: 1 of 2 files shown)


lua/lvim/lsp/init.lua
```diff
@@ -35,11 +40,6 @@ function M.common_capabilities()
     },
   }
 
-  local status_ok, cmp_nvim_lsp = pcall(require, "cmp_nvim_lsp")
```
kylo252, line 38: the deprecation message is a bit confusing, looking at the code this should work

```lua
  local capabilities = vim.lsp.protocol.make_client_capabilities()
  capabilities.textDocument.completion.completionItem.snippetSupport = true
  local status_ok, cmp_nvim_lsp = pcall(require, "cmp_nvim_lsp")
  if status_ok then
    capabilities = cmp_nvim_lsp.default_capabilities(capabilities)
  end
```

snapshots/default.json
```diff
@@ -15,7 +15,7 @@
     "commit": "3022dbc"
   },
   "cmp-nvim-lsp": {
```
kylo252, line 17: could you also update `neodev`? folke has pushed some [new commits](https://github.com/folke/neodev.nvim/commits/main) :smiley:

abzcoding: sure thing :)

lua/lvim/lsp/init.lua
abzcoding: @kylo252  if we have nvim-cmp-lsp, then we don't need to call `vim.lsp.protocol.make_client_capabilities`

if we don't have it, then we should call and do the overrides

kylo252: but this call is only using adding `.textDocument.completion`, see: https://github.com/hrsh7th/cmp-nvim-lsp/blob/389f06d3101fb412db64cb49ca4f22a67882e469/lua/cmp_nvim_lsp/init.lua#L24

abzcoding: oh that makes sense, let me fix it

kylo252: actually this should work since lspconfig has gotten updated as well, see https://github.com/hrsh7th/cmp-nvim-lsp/commit/389f06d3101fb412db64cb49ca4f22a67882e469 and https://github.com/neovim/nvim-lspconfig/commit/b6d9e427c9fafca5b84a1f429bef4df3ef63244b

kylo252 (approved):
Thanks! :heart:

PR #3261, opened 2022-10-16
feat(config): allow disabling reload-on-save

ChristianChiarulli:


lua/lvim/config/defaults.lua
```diff
@@ -1,5 +1,6 @@
 return {
   leader = "space",
+  reload_config_on_save = false,
```
kylo252, line 3:
```suggestion
  reload_config_on_save = true,
```
I don't see the need to disable this by default, as that will make this PR a breaking change.
+
we worked so hard to have this standout feature be stable, this will get it buried :smiling_face_with_tear:

lua/lvim/core/autocmds.lua
```diff
@@ -181,6 +163,23 @@ function M.toggle_format_on_save()
...
+  local user_config_file = require("lvim.config"):get_user_config_path()
+
+  if vim.loop.os_uname().version:match "Windows" then
+    -- autocmds require forward slashes even on windows
+    user_config_file = user_config_file:gsub("\\", "/")
+  end
+  vim.api.nvim_create_autocmd("BufWritePost", {
+    group = "_general_settings",
+    pattern = user_config_file,
+    desc = "Trigger LvimReload on saving " .. vim.fn.expand "%:~",
```
kylo252, line 176:
```suggestion
    desc = "Trigger LvimReload on saving config.lua",
```

ChristianChiarulli:
@kylo252 updated to true by default, I wanted to be able to save my config without compiling since we have that colorscheme bug that's been annoying me lately.

kylo252 (approved):

abzcoding (approved):

PR #3218, opened 2022-10-13
feat: display changelog on `:LvimUpdate`

Almo7aya:
Everytime I run `:LvimUpdate` I have to go to github and check the commits log on LunarVim repo, so I thought it would be nice if lvim displays them after running `:LvimUpdate`, the PR still WIP, waiting for your thoughts on this

image

PR #3270, opened 2022-10-17
feat: display changelog on :LvimUpdate

Almo7aya:
instead of https://github.com/LunarVim/LunarVim/pull/3218
Everytime I run `:LvimUpdate` I have to go to github and check the commits log on LunarVim repo, so I thought it would be nice if lvim displays them after running `:LvimUpdate`, the PR still WIP, waiting for your thoughts on this

image

### DIFF
--- a/lua/lvim/utils/hooks.lua
+++ b/lua/lvim/utils/hooks.lua
@@ -1,6 +1,7 @@
 local M = {}
 
 local Log = require "lvim.core.log"
+local Info = require "lvim.core.info"
 local in_headless = #vim.api.nvim_list_uis() == 0
 
 function M.run_pre_update()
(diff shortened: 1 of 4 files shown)


lua/lvim/utils/hooks.lua
```diff
@@ -68,14 +69,15 @@ function M.run_post_update()
   M.reset_cache()
 
   Log:debug "Syncing core plugins"
-  require("lvim.plugin-loader").sync_core_plugins()
+  require("lvim.plugin-loader").sync_core_plugins(function()
```
kylo252, line 72: this is the wrong function, you're probably thinking of  `update_base_lvim()`

lua/lvim/utils/hooks.lua
```diff
@@ -68,14 +69,15 @@ function M.run_post_update()
...
+    Info.make_changelog()
+  end)
 
   if not in_headless then
     vim.schedule(function()
       if package.loaded["nvim-treesitter"] then
         vim.cmd [[ TSUpdateSync ]]
       end
-      -- TODO: add a changelog
       vim.notify("Update complete", vim.log.levels.INFO)
```
kylo252, line 81:
```suggestion
		  Info.make_changelog()
      vim.notify("Update complete", vim.log.levels.INFO)
```

lua/lvim/utils/hooks.lua
Almo7aya: no I put it there on purpose, because putting `make_changelog` in `update_base_lvim()` or in line 81, displays the window before packer's window and when packer window kicks in it overlaps with changelog window, so showing it after packer is done makes more sense

PR #3298, opened 2022-10-19
fix: don't overide in terminal (c-l clears terminal)

ChristianChiarulli:


kylo252 (commented):
hmm, this will make it so you can't navigate back and forth from a _split_ terminal. 

I have two suggestions instead:
- unmap `<C-l>` if you're in a float mode
- add an additional binding in your shell that you can use as a back up, I personally use `<C-a><C-l>`

ChristianChiarulli:
@kylo252 is there an easy way to turn this off, adding this to the FAQ on the site may be enough

PR #3300, opened 2022-10-19
refactor(plugins)!: remove nvim-notify from core

### LINKED ISSUE #3294: remove nvim-notify from the core plugins
Unfortunately, [nvim-notify](https://github.com/rcarriga/nvim-notify) has been causing way too many issues and instabilities.

several plugins are now using `vim.notify` notice that would trigger in the middle of packer's operations, at that point `nvim-notify` lost its state or something, and you get crazy errors like this

<details>
<summary>expand</summary>

```gdb
Error executing vim.schedule lua callback: ...ker/start/nvim-notify/lua/notify/serv
ice/buffer/init.lua:98: Vim(set):E474: Invalid argument: colorcolumn=value
stack traceback:
        [C]: in function 'nvim_buf_set_option'
        ...ker/start/nvim-notify/lua/notify/service/buffer/init.lua:98: in function 'render'
        ...ack/packer/start/nvim-notify/lua/notify/service/init.lua:53: in function 'push'
        ...m/site/pack/packer/start/nvim-notify/lua/notify/init.lua:209: in function 'impl'
        ...start/structlog.nvim/lua/structlog/sinks/nvim_notify.lua:49: in function 'write'
        ...ack/packer/start/structlog.nvim/lua/structlog/logger.lua:58: in function 'log'
        ...ack/packer/start/structlog.nvim/lua/structlog/logger.lua:64: in function 'log'
        ...cedraz-/.local/share/lunarvim/lvim/lua/lvim/core/log.lua:155: in function 'add_entry'
        ...cedraz-/.local/share/lunarvim/lvim/lua/lvim/core/log.lua:198: in function 'info'
        ...raz-/.local/share/lunarvim/lvim/lua/lvim/utils/hooks.lua:22: in function 'run_on_packer_comp
lete'
        ...z-/.local/share/lunarvim/lvim/lua/lvim/plugin-loader.lua:50: in function ''
        vim/_editor.lua: in function <vim/_editor.lua:0>

```

</details>

it feels like so many plugins are spamming `vim.notify` to the point where it's hard to keep track of all of these popups.. :disappointed:  

also, we've had to create sketchy workarounds just for the sake of having this plugin

<details>
<summary>expand</summary>

https://github.com/LunarVim/LunarVim/blob/c015d0feeed35b34d84558a8e4a49549b3b67781/lua/lvim/core/nvimtree.lua#L155-L166

</details>

here are some related issues and complaints I've managed to find
- https://github.com/LunarVim/LunarVim/issues/3265
- https://github.com/LunarVim/LunarVim/issues/2713
- https://github.com/LunarVim/LunarVim/issues/2242
- https://github.com/LunarVim/LunarVim/issues/1821
- https://github.com/LunarVim/LunarVim/issues/2778
- https://github.com/LunarVim/LunarVim/issues/2960
- https://github.com/LunarVim/LunarVim/issues/2734
- https://github.com/LunarVim/LunarVim/issues/2738
- https://github.com/LunarVim/LunarVim/issues/2673

mhuggins7278:
# Description

removing nvim-notify 

fixes #3294

## How Has This Been Tested?

- Run command `:mycommand`
- Check logs
Tested locally 
The only issue i ran into is as follows. 

I had the following line in my config
```lua
lvim.buitlin.notify.enable = true
```
which resulted in the following error
2022-10-19 21:21:47 [WARN ] lvim: "Invalid configuration: /Users/MHuggins/.config/lvim/lua/user/options.lua:30: attempt to index field 'notify' (a nil value)" file="lua/lvim/config/init.lua", line=119
 

simply removing the line from my config resolved the issue not sure if there is a better way to handle that


kylo252:
> simply removing the line from my config resolved the issue not sure if there is a better way to handle that

we have a deprecation handler in `lvim/config/init.lua` that we need to use. 

I'm not sure if it's necessary to create a meta-table to handle nested options 🤔

@abzcoding, wdyt?

abzcoding:
> > simply removing the line from my config resolved the issue not sure if there is a better way to handle that
> 
> we have a deprecation handler in `lvim/config/init.lua` that we need to use.
> 
> I'm not sure if it's necessary to create a meta-table to handle nested options 🤔
> 
> @abzcoding, wdyt?

agreed, we should alert users about the deprecation of `nvim-notify`

mhuggins7278:
Ok I'll take a look and update the PR

mhuggins7278:
I've added the deprecation notice 
image

lua/lvim/core/nvimtree.lua
```diff
@@ -147,24 +147,11 @@ function M.setup()
     return
   end
 
-  local status_ok_1, utils = pcall(require, "nvim-tree.utils")
+  local status_ok_1 = pcall(require, "nvim-tree.utils")
   if not status_ok_1 then
     return
   end
```
kylo252, line 153:
```suggestion
```

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### <!-- 1 --> Features
 
+- _(notify)_ remove nvim.notify ([#3300](https://github.com/LunarVim/LunarVim/pull/3300))
```
kylo252, line 9:
```suggestion
```

you can skip this for now :)

lua/lvim/core/nvimtree.lua
mhuggins7278: cool ty.. not sure why i just didn't remove the whole function initially :-(

ChristianChiarulli (approved):

ChristianChiarulli:
LGTM

danielo515:
Why not just deactivate by default so it is easier to activate for those who want?

kylo252:
> Why not just deactivate by default so it is easier to activate for those who want?

because that means that lunarvim is _still_ maintaining compatibility, which is precisely what we want to avoid here. 

If it works, great! otherwise, it's easier for users to see the ~~problems~~ caveats with using this plugin, and then they're free to set whatever options fits their needs.

latipun7:
There's a leftover here:
https://github.com/LunarVim/LunarVim/blob/25c4241b07d2036fc5d0d4c18be43a16a5370ef0/lua/lvim/core/which-key.lua#L232

PR #3301, opened 2022-10-19
update readme

SingularisArt:


umop3plsdn:
maybe recenter it? Also I put a quick filler in for the tagline and think maybe "An IDE layer for Neovim with sane defaults. Completely free and community driven." is better in that spot?

SingularisArt:
Something like this:
Screenshot_2022-10-19_18-24-06

I'm still playing around with the colors. I'm thinking of just copying the colors from the tokyonight-night pallete. IDK, what are your thoughts?

OkelleyDevelopment:
@SingularisArt  thinking: can't imagine the colors would need to be adjusted much more. Also like the tag line idea that @umop3plsdn suggested :100:

PR #3319, opened 2022-10-22
feat(breadcrumbs): add exclude check for file path (buf name)

latipun7:
# Description

[Noice](https://github.com/folke/noice.nvim) is setting some of popup window with filetype `lua`, `vim`, `regex`, and `sh`. If I don't exclude that filetype, breadcrumbs and noice are error. But, I don't want to exclude that filetype. That's why I check with the buffer name instead.

So, this PR is dirty check from me (lack of experience with lua and neovim api) to also exclude breadcrumbs from certain file path or buffer name.

## How Has This Been Tested?

- check this in my local environment. `lua` file has working breadcrumbs and no error when I open noice cmdline_popup.

### DIFF
--- a/lua/lvim/core/breadcrumbs.lua
+++ b/lua/lvim/core/breadcrumbs.lua
@@ -8,6 +8,7 @@ M.config = function()
   lvim.builtin.breadcrumbs = {
     active = true,
     on_config_done = nil,
+    winbar_path_contains_exclude = { "noice://" },
     winbar_filetype_exclude = {
       "help",
       "startify",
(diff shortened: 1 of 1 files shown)


latipun7:
**Update:**

With this commit from noice (https://github.com/folke/noice.nvim/commit/df02aa2d4d805e1a87cd39b858f602be23147851), my use case is no longer relevant. But if this PR is still reviewed, it would be nice addition too, imo.

Please give feedback. Thank you.

abzcoding:
> Please give feedback, @abzcoding. Thank you. Sorry for the ping 🙏

hey man, I don't use breadcrumbs, @ChristianChiarulli is probably a better choice for giving feedback here :)

ChristianChiarulli:
let me try out noice, and I'll get back to you.

cpea2506:
How is this going on?

latipun7:
Now I don't use breadcrumbs anymore since this is too long to be merged, and I don't want to maintain branch for long time🙏
Barbecue is great replacement for me.

Issue #3311, opened 2022-10-21
LunarVim permanently and completely unusable until reinstall


twharmon:
### Problem description

LunarVim works well after fresh install, but after a few hours I always run into problems.

1. Use LunarVim
2. Wait for very long list of errors that look like this:

```
Unable to set logger's level to warn
Error detected while processing /Users/travis/.local/share/lunarvim/lvim/init.lu
a:
E5113: Error while calling lua chunk: ...ravis/.local/share/lunarvim/lvim/lua/lv
im/core/theme.lua:107: Vim(colorscheme):E185: Cannot find color scheme 'tokyonig
ht'
stack traceback:
        [C]: in function 'cmd'
        ...ravis/.local/share/lunarvim/lvim/lua/lvim/core/theme.lua:107: in func
tion 'setup'
        /Users/travis/.local/share/lunarvim/lvim/init.lua:19: in main chunk
Error detected while processing /Users/travis/.local/share/lunarvim/site/pack/pa
cker/start/cmp_luasnip/after/plugin/cmp_luasnip.lua:
E5113: Error while calling lua chunk: ...ck/packer/start/cmp_luasnip/after/plugi
n/cmp_luasnip.lua:1: module 'cmp' not found:
        no field package.preload['cmp']
        no file './cmp.lua'
        no file '/opt/homebrew/share/luajit-2.1.0-beta3/cmp.lua'
        no file '/usr/local/share/lua/5.1/cmp.lua'
        no file '/usr/local/share/lua/5.1/cmp/init.lua'
        no file '/opt/homebrew/share/lua/5.1/cmp.lua'
        no file '/opt/homebrew/share/lua/5.1/cmp/init.lua'
```

### LunarVim version

E492: Not an editor command: LvimVersion

### Neovim version (>= 0.8.0)

NVIM v0.8.0

### Operating system/version

macOS 12.5

### Steps to reproduce

1. Use LunarVim
2. Wait for very long list of errors that look like this:

```
Unable to set logger's level to warn
Error detected while processing /Users/travis/.local/share/lunarvim/lvim/init.lu
a:
E5113: Error while calling lua chunk: ...ravis/.local/share/lunarvim/lvim/lua/lv
im/core/theme.lua:107: Vim(colorscheme):E185: Cannot find color scheme 'tokyonig
ht'
stack traceback:
        [C]: in function 'cmd'
        ...ravis/.local/share/lunarvim/lvim/lua/lvim/core/theme.lua:107: in func
[...]

PR #3327, opened 2022-10-24
fix(theme): do not fallback to tokyonight if no user's theme found

cpea2506:
# Description
This fixes the problem that `tokyonight` colorscheme cannot be found.

Related #3311


lua/lvim/core/theme.lua
```diff
@@ -100,7 +100,7 @@ M.setup = function()
   local colors = vim.api.nvim_get_runtime_file(("colors/%s.*"):format(lvim.colorscheme), false)
   if #colors == 0 then
     Log:warn(string.format("Could not find '%s' colorscheme", lvim.colorscheme))
```
kylo252, line 102: just to be on the safe side :smiling_face_with_tear: 
```suggestion
    Log:debug(string.format("Could not find '%s' colorscheme", lvim.colorscheme))
```

kylo252 (approved):
Thanks!

PR #3332, opened 2022-10-24
fix(finders): use lunarvim basedir

matleh:
# Description

When LunarVim is cloned in ~/.config/nvim (to make it play with Neovide), the LunarVim which-key shortcuts `Lf` and `Lg` did not work, because a wrong base-directory was used.

fixes #(issue)

## How Has This Been Tested?

I tested the shortcuts in both settings - when LunarVim is cloned to ~/.config/nvim and when installed the usual way.

- Run command `:mycommand`
- Check logs
- ...


kylo252 (approved):
Thanks! :rocket:

PR #3337, opened 2022-10-25
fix: call proper log function in notify override

danielo515:
related to #3202

# Description

There is a wrong call when overriding vim.notify

There is no issue related, but there is a discussion #3202 

## How Has This Been Tested?

- Not tested


kylo252 (approved):
LGTM! 🚀

PR #3354, opened 2022-10-27
fix(lsp): do a nil check before string matching autocmd desc

abzcoding:
Apparently in the latest neovim build, filetype.lua or packer.nvim is working a bit different, 
for example when I want to load a plugin only in certain filetypes, this happens:
Screenshot 2022-10-27 at 12 49 21 PM

because some autocmds don't have a description:
Screenshot 2022-10-27 at 12 49 29 PM


kylo252 (approved):
LGTM! :rocket:

PR #3361, opened 2022-10-29
fix(lsp): don't highlight strings when treesitter is enabled

emxxjnm:
Do not highlight strings when treesitter is enabled

### DIFF
--- a/lua/lvim/lsp/utils.lua
+++ b/lua/lvim/lsp/utils.lua
@@ -3,6 +3,28 @@ local M = {}
 local tbl = require "lvim.utils.table"
 local Log = require "lvim.core.log"
 
+local function highlight_references()
+  local status_ok, ts_utils = pcall(function()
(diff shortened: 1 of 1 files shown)


kylo252:
Sorry for the delay!

This isn't necessary since:
- `lvim.lsp.document_highlight` has been disabled by default since https://github.com/LunarVim/LunarVim/commit/518b1d4167162a54a6e76784038d30191613b76d
- this is [currently controlled by `illuminate`](https://github.com/LunarVim/LunarVim/blob/eabf6c6990cb7ac9416b1733cf59c36c85f50777/lua/lvim/core/illuminate.lua#L9-L13), but since its future is not guaranteed you can still define your own autocmd
- lastly, we should try to keep LSP tooling independent from tree-sitter as much as possible

Please let me know if that still doesn't solve your use-case!

PR #3366, opened 2022-10-29
fix(lsp): launch setup only on installed server

cpea2506:
# Description

Do not run `lsp` setup on server that is not installed as describe in the error below.

Screenshot 2022-10-30 at 06 27 31

### DIFF
--- a/lua/lvim/lsp/manager.lua
+++ b/lua/lvim/lsp/manager.lua
@@ -119,10 +119,10 @@ function M.setup(server_name, user_config)
           end)
         end
       end)
-      return
     else
       Log:debug(server_name .. " is not managed by the automatic installer")
     end
+    return
   end
 
   local config = resolve_config(server_name, resolve_mason_config(server_name), user_config)


lua/lvim/lsp/manager.lua
```diff
@@ -119,10 +119,10 @@ function M.setup(server_name, user_config)
           end)
         end
       end)
-      return
     else
       Log:debug(server_name .. " is not managed by the automatic installer")
     end
+    return
```
kylo252, line 125: I'm not clear on the difference here, can you please provide a workflow example of how to test this?

it might've been that the internal mason hook that we're using is incorrect now, 
https://github.com/LunarVim/LunarVim/blob/b190f045e896fe5f1646e0807ab0bea008ee8950/lua/lvim/lsp/manager.lua#L113-L120

cpea2506: The difference here is we should return if server is not installed regardless of `automatic_íntallation` is enable or not.

```lua
if not registry.is_installed(pkg_name) then
    if should_auto_install(server_name) then
        -- snip
        return
    else
        Log:debug(server_name .. " is not managed by the automatic installer")
    end
end

local config = resolve_config(server_name, resolve_mason_config(server_name), user_config)
launch_server(server_name, config)
```

As you can see from the code above from the original implement. Whenever we fall into the `else` block it can still go out and run `launch_server` which will cause warning.

## How to test
1. Create and open typescript file.
2. See the warning popup.

kylo252: btw, the problem with `glint` that you're hitting is a bug caused by the CI not telling me when that server was duplicated, but that was fixed in https://github.com/LunarVim/LunarVim/commit/b190f045e896fe5f1646e0807ab0bea008ee8950, so you should run `LvimCacheReset` to re-generate those templates.

> The difference here is we should return if server is not installed regardless of `automatic_íntallation` is enable or not.

that sounds about right, we just need to make sure we aren't forgetting any (exotic) workflow here.

Issue #3257, opened 2022-10-16
fix highlighting for statusline and breadcrumbs when syncing & compiling config


ChristianChiarulli:
AC:

- after sync & compile the highlights for the statusline and breadcrumbs is unchanged

kylo252:
I'm not sure what to look for exactly, can you provide some screenshots?

PR #3374, opened 2022-10-31
feat: use our own colorscheme based on tokyonight

ChristianChiarulli:
# Description

summary of the change

fixes #(issue)

## How Has This Been Tested?

- Run command `:mycommand`
- Check logs
- ...

### DIFF
--- a/lua/lvim/config/defaults.lua
+++ b/lua/lvim/config/defaults.lua
@@ -1,7 +1,7 @@
 return {
   leader = "space",
   reload_config_on_save = true,
-  colorscheme = "tokyonight",
+  colorscheme = "lunar",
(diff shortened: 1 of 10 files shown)


ChristianChiarulli:
This PR will give us more control over the colorscheme and will also allow the user to use tokyonight however they want. This should also mostly fix: https://github.com/LunarVim/LunarVim/issues/3257

kylo252 (changes requested):
I'm afraid that swapping out tokyonight is rather a big breaking change especially considering it involves removing `lvim.builtin.theme`. 

That said, this PR is basically two parts, one related to the hooks and the other related to the theme itself, so let's try to break it as such.

PR #3376, opened 2022-10-31
fix: statusline and breadcrumbs hls reload with config reload

ChristianChiarulli:


lua/lvim/core/autocmds.lua
```diff
@@ -99,6 +99,28 @@ function M.load_defaults()
...
+          end
+          local statusline_hl = vim.api.nvim_get_hl_by_name("StatusLine", true)
+          local cursorline_hl = vim.api.nvim_get_hl_by_name("CursorLine", true)
+          local normal_hl = vim.api.nvim_get_hl_by_name("Normal", true)
+          vim.api.nvim_set_hl(0, "CmpItemKindCopilot", { fg = "#6CC644" })
+          vim.api.nvim_set_hl(0, "CmpItemKindTabnine", { fg = "#CA42F0" })
+          vim.api.nvim_set_hl(0, "CmpItemKindCrate", { fg = "#F64D00" })
+          vim.api.nvim_set_hl(0, "CmpItemKindEmoji", { fg = "#FDE030" })
+          vim.api.nvim_set_hl(0, "SLCopilot", { fg = "#6CC644", bg = statusline_hl.background })
+          vim.api.nvim_set_hl(0, "SLGitIcon", { fg = "#E8AB53", bg = cursorline_hl.background })
```
kylo252, line 118: are these colors specific to tokyonight? if so, we should move this into the `setup` function for it, although I feel like there should be an easier to do this in that case.

lua/lvim/core/autocmds.lua
```diff
@@ -99,6 +99,28 @@ function M.load_defaults()
         end,
       },
     },
+    {
+      "ColorScheme",
+      {
+        group = "_lvim_colorscheme",
+        callback = function()
```
kylo252, line 106: since there's no `ColorSchemePost` we should probably defer this function just to be safe

```suggestion
        callback = function(vim.schedule(function()
```

lua/lvim/core/autocmds.lua
```diff
@@ -99,6 +99,28 @@ function M.load_defaults()
...
+          if lvim.builtin.breadcrumbs.active then
```
kylo252, line 107: I wonder if this is always safe, maybe we should add it in a separate autocmd when we know that `lvim.builtin.breadcrumbs` is available

lua/lvim/core/autocmds.lua
ChristianChiarulli: these are more lunarvim specific than tokyonight specific

lua/lvim/core/autocmds.lua
ChristianChiarulli: I can't imagine a case where it's not safe, if breadcrumbs is on reload it real quick on colorscheme change, if not ignore it.

PR #3378, opened 2022-10-31
feat: use our own colorscheme, decouple from tokyonight

ChristianChiarulli:

Issue #3313, opened 2022-10-21
A plea for stability


mandreyel:
### Problem description

This is a general issue rather than a report for any specific bug, as there have been too many to count. I have been using lvim for the better part of this year, and it has mostly been a great experience and I applaud the team for creating such an easy to use plug-and-play IDE experience for vim. However, in recent months after almost every lvim update my setup broke in various ways.

These have usually been breaking changes, ranging from small to big ones. Just a few examples:
* change in the default keymaps (e.g. the recent remapping of `<C-t>` to `<C-\>` for `ToggleTerm`)
* breaking previous rust-tools.nvim setup https://github.com/LunarVim/LunarVim/issues/2894
* introducing taplo, an lsp for toml (likely came with the mason change as far as I could tell), which started auto-formatting Cargo.toml files out of the blue (I never asked for it)
* introducing nvim-notify that spammed notifications with entire window overlays (what happened to appreciating the subtlety of the vim ethos?) https://github.com/LunarVim/LunarVim/issues/3294
* the installer breaking https://github.com/LunarVim/LunarVim/issues/3290
* in a recent reinstall just now, with default config, the which-key key binding (leader/space) either doesn't work or has been remapped to a new key binding
* and many more, but I mostly forgot them or others have reported on them

With each release, I found myself spending about an afternoon fixing my config. The whole reason for my moving to lvim was that I no longer had time to maintain my own nvim config and so I had decided to leave it to people who are more passionate about and more involved with the recent developments of the nvim ecosystem. A dev's editor is their number one tool and it must be absolutely rock solid and never even give the faintest impression that it cannot be relied upon. This is what I'd desperately like LunarVim to be because I love the idea and have for the most part enjoyed using it.

### LunarVim version

master-36c8bde

### Neovim version (>= 0.8.0)

NVIM v0.8.0

### Operating system/version

macOS 12.6

### Steps to reproduce

default install will do

### support info

 Language client log: /Users/marcloedi/.local/state/nvim/lsp.log
 Detected filetype:   lua
 
 0 client(s) attached to this buffer: 
 
 Configured servers list: 

    __                          _    ___         
   / /   __  ______  ____ _____| |  / (_)___ ___ 
  / /   / / / / __ \/ __ `/ ___/ | / / / __ `__ \
 / /___/ /_/ / / / / /_/ / /   | |/ / / / / / / /
/_____/\__,_/_/ /_/\__,_/_/    |___/_/_/ /_/ /_/ 

[...]

PR #3384, opened 2022-11-01
refactor(theme)!: decouple tokyonight options

kylo252:
# Description

#### (refactor!) decouple tokyonight options

moves `lvim.builtin.theme.options` into `lvim.builtin.theme.tokyonight.options` since they are
specific to how `tokyonight` is configured.

#### (refactor) better deprecation handling

- all the deprecated options are now in a single file
- use `__newindex` to handle new entries
- this gives us much better flexibility (e.g. this can handle `lvim.builtin.theme.options.dummy`)
- and it allows us to generate a trace where that action was last defined

#### misc

- chore: remove merge duplicate

related #3313, #3378

## How Has This Been Tested?

```lua
-- add this somewhere in your config
lvim.builtin.theme.options.dim_inactive = true
```

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -10,7 +10,7 @@ end
 
 require("lvim.bootstrap"):init(base_dir)
 
-require("lvim.config"):load()
+require("lvim.config").load()
 
 local plugins = require "lvim.plugins"
 
--- a/lua/lvim/config/_deprecated.lua
+++ b/lua/lvim/config/_deprecated.lua
@@ -25,6 +25,15 @@ function M.handle()
   }
 
   ---@deprecated
+  lvim.builtin.theme.options = {}
+  setmetatable(lvim.builtin.theme.options, {
+    __newindex = function(_, k, v)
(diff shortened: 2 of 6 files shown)


LostNeophyte:
`lvim.builtin.dashboard.abc = 1` does work,
 but `lvim.builtin.dashboard.abc.def = 1` does not
We could use [automagic tables](http://lua-users.org/wiki/AutomagicTables) to handle it
They allow this:
```lua
a = AutomagicTable()
a.b.c.d = "a.b and a.b.c are automatically created"
```

ChristianChiarulli (approved):

abzcoding (approved):

kylo252:
> We could use [automagic tables](http://lua-users.org/wiki/AutomagicTables) to handle it

@LostNeophyte, this seems quite useful. 

Let's iterate on that in a separate PR, in case there are some (performance) gotchas involved.

PR #3386, opened 2022-11-01
chore(dap-ui): move dap-ui config to lvim.builtin.dap

igorstr:
# Description
Move `dap-ui` config to `lvim.builtin.dap` to make it configurable for user.


kylo252 (approved):
Thanks!

PR #3408, opened 2022-11-03
fix(installer): use quotes in set-alias

smarttommyau:
This can prevent spacing in the path causeing the Set-Alias error.

# Description

summary of the change
Simply add a '' to prevent the error/bug.

fixes #(issue)
Fix error when the path to the lvim.ps1 have spacing.(e.g. the username has spaces)
## How Has This Been Tested?

- Run the setup script, then run the```. $profile```, see if there is any error message.


kylo252 (approved):
Thanks!

PR #3419, opened 2022-11-04
refactor(config): better deprecation handling

kylo252:
# Description

### (refactor) better deprecation handling

- all the deprecated options are now in a single file
- use `__newindex` to handle new entries
- this gives us much better flexibility (e.g. this can handle `lvim.builtin.theme.options.dummy`)
- and it allows us to generate a trace where that action was last defined

originally added in #3384

### How Has This Been Tested?

```lua
-- add this somewhere in your config
lvim.builtin.theme.options.dim_inactive = true
```


ChristianChiarulli (approved):
tested LGTM

PR #3427, opened 2022-11-05
fix(config): use a minimal bootstrap for mason

kylo252:
# Description

fixes the issue with early launching of servers installed by mason

## How Has This Been Tested?

```lua
-- in config.lua
print(vim.env.PATH)
```


LostNeophyte (approved):
LGTM :)

PR #3435, opened 2022-11-07
feat: portal integration to jumping between  buffers history

edenac:
# Description

portal plugin added to go to back or forward in buffers history with alt + i and alt + o  beeing like easy-motion of buffers history:
https://github.com/cbochs/portal.nvim

## How Has This Been Tested?
i just added portal plugin and i set the keymaps (alt + i , alt + o) to jumping between buffers history

- Run command `:lua require("portal").jump_backward()`
- Run command `:lua require("portal").jump_forward()`
- I just assigned that both functions in keymapps alt + i  and alt + o

### DIFF
--- a/lua/lvim/keymappings.lua
+++ b/lua/lvim/keymappings.lua
@@ -59,6 +59,10 @@ local defaults = {
     ["<A-j>"] = ":m .+1<CR>==",
     ["<A-k>"] = ":m .-2<CR>==",
 
+    -- portal to jump between buffers history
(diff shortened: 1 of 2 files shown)


edenac:
sc

cpea2506:
Do you think it should be managed to be put into core plugins?

LostNeophyte:
It looks like it's very easy to add it from config.lua. I don't think this should be in core, maybe add it to the website in extra plugins

edenac:
ok, thanks  for the feedback guys

PR #3471, opened 2022-11-13
fix(lsp): some servers have dynamic commands

opalmay:
Fixes regression as a result of #3366
This prevented `powershell-editor-services` lsp and possibly others from attaching


kylo252 (approved):
LGTM! :rocket:

PR #3508, opened 2022-11-22
fix(uninstaller): add separate flag to remove user config

Vaisakhkm2625:
# Description

By default when we uninstall lunarvim, the uninstaller script removes user configuration,which normally located in .config/lvim/  and unlike installer script, it will not make a *.old file

This is means those configuration made by the user is completely gone, and is not reversible, if  anytime user wants to install lvim back. ( :*(  just now i lost it, glad i had a backup of system, but it was old).

This is unexpected because most package managers don't remove user configurations

So i made default behaviour to not to remove user config files, and  but added it as a  additional option for those who wanted it. @Lost Neophyte from discord proposed this idea to make it additional flag rather than (y/n) question

fixes #(issue)

## How Has This Been Tested?

- Run command `bash uninstall.sh` for not removing .config/lvim directory
- Run command `bash uninstall.sh --remove-config` to remove .config/lvim directory

### DIFF
--- a/utils/installer/uninstall.sh
+++ b/utils/installer/uninstall.sh
@@ -2,6 +2,7 @@
 set -eo pipefail
 
 ARGS_REMOVE_BACKUPS=0
+ARGS_REMOVE_CONFIG=0
 
 declare -r XDG_DATA_HOME="${XDG_DATA_HOME:-"$HOME/.local/share"}"
 declare -r XDG_CACHE_HOME="${XDG_CACHE_HOME:-"$HOME/.cache"}"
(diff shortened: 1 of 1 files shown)


ChristianChiarulli (approved):

Vaisakhkm2625:
:*) Thanks for merging this

This is my first pull request in my life ever.... 
glad i can started it with your project

Vaisakhkm2625:
`echo "    --remove-config                  Remove old backup folders as well"`

should have been 
`echo "    --remove-config                  Remove user config as well"`

better i will give one more pull request as well

PR #3511, opened 2022-11-22
fix(uninstaller): correct a sentence in help

Vaisakhkm2625:
# Description
in pr #3508, i have given wrong description for --remove-config flag in help menu fixing it

summary of the change
 `Remove old backup folders as well` changed to ` Remove user config files  as well` 

fixes #(issue)

### DIFF
--- a/utils/installer/uninstall.sh
+++ b/utils/installer/uninstall.sh
@@ -24,7 +24,7 @@ function usage() {
   echo ""
   echo "Options:"
   echo "    -h, --help                       Print this help message"
-  echo "    --remove-config                  Remove user config files  as well"
+  echo "    --remove-config                  Remove user config files as well"
   echo "    --remove-backups                 Remove old backup folders as well"
 }
 
--- a/utils/installer/uninstall.sh
+++ b/utils/installer/uninstall.sh
@@ -24,7 +24,7 @@ function usage() {
   echo ""
   echo "Options:"
   echo "    -h, --help                       Print this help message"
-  echo "    --remove-config                  Remove old backup folders as well"
+  echo "    --remove-config                  Remove user config files  as well"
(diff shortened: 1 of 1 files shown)


Vaisakhkm2625:
sorry for not checking it, i only tested if the install and uninstall works correctly that time

utils/installer/uninstall.sh
```diff
@@ -24,7 +24,7 @@ function usage() {
   echo ""
   echo "Options:"
   echo "    -h, --help                       Print this help message"
-  echo "    --remove-config                  Remove old backup folders as well"
+  echo "    --remove-config                  Remove user config files  as well"
```
LostNeophyte, line 27:
```suggestion
  echo "    --remove-config                  Remove user config files as well"
```

LostNeophyte (changes requested):

LostNeophyte (approved):
Thanks :)

utils/installer/uninstall.sh
Vaisakhkm2625: that double space was accidental

but i left there because 'as well' in both lines aligned correctly😅
image

but better to remove

Vaisakhkm2625:
```
  echo "    --remove-config                  Remove user config files as well"
  echo "    --remove-backups                 Remove old backup folders as well"
  echo ""

  echo "    --remove-config                  Remove user config folder as well"
  echo "    --remove-backups                 Remove old backup folders as well"

  echo "    --remove-config                  Remove user config files  as well"
  echo "    --remove-backups                 Remove old backup folders as well"

```

folder would have been better.. :)

LostNeophyte: doesn't matter it's just the uninstaller, it doesn't have to align
it's merged, don't worry about it, thanks for the pr :)

PR #3357, opened 2022-10-28
fix(breadcrumbs) Performance issue, change CursorMoved event to CursorHoldI

liujoey:
# Description

Make breadcrumb update less frequent by changing the `CursorMoved` event to `CursorHoldI`
I think having a short delay on updating the breadcrumbs is more acceptable than having a huge lag on moving cursor. 

fixes #3353

PR #3544, opened 2022-11-28
fix(breadcrumbs): add CursorMoved makes winbar seems 10 times faster

nullchilly:
# Description

Related: https://github.com/glepnir/lspsaga.nvim/issues/334

## How Has This Been Tested?

Enabled `lvim.builtin.breadcrumbs.active = true` in `config.lua`


LostNeophyte:
It was removed for performance reasons https://github.com/LunarVim/LunarVim/pull/3357

nullchilly:
In lspsaga this is fine because the output is cached

LostNeophyte:
We don't use lspsaga

nullchilly:
Alright

PR #3589, opened 2022-12-09
feat(cmp): add on_config_done callback

mcartaud:
# Description

User can use a `on_config_done` property in nvim-cmp config. It allow user to hook on the plugin and make additionnal configuration after the plugin is setup.

## How has this been tested

1. See that completion shows up when writing a commit for exemple
2. Add 
```lua
lvim.builtin.cmp.on_config_done = function(cmp)
  cmp.setup.filetype("gitcommit", {
    enabled = false,
  })
end
```
In the nvim config file
4. Reload LunarVim
5. Commit something in a repo and see that completion doesn't open

Signed-off-by: Mathieu Cartaud <dev@mcartaud.com>

### DIFF
--- a/lua/lvim/core/cmp.lua
+++ b/lua/lvim/core/cmp.lua
@@ -129,6 +129,7 @@ M.config = function()
 
   lvim.builtin.cmp = {
     active = true,
+    on_config_done = nil,
     enabled = function()
       local buftype = vim.api.nvim_buf_get_option(0, "buftype")
       if buftype == "prompt" then
@@ -366,6 +367,9 @@ function M.setup()
         sources = option.sources,
       })
     end
+  end
+  if lvim.builtin.cmp.on_config_done then
+    lvim.builtin.cmp.on_config_done(cmp)
   end
 end
 
--- a/lua/lvim/core/cmp.lua
+++ b/lua/lvim/core/cmp.lua
@@ -368,6 +368,7 @@ function M.setup()
       })
     end
   end
+  
   if lvim.builtin.cmp.on_config_done then
     lvim.builtin.cmp.on_config_done(cmp)
   end


LostNeophyte:
The `enabled` function is configurable so it's already possible

LostNeophyte:
The cmp.setup.filetypes may not be possible to set but i think having an on_config_done callback like with other plugins would be a better solution

mcartaud:
> The `enabled` function is configurable so it's already possible

Ok, I though that it would be best to have this case available by configuration instead of by overriding all work done in the `enabled` function, because it might be harder to upgrade for the user if LunarVim code changes

> The cmp.setup.filetypes may not be possible to set but i think having an on_config_done callback like with other plugins would be a better solution

I was not aware of this mechanism, will rework the PR

mcartaud:
This new version adds an `on_config_done` property on the cmp object and removes the configuration to disable cmp in comments

LostNeophyte:
Is it ready for review?

lua/lvim/core/cmp.lua
```diff
@@ -367,6 +368,9 @@ function M.setup()
       })
     end
   end
```
LostNeophyte, line 370:
```suggestion
  end
  
```

mcartaud:
Thanks for the review

LostNeophyte:
Update the pr description and all looks good to me, i've never seen the checks take so long btw

mcartaud:
> i've never seen the checks take so long btw

Yes, lot's of things to test in this review :)

It's all done, thanks

LostNeophyte (approved):
LGTM, thanks :)

kylo252 (approved):
Thanks!

PR #3647, opened 2022-12-22
refactor: migrate to lazy.nvim

### LINKED ISSUE #3645: feat: Switch plugin management system to `lazy.nvim`
### Feature Description

I recently found [`lazy.nvim`](https://github.com/folke/lazy.nvim). I tried it out, and it works great! Though `packer.nvim` is great, I feel `lazy.nvim` would be a better fit. Here are some reasons why:

- Mason-like install menu: `lazy.nvim` provides easy install and update overviews, which can be very helpful in some cases, but `packer.nvim` lacks this feature.	
- Replacement for `packer.nvim`: Packer is a great package manager, and provides many useful tools. The thing is, with the abstractions LunarVim provides, it could be a simple change for the more friendly `lazy.nvim` to replace it.

-----

- **Some features that would make LunarVim even better:** (from https://github.com/folke/lazy.nvim/blob/main/README.md)

	- 🔒 Lockfile lazy-lock.json to keep track of installed plugins: This could be helpful for automatically adding core plugins.
	- 📁 Configurable in multiple files: This might make it easier to integrate into the abstractions that LunarVim provides.
	- 💻 Dev options and patterns for using local plugins: This could be helpful for loading core plugins.
	- 🚀 Fast startup times thanks to automatic caching and bytecode compilation of lua modules: This could improve LunarVim's performance.
	- 🔌 Automatic lazy-loading of lua modules and lazy-loading on events, commands, filetypes, and key mappings: This could improve LunarVim's performance.
	- ⏳ Automatically install missing plugins before starting up Neovim, allowing you to start using it right away: This could make sure there are no errors on startup due to missing or deleted plugins
	- 💪 Async execution for improved performance: This could improve LunarVim's performance.

- Keep `packer.nvim`
- Writing a custom plugin manager (unlikely)

### Support information

- link to `lazy.nvim`: https://github.com/folke/lazy.nvim
- screenshot of interface:
image

LostNeophyte:
# Description

fixes #3645

# Tasks (probably best to do in order)

- [x] figure out how to pin core plugins
- [x] bootstrap lazy instead of packer
- [x] change plugin specs to lazy format
- [x] loading of plugins
- [x] reloading
- [x] hooks? (if needed)
- [x] pin core plugins
- [x] correct runtimepath
- [x] pin lazy.nvim commit
- [ ] todos from comments in the code
- [x] **[not fully tested]** backwards compability for lvim.plugins 
- [x] (test) headless bootstrap
- [x] test startup and bootstrap time lazy vs packer
- [x] update LunarVim CI/scripts
- [ ] update LunarVim.org CI/scripts
- [x] packaging
- [x] whichkey bindings
- [x] update docs
- [ ] test (help with testing highly appreciated, I want to make sure this PR works with all lvim configs) 

## how to install this version of lunarvim 
1st method: grab lunarvim from [releases](https://github.com/LunarVim/LunarVim-mono/releases/tag/nightly), I'll try to keep them up to date, (appimage is probably the easiest) when you run it the first time plugins should get copied and when it's done the dashboard should open. If you encounter any errors, report them - they are bugs, the refactor should be backwards compatible. 

2nd method (linux/mac only): use [this script](https://gist.github.com/LostNeophyte/c3ca8f0325034353d0cd4d3e4e327e45) to install lunarvim to /tmp


BlueFalconHD:
Maybe, to help report progress of this, could you create a todo list of tasks needed to be accomplished. That way myself or others could help out?

PS: For GitHub to integrate this into the PR status, you must edit the original post and add a todo list there rather than in a new comment.

LostNeophyte:
> Maybe, to help report progress of this, could you create a todo list of tasks needed to be accomplished. That way myself or others could help out?
> 
> PS: For GitHub to integrate this into the PR status, you must edit the original post and add a todo list there rather than in a new comment.

I don't really know which issues I'll have before i face them tbh so i wouldn't know what to add to the list, but maybe I'll think of something, thanks for the suggestion

lua/lvim/plugin-loader.lua
```diff
@@ -3,174 +3,178 @@ local plugin_loader = {}
...
-end
+  Config.spec = spec
 
-function plugin_loader.reload(configurations)
-  _G.packer_plugins = _G.packer_plugins or {}
-  for k, v in pairs(_G.packer_plugins) do
-    if k ~= "packer.nvim" then
-      _G.packer_plugins[v] = nil
-    end
+  require("lazy.core.plugin").load(true)
```
folke, line 56: you really shoudn't do all of this.

lazy automatically installs missing plugins during startup

lua/lvim/plugin-loader.lua
```diff
@@ -3,174 +3,178 @@ local plugin_loader = {}
...
 
--- packer expects a space separated list
-local function pcall_packer_command(cmd, kwargs)
-  local status_ok, msg = pcall(function()
-    require("packer")[cmd](unpack(kwargs or {}))
+  vim.opt.runtimepath:prepend(lazy_install_dir)
+
+  -- Add plugins to rtp (needed for config:init)
+  -- TODO: is there a better way to do this?
+  require("lazy.core.util").ls(plugins_dir, function(path, name, type)
```
folke, line 41: What are you trying to achieve here? should not be needed?

lua/lvim/plugin-loader.lua
```diff
@@ -3,174 +3,178 @@ local plugin_loader = {}
...
 local join_paths = utils.join_paths
-local in_headless = #vim.api.nvim_list_uis() == 0
 
--- we need to reuse this outside of init()
-local compile_path = join_paths(get_config_dir(), "plugin", "packer_compiled.lua")
-local snapshot_path = join_paths(get_cache_dir(), "snapshots")
-local default_snapshot = join_paths(get_lvim_base_dir(), "snapshots", "default.json")
+local plugins_dir = join_paths(get_runtime_dir(), "lazy", "plugins")
+
+local function remove_rtp_paths()
```
folke, line 9: not needed with lazy. lazy already resets the rtp

lua/lvim/plugin-loader.lua
LostNeophyte: That's for hot reload, so it will install missing plugins when the config is saved (we have an autocommand)

folke: ah ok. Lazy also does change detection that works even when you make changes outside of Neovim. I can provide a user event that triggers when changes are detected.

lua/lvim/plugin-loader.lua
LostNeophyte: it adds `vim.fn.stdpath("config")`, which is undesirable in lunarvim, but I've realized that my removing from rtp doesn't take effect, since lazy resets it after I remove the paths and loads .config/nvim/after anyway
@kylo252 what about adding `config` to [our override of vim.fn.stdpath](https://github.com/LunarVim/LunarVim/blob/62c2c0e15706be48abd2e3d353aaf79b00548d58/lua/lvim/bootstrap.lua#L66-L71)?

folke: you can disable rtp reset or add extra paths if you like. check the docs on config

lua/lvim/plugin-loader.lua
LostNeophyte: this adds the ability to `require` plugins before loading lazy.nvim, we need it currently because user's config is loaded before loading lazy.nvim, so that the user can change plugins's configuration and requiring plugins is needed to to that in some cases. It may not be needed in the future though, we might switch to lazy-requiring the plugins

lua/lvim/plugin-loader.lua
LostNeophyte: I'll probably do that, I though it would be nice to have it all in lazy.nvim, but lunarvim's use case is quite unusual. Thank you for taking the time to review this PR

lua/lvim/plugin-loader.lua
LostNeophyte: that automatic change detection requires polling a lua module as far as I know, and currently lunarvim and user plugins are split, they could be joined, but the polling would be a bit expensive, since I currently parse a json file to set `commit` in plugin specs, we are working on packaging lunarvim and that step could be done during packaging and then it wouldn't be a problem, but we need the autocommand anyway and doing it here also means no delay, if we switch to a lua module for the specs, we could disable change detection timer and trigger checking from the autocmd, I'm not sure if it's worth the effort though

LostNeophyte:
If someone wants to help, plugins in [lunarvim.org/extra-plugins.md](https://github.com/LunarVim/lunarvim.org/blob/master/docs/plugins/02-extra-plugins.md) need to be converted to lazy's format, it should be quite easy, there's just a lot of plugins there, if someone wants to do it then leave a reply here so that there's no duplicated work

roobert:
> If someone wants to help, then here's a list of easier tasks:
> 
> 1. plugins in [lunarvim.org/extra-plugins.md](https://github.com/LunarVim/lunarvim.org/blob/master/docs/plugins/02-extra-plugins.md) need to be converted to lazy's format, it should be quite easy, there's just a lot of plugins there,
> 
> if you want to do it then leave a reply here with the task you chose so that there's no duplicated work

https://github.com/LunarVim/lunarvim.org/pull/327 - hope this helps!

luxus:
does it supports `url =` (https://github.com/folke/lazy.nvim#-plugin-spec) for non-github repos? we didn't get it working

LostNeophyte:
> does it supports `url =` (https://github.com/folke/lazy.nvim#-plugin-spec) for non-github repos? we didn't get it working 
> 

I have a plugin from sourcehut and lazy downloaded it fine

lua/lvim/utils/hooks.lua
```diff
@@ -12,31 +12,22 @@ function M.run_pre_reload()
...
-  -- -- FIXME(kylo252): nvim-tree.lua/lua/nvim-tree/view.lua:442: Invalid window id
-  -- vim.g.colors_name = lvim.colorscheme
-  -- pcall(vim.cmd.colorscheme, lvim.colorscheme)
-
-  if M._reload_triggered then
-    Log:debug "Reloaded configuration"
-    M._reload_triggered = nil
-  end
-end
+-- TODO: convert to lazy.nvim
```
BlueFalconHD, line 15: This might be the relevant event to use:
image

LostNeophyte: the event would be a combination of LazyInstall and LazyUpdate, but these lines were commented out earlier and I'm not sure what they're supposed to do so I'll leave it for now

Uzaaft:
@LostNeophyte  I opened up a PR to add more Lazy specific keybindings to Lunarvim.

Uzaaft:
Do we have a plan for lazy loading things @LostNeophyte  ? I just changed one plugins, and I managed to reduce loading time with about 7ms. I believe some gains can be made here.

LostNeophyte:
> Do we have a plan for lazy loading things @LostNeophyte ? I just changed one plugins, and I managed to reduce loading time with about 7ms. I believe some gains can be made here.

All the plugins that can, should be lazy loaded

Uzaaft:
I'll go ahead then and add `lazy = true` to plugins in `lua/lvim/lsp/null-ls/code_actions.lua`. One after one

Uzaaft:
Added lazy loading in #3666.
Got pretty significant gains.

LostNeophyte:
> Added lazy loading in #3666. Got pretty significant gains.

completion doesn't work for me with that pr

folke:
For cmp, you can use   event = "InsertEnter",

Uzaaft:
> 

Thanks!

Uzaaft:
> For cmp, you can use `event = "InsertEnter"`

I just have to say, thank you folke for all the help you've given. Just give me a few weeks, and you've got yourself a new sponser. <33

folke:
Ha, thank you but I don't need any sponsoring :)

If you really want to get rid of some money, feel free to donate somthing to a good cause instead.

Uzaaft:
> Ha, thank you but I don't need any sponsoring :)
> 
> If you really want to get rid of some money, feel free to donate somthing to a good cause instead.

Then give me a good cause YOU like, and I'll donate. <3

opalmay:
Amazing work on this :)

I'm facing two issues:
1. My TS stopped working and it wouldn't let me install any language, not sure why. For example running `:TSInstall lua` shows a success message, but it doesn't actually seem to do anything.
2. Cmp needs a different event than `InsertEnter` to also include the cmd line

folke:
For cmp you can do `event = {"InsertEnter", "CmdlineEnter"}`

opalmay:
> Amazing work on this :)
> 
> I'm facing two issues:
> 
> 1. My TS stopped working and it wouldn't let me install any language, not sure why. For example running `:TSInstall lua` shows a success message, but it `:TSInstallInfo` still reports it as not installed.
>    ~2. Cmp needs a different event than `InsertEnter` to also include the cmd line~

I've removed all Packer/Lazy data folders, let Lazy set itself up again and now TS works fine. Not sure what that was about

LostNeophyte:
https://github.com/LunarVim/LunarVim/pull/3647/commits/f714a6f22a815e538143daaf50997187bdfdbb12 is hacky, but I couldn't find a better way to do what it does. With that change startup is `1.16 ± 0.04 times faster packer_lvim` , before it was `1.04 ± 0.03 times faster than packer_lvim`

LostNeophyte:
For anyone who wants to help with testing: I included instructions on how to install this lvim version to the description of the PR

folke:
@LostNeophyte wouldn't it be better to use lazy's lockfile directly?
I've actually made a change today that when installing missing plugins it will always install those commits from the lockfile if available.

If you use the api, you can add `lockfile=true` to the `install()` command.

LostNeophyte:
> @LostNeophyte wouldn't it be better to use lazy's lockfile directly? I've actually made a change today that when installing missing plugins it will always install those commits from the lockfile if available.
> 
> If you use the api, you can add `lockfile=true` to the `install()` command.

I though it was impossible to use with lunarvim, but now that I think about it if we add `pin = true` to core plugins then `:Lazy sync` won't update them, but then there's a different problem: the lockfile would probably reside in .config, and then if the user updates lunarvim the lockfile will be outdated (updates shouldn't touch .config), if we put the lockfile in .local/share then we can assume that the user won't modify it but then the user will loose the ability to use it for custom plugins

roobert:
> For anyone who wants to help with testing: I included instructions on how to install this lvim version to the description of the PR

Tested on OSX and works great for me, thanks for your effort on this and to folke for the awesome plugin!

My config is here: https://github.com/roobert/dotfiles/tree/master/.config/lvim

luxus:
> For anyone who wants to help with testing: I included instructions on how to install this lvim version to the description of the PR

works great, I use a personalized version of @abzcoding config. no issues so far

ChristianChiarulli:
Just tested, so far it looks good other than the bufferline plugin flashes briefly on startup. Not sure if that's just me experiencing that.

folke:
@ChristianChiarulli the flashing is probably because bufferline loads before the colorscheme. https://github.com/LunarVim/LunarVim/blob/e986f53fa770959396670ed9a4c8e62da39ebec2/lua/lvim/plugins.lua#L269 should set a priority and load the colorscheme.

See https://github.com/folke/lazy.nvim/tree/release-please--branches--main--components--lazy.nvim#-colorschemes

LostNeophyte:
@ChristianChiarulli @folke i've set bufferline to load on VeryLazy, that's the cause. before bufferline loads the default tabline is visible and It's doing the flashing, should be fixed now

brneor:
Tested with the _AppImage_. Got a bunch of errors that disappeared to quickly but (almost) everything worked just fine after a restart (had to adjust a `module` entry in a plugin setting).

Funny thing is that I'm seeing treesitter installing everything everytime I open LunarVim. Not sure if is related to this change or the AppImage.

opalmay:
> Tested with the _AppImage_. Got a bunch of errors that disappeared too quickly but (almost) everything worked just fine after a restart (had to adjust a `module` entry in a plugin setting).
> 
> Funny thing is that I'm seeing treesitter installing everything everytime I open LunarVim. Not sure if is related to this change or the AppImage.

The TS thing sounds similar to what I've had, see my comment.
I was not using the AppImage, it's something with the change from packer to lazy

brneor:
Another thing that I just noticed is that everytime I open `Lazy` all the plugins are marked with "_updates available_".

LostNeophyte:
> Edit: just to add, I cant get it working normally with "install to /tmp" approach.

Chris used it on stream and all seemed to work fine

> Another thing that I just noticed is that everytime I open Lazy all the plugins are marked with "updates available".

I think it's a problem with lazy, it shows that updates are available even if the plugins are pinned to a version with `commit` and that commit is currently checked out

I'll have to look into that treesitter issue, but if I remember correctly I don't experience it on my machine

folke:
@LostNeophyte was indeed a Lazy issue. The problem was that I use long git commit hashes internally and compared those to the short ones you add as `commit=`. Just pushed a fix for this.

svirschevskiy:
I couldn't install any tree-sitter languages until I removed  `~/.local/share/lunarvim/site/pack`. It solved the problem for me.

It seems like `nvim-treesitter` was pointed to 
`~/.local/share/lunarvim/site/pack/packer/start/nvim-treesitter/parser`
 instead of  
`~/.local/share/lunarvim/site/pack/lazy/opt/nvim-treesitter/parser`.

Other than that it works great, thank you!

folke:
@svirschevskiy normally `:checkhealth lazy` should have reported that. Did it show any warnings?

svirschevskiy:
@folke there are not warnings

I managed to reproduce it by following steps:
1. run lvim from `master` branch, make sure that Packer installs all the dependencies
2. switch to `refactor/lazy` make sure that Lazy installs ann the dependencies
3. try to install any tree-sitter language, i.e.  `TSInstall astro`
4. the language is not listed in `TSInstallInfo`

```
~/.local/share/lunarvim/site/pack
❯ tree -L 2
.
├── lazy
│   └── opt
└── packer
    ├── opt
    └── start
```

```
lazy: require("lazy.health").check()
========================================================================
## lazy.nvim
  - OK: no existing packages found by other package managers
  - OK: packer_compiled.lua not found
```

folke:
That is really odd. I've trying to firgure this one out, but haven't found the reason yet. What happens if you delete the lazy cache file? Does that solve it somehow?

svirschevskiy:
@folke I think that checkhealth is checking `~/.local/share/nvim/site/pack`
However, Lunarvim installs its packages into `~/.local/share/lunarvim/site/pack`

folke:
yep, you're right. Will change it to check all sites on the rtp

opalmay:
I have an issue with `mason-lspconfig.nvim`, `mason-lspconfig.nvim`, `nvim-lspconfig` not loading in any python file.
Removing lazy loading fixes it.
image

LostNeophyte:
> I have an issue with `mason-lspconfig.nvim`, `mason-lspconfig.nvim`, `nvim-lspconfig` not loading in any python file. Removing lazy loading fixes it. image

nvim-lspconfig is loaded, require("lspconfig") works fine but lazy somehow doesn't know that it was required, but i've updated lazy and it works

opalmay:
> nvim-lspconfig is loaded, require("lspconfig") works fine but lazy somehow doesn't know that it was required, but i've updated lazy and it works

Sorry, I don't think I understand your comment.
If I run `:lua require("lspconfig")` it obviously loads, but it should be automatic.
Running update didn't fix it

opalmay:
> actually, I don't think it was the update that helped, I removed cache at the same time and I think that did it

It didn't help for me. Tried removing the Lazy data and cache folders and that didn't fix it either.
Why would `Lazy = true` work for these plugins? What requires them?

folke:
I introduced a bug yesterday that resulted in the behavior you describe. Updating Lazy should fix it.No need to clear the cache.

opalmay:
> I introduced a bug yesterday that resulted in the behavior you describe. Updating Lazy should fix it.No need to clear the cache.

Right now I'm running 7.11.0 and I still get that behavior

folke:
@LostNeophyte how is lspconfig loaded? I don't see any triggers.

I personally load lsp-config on `BufReadPre`.

You can check [LazyVim](https://github.com/folke/LazyVim/blob/main/lua/lazyvim/plugins/lsp/init.lua) for an example on how to properly setup lspconfig, mason etc

folke:
@opalmay in that screenshot, is that a new file? Did you save it and open it again with `:e`? Otherwise Neovim's filetype detection probably won't run.

LostNeophyte:
It's loaded by `require("lspconfig")` in an autogenerated ftplugin

folke:
So it is loaded as I can see in your screenshot?

@opalmay if you do `:lua require("lspconfig")` does it then work?

opalmay:
> So it is loaded as I can see in your screenshot?
> 
> @opalmay if you do `:lua require("lspconfig")` does it then work?

Running require works.
Neophyte testing with a Lua file. For me it works in Lua but not Python. He is able to reproduce it in python but not consistently like me.

folke:
if it works after the require, then something is up with that ftplugin probably?

opalmay:
> if it works after the require, then something is up with that ftplugin probably?

The ftplugin loads, I've checked.
But even if I change its contents to `require("lspconfig")` it still doesn't work

folke:
I'll need to have a mininal init.lua that can reproduce this if you want me to look at this.

LostNeophyte:
> I couldn't install any tree-sitter languages until I removed `~/.local/share/lunarvim/site/pack`. It solved the problem for me.
> 
> It seems like `nvim-treesitter` was pointed to `~/.local/share/lunarvim/site/pack/packer/start/nvim-treesitter/parser` instead of `~/.local/share/lunarvim/site/pack/lazy/opt/nvim-treesitter/parser`.

yea, `:lua =require("nvim-treesitter.configs").get_parser_install_dir()` returns packer's path
`:lua =require("nvim-treesitter.utils").get_package_path()` returns packer's path
`:lua =debug.getinfo(require("nvim-treesitter.utils").get_package_path, "S")` returns packer's path

I removed site directory from rtp and it fixed it, I'm not sure it that has any unwanted effects

LostNeophyte:
> So it is loaded as I can see in your screenshot?

yes, it works for me, before it didn't and even requiring didn't load the plugin and dependencies but now everything is fine for me, even checked on a different pr with a clean install

lua/lvim/config/init.lua
```diff
@@ -56,13 +56,17 @@ function M:load(config_path)
       vim.notify_once(
         string.format("User-configuration not found. Creating an example configuration in %s", config_path)
       )
-      local example_config = join_paths(get_lvim_base_dir(), "utils", "installer", "config.example.lua")
+      local config_name = vim.loop.os_uname().version:match "Windows" and "config_win" or "config"
+      local example_config = join_paths(get_lvim_base_dir(), "utils", "installer", config_name .. ".example.lua")
+      vim.fn.mkdir(user_config_dir, "p")
```
kylo252, line 61: please move this to a separate PR

lua/lvim/config/init.lua
```diff
@@ -93,6 +97,8 @@ function M:reload()
 
     plugin_loader.reload { plugins, lvim.plugins }
     reload("lvim.core.theme").setup()
+
+    Log:debug "Reloaded configuration"
```
kylo252, line 101: the hooks are already adding a log entry 

```suggestion
```

lua/lvim/bootstrap.lua
```diff
@@ -3,6 +3,7 @@ local M = {}
 if vim.fn.has "nvim-0.8" ~= 1 then
   vim.notify("Please upgrade your Neovim base installation. Lunarvim requires v0.8+", vim.log.levels.WARN)
   vim.wait(5000, function()
+    ---@diagnostic disable-next-line: redundant-return-value
```
kylo252, line 6: you could also move these as well to that same PR

LostNeophyte: all the `---@diagnostic` ?

ChristianChiarulli (approved):

luxus:
Congratulation @LostNeophyte and all the people that helped.

LostNeophyte:
Thank you @folke, @kylo252, @opalmay, @Uzaaft, @roobert , @BlueFalconHD, and everyone else for all the help with this! :heart:

folke:
Oh wow, that's awesome. Congrats! :)

PR #3649, opened 2022-12-23
fix(telescope): ignore .git/*

meijieru:
Better default for telescope find_files.

### DIFF
--- a/lua/lvim/core/telescope.lua
+++ b/lua/lvim/core/telescope.lua
@@ -60,7 +60,7 @@ function M.config()
           ["<C-q>"] = actions.smart_send_to_qflist + actions.open_qflist,
         },
       },
-      file_ignore_patterns = {},
+      file_ignore_patterns = { ".git/" },
       path_display = { "smart" },
       winblend = 0,
       border = {},


cpea2506 (approved):

kylo252 (changes requested):
This option is not passed directly to `fd` or `ripgrep`, see https://github.com/nvim-telescope/telescope.nvim/blob/0b1c41ad8052badca6e72eafa4bc5481152e483e/doc/telescope.txt#L618
```help
     If you only want `file_ignore_patterns` for `find_files` and
    `grep_string`/`live_grep` it is suggested that you setup `gitignore`
    and have fd and or ripgrep installed because both tools will not show
    `gitignore`d files on default.
```

instead, maybe it's recommended to add `~/.config/fd/ignore`

```gitconfig
.git/*
*.git/*
.node-modules/*
```

meijieru:
Adding them to `~/.config/fd/ignore` doesn't work in `find_files`.

opalmay:
`.git/` is automatically gitignored so it doesn't show in either `fd` or `rg`. This is the default behavior

LostNeophyte:
> `.git/` is automatically gitignored so it doesn't show in either `fd` or `rg`. This is the default behavior

files in .git do show up using `Telescope find_files`, but not with `Telescope git_files`

meijieru:
This can be done by manually set
```lua
  lvim.builtin.telescope.pickers.find_files = {
    find_command = { "rg", "--files", "--hidden", "--glob", "!**/.git/*" },
  }
```

PR #3665, opened 2022-12-29
feat(lazy): add keymappings for profile, log, and debug

Uzaaft:
Added more which-key mappings for more of Lazy's functions, like cleaning, profiling etc to make it easier for the end-user to discover these.


LostNeophyte:
Thanks!

PR #3666, opened 2022-12-29
refactor: Add lazy loading to as plugins as possible

Uzaaft:
Added lazy loading for as many plugins as possible. 
There are quite a lot of commented out code in the file, perhaps we should take the opportunity and clean up things?


LostNeophyte:
I've modified this PR a bit and co-authored you

PR #3670, opened 2022-12-31
fix(cmp): jump for leading exit nodes and jump after confirm

chase:
# Description

- At some point the ability to confirm and jump to the next part of a snippet using `<CR>` was broken, this restores that behavior
- Fixes jumping after confirm for some cases where a luasnip exit node is intentionally placed prior to the cursor, such as the `import statement` snippet for `typescript`: `import { $0 } from "$1"`
- Also fixes a few LSP warnings

## How Has This Been Tested?

- Type something that prompts cmp for a snippet
- Confirm the snippet with `<CR>`
- Type something in one of the fields that may be completed
- Either abort the completion and press `<CR>` to jump or confirm it and jump to the next snippet section with `<CR>`


lua/lvim/core/cmp.lua
```diff
@@ -23,7 +23,7 @@ end
 M.methods.feedkeys = feedkeys
 
 ---when inside a snippet, seeks to the nearest luasnip field if possible, and checks if it is jumpable
----@param dir number 1 for forward, -1 for backward; defaults to 1
+---@param dir? number 1 for forward, -1 for backward; defaults to 1
 ---@return boolean true if a jumpable luasnip field is found while inside a snippet
 local function jumpable(dir)
```
kylo252, line 28: do we actually still need this btw? have you checked their new API since?  https://github.com/L3MON4D3/LuaSnip/wiki/Nice-Configs#only-jump-in-the-current-snippet

lua/lvim/core/cmp.lua
```diff
@@ -109,9 +106,9 @@ local function jumpable(dir)
   end
 
   if dir == -1 then
-    return luasnip.in_snippet() and luasnip.jumpable(-1)
+    return luasnip.in_snippet() == true and luasnip.jumpable(-1)
```
kylo252, line 109:
```suggestion
    return luasnip.in_snippet() and luasnip.jumpable(-1)
```

lua/lvim/core/cmp.lua
```diff
@@ -109,9 +106,9 @@ local function jumpable(dir)
...
   else
-    return luasnip.in_snippet() and seek_luasnip_cursor_node() and luasnip.jumpable(1)
+    return luasnip.in_snippet() == true and seek_luasnip_cursor_node() and luasnip.jumpable(1)
```
kylo252, line 111:
```suggestion
    return luasnip.in_snippet() and seek_luasnip_cursor_node() and luasnip.jumpable(1)
```

lua/lvim/core/cmp.lua
```diff
@@ -329,10 +326,15 @@ M.config = function()
             confirm_opts.behavior = cmp.ConfirmBehavior.Insert
           end
           if cmp.confirm(confirm_opts) then
+            if jumpable() then
```
kylo252, line 329:
```suggestion
            if jumpable(1) then
```

lua/lvim/core/cmp.lua
```diff
@@ -329,10 +326,15 @@ M.config = function()
...
+              luasnip.jump(1)
+            end
             return -- success, exit early
           end
         end
-        fallback() -- if not exited early, always fallback
+        if not jumpable() or not luasnip.jump(1) then
```
kylo252, line 335:
```suggestion
        if not jumpable(1) or not luasnip.jump(1) then
```

lua/lvim/core/cmp.lua
```diff
@@ -23,7 +23,7 @@ end
 M.methods.feedkeys = feedkeys
 
 ---when inside a snippet, seeks to the nearest luasnip field if possible, and checks if it is jumpable
----@param dir number 1 for forward, -1 for backward; defaults to 1
+---@param dir? number 1 for forward, -1 for backward; defaults to 1
```
kylo252, line 26:
```suggestion
---@param dir number|nil 1 for forward, -1 for backward [default: 1]
```

lua/lvim/core/cmp.lua
chase: Without `== true` it is null or true or false.

Based on the linting or type checking, this throws a warning for me which is why it was changed. Does this not cause a warning for you?

lua/lvim/core/cmp.lua
chase: 1 is optional

lua/lvim/core/cmp.lua
chase: Is this a style preference to not use `?` but `number|nil`?

lua/lvim/core/cmp.lua
chase: I'll have to review, but at a glance: yes, because it likely doesn't support resuming the cursor to the nearest modification after leaving insert mode for an unfinished snippet.

chase: After reviewing the code, yes. It still is required but uses the updated code which simply calls `in_snippet()` and `jumpable()`, which is done at the end of the function.

lua/lvim/core/cmp.lua
kylo252: yeah, I know but it's a bit easier to read, i.e. more consistent, since the very next command is `luasnip.jump(1)`

lua/lvim/core/cmp.lua
kylo252: > Is this a style preference to not use `?` but `number|nil`?

basically yes, at least that's the pattern I've seen preferred in neovim core.

lua/lvim/core/cmp.lua
kylo252: oh I forgot to say that we should change the type annotation for the return value instead.

```lua
---@return boolean|nil true if a jumpable luasnip field is found while inside a snippet
```

lua/lvim/core/cmp.lua
kylo252: > I'll have to review, but at a glance: yes, because it likely doesn't support resuming the cursor to the nearest modification after leaving insert mode for an unfinished snippet.

how do you feel about upstreaming this API? I remember asking @L3MON4D3 about it a while back.

chase: After further testing, I've removed the code and deprecated the exported function :+1:

chase: > > I'll have to review, but at a glance: yes, because it likely doesn't support resuming the cursor to the nearest modification after leaving insert mode for an unfinished snippet.
> 
> how do you feel about upstreaming this API? I remember asking @L3MON4D3 about it a while back.

As stated above, shouldn't be necessary.

kylo252 (dismissed):

chase:
After some extensive testing, it seems like there are occasions where the snippet doesn't unlink even after it is completed. Probably shouldn't be merged.

lua/lvim/core/cmp.lua
```diff
@@ -329,10 +242,15 @@ M.config = function()
             confirm_opts.behavior = cmp.ConfirmBehavior.Insert
           end
           if cmp.confirm(confirm_opts) then
+            if luasnip.locally_jumpable() then
+              luasnip.jump()
+            end
```
LostNeophyte, line 247: I think this should be in callback of `cmp.confirm`, it overwrites parenthesis for me

LostNeophyte:
> After some extensive testing, it seems like there are occasions where the snippet doesn't unlink even after it is completed. Probably shouldn't be merged.

Can you check if my commit fixed the problem?

kylo252:
> > After some extensive testing, it seems like there are occasions where the snippet doesn't unlink even after it is completed. Probably shouldn't be merged.
> 
> Can you check if my commit fixed the problem?

it would also be good to document these tests here for posterity

PR #3694, opened 2023-01-07
refactor: use lir fork in the org

opalmay:
Switch from the one in Chris's account.


LostNeophyte (approved):
LGTM

opalmay:
Made a PR upstream to add our change as a config option: https://github.com/tamago324/lir.nvim/pull/76
Hopefully this gets merged and we can use upstream directly.

PR #3715, opened 2023-01-12
fix(breadcrumbs): updating extension to empty string if it contains c…

### LINKED ISSUE #3713: Bug: Octo comment throws error on breadcrumbs file
### Problem description

Install and configure Octo and proceed to creating a comment via review:

1. `:Octo review start`
2. Find a hunk and add a comment: `:Octo comment add`
3. It will throw the following error:
image

The reason it seems is that the `extension` of the file comes back in this format:

image

`php:507` where is should just be `php`. 

### LunarVim version

8f4a7bde

### Neovim version (>= 0.8.0)

8.2

### Operating system/version

MacOs

### Steps to reproduce

_No response_

### Screenshots

_No response_

RizaHKhan:
# Description

Extension contains special character (colon) which is causing errors on nvim function `nvim_set_hl`. 

fixes #3713

## How Has This Been Tested?

Setup Octo and its dependencies.
Start a PR review and place cursor on hunk. 
Run `:Octo comment add` and the file should open where a comment can be added.

### DIFF
--- a/lua/lvim/core/breadcrumbs.lua
+++ b/lua/lvim/core/breadcrumbs.lua
@@ -95,12 +95,12 @@ end
 
 M.get_filename = function()
   local filename = vim.fn.expand "%:t"
-    
+
   local extension = vim.fn.expand "%:e"
   if string.find(extension, ":") then
-    extension = ""    
+    extension = ""
   end
-    
+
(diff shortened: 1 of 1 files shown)


LostNeophyte:
the issue is already fixed in master

PR #3719, opened 2023-01-13
feat: cmdline config option enables cmp-cmdline plugin

opalmay:
We already have the configuration for cmdline completions (disabled by default). But the plugin has to be installed as a user plugin after enabling it.
With this commit setting `lvim.builtin.cmp.cmdline.enable = true` will also enable the `cmp-cmdline` plugin.


LostNeophyte (approved):
LGTM

PR #3721, opened 2023-01-13
refactor(quit)!: use native quit confirm

opalmay:
Use the native confirm quit functionality of Neovim `:confirm q` instead of our custom implementation.
image
This is a soft breaking change due to `N` now exiting without saving changes, instead of `N<CR>` canceling.
Presumably, `ESC` is mostly used to cancel rather than `N<CR>`. In this case the behavior is the same.


LostNeophyte (approved):
LGTM

PR #3727, opened 2023-01-14
fix(breadcrumbs): refresh on TabEnter

opalmay:
To reproduce the issue:
1. Open a new tab
2. Switch back to an existing tab
3. Tab count (right side of the breadcrumbs) will show the old tab count for a brief moment.
As a fix, add TabEnter as an additional event for refresh autocmd.


LostNeophyte (approved):
LGTM

PR #3730, opened 2023-01-15
chore(devicons): rename devicons owner & update

latipun7:
# Description

`kyazdani42/nvim-web-devicons` is renamed to its org: https://github.com/nvim-tree/nvim-web-devicons. Also, update the plugin to its latest commit.

## How Has This Been Tested?

- Updated the plugin myself
- Check with additional new feature to set default icons:
  ```lua
  local ok, devicons = pcall(require, "nvim-web-devicons")
  if ok then devicons.set_default_icon("", "#6d8086") end -- this func is not available in the previous commit, only the latest
  ```
- Works!

### DIFF
--- a/lua/lvim/plugins.lua
+++ b/lua/lvim/plugins.lua
@@ -177,7 +177,7 @@ local core_plugins = {
 
   -- Icons
   {
-    "kyazdani42/nvim-web-devicons",
+    "nvim-tree/nvim-web-devicons",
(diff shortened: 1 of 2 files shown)


LostNeophyte:
Where did you get this information? Latest commit is five days old and I can't see anything about it being moved

LostNeophyte (approved):
LGTM, Thanks!

PR #3750, opened 2023-01-18
perf: lazy load most plugins

LostNeophyte:
faster startup
image


lua/lvim/plugins.lua
```diff
@@ -113,21 +116,23 @@ local core_plugins = {
...
   -- NvimTree
   {
     "kyazdani42/nvim-tree.lua",
-    -- event = "BufWinOpen",
-    -- cmd = "NvimTreeToggle",
     config = function()
       require("lvim.core.nvimtree").setup()
     end,
     enabled = lvim.builtin.nvimtree.active,
+    cmd = { "NvimTreeToggle", "NvimTreeOpen", "NvimTreeFocus" },
```
pr-313, line 134: Could we perhaps add NvimTreeFindFileToggle here as well? 
Have been testing this PR today and ran into this limitation with my config as I turn the autochdir equivalent of nvim-tree off

lua/lvim/plugins.lua
```diff
@@ -159,10 +165,10 @@ local core_plugins = {
   -- Comments
   {
     "numToStr/Comment.nvim",
```
pr-313, line 167: Load order might be a problem here
When doing the following
Open lvim > find file using Telescope > hit the comment line keybind <leader>/

Which-key will need to list comment.nvim as a dependency for <leader>/ to work.

gcc,gcb work well though.

lua/lvim/lsp/config.lua
```diff
@@ -96,12 +96,12 @@ return {
   },
   buffer_mappings = {
     normal_mode = {
-      ["K"] = { vim.lsp.buf.hover, "Show hover" },
-      ["gd"] = { vim.lsp.buf.definition, "Goto Definition" },
-      ["gD"] = { vim.lsp.buf.declaration, "Goto declaration" },
-      ["gr"] = { vim.lsp.buf.references, "Goto references" },
-      ["gI"] = { vim.lsp.buf.implementation, "Goto Implementation" },
-      ["gs"] = { vim.lsp.buf.signature_help, "show signature help" },
+      ["K"] = { "<cmd>lua vim.lsp.buf.hover()", "Show hover" },
```
pr-313, line 99: Will need a \<CR\> at the end of these commands, its erroring out currently.

lua/lvim/core/which-key.lua
```diff
@@ -205,20 +205,20 @@ M.config = function()
         a = { "<cmd>lua vim.lsp.buf.code_action()<cr>", "Code Action" },
         d = { "<cmd>Telescope diagnostics bufnr=0 theme=get_ivy<cr>", "Buffer Diagnostics" },
         w = { "<cmd>Telescope diagnostics<cr>", "Diagnostics" },
-        f = { require("lvim.lsp.utils").format, "Format" },
+        f = { "<cmd>lua require('lvim.lsp.utils').format()", "Format" },
```
pr-313, line 208: Same as previous comment, need \<CR\> at the end of \<cmd\>

pr-313:
Raised PR https://github.com/LunarVim/LunarVim/pull/3754
To fix the issue raised

lua/lvim/lsp/config.lua
```diff
@@ -96,12 +96,12 @@ return {
   },
   buffer_mappings = {
     normal_mode = {
-      ["K"] = { vim.lsp.buf.hover, "Show hover" },
-      ["gd"] = { vim.lsp.buf.definition, "Goto Definition" },
-      ["gD"] = { vim.lsp.buf.declaration, "Goto declaration" },
-      ["gr"] = { vim.lsp.buf.references, "Goto references" },
-      ["gI"] = { vim.lsp.buf.implementation, "Goto Implementation" },
-      ["gs"] = { vim.lsp.buf.signature_help, "show signature help" },
+      ["K"] = { "<cmd>lua vim.lsp.buf.hover()<cr>", "Show hover" },
```
kylo252, line 99: what's the reason behind this change? is it something to do with string dump?

lua/lvim/core/theme.lua
```diff
@@ -85,8 +85,12 @@ M.setup = function()
   vim.g.colors_name = lvim.colorscheme
   vim.cmd("colorscheme " .. lvim.colorscheme)
 
-  require("lvim.core.lualine").setup()
-  require("lvim.core.lir").icon_setup()
+  if package.loaded.lualine then
```
kylo252, line 88: isn't it better to move this line into the plugin's `config` section instead?

https://github.com/LunarVim/LunarVim/blob/edc5e0e93f10cdd7ebd0de936b22ec4bfb2e852c/init.lua#L19

lua/lvim/core/telescope.lua
```diff
@@ -7,21 +7,22 @@ local M = {}
 ---| "center"   # retain the default telescope theme
 
 function M.config()
-  -- Define this minimal config so that it's available if telescope is not yet available.
+  local lazy_action = function(name)
```
kylo252, line 10: is this necessary? I thought it would've been covered by one of the utility functions

lua/lvim/core/autocmds.lua
```diff
@@ -121,6 +121,35 @@ function M.load_defaults()
         end,
       },
     },
+    { -- taken from AstroNvim
```
kylo252, line 124: I'm not quite sold on these, at least not the DirOpened one, what happens if you try to open nvim-tree from the dashboard? :thinking:

lua/lvim/core/telescope.lua
LostNeophyte: it's necessary unless I put `actions.smart_send_to_qflist + actions.open_qflist` in a function which I guess is a more readable and easier approach

lua/lvim/core/autocmds.lua
LostNeophyte: then nvimtree will load by `:NvimTreeToggle` cmd

lua/lvim/core/theme.lua
LostNeophyte: I think it needs to be here so that it's run when the colorscheme is changed

lua/lvim/utils/modules.lua
```diff
@@ -95,4 +95,59 @@ M.reload = function(mod)
...
+  return setmetatable({}, {
+    __index = function(_, k)
+      return function(...)
+        return require(require_path)[k](...)
+      end
+    end,
+  })
+end
+
+function M.lazy_set(require_path, fn_name)
```
kylo252, line 122: did you write this manually? isn't this one missing `__call`?

LostNeophyte: yes and yes

kylo252: > although it doesn't need __call, it makes more sense to use `require_on_exported_call` in that case

I think you forgot to add that one :eyes:

[wouldn't it be the one to use with telescope?](https://github.com/tjdevries/lazy-require.nvim/blob/bb626818ebc175b8c595846925fd96902b1ce02b/lua/lazy-require.lua#L56-L61)

lua/lvim/lsp/config.lua
kylo252: > it's so that `vim.lsp` won't be loaded when this file is run

hmm, can't we overwrite `vim.lsp` once?

lua/lvim/utils/modules.lua
LostNeophyte: I did change telescope to use `require_on_exported_call`

`lazy_set` is only used for dashboard config table so that it's loaded with the plugin or if the user wants to change something in the table

lua/lvim/core/cmp.lua
```diff
@@ -277,24 +278,24 @@ M.config = function()
...
-      ["<C-f>"] = cmp.mapping.scroll_docs(4),
-      ["<C-y>"] = cmp.mapping {
-        i = cmp.mapping.confirm { behavior = cmp.ConfirmBehavior.Replace, select = false },
+    mapping = cmp_mapping.preset.insert {
+      ["<C-k>"] = cmp_mapping(cmp_mapping.select_prev_item(), { "i", "c" }),
+      ["<C-j>"] = cmp_mapping(cmp_mapping.select_next_item(), { "i", "c" }),
+      ["<Down>"] = cmp_mapping(cmp_mapping.select_next_item { behavior = cmp_types.SelectBehavior.Select }, { "i" }),
+      ["<Up>"] = cmp_mapping(cmp_mapping.select_prev_item { behavior = cmp_types.SelectBehavior.Select }, { "i" }),
+      ["<C-d>"] = cmp_mapping.scroll_docs(-4),
+      ["<C-f>"] = cmp_mapping.scroll_docs(4),
```
opalmay, line 287: How is this gonna work in user configs?

LostNeophyte: Unfortunately it won't, the user would need to do the same thing, but i think that most don't edit these mappings

opalmay: Is it a breaking change, or would it only prevent lazy loading?

LostNeophyte: not breaking, if the user requires cmp in their config it will load and that's all

opalmay (approved):

lua/lvim/plugins.lua
```diff
@@ -6,25 +6,28 @@ local core_plugins = {
...
+  { "williamboman/mason-lspconfig.nvim", lazy = true, dependencies = "mason.nvim" },
   { "tamago324/nlsp-settings.nvim", lazy = true },
   { "jose-elias-alvarez/null-ls.nvim", lazy = true },
   {
     "williamboman/mason.nvim",
     config = function()
       require("lvim.core.mason").setup()
     end,
+    lazy = true,
   },
```
cshuaimin, line 18: I can't use `:Mason` cmd now, could you please add `cmd = "Mason"` here?

LostNeophyte: thanks for the review, I missed that, I'll also need to add LspInstall

ChristianChiarulli:
image

After initial install alpha doesn't seem to load correctly.

LostNeophyte:
> After initial install alpha doesn't seem to load correctly.

I can't reproduce

ChristianChiarulli:
probably some weird state on my end let me try clearing things out a little harder.

ChristianChiarulli (approved):

lua/lvim/core/bufferline.lua
```diff
@@ -152,6 +152,8 @@ M.setup = function()
     return
   end
 
+  vim.opt.showtabline = 2
+
```
kylo252, line 156: why not move is settings,lua?

ChristianChiarulli:
Actually I'm also noticing some issues with using copilot and tabnine, seems like suggestion don't show up when they're loaded

lua/lvim/core/cmp.lua
```diff
@@ -118,14 +118,15 @@ end
 M.methods.jumpable = jumpable
 
 M.config = function()
-  local status_cmp_ok, cmp = pcall(require, "cmp")
+  local status_cmp_ok, cmp_types = pcall(require, "cmp.types.cmp")
```
kylo252, line 121: wouldn't this require `cmp` anyway? I think we would need to get rid of this condition

lua/lvim/core/cmp.lua
```diff
@@ -138,7 +139,7 @@ M.config = function()
       return lvim.builtin.cmp.active
     end,
     confirm_opts = {
-      behavior = cmp.ConfirmBehavior.Replace,
+      behavior = cmp_types.ConfirmBehavior.Replace,
```
kylo252, line 142: this is still in `cmp`, see https://github.com/hrsh7th/nvim-cmp/blob/e7e2ef7031db8e0cfad7ad9916ec766ddd9dbb3b/lua/cmp/init.lua#L188

```suggestion
      behavior = cmp.ConfirmBehavior.Replace,
```

lua/lvim/core/cmp.lua
```diff
@@ -277,24 +278,24 @@ M.config = function()
...
-      ["<Up>"] = cmp.mapping(cmp.mapping.select_prev_item { behavior = cmp.SelectBehavior.Select }, { "i" }),
-      ["<C-d>"] = cmp.mapping.scroll_docs(-4),
-      ["<C-f>"] = cmp.mapping.scroll_docs(4),
-      ["<C-y>"] = cmp.mapping {
-        i = cmp.mapping.confirm { behavior = cmp.ConfirmBehavior.Replace, select = false },
+    mapping = cmp_mapping.preset.insert {
+      ["<C-k>"] = cmp_mapping(cmp_mapping.select_prev_item(), { "i", "c" }),
+      ["<C-j>"] = cmp_mapping(cmp_mapping.select_next_item(), { "i", "c" }),
+      ["<Down>"] = cmp_mapping(cmp_mapping.select_next_item { behavior = cmp_types.SelectBehavior.Select }, { "i" }),
+      ["<Up>"] = cmp_mapping(cmp_mapping.select_prev_item { behavior = cmp_types.SelectBehavior.Select }, { "i" }),
```
kylo252, line 285:
```suggestion
      ["<Down>"] = cmp_mapping(cmp_mapping.select_next_item { behavior = cmp.SelectBehavior.Select }, { "i" }),
      ["<Up>"] = cmp_mapping(cmp_mapping.select_prev_item { behavior = cmp.SelectBehavior.Select }, { "i" }),
```

lua/lvim/core/cmp.lua
```diff
@@ -277,24 +278,24 @@ M.config = function()
...
+      ["<C-d>"] = cmp_mapping.scroll_docs(-4),
+      ["<C-f>"] = cmp_mapping.scroll_docs(4),
+      ["<C-y>"] = cmp_mapping {
+        i = cmp_mapping.confirm { behavior = cmp_types.ConfirmBehavior.Replace, select = false },
```
kylo252, line 289:
```suggestion
        i = cmp_mapping.confirm { behavior = cmp.ConfirmBehavior.Replace, select = false },
```

lua/lvim/core/cmp.lua
```diff
@@ -277,24 +278,24 @@ M.config = function()
...
         c = function(fallback)
           if cmp.visible() then
-            cmp.confirm { behavior = cmp.ConfirmBehavior.Replace, select = false }
+            cmp.confirm { behavior = cmp_types.ConfirmBehavior.Replace, select = false }
```
kylo252, line 292:
```suggestion
            cmp.confirm { behavior = cmp.ConfirmBehavior.Replace, select = false }
```

lua/lvim/core/cmp.lua
```diff
@@ -317,16 +318,16 @@ M.config = function()
...
+      ["<C-e>"] = cmp_mapping.abort(),
+      ["<CR>"] = cmp_mapping(function(fallback)
         if cmp.visible() then
           local confirm_opts = vim.deepcopy(lvim.builtin.cmp.confirm_opts) -- avoid mutating the original opts below
           local is_insert_mode = function()
             return vim.api.nvim_get_mode().mode:sub(1, 1) == "i"
           end
           if is_insert_mode() then -- prevent overwriting brackets
-            confirm_opts.behavior = cmp.ConfirmBehavior.Insert
+            confirm_opts.behavior = cmp_types.ConfirmBehavior.Insert
```
kylo252, line 330:
```suggestion
            confirm_opts.behavior = cmp.ConfirmBehavior.Insert
```

opalmay:
> image
> 
> After initial install alpha doesn't seem to load correctly.

Just noticed this yesterday too after upgrading nvim.
I can reproduce on `NVIM v0.9.0-dev-834+g4c90a84c6` but not on `NVIM v0.8.2`

> Actually I'm also noticing some issues with using copilot and tabnine, seems like suggestion don't show up when they're loaded

I'm using Copilot and can't reproduce. Try checking the output of `:Copilot status`

lua/lvim/core/cmp.lua
LostNeophyte: no, cmp is not required with this, I checked with --startuptime

lua/lvim/core/bufferline.lua
LostNeophyte: when it was in settings, there was flickering when default tabline changed to bufferline when it's loaded

lua/lvim/core/cmp.lua
kylo252: we don't need it anyways, so let's remove this artificial guard.

PR #3754, opened 2023-01-19
fix(lazy): suggested fixes for pref/lazyloading branch

pr-313:
### Description

Adds some edits suggested in the comments for PR #(3750)


lua/lvim/plugins.lua
```diff
@@ -158,8 +158,10 @@ local core_plugins = {
     config = function()
       require("lvim.core.which-key").setup()
     end,
-    event = "VeryLazy",
+    event = {"User FileOpened"},
+    keys = {"<leader>"},
     enabled = lvim.builtin.which_key.active,
+    dependencies={"Comment.nvim"}
```
LostNeophyte, line 164:
```suggestion
```
that's not a good solution, comment.nvim should load when opening a file, it that doesn't work then the autocmd needs fixing

lua/lvim/plugins.lua
```diff
@@ -158,8 +158,10 @@ local core_plugins = {
     config = function()
       require("lvim.core.which-key").setup()
     end,
-    event = "VeryLazy",
+    event = {"User FileOpened"},
+    keys = {"<leader>"},
```
LostNeophyte, line 162:
```suggestion
    event = "VeryLazy",
```
whichkey should be always loaded on startup, it works for more keys than just leader

LostNeophyte (changes requested):

LostNeophyte:
Thank you!

pr-313:
@LostNeophyte 
Thank you for your feedback! 
I added a commit to account for your suggestions.
And apologies for the repeated force commits... it's my first time contributing :D

PR #3759, opened 2023-01-20
fix!: disable strange autopairs rules

meijieru:
Existing autopairs rules should be opt-in instead of as a default.

### DIFF
--- a/lua/lvim/core/autopairs.lua
+++ b/lua/lvim/core/autopairs.lua
@@ -55,7 +55,6 @@ M.setup = function()
   if not status_ok then
     return
   end
-  local Rule = require "nvim-autopairs.rule"
 
   autopairs.setup {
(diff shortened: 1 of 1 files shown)


LostNeophyte:
these were actually quite annoying for me personally

as for the todo in the code: yes, you can add rules from config.lua
```lua
lvim.builtin.autopairs.on_config_done = function()
  local npairs = require("nvim-autopairs")
  local Rule = require("nvim-autopairs.rule")
  npairs.add_rules({
    Rule("<", ">"),
  })
end
```

LostNeophyte (approved):
LGTM

LostNeophyte:
the rules were added in https://github.com/LunarVim/LunarVim/pull/315 which makes no sense to me

opalmay (approved):

LostNeophyte:
Thanks!

PR #3767, opened 2023-01-22
feat(alpha): add quit button to dashboard

anchal270899:
# Description

Added option to modify the autocommand that declares \<ESC\> as :qa when in filetype=alpha 
Added default as true to alpha.lua's main config, to not have any change to the current settings. 
Added "q to exit" to the alpha (dashboard mode) prompt list to make the q to :qa\<CR\> remap apparent.

## How Has This Been Tested?

- Tested with and without newly added setting `lvim.builtin.alpha.esc_to_quit = true/false` to only change the action if setting is added to personal config.

### DIFF
--- a/lua/lvim/core/alpha.lua
+++ b/lua/lvim/core/alpha.lua
@@ -15,6 +15,7 @@ function M.config()
       opts = { autostart = true },
     },
     active = true,
+    esc_to_quit = true,
     mode = "dashboard",
   }
 end
--- a/lua/lvim/core/alpha.lua
+++ b/lua/lvim/core/alpha.lua
@@ -15,7 +15,6 @@ function M.config()
       opts = { autostart = true },
     },
     active = true,
-    esc_to_quit = true,
     mode = "dashboard",
   }
 end
--- a/lua/lvim/core/autocmds.lua
+++ b/lua/lvim/core/autocmds.lua
@@ -80,11 +80,12 @@ function M.load_defaults()
         group = "_filetype_settings",
         pattern = "alpha",
         callback = function()
-          vim.cmd [[
-            nnoremap <silent> <buffer> q :qa<CR>
+          vim.opt.buflisted = false
+          if lvim.builtin.alpha.esc_to_quit then
+            vim.cmd [[
             nnoremap <silent> <buffer> <esc> :qa<CR>
-            set nobuflisted
           ]]
+          end
         end,
       },
     },
--- a/lua/lvim/core/autocmds.lua
+++ b/lua/lvim/core/autocmds.lua
@@ -82,7 +82,6 @@ function M.load_defaults()
         callback = function()
           vim.cmd [[
             set nobuflisted
-            nnoremap <silent> <buffer> <esc> :qa<CR>
           ]]
         end,
       },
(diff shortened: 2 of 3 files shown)


LostNeophyte (changes requested):
Can you remove the autocmd change and leave the "q to exit" one?

I plan to make all autocommands configurable in the future, I need to figure out the best way to do it, but I think that putting options in `lvim.builtin.plugin` will be problematic

for now you can use 
`vim.api.nvim_clear_autocmds { pattern = "alpha", group = "_filetype_settings" }`
to remove that autocomand

lua/lvim/core/alpha.lua
```diff
@@ -15,6 +15,7 @@ function M.config()
       opts = { autostart = true },
     },
     active = true,
+    esc_to_quit = true,
```
pr-313, line 18:
```suggestion
```

lua/lvim/core/autocmds.lua
```diff
@@ -80,11 +80,12 @@ function M.load_defaults()
         group = "_filetype_settings",
         pattern = "alpha",
         callback = function()
-          vim.cmd [[
-            nnoremap <silent> <buffer> q :qa<CR>
+          vim.opt.buflisted = false
+          if lvim.builtin.alpha.esc_to_quit then
+            vim.cmd [[
```
pr-313, line 85:
```suggestion
          vim.cmd [[
            set nobuflisted
            nnoremap <silent> <buffer> q :qa<CR>
```

lua/lvim/core/autocmds.lua
```diff
@@ -81,10 +81,11 @@ function M.load_defaults()
         pattern = "alpha",
         callback = function()
           vim.cmd [[
+            set nobuflisted
             nnoremap <silent> <buffer> q :qa<CR>
             nnoremap <silent> <buffer> <esc> :qa<CR>
-            set nobuflisted
           ]]
+          end
```
LostNeophyte, line 88:
```suggestion
```

that's a syntax error

lua/lvim/core/autocmds.lua
```diff
@@ -81,10 +81,11 @@ function M.load_defaults()
         pattern = "alpha",
         callback = function()
           vim.cmd [[
+            set nobuflisted
             nnoremap <silent> <buffer> q :qa<CR>
```
LostNeophyte, line 85:
```suggestion
```

you already added it as a button

lua/lvim/core/autocmds.lua
```diff
@@ -81,9 +81,8 @@ function M.load_defaults()
         pattern = "alpha",
         callback = function()
           vim.cmd [[
-            nnoremap <silent> <buffer> q :qa<CR>
-            nnoremap <silent> <buffer> <esc> :qa<CR>
             set nobuflisted
+            nnoremap <silent> <buffer> <esc> :qa<CR>
```
LostNeophyte, line 85: let's just get rid of ESC
```suggestion
```

LostNeophyte (approved):
Thanks!

I removed the ESC keybind completely, it's not that intuitive and `q` is enough

Issue #3164, opened 2022-10-06
Several plugins are getting values they don't care in their configurations


danielo515:
### Problem description

I noticed that several plugins are being called their setup functions with a lot of values they don't care about.
This is the result of too general programming and the lack of type checking on Lua.
Two good examples are Telescope and nvim-cmp:
https://github.com/LunarVim/LunarVim/blob/rolling/lua/lvim/core/telescope.lua#L54-L169
https://github.com/LunarVim/LunarVim/blob/rolling/lua/lvim/lsp/null-ls/init.lua#L12-L13

Note how Telescope is getting the values of on_config_done and disabled, and a lot of options that do not apply to nvim-cmp are being passed because there is a generic function to build the options

I propose to create an specific interface for plugin configurations, where the plugin specific configs are in a defined place (eg telescope.config) and the lvim related things in a separate one.

### LunarVim version

rolling

### Neovim version (>= 0.8.0)

NVIM v0.7.0 Build type: Release LuaJIT 2.1.0-beta3 Compiled by brew@HMBRW-A-001-M1-004.local  Features: +acl +iconv +tui See ":help feature-compile"     system vimrc file: "$VIM/sysinit.vim"   fall-back for $VIM: "/opt/homebrew/Cellar/neovim/0.7.0/share/nvim"  Run :checkhealth for more info

### Operating system/version

macOS 12.4

### Steps to reproduce

There is nothing to reproduce

### support info

This is not a user problem, just code to be fixed

danielo515:
Bump

LostNeophyte:
that would introduce a breaking change with no benefit (currently) and there were too many of these recently so I think that's something we'll have to deal with when it starts giving us issues

danielo515:
> that would introduce a breaking change with no benefit (currently) and there were too many of these recently so I think that's something we'll have to deal with when it starts giving us issues

Ok then, that's fair.
Let's wait for the next round of breaking changes 😂

LostNeophyte:
> Ok then, that's fair. Let's wait for the next round of breaking changes joy

well, here they come https://github.com/LunarVim/LunarVim/pull/3776

this issue will be fixed by https://github.com/LunarVim/LunarVim/pull/3776/commits/75b653cc623a8cd6de397e42f21b838a065eb0e0, [the pr](https://github.com/LunarVim/LunarVim/pull/3776) also includes more breaking changes, what do you think about this it?

PR #3776, opened 2023-01-25
refactor: use callbacks for configuring builtins

LostNeophyte:
problems with our current config architecture:

1. The whole config is loaded on startup:
   - We don't take full advantage of lazy loading
     (all builtin plugins' config tables are populated on startup)
   - It's limiting, not everything can be configured (e.g. icons, lazy.nvim)
   - high startup time
2. No consistency (builtin.plugin.option vs builtin.plugin.setup.option)
   We pass things like on_config_done and active to the setup
   function of plugins (#3164)

## The solution
```lua
lvim.builtin.pluginA.setting1 = value
lvim.builtin.pluginA.setting2 = value

lvim.builtin.pluginB.setup.setting1 = value
```
would become
```lua
lvim.builtin.pluginA.on_config = function()
  lvim.builtin.pluginA.opts.setting1 = value
  lvim.builtin.pluginA.opts.setting2 = value
end

lvim.builtin.pluginB.on_config = function()
  lvim.builtin.pluginB.opts.setting1 = value
end
```

benefits:
- faster startup
- ability to configure more things in the future (e.g. icons, lazy.nvim)
- more consistency

drawbacks:
- breaking change


danielo515:
Why not expect that function to just return the expected config rather than having it to modify the global lvim object?

LostNeophyte:
> Why not expect that function to just return the expected config rather than having it to modify the global lvim object? 

This way lsp completion works

PR #3811, opened 2023-02-04
perf(toggleterm): on-demand lazy load

opalmay:
Set keybinds in `init`
Load by any of them

### DIFF
--- a/lua/lvim/plugins.lua
+++ b/lua/lvim/plugins.lua
@@ -287,6 +287,14 @@ local core_plugins = {
     config = function()
       require("lvim.core.terminal").setup()
     end,
+    cmd = {
+      "ToggleTerm",
+      "TermExec",
+      "ToggleTermToggleAll",
+      "ToggleTermSendCurrentLine",
+      "ToggleTermSendVisualLines",
+      "ToggleTermSendVisualSelection",
+    },
     keys = lvim.builtin.terminal.open_mapping,
     lazy = true,
     enabled = lvim.builtin.terminal.active,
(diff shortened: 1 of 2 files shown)


lua/lvim/plugins.lua
```diff
@@ -265,11 +265,15 @@ local core_plugins = {
...
     "akinsho/toggleterm.nvim",
-    event = "VeryLazy",
     branch = "main",
+    init = function()
+      require("lvim.core.terminal").init()
+    end,
     config = function()
       require("lvim.core.terminal").setup()
     end,
+    keys = lvim.builtin.terminal.keys,
```
LostNeophyte, line 275: where is `lvim.builtin.terminal.keys` set?

opalmay: My bad, fixed

opalmay:
Note that after my planned execs refactor, the logic for init will change. It'll only set the keymaps without any further logic.

lua/lvim/plugins.lua
```diff
@@ -265,11 +265,23 @@ local core_plugins = {
...
+    cmd = {
+      "ToggleTerm",
+      "TermExec",
+      "ToggleTermToggleAll",
+      "ToggleTermSendCurrentLine",
+      "ToggleTermSendVisualLines",
+      "ToggleTermSendVisualSelection",
+    },
+    keys = lvim.builtin.terminal.open_mapping,
+    lazy = true,
```
LostNeophyte, line 284:
```suggestion
```
no need to set lazy to true if cmd/keys is set

LostNeophyte (approved):
LGTM

PR #3812, opened 2023-02-04
Refactor toggleterm exec

opalmay:
Users can now add execs more easily to `lvim.builtin.terminal.execs`
eg:
```lua
local execs = {
	{ keymap = "<leader>u", cmd = "sudo dmesg | less", desc = "Show Kernel Logs", size = 1 },
	{ keymap = "<C-4>", cmd = "neofetch && read", size = 0.6 },
	{ keymap = "<C-5>", cmd = "ls -l && read", direction = "horizontal" },
}
lvim.builtin.terminal.execs = vim.tbl_deep_extend("force", lvim.builtin.terminal.execs, execs)
```

Also refactored lazygit to use this.
Still needs deprecation handling.


opalmay:
@LostNeophyte How do you think we should handle deprecation here?
Probably the easiest would be to call this something different than execs, so it can be addressed in `_deprecated.lua` by the old table name

kylo252 (changes requested):
I'm not sure we want to add a breaking change to tweak this wrapper, as we mostly use this table as only to alter some common defaults.

At the very least, we can soft-deprecate the original table without doing a breaking change.

Alternatively, we can leave this as is, and suggest using the full API from toggle-term instead.

lua/lvim/core/which-key.lua
```diff
@@ -176,7 +176,6 @@ M.config = function()
       -- " Debugging
       g = {
         name = "Git",
-        g = { "<cmd>lua require 'lvim.core.terminal'.lazygit_toggle()<cr>", "Lazygit" },
         j = { "<cmd>lua require 'gitsigns'.next_hunk({navigation_message = false})<cr>", "Next Hunk" },
         k = { "<cmd>lua require 'gitsigns'.prev_hunk({navigation_message = false})<cr>", "Prev Hunk" },
```
kylo252, line 180: I was actually thinking that `[c` and `]c` instead for these. It would also be nice to have a faster keybind for a such common operation.

lua/lvim/core/terminal.lua
```diff
@@ -77,49 +78,69 @@ local function get_dynamic_terminal_size(direction, size)
...
-  end
+local function keybind_toggle(keybind)
+  local Terminal = require("toggleterm.terminal").Terminal
+  local term = Terminal:new(keybind)
+  term:toggle(keybind.size, keybind.direction)
 end
 
-M.add_exec = function(opts)
-  local binary = opts.cmd:match "(%S+)"
+local function add_keybind_keymap(keybind)
```
kylo252, line 87: any specific reason for renaming functions? keep in mind that this would break existing API

lua/lvim/core/terminal.lua
```diff
@@ -77,49 +78,69 @@ local function get_dynamic_terminal_size(direction, size)
...
+  local binary = keybind.cmd:match "(%S+)"
```
kylo252, line 88: I think calling this a `keybind` is confusing, you can go for `term_opts` instead if you just want something more descriptive than `opts`

lua/lvim/core/terminal.lua
opalmay: These functions are only used locally, so it makes sense to make them local.  The only code that uses it was also refactored. We could make this one public, but same name or not the opts are now passed directly to toggleterm:new, not the same API
More on the naming, see below

lua/lvim/core/terminal.lua
opalmay: I renamed the functions and parameters to match the new option name. As discussed on the Discord I'm not very keen on the new name. If you have a better suggestion let me know, and I could also keep the same name and handle deprecation with a loop instead.
Either way, I can change the parameter as you suggested

lua/lvim/core/which-key.lua
kylo252: it looked like you added those lines 😅

lua/lvim/core/terminal.lua
```diff
@@ -146,24 +166,4 @@ M.toggle_log_view = function(logfile)
...
-    cmd = "lazygit",
-    hidden = true,
-    direction = "float",
-    float_opts = {
-      border = "none",
-      width = 100000,
-      height = 100000,
-    },
-    on_open = function(_)
-      vim.cmd "startinsert!"
```
kylo252, line 161: we don't need this one anymore?

lua/lvim/core/terminal.lua
```diff
@@ -77,23 +78,60 @@ local function get_dynamic_terminal_size(direction, size)
...
+
+local function add_term_keymap(term_opts)
+  local binary = term_opts.cmd:match "(%S+)"
+  if vim.fn.executable(binary) ~= 1 then
+    Log:debug("Skipping configuring executable " .. binary .. ". Please make sure it is installed properly.")
+    return
+  end
+
+  local modes = { "n" }
+  if not term_opts.keymap:find "<leader>" then
```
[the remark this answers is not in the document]
opalmay: It's so that space wouldn't have a delay inside a terminal. I've added `<space>` too.

lua/lvim/core/terminal.lua
opalmay: Nope, lazygit is just another terminal in `lvim.builtin.terminal.keybinds` and handled like the others :)

lua/lvim/core/terminal.lua
```diff
@@ -77,23 +78,60 @@ local function get_dynamic_terminal_size(direction, size)
...
-      direction = direction,
-      size = function()
-        return get_dynamic_terminal_size(direction, exec[5])
-      end,
-    }
+    term_opts.direction = term_opts.direction or lvim.builtin.terminal.terminals_defaults.direction
+    term_opts.size = term_opts.size or lvim.builtin.terminal.terminals_defaults[term_opts.direction .. "_size"]
+    -- size is calculated dynamically as a percentage of the current buffer
+    term_opts.size = get_dynamic_terminal_size(term_opts.direction, term_opts.size)
+    term_opts.cmd = term_opts.cmd or lvim.builtin.terminal.shell
```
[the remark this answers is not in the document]
opalmay: Done. I wrote this before #3867

lua/lvim/core/terminal.lua
```diff
@@ -77,23 +78,60 @@ local function get_dynamic_terminal_size(direction, size)
...
-      count = i + 100,
-      direction = direction,
-      size = function()
-        return get_dynamic_terminal_size(direction, exec[5])
-      end,
-    }
+    term_opts.direction = term_opts.direction or lvim.builtin.terminal.terminals_defaults.direction
+    term_opts.size = term_opts.size or lvim.builtin.terminal.terminals_defaults[term_opts.direction .. "_size"]
+    -- size is calculated dynamically as a percentage of the current buffer
+    term_opts.size = get_dynamic_terminal_size(term_opts.direction, term_opts.size)
```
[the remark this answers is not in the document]
opalmay: Are you sure? It doesn't work and it works without a callback

lua/lvim/core/terminal.lua
```diff
@@ -77,23 +78,60 @@ local function get_dynamic_terminal_size(direction, size)
...
+
+--- Setup the terminal cmds
 M.init = function()
-  for i, exec in pairs(lvim.builtin.terminal.execs) do
-    local direction = exec[4] or lvim.builtin.terminal.direction
+  for i, term_opts in ipairs(lvim.builtin.terminal.terminals) do
+    -- size == 1 is a special case for full screen
+    if term_opts.size == 1 then
+      term_opts.direction = "float"
+      term_opts.float_opts = {
```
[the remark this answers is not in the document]
opalmay: This is actually pointless because it's handling `size = 1` which is a special case for fullscreen.
Nothing prevents you from setting `float_opts` and `direction = float` normally

lua/lvim/core/terminal.lua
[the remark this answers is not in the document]
kylo252: @opalmay, it should be able to take a function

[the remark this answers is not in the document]
kylo252: @opalmay, it should be able to take a function

https://github.com/akinsho/toggleterm.nvim/blob/31d38d11390bcd35a568fcc65a79b7d6ec89de62/lua/toggleterm/ui.lua#L258-L260

check their README for an example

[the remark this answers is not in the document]
kylo252: basically, it won't be _dynamically_ calculated if it's not a function.

PR #3827, opened 2023-02-08
docs(ts): add hint about `ensure_installed`

kylo252:
# Description

parsers without a strict filetype, e.g. "comment", are not handled when `auto_install = true`


LostNeophyte (approved):
LGTM

PR #3832, opened 2023-02-10
fix(alpha): rollback to older commit

LostNeophyte:
latest version of alpha breaks if you open a file first and then open the dashboard 
https://github.com/goolord/alpha-nvim/issues/178

### DIFF
--- a/snapshots/default.json
+++ b/snapshots/default.json
@@ -6,7 +6,7 @@
     "commit": "d404ec3"
   },
   "alpha-nvim": {
-    "commit": "d5fdeb2"
+    "commit": "21a0f25"
   },
   "bigfile.nvim": {
     "commit": "c1bad34"


opalmay (approved):

goolord:
the latest commit https://github.com/goolord/alpha-nvim/commit/0349fc0aa0c1d940ec3be395cb110483b416bc84 should be working fine, you can pin to this commit instead of rolling back

LostNeophyte:
> the latest commit [goolord/alpha-nvim@0349fc0](https://github.com/goolord/alpha-nvim/commit/0349fc0aa0c1d940ec3be395cb110483b416bc84) should be working fine, you can pin to this commit instead of rolling back

you're right, I should've checked it, we had a plugin bump recently and I didn't check if that was the latest version, I checked now and no problems with latest

Thanks for the fix and letting us know!

goolord:
nvm, I was actually able to reproduce this, fixed in https://github.com/goolord/alpha-nvim/commit/46f6cd41a701be55cce669d99bbf984b32e2f81a

PR #3841, opened 2023-02-12
fix(lsp): lua lsp renamed

meijieru:
The lua lsp is renamed to `lua_ls`.


opalmay:
Thanks! Can you fix the lint please?

@LostNeophyte Is the regeneration of ftplugins handled by `:LvimUpdate`?

PR #3846, opened 2023-02-14
fix: LvimUpdate should bump core plugins

LostNeophyte:


kylo252 (approved):
Let's see if this fixes the CI issues for https://github.com/LunarVim/LunarVim/pull/3836

PR #3853, opened 2023-02-16
fix(cmp): handle deprecated tree-sitter api

kylo252:
# Description

Override deprecated API from `nvim-treesitter` as a workaround for https://github.com/LunarVim/LunarVim/pull/3836#issuecomment-1427613551

### How has this been tested?

not sure how to trigger the error in the first place..


LostNeophyte (approved):
LGTM

Issue #3840, opened 2023-02-12
Stop using `$XDG_STATE_HOME/nvim`


REmerald:
### Feature Description

Logs and stuff must be separated from nvim and written to e.g. `$HOME/.local/state/lunarvim/` and NOT `$HOME/.local/state/nvim/`.

### Describe the alternatives you have considered

Use `$LUNARVIM_CACHE_DIR` instead?

### Support information

I also use my own neovim config in addition to lunarvim, and when I was trying to debug a lsp server in lunarvim, I realized that files like `lsp.log`, `mason.log` are written to the nvim's dir, not lunarvim's!

LostNeophyte:
related https://github.com/LunarVim/LunarVim/issues/3541

jeeeem:
maybe this can be useful - https://github.com/neovim/neovim/pull/22128

PR #3855, opened 2023-02-16
refactor(bootstrap)!: improve stdpath override

### LINKED ISSUE #3541: plenary logs are written to .cache/nvim and not .cache/lvim
`:lua print(vim.api.nvim_call_function("stdpath", { "cache" }))` (that's what plenary is using) returns `.cache/nvim`
for this reason :NullLsLog opens the wrong file which is empty

opalmay:
Fixes #3541 #3840 
Now overriding `stdpath` when called through `nvim_call_function` too, used to only apply to `vim.fn.stdpath`
It used to only have `config`, I've added `config` and `data` overrides too.
As per #2507, there was originally an issue with lsp-installer which is not used anymore. And after a bit of testing, nothing else seems broken.

Down side is, everything stored in `nvim` path would be lost. E.g. Mason binaries.
Thinking about making the move as smooth as possible, maybe it makes sense to copy the old data folder to the new location as part of the update process

### DIFF
--- a/lua/lvim/bootstrap.lua
+++ b/lua/lvim/bootstrap.lua
@@ -64,11 +64,14 @@ function M:init(base_dir)
   self.lazy_install_dir = join_paths(self.pack_dir, "lazy", "opt", "lazy.nvim")
 
   ---@meta overridden to use LUNARVIM_CACHE_DIR instead, since a lot of plugins call this function internally
-  ---NOTE: changes to "data" are currently unstable, see #2507
   ---@diagnostic disable-next-line: duplicate-set-field
   vim.fn.stdpath = function(what)
     if what == "cache" then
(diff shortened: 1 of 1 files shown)


LostNeophyte:
Nvim appname would be a better way to do this, we just need to test it https://twitter.com/Neovim/status/1626198815467266049?t=ppcV-o6QrjpROE84W5uPvw&s=19

opalmay:
> Nvim appname would be a better way to do this, we just need to test it https://twitter.com/Neovim/status/1626198815467266049?t=ppcV-o6QrjpROE84W5uPvw&s=19

I knew it was planned but amazing timing that it got merged today, lol.
Definitely needs testing, looking at the code we might still need to deal with changing paths

PR #3856, opened 2023-02-16
refactor: utilize $NVIM_APPNAME

opalmay:
New lvim launching script looks like this:
```bash
#!/usr/bin/env bash

export NVIM_APPNAME="${NVIM_APPNAME:-"lvim"}"
export LUNARVIM_BASE_DIR="/root/.local/share/$NVIM_APPNAME/core"

exec -a lvim nvim -u "$LUNARVIM_BASE_DIR/init.lua" "$@"
```

TODO:
- [x] ~~Bootstrap and other cleanups (plugin-loader.lua?)~~ We keep supporting the old env vars if set
- [x] Refactor install/uninstall bash scripts. $LUNARVIM_BASE_DIR for new installs is different, see above
- [x] ~~Powershell installer?~~ This shouldn't break the PS installer
- [x] ~~Implement handling of the changing data dirs during the update process (`.local/share/lunarvim` and `.local/share/nvim` to `.local/share/lvim`)~~
- [x] ~~Handle old installs that launch without `$NVIM_APPNAME`. Defaults to `lvim` if unset, but still breaking custom paths that don't follow `$NVIM_APPNAME` logic~~
- [ ] Bump nvim requirement to `0.9` #3858 (this branch's base)
- [ ] Fix CI tests
- [ ] Docs


lua/lvim/bootstrap.lua
```diff
@@ -63,14 +50,9 @@ function M:init(base_dir)
...
-  ---NOTE: changes to "data" are currently unstable, see #2507
-  ---@diagnostic disable-next-line: duplicate-set-field
-  vim.fn.stdpath = function(what)
-    if what == "cache" then
-      return _G.get_cache_dir()
-    end
-    return vim.call("stdpath", what)
+  -- for older installs. NVIM_APPNAME defaults to "lvim"
+  if not vim.env.NVIM_APPNAME then
+    vim.env.NVIM_APPNAME = "lvim"
```
LostNeophyte, line 55: does it work if it's under the lines above? it looks like they get the directories before APPNAME would be set so self.runtime_dir, self.config_dir and self.cache_dir would point to nvim dirs

opalmay:
@kylo252 please take a look at the unit tests. I think some tests are irrelevant now (commented).
plus, running `make test` locally overrides your config.lua because it can't use temp cache/config dirs now. Not sure how to deal with it (could make a copy then copy it back)

lua/lvim/bootstrap.lua
```diff
@@ -66,33 +53,33 @@ function M:init(base_dir)
...
     return base_dir
   end
 
-  if os.getenv "LUNARVIM_RUNTIME_DIR" then
-    vim.opt.rtp:remove(join_paths(vim.call("stdpath", "data"), "site"))
-    vim.opt.rtp:remove(join_paths(vim.call("stdpath", "data"), "site", "after"))
-    -- vim.opt.rtp:prepend(join_paths(self.runtime_dir, "site"))
-    vim.opt.rtp:append(join_paths(self.runtime_dir, "lvim", "after"))
-    vim.opt.rtp:append(join_paths(self.runtime_dir, "site", "after"))
+  -- if os.getenv "LUNARVIM_RUNTIME_DIR" then
```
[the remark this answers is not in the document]
kylo252: @opalmay, oh and we shouldn't need to bump the required version just yet, at least not for this for specific case.

utils/ci/run_test.sh
```diff
@@ -1,21 +1,23 @@
...
 
-export LUNARVIM_CONFIG_DIR LUNARVIM_CACHE_DIR
+export XDG_CACHE_HOME XDG_CONFIG_HOME
 
-printf "cache_dir: %s\nconfig_dir: %s\n" "$LUNARVIM_CACHE_DIR" "$LUNARVIM_CONFIG_DIR"
+printf "cache_dir: %s\nconfig_dir: %s\n" "$XDG_CACHE_HOME" "$XDG_CONFIG_HOME"
 
 lvim() {
-  nvim -u "$LUNARVIM_BASE_DIR/tests/minimal_init.lua" --cmd "set runtimepath+=$LUNARVIM_BASE_DIR" "$@"
+  nvim -u "$LUNARVIM_BASE_DIR/tests/minimal_init.lua" "$@"
```
kylo252, line 20: This will completely break other tools that otherwise works correctly now. For example, [ripgrep expects its config file under `XDG_CONFIG_DIR/ripgrep/ripgreprc`](https://github.com/BurntSushi/ripgrep/blob/master/GUIDE.md#configuration-file).

lua/lvim/bootstrap.lua
```diff
@@ -57,6 +57,11 @@ end
 ---Initialize the `&runtimepath` variables and prepare for startup
 ---@return table
 function M:init(base_dir)
+  -- for older installs. NVIM_APPNAME defaults to "lvim"
+  if not vim.env.NVIM_APPNAME then
+    vim.env.NVIM_APPNAME = "lvim"
+  end
```
kylo252, line 63: This will break the ability of using lvim's config as a neovim configuration, which used to be the best way to run on Windows.

utils/bin/lvim.template
```diff
@@ -1,9 +1,6 @@
 #!/usr/bin/env bash
 
-export LUNARVIM_RUNTIME_DIR="${LUNARVIM_RUNTIME_DIR:-RUNTIME_DIR_VAR}"
```
kylo252, line 3: Users have been using and expecting `LUNARVIM_xxx_DIR` variables to work for:
- customizing the installation path 
- have multiple installations
- have multiple configs (same base dir)
- maybe something else

kylo252 (changes requested):
We should introduce using `NVIM_APPNAME` without removing existing code, just to control the last few loose ends with `vim.lsp.log` and shada.

I'll create a new minimal draft to show how to do it without introducing a massive breaking change.

PR #3864, opened 2023-02-18
feat: regenerate lsp templates after LvimUpdate

LostNeophyte:


cpea2506:
https://github.com/LunarVim/LunarVim/blob/0a0cf29c8aa1fafb09dac1a36e30dec642ed24a9/lua/lvim/utils/hooks.lua#L59

This line already does the template regeneration for us.

LostNeophyte:
> https://github.com/LunarVim/LunarVim/blob/0a0cf29c8aa1fafb09dac1a36e30dec642ed24a9/lua/lvim/utils/hooks.lua#L59
> 
> This line already does the template regeneration for us.

but it does it before updating plugins

cpea2506:
> > https://github.com/LunarVim/LunarVim/blob/0a0cf29c8aa1fafb09dac1a36e30dec642ed24a9/lua/lvim/utils/hooks.lua#L59
> > 
> > This line already does the template regeneration for us.
> 
> but it does it before updating plugins

I couldn't see the difference here 😅. I have two options:
1) I think we can re-order them: Plugin update -> Reset cache
2) Remove this line instead
https://github.com/LunarVim/LunarVim/blob/0a0cf29c8aa1fafb09dac1a36e30dec642ed24a9/lua/lvim/utils/hooks.lua#L39

LostNeophyte:
>     1. Re-order them: Plugin update -> Reset cache
clearing cache before updating may have benefits, if we change `enabled` flag of some plugins I'm not sure if lazy will pick this up without clearing cache

>     2. Remove this line and let `post-hook` handle it
>        https://github.com/LunarVim/LunarVim/blob/0a0cf29c8aa1fafb09dac1a36e30dec642ed24a9/lua/lvim/utils/hooks.lua#L39

we can't remove that line, it's used in LvimCacheReset

cpea2506:
> > ```
> > 1. Re-order them: Plugin update -> Reset cache
> > ```
> 
> 
> clearing cache before updating may have benefits, if we change `enabled` flag of some plugins I'm not sure if lazy will pick this up without clearing cache
> 
> 
> > ```
> > 2. Remove this line and let `post-hook` handle it
> >    https://github.com/LunarVim/LunarVim/blob/0a0cf29c8aa1fafb09dac1a36e30dec642ed24a9/lua/lvim/utils/hooks.lua#L39
> > ```
> 
> 
> we can't remove that line, it's used in LvimCacheReset

I almost forgot it was used in a command. 🥲

lua/lvim/utils/hooks.lua
```diff
@@ -61,6 +61,7 @@ function M.run_post_update()
   Log:debug "Syncing core plugins"
   plugin_loader.reload { reload "lvim.plugins", lvim.plugins }
   plugin_loader.sync_core_plugins()
+  require("lvim.lsp.templates").generate_templates()
```
kylo252, line 64:
```suggestion
  M.reset_cache() -- force cache clear and templates regen once more  
```

PR #3867, opened 2023-02-18
fix(terminal): don't set  the shell by default

cpea2506:
# Description

I have an issue on Windows when my shell settings like `vim.opt.shell` or `lvim.builtin.terminal.shell` has no effect with `open_mapping` key.

Tested on other OSes as well.

## How Has This Been Tested?

- Setting `vim.opt.shell = "pwsh"` or `lvim.builtin.terminal.shell = "pwsh"`
- See that the shell change when pressing with `open_mapping` key like `<C-\>`

### DIFF
--- a/lua/lvim/core/terminal.lua
+++ b/lua/lvim/core/terminal.lua
@@ -78,13 +78,11 @@ local function get_dynamic_terminal_size(direction, size)
 end
 
 M.init = function()
-  lvim.builtin.terminal.shell = lvim.builtin.terminal.shell or vim.o.shell
-
   for i, exec in pairs(lvim.builtin.terminal.execs) do
     local direction = exec[4] or lvim.builtin.terminal.direction
 
     local opts = {
-      cmd = exec[1] or lvim.builtin.terminal.shell,
+      cmd = exec[1] or lvim.builtin.terminal.shell or vim.o.shell,
       keymap = exec[2],
       label = exec[3],
       -- NOTE: unable to consistently bind id/count <= 9, see #2146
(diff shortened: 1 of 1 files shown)


lua/lvim/core/terminal.lua
```diff
@@ -78,6 +78,8 @@ local function get_dynamic_terminal_size(direction, size)
 end
 
 M.init = function()
+  lvim.builtin.terminal.shell = lvim.builtin.terminal.shell or vim.o.shell
+
```
LostNeophyte, line 82:
```suggestion
```

vim.o.shell is toggleterm's default so it will use it if we pass nil

LostNeophyte (changes requested):

lua/lvim/core/terminal.lua
cpea2506: Screenshot 2023-02-18 at 21 18 20

There was an error because of this line 
https://github.com/LunarVim/LunarVim/blob/f43cc8e20b8443f052c8bd1dd74066eec514587b/lua/lvim/core/terminal.lua#L109

LostNeophyte: In that case I'd change line 85 https://github.com/LunarVim/LunarVim/blob/f43cc8e20b8443f052c8bd1dd74066eec514587b/lua/lvim/core/terminal.lua#L85
to
`cmd = exec[1] or lvim.builtin.terminal.shell or vim.o.shell,`

cpea2506: You are right, that's more concise

LostNeophyte (approved):
LGTM, thanks

I'd change the pr title to something like "don't set the shell by default"

PR #3869, opened 2023-02-18
fix(config-loader): defer invalid configuration warning

LostNeophyte:
before this pr the error message after loading invalid config disappeared after loading the theme and it was easily missed

### DIFF
--- a/lua/lvim/config/init.lua
+++ b/lua/lvim/config/init.lua
@@ -51,7 +51,9 @@ function M:load(config_path)
   local ok, err = pcall(dofile, config_path)
   if not ok then
     if utils.is_file(user_config_file) then
-      Log:warn("Invalid configuration: " .. err)
+      vim.schedule(function()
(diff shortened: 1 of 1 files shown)


lua/lvim/config/init.lua
```diff
@@ -51,7 +51,9 @@ function M:load(config_path)
   local ok, err = pcall(dofile, config_path)
   if not ok then
     if utils.is_file(user_config_file) then
-      Log:warn("Invalid configuration: " .. err)
+      vim.schedule(function()
+        Log:warn("Invalid configuration: " .. err)
+      end)
     else
       vim.notify_once(
```
kylo252, line 58: it's not needed here as well then?

LostNeophyte: yup, done

kylo252 (approved):
LGTM! :rocket:

PR #3892, opened 2023-02-24
fix: lazy cache

opalmay:
Calling `lazy.cache.enable()` in `plugin_loader.init()` doesn't work properly, it is before calling lazy setup and then the lazy setup calls `cache.enable` again on it's own.
Instead only override the path and let lazy do the rest. (already enabled in default lazy config)
Also remove the `pcall`, this is an internal variable and we would wanna know if it breaks in the future.
`reset_cache` is also removed as looks like that's not necessary.

### DIFF
--- a/lua/lvim/plugin-loader.lua
+++ b/lua/lvim/plugin-loader.lua
@@ -47,9 +47,11 @@ function plugin_loader.init(opts)
   vim.opt.runtimepath:append(lazy_install_dir)
   vim.opt.runtimepath:append(join_paths(plugins_dir, "*"))
 
-  -- set a custom path for lazy's cache
-  local lazy_cache = require "lazy.core.cache"
-  lazy_cache.path = join_paths(get_cache_dir(), "lazy", "luac")
+  pcall(function()
+    -- set a custom path for lazy's cache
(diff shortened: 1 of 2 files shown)


LostNeophyte (changes requested):
> it is before calling lazy setup and then the lazy setup calls cache.enable again on it's own.

it's not a problem, lazy checks if it was already enabled

> Also remove the pcall, this is an internal variable and we would wanna know if it breaks in the future.

we need the pcall to support older versions of lazy

before this change lvim started 1.14x faster, but we unfortunately need it
image

opalmay:
> > it is before calling lazy setup and then the lazy setup calls cache.enable again on it's own.
> 
> it's not a problem, lazy checks if it was already enabled
> 
> > Also remove the pcall, this is an internal variable and we would wanna know if it breaks in the future.
> 
> we need the pcall to support older versions of lazy
> 
> before this change lvim started 1.14x faster, but we unfortunately need it image

Added the pcall back.
If it already checks then root cause might've been `enable()` being called before `setup()`. Not exactly sure, but it works this way.

LostNeophyte (approved):
LGTM

PR #2540, opened 2022-04-29
fix(lsp): undo stdpath overload to avoid datarace

kylo252:
# Description

~~Changes to "data" are currently unstable~~

- fix(lsp): undo stdpath overload to avoid datarace
- chore(linting): minor fixes

If the user adds an LSP setup call in their `config.lua`

```lua 
require("lvim.lsp.manager").setup("ccls", {})
```
then that will be executed before we've had a chance to call `require("nvim-lsp-installer").setup {}`, which happens in `require("lvim.lsp").setup()` which in turns relies on values from `config.lua`.. 

so it seems currently difficult to allow for both, instead, we'll revert the stdpath overload to keep things consistent with the old settings 😞 

Fixes #2507

## How Has This Been Tested?

It's already the default behavior, and having this in `config.lua` no longer throws an error

```lua
require("lvim.lsp.manager").setup("emmet_ls", {})
```

PR #3893, opened 2023-02-24
refactor(bootstrap): better overrides

opalmay:
Same idea as #3855
As we are not going to deprecate not using $NVIM_APPNAME, this is still relevant.
I removed the `data` override so that it's not breaking. But at least things like `NullLsLog` will show correctly.

### DIFF
--- a/lua/lvim/bootstrap.lua
+++ b/lua/lvim/bootstrap.lua
@@ -64,11 +64,14 @@ function M:init(base_dir)
   self.lazy_install_dir = join_paths(self.pack_dir, "lazy", "opt", "lazy.nvim")
 
   ---@meta overridden to use LUNARVIM_CACHE_DIR instead, since a lot of plugins call this function internally
-  ---NOTE: changes to "data" are currently unstable, see #2507
   ---@diagnostic disable-next-line: duplicate-set-field
   vim.fn.stdpath = function(what)
     if what == "cache" then
       return _G.get_cache_dir()
+    elseif what == "config" then
+      return _G.get_config_dir()
+    elseif what == "data" then
+      return _G.get_runtime_dir()
     end
     return vim.call("stdpath", what)
   end
--- a/lua/lvim/bootstrap.lua
+++ b/lua/lvim/bootstrap.lua
@@ -73,6 +73,13 @@ function M:init(base_dir)
     return vim.call("stdpath", what)
   end
 
+  vim.api.nvim_call_function = function(fn, args)
+    if fn == "stdpath" then
+      return vim.fn.stdpath(args[1])
+    end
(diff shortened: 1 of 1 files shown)


lua/lvim/bootstrap.lua
```diff
@@ -64,15 +64,25 @@ function M:init(base_dir)
...
       return _G.get_cache_dir()
+    elseif what == "config" then
+      return _G.get_config_dir()
+      -- elseif what == "data" then
+      --   return _G.get_runtime_dir()
     end
     return vim.call("stdpath", what)
   end
 
+  vim.api.nvim_call_function = function(fn, args)
```
kylo252, line 79: is this function commonly used? or is it for covering all the basis? I'm not sure it's worth it if it's the latter, since there are quite a few ways that can still access these values.

lua/lvim/bootstrap.lua
```diff
@@ -64,15 +64,25 @@ function M:init(base_dir)
...
   ---@meta overridden to use LUNARVIM_CACHE_DIR instead, since a lot of plugins call this function internally
-  ---NOTE: changes to "data" are currently unstable, see #2507
   ---@diagnostic disable-next-line: duplicate-set-field
   vim.fn.stdpath = function(what)
     if what == "cache" then
       return _G.get_cache_dir()
+    elseif what == "config" then
+      return _G.get_config_dir()
+      -- elseif what == "data" then
+      --   return _G.get_runtime_dir()
```
kylo252, line 74:
```suggestion
```

lua/lvim/bootstrap.lua
```diff
@@ -64,15 +64,25 @@ function M:init(base_dir)
   self.lazy_install_dir = join_paths(self.pack_dir, "lazy", "opt", "lazy.nvim")
 
   ---@meta overridden to use LUNARVIM_CACHE_DIR instead, since a lot of plugins call this function internally
-  ---NOTE: changes to "data" are currently unstable, see #2507
```
kylo252, line 67: we should keep this note :smiling_face_with_tear:

lua/lvim/bootstrap.lua
```diff
@@ -64,15 +64,25 @@ function M:init(base_dir)
...
   ---@diagnostic disable-next-line: duplicate-set-field
   vim.fn.stdpath = function(what)
     if what == "cache" then
       return _G.get_cache_dir()
+    elseif what == "config" then
```
kylo252, line 71: Don't forget to update the corresponding test, see https://github.com/LunarVim/LunarVim/pull/2540/files#diff-e02939ff80b05af32ad541b952d46570ab42ca3386e9eeab4bd4990955af4687

PR #3903, opened 2023-02-28
chore(custom-finders): solve deprecation warning

mamaraddio:
# Description

`_close` in `telescope.actions` is deprecated, telescope docs says to use `close` instead


LostNeophyte (approved):
Thanks!

PR #3904, opened 2023-02-28
add commit, vim-dadbot-completion, cmp-tw2css plugins to cmp config

thyseus:


rebuilt:
Is it for performance reasons that we haven't added everything on this list?  https://github.com/hrsh7th/nvim-cmp/wiki/List-of-sources

If the performance hit is negligable, I don't see why we can't add more items to the list of sources.  
If adding more sources is noticeable, especially on slow machines, then we should document the expected way completion sources should be added:
```lua
lvim.builtin.cmp.formatting.source_names["copilot"] = "(Copilot)"
table.insert(lvim.builtin.cmp.sources, 1, { name = "copilot" })
```

thyseus:
A similar discussion is going on here:

https://github.com/LunarVim/LunarVim/pull/2542

unfortunately my lua skills are not high enough to do it :(
thanks a lot for that lines, @rebuilt thats what i needed !

PR #3915, opened 2023-03-03
ci(issue_template): require additional terminal information

cpea2506:
Request the user to provide the terminal name for more effortless issue resolve

### DIFF
--- a/.github/ISSUE_TEMPLATE/general-issue-form.yaml
+++ b/.github/ISSUE_TEMPLATE/general-issue-form.yaml
@@ -41,6 +41,15 @@ body:
     validations:
       required: true
   - type: input
+    id: terminal-name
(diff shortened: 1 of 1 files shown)


rebuilt:
I can see how it would be useful.   Should it be required?  I'm not sure but I'm open to being convinced.

cpea2506:
> I can see how it would be useful. Should it be required? I'm not sure but I'm open to being convinced.

Yes, I think so. If we know the user's terminal, we can easily find out their related issues or point the user to how they can go to the settings of their terminal directly.

rebuilt (approved):
@cpea2506  I performed a search of closed issues that deal with the terminal in some way an there were more than I expected so I'll approve this.

PR #3930, opened 2023-03-07
feat(terminal): fire neo tree git event when closing lazygit

MisterFISHUP:
Hi,

Thank you for the awesome project. I use neo-tree and found that the git status in neo-tree is not in sync after performing some git actions in lazygit through toggleterm. I then found a way to trigger git status update in neo-tree when closing lazygit (cf https://github.com/nvim-neo-tree/neo-tree.nvim/issues/337#issuecomment-1114340769). I think it would be useful to add this in lunarvim. Protected call is used so the code won't do anything if neo-tree is not used.

### DIFF
--- a/lua/lvim/core/terminal.lua
+++ b/lua/lvim/core/terminal.lua
@@ -160,7 +160,13 @@ M.lazygit_toggle = function()
     on_open = function(_)
       vim.cmd "startinsert!"
     end,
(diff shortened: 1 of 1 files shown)


LostNeophyte:
neotree is not a core plugin, so this code shouldn't be in lunarvim, you can add it to your config.lua

PR #3951, opened 2023-03-12
fix(dap): invalid border value

meijieru:
`vim.g.border_chars` is not defined.


rebuilt (approved):
`:lua print(vim.inspect(vim.g.border_chars))` results in nil.

rebuilt:
Thanks!

PR #3953, opened 2023-03-12
feat(lang): yaml use schemastore

meijieru:
https://github.com/b0o/SchemaStore.nvim/issues/3 get merged.


lvimuser:
Why do we need to remove the `schemaStore` section? Wouldn't `schemas` be enough?

meijieru:
Fixed.

kylo252:
> Is it necessary to remove the schemaStore section?

agreed!

@meijieru, did you check that the schemas aren't conflicting/duplicated with the official store?

ChristianChiarulli (approved):

PR #3975, opened 2023-03-24
feat(cmp): enable the user to easily add new completion sources

ww-daniel-mora:
# Description
* Consolidate lvim.builtin.cmp.formatting.source_names and lvim.builtin.cmp.formatting.duplicates into a new lvim.builtin.cmp.formatting.source_settings table
* Update lvim.builtin.cmp.formatting.format function to use source_settings table to determine the menu name, icon, highlight group, and duplicate flag
* Add menu names for crates, and lab.quick_data
* These changes make it easy for users to add new completion sources by adding a new entry to the lvim.builtin.cmp.formatting.source_settings table
summary of the change
* This change will break any users who accessed lvim.builtin.cmp.formatting.source_names or lvim.builtin.cmp.formatting.duplicates directly

## How Has This Been Tested?
Manual testing in config.lua
```
lvim.plugins = {
...
  {
    "jcdickinson/codeium.nvim",
    dependencies = { "hrsh7th/nvim-cmp" },
    config = function()
      require("codeium").setup({})
    end
  },
}
table.insert(lvim.builtin.cmp.sources, { name = "codeium" })
lvim.builtin.cmp.source_settings.codeium = {
  menu_name = "(Codeium)",
  icon = "",
  hl_group = "CmpItemKindTabnine",
}
```

### DIFF
--- a/lua/lvim/core/cmp.lua
+++ b/lua/lvim/core/cmp.lua
@@ -156,62 +156,66 @@ M.config = function()
       fields = { "kind", "abbr", "menu" },
       max_width = 0,
       kind_icons = lvim.icons.kind,
-      source_names = {
-        nvim_lsp = "(LSP)",
-        emoji = "(Emoji)",
-        path = "(Path)",
-        calc = "(Calc)",
-        cmp_tabnine = "(Tabnine)",
-        vsnip = "(Snippet)",
-        luasnip = "(Snippet)",
-        buffer = "(Buffer)",
-        tmux = "(TMUX)",
-        copilot = "(Copilot)",
-        treesitter = "(TreeSitter)",
-      },
-      duplicates = {
-        buffer = 1,
-        path = 1,
-        nvim_lsp = 0,
-        luasnip = 1,
+      source_settings = {
+        copilot = {
+          menu_name = "(Copilot)",
+          icon = lvim.icons.git.Octoface,
+          hl_group = "CmpItemKindCopilot",
+        },
+        emoji = {
+          menu_name = "(Emoji)",
+          icon = lvim.icons.misc.Smiley,
+          hl_group = "CmpItemKindEmoji",
+        },
+        cmp_tabnine = {
+          menu_name = "(Tabnine)",
+          icon = lvim.icons.misc.Robot,
+          hl_group = "CmpItemKindTabnine",
+        },
+        crates = {
+          menu_name = "(Crates)",
+          icon = lvim.icons.misc.Package,
+          hl_group = "CmpItemKindCrate",
+        },
+        ["lab.quick_data"] = {
+          menu_name = "(QuickData)",
+          icon = lvim.icons.misc.CircuitBoard,
+          hl_group = "CmpItemKindConstant",
+        },
+        nvim_lsp = {
+          menu_name = "(LSP)",
+          duplicates = 0,
+        },
+        path = {
+          menu_name = "(Path)",
+          duplicates = 1,
+        },
+        luasnip = {
+          menu_name = "(Snippet)",
+          duplicates = 1,
+        },
+        buffer = {
+          menu_name = "(Buffer)",
+          duplicates = 1,
+        },
+        calc = { menu_name = "(Calc)" },
+        vsnip = { menu_name = "(Snippet)" },
+        treesitter = { menu_name = "(TreeSitter)" },
+        tmux = { menu_name = "(TMUX)" },
       },
       duplicates_default = 0,
       format = function(entry, vim_item)
         local max_width = lvim.builtin.cmp.formatting.max_width
+        local source_settings = lvim.builtin.cmp.formatting.source_settings[entry.source.name]
         if max_width ~= 0 and #vim_item.abbr > max_width then
           vim_item.abbr = string.sub(vim_item.abbr, 1, max_width - 1) .. lvim.icons.ui.Ellipsis
         end
         if lvim.use_icons then
-          vim_item.kind = lvim.builtin.cmp.formatting.kind_icons[vim_item.kind]
-
-          if entry.source.name == "copilot" then
-            vim_item.kind = lvim.icons.git.Octoface
-            vim_item.kind_hl_group = "CmpItemKindCopilot"
-          end
-
-          if entry.source.name == "cmp_tabnine" then
-            vim_item.kind = lvim.icons.misc.Robot
-            vim_item.kind_hl_group = "CmpItemKindTabnine"
-          end
-
-          if entry.source.name == "crates" then
-            vim_item.kind = lvim.icons.misc.Package
-            vim_item.kind_hl_group = "CmpItemKindCrate"
-          end
-
-          if entry.source.name == "lab.quick_data" then
-            vim_item.kind = lvim.icons.misc.CircuitBoard
-            vim_item.kind_hl_group = "CmpItemKindConstant"
-          end
-
-          if entry.source.name == "emoji" then
-            vim_item.kind = lvim.icons.misc.Smiley
-            vim_item.kind_hl_group = "CmpItemKindEmoji"
-          end
+          vim_item.kind = source_settings.icon or lvim.builtin.cmp.formatting.kind_icons[vim_item.kind]
+          vim_item.kind_hl_group = source_settings.hl_group
         end
-        vim_item.menu = lvim.builtin.cmp.formatting.source_names[entry.source.name]
-        vim_item.dup = lvim.builtin.cmp.formatting.duplicates[entry.source.name]
-          or lvim.builtin.cmp.formatting.duplicates_default
+        vim_item.menu = source_settings.menu_name
+        vim_item.dup = source_settings.duplicates or lvim.builtin.cmp.formatting.duplicates_default
         return vim_item
       end,
     },


kylo252 (changes requested):
I'm not sure what the overall benefit here is, since you can already adjust/add/delete sources directly by accessing `lvim.builtin.cmp.sources`.

I suggest:
- consolidating `source_settings` to the `sources` table instead, and avoid adding new options
- deprecating any redundant _and_ incompatible options that we already have (be sure **not** to break backwards compatibility)

ww-daniel-mora:
Hey @kylo252 thank you for taking a quick look.
As I mentioned in the description this change consolidates `lvim.builtin.cmp.formatting.source_names` `lvim.builtin.cmp.formatting.duplicates` into a new table `lvim.builtin.cmp.formatting.source_settings`

While you are correct that users can add new sources via the `lvim.builtin.cmp.sources` table, users cannot currently easily configure how the sources are displayed in the completion menu, as menu styling is set in the `lvim.builtin.cmp.formatting.format` function, which has special logic for styling `copilot`, `tabnine`, `crates` and a few other sources.

This change replaces the source specific logic with more generic logic that retrieves all the completion menu settings from the `source_settings` table for every source. This means that users can now if the user adds a new source via the `lvim.builtin.cmp.sources` table they can optionally add an entry to `lvim.builtin.cmp.formatting.source_settings` with the completion menu icon and completion menu name, among other settings.

If you take a look at the testing example in the PR it might help demonstrate what this PR enables.

As for backwards compatibility I did not find either of the deleted tables in any documentation so I believe they are internal. Please let me know if this is incorrect and they are considered part of the public API and I would be happy to deprecate the tables rather than remove them.

Thank you again. I'm looking forward to your response.

ww-daniel-mora:
@kylo252  I wanted to check back in before I make additional changes. I had some questions above and your guidance would be greatly appreciated.

1. I want to make sure the benefit of the change is clear (easily add menu formatting for new sources)
2. I wanted to clarify consolidation `source_settings` is already a consolidation of 2 tables. Should I collapse them into `sources` instead
3. Can I delete `source_names` and `duplicates` since they are not documented or are they considered part of the public API and should therefore should be deprecated first.

kylo252:
> While you are correct that users can add new sources via the lvim.builtin.cmp.sources table, users cannot currently easily configure how the sources are displayed in the completion menu, as menu styling is set in the lvim.builtin.cmp.formatting.format function, which has special logic for styling copilot, tabnine, crates and a few other sources.

is there no additional API that we can use from cmp's side?

cc: @abzcoding, @LostNeophyte

ww-daniel-mora:
Great question @kylo252  that I had not considered. The cmp way to control menu formatting is via the `formatting.format` function passed to setup ([cmp wiki example](https://github.com/hrsh7th/nvim-cmp/wiki/Menu-Appearance#show-devicons-as-kind-field)).

Looking through cmp.lua the `lvim.builtin.cmp` object we are building up is actually what we pass directly to [cmp setup](https://github.com/LunarVim/LunarVim/blob/476743d2d174c11541eb6c4dd70ab3e0dc521049/lua/lvim/core/cmp.lua#L370)

So this function `formatting.format` **is** the cmp way of setting the icons. Cmp has not documented a way of setting dev icons outside of setup.

There is an available workaround for this use case if the team choses not to accept the PR. LunarVim users can wrap and override the formatting function.

To satisfy my particular use case the following code in config works:
```lua
-- codeium setup
table.insert(lvim.builtin.cmp.sources, { name = "codeium" })
lvim.builtin.cmp.formatting.source_names.codeium = "(Codeium)"
local default_format = lvim.builtin.cmp.formatting.format
lvim.builtin.cmp.formatting.format = function(entry, vim_item)
  vim_item = default_format(entry, vim_item)
  if entry.source.name == "codeium" then
    vim_item.kind = ""
    vim_item.kind_hl_group = "CmpItemKindTabnine"
  end
  return vim_item
end
```

That being said I still believe the table approach in this PR is an improvement. Besides enabling new completion sources with dev icons it simplifies the `formatting.format` function replacing many if checks with a single table lookup.

ww-daniel-mora:
Hey @kylo252 just wanted to followup here. To succinctly reiterate my earlier responses & questions:

1. No, there is no other cmp way to control menu formatting
2. This is already a consolidation of two existing tables
3. Do undocumented tables need to be deprecated before being deleted or can we just drop them?

If accepted, as a followup to this PR I would also like to contribute updates to the documentation on how to add new cmp sources with proper menu formatting.

lua/lvim/core/cmp.lua
```diff
@@ -156,62 +156,66 @@ M.config = function()
...
+        },
+        calc = { menu_name = "(Calc)" },
+        vsnip = { menu_name = "(Snippet)" },
+        treesitter = { menu_name = "(TreeSitter)" },
+        tmux = { menu_name = "(TMUX)" },
       },
       duplicates_default = 0,
       format = function(entry, vim_item)
         local max_width = lvim.builtin.cmp.formatting.max_width
+        local source_settings = lvim.builtin.cmp.formatting.source_settings[entry.source.name]
```
kylo252, line 209: you need to make sure this exists

```suggestion
        local source_settings = vim.tbl_get(lvim.builtin.cmp.formatting.source_settings, entry.source.name) or {}
```

lua/lvim/core/cmp.lua
```diff
@@ -156,62 +156,66 @@ M.config = function()
...
+        cmp_tabnine = {
+          menu_name = "(Tabnine)",
+          icon = lvim.icons.misc.Robot,
+          hl_group = "CmpItemKindTabnine",
+        },
+        crates = {
+          menu_name = "(Crates)",
+          icon = lvim.icons.misc.Package,
+          hl_group = "CmpItemKindCrate",
+        },
```
kylo252, line 179:
```suggestion
        copilot = {
          label = "(Copilot)",
          kind = lvim.icons.git.Octoface,
          kind_hl_group = "CmpItemKindCopilot",
        },
        emoji = {
          label = "(Emoji)",
          kind = lvim.icons.misc.Smiley,
          kind_hl_group = "CmpItemKindEmoji",
        },
        cmp_tabnine = {
          label = "(Tabnine)",
          kind = lvim.icons.misc.Robot,
          kind_hl_group = "CmpItemKindTabnine",
        },
        crates = {
          label = "(Crates)",
          kind = lvim.icons.misc.Package,
          kind_hl_group = "CmpItemKindCrate",
        },
```

lua/lvim/core/cmp.lua
```diff
@@ -156,62 +156,66 @@ M.config = function()
...
-          if entry.source.name == "lab.quick_data" then
-            vim_item.kind = lvim.icons.misc.CircuitBoard
-            vim_item.kind_hl_group = "CmpItemKindConstant"
-          end
-
-          if entry.source.name == "emoji" then
-            vim_item.kind = lvim.icons.misc.Smiley
-            vim_item.kind_hl_group = "CmpItemKindEmoji"
-          end
+          vim_item.kind = source_settings.icon or lvim.builtin.cmp.formatting.kind_icons[vim_item.kind]
```
kylo252, line 214:
```suggestion
          vim_item.kind = source_settings.kind or lvim.builtin.cmp.formatting.kind_icons[vim_item.kind]
```

lua/lvim/core/cmp.lua
```diff
@@ -156,62 +156,66 @@ M.config = function()
...
+          vim_item.kind_hl_group = source_settings.hl_group
         end
-        vim_item.menu = lvim.builtin.cmp.formatting.source_names[entry.source.name]
-        vim_item.dup = lvim.builtin.cmp.formatting.duplicates[entry.source.name]
-          or lvim.builtin.cmp.formatting.duplicates_default
+        vim_item.menu = source_settings.menu_name
```
kylo252, line 217:
```suggestion
        vim_item.menu = source_settings.label or ("(" .. entry.source.name .. ")")
```

kylo252 (changes requested):
Sorry for the delay!

lua/lvim/core/cmp.lua
```diff
@@ -156,62 +156,66 @@ M.config = function()
...
-        tmux = "(TMUX)",
-        copilot = "(Copilot)",
-        treesitter = "(TreeSitter)",
-      },
-      duplicates = {
-        buffer = 1,
-        path = 1,
-        nvim_lsp = 0,
-        luasnip = 1,
+      source_settings = {
```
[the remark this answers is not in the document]
ww-daniel-mora: @kylo252 I actually really like because there are already a few source settings sprinkled into the sources list, see copilot for an example:

```lua
sources = {
      {
        name = "copilot",
        -- keyword_length = 0,
        max_item_count = 3,
        trigger_characters =
-- etc.
```
So consolidating the rest of the settings in there make sense. However, as you can see, the sources object is a list so we can't index in as you suggested `cmp.sources[foo].options`. We could change it to that format i.e.

```lua
sources = {
  copilot = { 
    -- options for copilot
  }
  other_source = {
   -- other options
  }
```
This would give us a single place to add sources and configure them in a way that I thinks makes sense, of course we would need to update everything that access the sources list to handle the new format.

ww-daniel-mora (commented):
Hey @kylo252  I've accepted all your suggestions, thank you, and I responded to your consolidation comment with some additional questions. Let me know what you think.

PR #3986, opened 2023-03-27
fix(nvimtree): don't overwrite update_focused_file.ignore_list

### LINKED ISSUE #3985: NvimTree Configs Overwritten
Hello,

I have a nvim-tree config that get overwritten by the following line of code, specifically:
> *nvim-tree.update_focused_file.ignore_list*
    List of buffer names and filetypes that will not update the root dir
    of the tree if the file isn't found under the current root directory.
    Only relevant when `update_focused_file.update_root` and
    `update_focused_file.enable` are `true`.
      Type: {string}, Default: `{}`

Is there a workaround for this or is there a bug?

https://github.com/LunarVim/LunarVim/blob/83aadfe66e9eac8f6441985d988eaedfe63fcf83/lua/lvim/core/nvimtree.lua#L248

I changed it to the following and it fixed my problem in my local `~/.local/share/lunarvim/lvim/lua/lvim/core/nvimtree.lua`:
```lua
lvim.builtin.nvimtree.setup.update_focused_file.enable = true
lvim.builtin.nvimtree.setup.update_focused_file.update_cwd = true
```

gtscherer:
# Description

Prevents overwrite of `nvim-tree.update_focused_file.ignore_list` configuration.

Fixes #3985

## How Has This Been Tested?

- Create a configuration with an ignore_list for nvim-tree. That list should be respected after the fix is applied.
- Specifically I use a plugin called DiffView, and every time I execute `DiffviewOpen` when nvim-tree is open, I get a nasty error taking up my whole screen. With the following line in my `config.lua`, it is supposed to prevent the error from appearing: `lvim.builtin.nvimtree.setup.update_focused_file.ignore_list = { 'DiffViewFiles' }`. However, without this fix the error still appears.

### DIFF
--- a/lua/lvim/core/nvimtree.lua
+++ b/lua/lvim/core/nvimtree.lua
@@ -245,7 +245,8 @@ function M.setup()
   if lvim.builtin.project.active then
     lvim.builtin.nvimtree.setup.respect_buf_cwd = true
     lvim.builtin.nvimtree.setup.update_cwd = true
-    lvim.builtin.nvimtree.setup.update_focused_file = { enable = true, update_cwd = true }
+    lvim.builtin.nvimtree.setup.update_focused_file.enable = true
+    lvim.builtin.nvimtree.setup.update_focused_file.update_cwd = true
   end
 
   local function telescope_find_files(_)


LostNeophyte:
thanks!

LostNeophyte (approved):

PR #3991, opened 2023-03-29
feat(mason): add on_config_done option to lvim.builtin.mason

choplin:
# Description

This PR adds the `on_config_done` option to `lvim.builtin.mason`. This behaves like `on_config_done` for other plugins, such as `lvim.builtin.nvimtree.on_config_done`. That is, if a function is assigned to `on_config_done`, it will be called at the end of the `lvim.core.mason.setup()` with an argument of the loaded `mason` module.

This enables users to define setup logic in `config.lua` using the correctly set up `mason` module. 
For example, `require("mason-registry").get_installed_packages()` does not work as expected before `mason` is set up.

## How Has This Been Tested?

- Assign a function to `lvim.builtin.mason.on_config_done`,  and check if the `mason` module has been setup properly when that function is called.


LostNeophyte (approved):
LGTM, thanks!

ChristianChiarulli (approved):

PR #4004, opened 2023-04-04
chore(plugins): remove `popup.nvim` as it is deprecated

hehelego:
According to https://github.com/nvim-lua/plenary.nvim/pull/209 , `nvim-lua/popup.nvim` is deprecated and was fully merged into `plenary.nvim`.   
`popup.nvim` is originally installed as a dependency of `telescope.nvim` but it is nolonger required, see https://github.com/nvim-telescope/telescope.nvim#required-dependencies

I've removed `popup.nvim` from the plugin list.

### DIFF
--- a/lua/lvim/plugins.lua
+++ b/lua/lvim/plugins.lua
@@ -38,8 +38,6 @@ local core_plugins = {
     lazy = lvim.colorscheme ~= "lunar",
   },
   { "Tastyep/structlog.nvim", lazy = true },
-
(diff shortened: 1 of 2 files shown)


TheBearodactyl (approved):

ChristianChiarulli (approved):

ChristianChiarulli:
actually we should remove this from the snapshot as well

kylo252 (approved):
Thanks!

Issue #4037, opened 2023-04-14
Error on install in MacOS: failed to run `config` for mason.nvim


IanQS:
### Problem description

When installing a fresh lunarvim (following the instructions) and on the most recent version of nvim, I run into the following error:

```
Initializing first time setupError detected while processing /Users/ianquah/.local/share/lunarvim/lvim/init.lua:
Failed to run `config` for mason.nvim
...nquah/.local/share/lunarvim/lvim/lua/lvim/core/mason.lua:113: attempt to call field 'setup' (a nil value)
# stacktrace:
  - ~/.local/share/lunarvim/lvim/lua/lvim/core/mason.lua:113 _in_ **setup**
  - ~/.local/share/lunarvim/lvim/lua/lvim/plugins.lua:27 _in_ **config**
  - ~/.local/share/lunarvim/lvim/lua/lvim/plugin-loader.lua:97
  - ~/.local/share/lunarvim/lvim/lua/lvim/plugin-loader.lua:95 _in_ **load**
  - ~/.local/share/lunarvim/lvim/init.lua:17Lazy setup complete
```

### LunarVim version

master-23ace671

### Neovim version (>= 0.8.0)

NVIM v0.10.0-dev-63+gf64f6706e-Homebrew

### Terminal name

iTerm

### Operating system/version

Ventura 13.2.1

### Steps to reproduce

0) Uninstall current lunarvim via

`bash ~/.local/share/lunarvim/lvim/utils/installer/uninstall.sh`

1) Clear your current configs

`mv ~/.config/lvim ~/.config/lvim_bu`

2) uninstall current nvim

`brew uninstall neovim`

3) install dev nvim

`brew install --HEAD neovim`

4) Install lvim

`bash <(curl -s https://raw.githubusercontent.com/lunarvim/lunarvim/master/utils/installer/install.sh)`

and you should run into the issue I'm facing.

### Screenshots

<img width="937" alt="Screenshot 2023-04-13 at 4 54 05 PM" src="https://user-images.githubusercontent.com/6968573/231906983-bc691ddf-80c5-46c3-8cb6-c1110c4efbc6.png">

kipintouch:
Experiencing the same issue on `linux` with `neovim-0.9-release` and `lunarvim - master` branch
### Stacktrace
```
Failed to run `config` for mason.nvim
...nnnei/.local/share/lunarvim/lvim/lua/lvim/core/mason.lua:113: attempt to call field 'setup' (a nil value)
[...]

Issue #4024, opened 2023-04-11
Error complaining when install nightly version 


YinAqu:
### Problem description

An error occurs when installing nightly version, but it just seems to be an insignificant issue, which does not affect the use after installation.

### LunarVim version

Master 2d373036493b3a61ef24d98efdeecbe8e74467be

### Neovim version (>= 0.8.0)

NVIM  v0.9.0

### Terminal name

suckless st

### Operating system/version

Arch linux

### Steps to reproduce

1. Just execute `bash <(curl -s https://raw.githubusercontent.com/lunarvim/lunarvim/master/utils/installer/install.sh)`

### Screenshots

![image](https://user-images.githubusercontent.com/16650765/231111653-1a749bc6-96a7-4d2c-98d6-aa8a1ae8afee.png)

YinAqu:
@kobzar actually I've tried to reinstall many times after clean all old files, every time this error occurs

YinAqu:
After some investigation, I figure out that when `build` giving,  `config` will be executed even `lazy = true`,  but in this `config` callback, `mason` has not loaded yet as  [lazy plugin doc](https://github.com/folke/lazy.nvim#-plugin-spec) says on a first fresh installing.
```
  {
    "williamboman/mason.nvim",
    config = function()
      require("lvim.core.mason").setup()
    end,
    cmd = { "Mason", "MasonInstall", "MasonUninstall", "MasonUninstallAll", "MasonLog" },
    build = function()
      pcall(function()
        require("mason-registry").refresh()
      end)
    end,
    enabled = true,
    lazy = true,
  },

```

YinAqu:
After some investigation, I figure out that when `build` giving,  `config` will be executed even `lazy = true` on a first fresh installing,  but in this `config` callback, `mason` has not been loaded yet as  [lazy plugin doc](https://github.com/folke/lazy.nvim#-plugin-spec) says .
```
  {
    "williamboman/mason.nvim",
    config = function()
      require("lvim.core.mason").setup()
    end,
    cmd = { "Mason", "MasonInstall", "MasonUninstall", "MasonUninstallAll", "MasonLog" },
    build = function()
      pcall(function()
        require("mason-registry").refresh()
      end)
    end,
    enabled = true,
    lazy = true,
  },

```

LostNeophyte:
[...]

PR #4038, opened 2023-04-14
fix: use `require` instead of `reload` in pcalls 

YinAqu:
# Description

When no package loaded by `require_safe` , it signals no error, which results `local status_ok, xxx = pcall(reload, xxx)`  receives a wrong `status_ok` when `xxx` module not found.

fixes #(issue)
#4024 #4037

### DIFF
--- a/lua/lvim/core/illuminate.lua
+++ b/lua/lvim/core/illuminate.lua
@@ -54,7 +54,7 @@ M.config = function()
 end
 
 M.setup = function()
-  local status_ok, illuminate = pcall(reload, "illuminate")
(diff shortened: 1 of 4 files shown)


LostNeophyte (changes requested):
https://github.com/LunarVim/LunarVim/pull/4041 will fix the issue, but you're right, `pcall(reload...` doesn't actually guarantee that the require was successful. `reload` isn't actually needed in those places, can you change `reload` to `require` instead of doing `type(x) ~= "table"`?

LostNeophyte (approved):
Thanks!

PR #4043, opened 2023-04-15
feat: allow user to not spawn git at startup

sigmaSd:
Currently lvim always spawns git at startup (doesn't matter if alpha is active or not) because of `get_lvim_version`

This PR allows setting alpha section as functions, this differs execution of `git` and allows the user to remove it completly by overriding footer section for example `lvim.builtin.alpha.dashboard.section.footer = {}`

The motivation is I feel like this impacts startuptime (didn't measure it) and I do from time to time get timeout errors at startup waiting for git to finish

### DIFF
--- a/lua/lvim/core/alpha.lua
+++ b/lua/lvim/core/alpha.lua
@@ -49,6 +49,9 @@ local function resolve_config(theme_name)
   local section = lvim.builtin.alpha[theme_name].section
 
   for name, el in pairs(section) do
+    if type(el) == "function" then
+      el = el()
+    end
     for k, v in pairs(el) do
       if name:match "buttons" and k == "entries" then
         resolved_section[name].val = resolve_buttons(theme_name, el)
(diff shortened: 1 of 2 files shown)


sigmaSd:
Random note: the lint-pr-title seems wrong , I imagine its supposed to lint the commit, but it actually lints the pr title (which is not the same, the title can be set from github ui and have no effect on the commit msg)

LostNeophyte:
> Random note: the lint-pr-title seems wrong , I imagine its supposed to lint the commit, but it actually lints the pr title (which is not the same, the title can be set from github ui and have no effect on the commit msg)

we want it to line the pr title, it will become the title of the squashed commit merged to master

LostNeophyte:
> The motivation is I feel like this impacts startuptime (didn't measure it) and I do from time to time get timeout errors at startup waiting for git to finish

please measure it, the feature doesn't make sense if it doesn't impact it

sigmaSd:
I doubt numbers can back me up, here but I do from time to time get an error on lvim timeout on waiting on git, I guess I'll just debug it next time, and see if it was because calling git was that slow

LostNeophyte:
> I doubt numbers can back me up, here but I do from time to time get an error on lvim timeout on waiting on git, I guess I'll just debug it next time, and see if it was because calling git was that slow

open an issue about that

PR #4075, opened 2023-04-23
fix: add missing LF to install script

wintermi:
When running the installation script, the following appear as a single line. "Initializing first time setupLazy setup complete"

image

Added an extra LF to the script to resolve

image

### DIFF
--- a/utils/installer/install.sh
+++ b/utils/installer/install.sh
@@ -439,7 +439,7 @@ function setup_lvim() {
 
   "$INSTALL_PREFIX/bin/$NVIM_APPNAME" --headless -c 'quitall'
 
-  echo "Lazy setup complete"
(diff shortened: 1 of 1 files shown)


utils/installer/install.sh
```diff
@@ -439,6 +439,7 @@ function setup_lvim() {
 
   "$INSTALL_PREFIX/bin/$NVIM_APPNAME" --headless -c 'quitall'
 
+  echo ""
   echo "Lazy setup complete"
```
cpea2506, line 443: Not sure, but I think we could use `printf` here.

```suggestion
  printf "\nLazy setup complete"
```

cpea2506 (changes requested):

cpea2506 (approved):
Thank you! 😁

PR #4079, opened 2023-04-24
fix(lsp)!: switch to csharp_ls to avoid startup errors

joro550:
# Description

summary of the change

When editing files omnisharp seems to constantly error - there is a work around documented in their repo - https://github.com/OmniSharp/omnisharp-roslyn/issues/2483 but honestly I feel like I shouldn't have to scour the internet for these types of fixes when I download the editor it should just work. This isn't the fault of lunar vim but omnisharp itself so I'm suggesting that lvim moves to csharp_ls until either 
- Omnisharp fixes it's crap 
- You can find a work around that makes it not error

## How Has This Been Tested?
Changed it locally loaded up a csharp project saw that it installed cs_ls and not omnisharp


kylo252 (approved):
Thanks! 

We should keep an eye on the original issue in case it gets addressed.

PR #4090, opened 2023-04-28
fix(nvim-tree): use local buffer keymaps

cpea2506:
# Description

Due to #4054, there was a change to `nvim-tree` custom keymappings to suppress the deprecation warning. These changes override the keymaps outside of the tree causing error when pressing `h,l,o`.

## How Has This Been Tested?

- Open buffer with `lvim`
- Press `h,l,o` inside buffer won't log error of invalid window id from `nvim-tree`


kylo252 (approved):
LGTM! :rocket:

PR #4104, opened 2023-04-30
feat(breadcrumbs): replace Circle with Plus

fdelacruz:
This PR simply changes the modified-file sign from a Circle to a Plus sign.

Exhibit A: Circle
Screen Shot 2023-04-30 at 1 42 37 PM

Exhibit B: Plus
Screen Shot 2023-05-01 at 6 55 56 AM

I think this is more vim-like and even more aesthetically pleasing to have the Plus sign.

### DIFF
--- a/lua/lvim/core/breadcrumbs.lua
+++ b/lua/lvim/core/breadcrumbs.lua
@@ -185,8 +185,7 @@ M.get_winbar = function()
   end
 
   if not f.isempty(value) and f.get_buf_option "mod" then
-    -- TODO: replace with circle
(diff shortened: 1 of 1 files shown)


cpea2506:
It just makes me feel like a button to add new tab instead of a modified sign 🥲

antonkesy:
> It just makes me feel like a button to add new tab instead of a modified sign smiling_face_with_tear

I have to agree with that :/

LostNeophyte:
I think that circle is more widely used to indicate modified files than a plus e.g. in vs code, git diff modified nerd font icon, notepad on windows uses `*` iirc. and the plus could be easily mistaken as a button for adding tabs as cpea said

PR #4128, opened 2023-05-05
feat: simplify example config, add links for user instead

ChristianChiarulli:


LostNeophyte (approved):
LGTM, I think now the installer shouldn't move the user's config if it already exists, it would make updates easier

ChristianChiarulli:
@LostNeophyte I agree that was a source of confusion I had to explain a couple times to users who were upgrading.

jgoulah:
Hi, this change is very confusing especially when the [docs](https://www.lunarvim.org/docs/configuration) say to copy this file to get started and then the file only contains links _back_ to that same documentation.  Its extremely confusing for new users.  I had to go back in git history to understand what config to use.  IMO this should be **reverted**.

PR #3700, opened 2023-01-09
refactor: remove old colorschemes

ChristianChiarulli:
# Description

summary of the change

fixes #(issue)

## How Has This Been Tested?

- Run command `:mycommand`
- Check logs
- ...

PR #4134, opened 2023-05-07
refactor: remove non default colorschemes

ChristianChiarulli:


LostNeophyte:
can we use #3700 instead? has a deprecation warning

kylo252 (commented):
I agree with @LostNeophyte, we should at least have a clear deprecation warning, ~~and a fallback by checking if the plugin exists~~ we already implement a fallback in https://github.com/LunarVim/LunarVim/pull/2982 :smile: .

Issue #3842, opened 2023-02-12
Lunarvim doesn't complete macros in C (clangd)


REmerald:
### Problem description

Macros for the C language don't show up in the nvim-cmp: `EIO`, `EXIT_FAILURE`, etc. and user-defined macros:
![image](https://user-images.githubusercontent.com/55359236/218329667-018910b5-9065-4cec-8f53-4510f7101ff9.png)

As opposed to my basic neovim config (also `nvim-lspconfig`, `nvim-cmp` and `clangd` but without `mason`):
![image](https://user-images.githubusercontent.com/55359236/218329573-1b83651d-a052-4e9c-8ab1-68e499514f4b.png)

### LunarVim version

master-fde46c4b

### Neovim version (>= 0.8.0)

NVIM v0.9.0-dev

### Operating system/version

Arch Linux

### Affected language servers

clangd

### Steps to reproduce

1. Open lvim with installed clangd.
2. Open `.c` file.
3. Add `#include <stdlib.h>`.
4. Add `main()` function.
5. Type `return ` `;`.
6. Try to complete `E` -> `EXIT_SUCCESS`.
7. The lsp completion doesn't show up inside nvim-cmp.

### Actual behavior

The lsp completion doesn't show up inside nvim-cmp.

### support info

lvim: [lvim-lsp.log](https://github.com/LunarVim/LunarVim/files/10716901/lvim-lsp.log)

nvim: [nvim-lsp.log](https://github.com/LunarVim/LunarVim/files/10716903/nvim-lsp.log)

LostNeophyte:
use this, we filter out text from lsp suggestions by default 
```lua
  lvim.builtin.cmp.sources[2].entry_filter = nil
```

tomkark:
> ```lua
> lvim.builtin.cmp.sources
> ```

Did not fix it.

PR #4144, opened 2023-05-08
fix: support text if marksman lang server

ChristianChiarulli:


LostNeophyte:
i think we should allow text for all servers, (or have a deny-list instead of an allow-list) here's a problem with c++ caused by this filter https://github.com/LunarVim/LunarVim/issues/3842

ChristianChiarulli:
my worry is that it's pretty noisy, I'm not sure it's super helpful in most cases. I wasn't aware of the C++ macro issue tho... not sure what they best thing is here. 

I guess a configurable deny list is a good idea.

ChristianChiarulli:
I'm just going to remove the text filter for now, we can work on config for this later: https://github.com/LunarVim/LunarVim/issues/4147

ravilock:
> i think we should allow text for all servers, (or have a deny-list instead of an allow-list) here's a problem with c++ caused by this filter #3842

Just for added context, I also found some problem but with SQL caused by this filter.
https://github.com/joe-re/sql-language-server/issues/219

Issue #3612, opened 2022-12-15
Error while lunarvim installation `unknown flag 'u'` (read workaround at the top of the issue)


hamzamohdzubair:
## **how to fix this issue (maintainer edit**)
https://github.com/LunarVim/LunarVim/issues/3612#issuecomment-1379895972
Remove `exec -a "$NVIM_APPNAME"` from ~/.local/bin/lvim as a workaround

### Problem description

I get the following error as the last line of the installation process (see screenshots)
And i get the same error when i try to run lunarvim later using `lvim`
```
error: unknown flag `u'
```

### LunarVim version

1.2

### Neovim version (>= 0.8.0)

NVIM v0.8.1

### Operating system/version

CentOS Linux 7 (Core)

### Steps to reproduce

1. Install neovim

### Screenshots
![image](https://user-images.githubusercontent.com/13695228/207779006-c935374d-1e75-4e04-8d62-bb719007cc09.png)
![image](https://user-images.githubusercontent.com/13695228/207777515-f35d391a-f341-409e-a1f9-7b3513dbbe38.png)
![image](https://user-images.githubusercontent.com/13695228/207779214-e638c721-f42e-4e28-aa2f-fca9530afa52.png)

LostNeophyte:
Send the output of `nvim --help`

hamzamohdzubair:
> Send the output of `nvim --help`

![image](https://user-images.githubusercontent.com/13695228/207793235-25ca3340-0f4c-4d9b-949c-a82ce0ed7722.png)

LostNeophyte:
interesting, `-u` is in the list. Create a `file.lua` and run `nvim -u file.lua`, will it work?

hamzamohdzubair:
> interesting, `-u` is in the list. Create a `file.lua` and run `nvim -u file.lua`, will it work?

yes it worked

hamzamohdzubair:
> does `lvim -u file.lua` work?

![image](https://user-images.githubusercontent.com/13695228/207808747-e25178d0-9351-4fe0-80bb-4da2fad000b7.png)

LostNeophyte:
send the output of `cat ~/.local/bin/lvim`

hamzamohdzubair:
> send the output of `cat ~/.local/bin/lvim`

![image](https://user-images.githubusercontent.com/13695228/207813558-2359ab3a-f80c-43a1-b19d-625fc50d7161.png)

LostNeophyte:
[...]

PR #4149, opened 2023-05-09
fix: lvim bin `-u` issue

LostNeophyte:
fixes https://github.com/LunarVim/LunarVim/issues/3612, https://github.com/MordechaiHadad/bob/issues/132

### DIFF
--- a/utils/bin/lvim.template
+++ b/utils/bin/lvim.template
@@ -8,4 +8,4 @@ export LUNARVIM_CACHE_DIR="${LUNARVIM_CACHE_DIR:-CACHE_DIR_VAR}"
 
 export LUNARVIM_BASE_DIR="${LUNARVIM_BASE_DIR:-BASE_DIR_VAR}"
 
-exec -a "$NVIM_APPNAME" nvim -u "$LUNARVIM_BASE_DIR/init.lua" "$@"
+exec -a "$NVIM_APPNAME" sh -c "nvim -u \"$LUNARVIM_BASE_DIR/init.lua\" \"$@\""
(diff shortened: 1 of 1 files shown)


LostNeophyte:
> This has worked for me, but I guess the solution could be more specific:
> 
> It looks like the problem is that $NVIM_APPNAME is empty when installing `nvim` with snap. Then, the following target precisely and only that (by fallback $NVIM_APPNAME to "nvim" as a default value)
> 
> ```diff
> # ~/.local/bin/lvim
> ...
> +++ [[ ! -z NVIM_APPNAME ]] && export NVIM_APPNAME="nvim"
> export NVIM_APPNAME="${NVIM_APPNAME:-"lvim"}"
> ```

the problem has nothing to do with NVIM_APPNAME, it existed before we added NVIM_APPNAME support

kylo252 (changes requested):
how did you test this btw? I'm worried about it's actually expanding `nvim` that's causing the issue in this case.

LostNeophyte:
> how did you test this btw? I'm worried about it's actually expanding `nvim` that's causing the issue in this case.

you're right, my fix broke -a, I think we should just drop it

kylo252:
> > how did you test this btw? I'm worried about it's actually expanding `nvim` that's causing the issue in this case.
> 
> you're right, my fix broke -a, I think we should just drop it

@LostNeophyte, just to clarify I meant that this would not really handle `nvim` being a shim itself all that well, which I suspect is the case with snaps judging by https://github.com/neovim/neovim-snap/blob/2fdf079b70a8dd594addbf436ab2b9ad3990ba2d/snap/local/nvim-wrapper

argparsing can be a pain, especially in bash when it's multiple layers (of indirection)

maybe we could add a note about using it with snap installations (refer them to use https://github.com/LunarVim/LunarVim-mono/releases instead)

PR #4193, opened 2023-05-21
refactor(lualine): use `string.format` to return unique names

rafamadriz:
To purpose of this change is to make code more redeable by making `unique_client_names` a string with the unique names instead of a table and later on calling `table.concat`. Then use `string.format` instead of string concatenation with `..`

### DIFF
--- a/lua/lvim/core/lualine/components.lua
+++ b/lua/lvim/core/lualine/components.lua
@@ -123,9 +123,8 @@ return {
       local supported_linters = linters.list_registered(buf_ft)
       vim.list_extend(buf_client_names, supported_linters)
 
-      local unique_client_names = vim.fn.uniq(buf_client_names)
-
-      local language_servers = "[" .. table.concat(unique_client_names, ", ") .. "]"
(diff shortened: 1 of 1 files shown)


lua/lvim/core/lualine/components.lua
```diff
@@ -123,9 +123,8 @@ return {
       local supported_linters = linters.list_registered(buf_ft)
       vim.list_extend(buf_client_names, supported_linters)
 
-      local unique_client_names = vim.fn.uniq(buf_client_names)
-
-      local language_servers = "[" .. table.concat(unique_client_names, ", ") .. "]"
+      local unique_client_names = table.concat(vim.fn.uniq(buf_client_names), ", ")
```
LostNeophyte, line 126: while we're at it, `vim.fn.uniq` might be unnecessary, I don't think `get_active_clients` returns duplicates

rafamadriz: done

LostNeophyte (approved):

PR #4243, opened 2023-06-17
fix(lsp): skip auto config server even if it's ensure installed

cpea2506:
`lsp.installer.setup.ensure_installed` takes priority over `lvim.lsp.automatic_configuration.skipped_servers` causes users cannot skip the default template generations. So if they want to customize their server options they need to skip the entire filetype instead (it's not expected). 

This makes sure the server always be skipped if it's in `skipped_server` list.

### DIFF
--- a/lua/lvim/lsp/templates.lua
+++ b/lua/lvim/lsp/templates.lua
@@ -17,13 +17,11 @@ end
 
 local skipped_filetypes = lvim.lsp.automatic_configuration.skipped_filetypes
 local skipped_servers = lvim.lsp.automatic_configuration.skipped_servers
-local ensure_installed_servers = lvim.lsp.installer.setup.ensure_installed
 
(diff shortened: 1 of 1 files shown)


opalmay:
Thanks, this was very confusing.

Is the macos check failing related to the change?

LostNeophyte:
> Is the macos check failing related to the change?

probably not, it's failing very often, I have no idea why

opalmay:
> > Is the macos check failing related to the change?
> 
> probably not, it's failing very often, I have no idea why

That was my initial assumption too, however running it a few times it keeps failing

LostNeophyte:
every time there's either no error (I assume it times out) or the "lsp workflow should be able to generate ftplugin templates" test fails. I think the issue is related to the semi recent changes in mason registry, I think that's when this issue started. Vacation will start soon, I'll have a lot of free time and I'll look more into it

PR #4274, opened 2023-07-05
fix(lsp): restore float border

svirschevskiy:
# lsp border is not shown in some cases

The string matching fails when `vim.env.VIMRUNTIME` contains a hyphen.

For example:
`/opt/homebrew/Cellar/neovim/HEAD-e8b3ed7/share/nvim/runtime`


LostNeophyte (commented):
Thanks!

LostNeophyte (approved):

PR #4297, opened 2023-07-16
fix(deprecation): only deprecate `tag == "*"` in lvim.plugins

### LINKED ISSUE #4237: "tag" is not deprecated in lazy.nvim
### Problem description

tried to pin fidget.nvim (j-hui/fidget.nvim) to a tagged branch:
 `"j-hui/fidget.nvim", tag = "legacy"`
I got the deprecation warning in the screenshot
still:
`"j-hui/fidget.nvim", version = "legacy"` doesn't register.

### LunarVim version

master

### Neovim version (>= 0.9.0)

NVIM 0.9.0

### Terminal name

iTerm2

### Operating system/version

macOS 10.15.7

### Steps to reproduce

_No response_

### Screenshots

Screen Shot 2023-06-13 at 7 56 22 PM

LostNeophyte:
fixes #4237
image

### DIFF
--- a/lua/lvim/config/_deprecated.lua
+++ b/lua/lvim/config/_deprecated.lua
@@ -110,7 +110,6 @@ function M.post_load()
       opt = "lazy",
       run = "build",
       lock = "pin",
-      tag = "version",
       requires = "dependencies",
     }
 
--- a/lua/lvim/config/_deprecated.lua
+++ b/lua/lvim/config/_deprecated.lua
@@ -114,6 +114,13 @@ function M.post_load()
       requires = "dependencies",
     }
 
+    alternatives.tag = function()
+      if spec.tag == "*" then
+        spec.version = "*"
+        return [[version = "*"]]
+      end
+    end
+
     alternatives.disable = function()
       if type(spec.disabled) == "function" then
         spec.enabled = function()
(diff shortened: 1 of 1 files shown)


lua/lvim/config/_deprecated.lua
```diff
@@ -114,6 +114,13 @@ function M.post_load()
       requires = "dependencies",
     }
 
+    alternatives.tag = function()
+      if spec.tag == "*" then
+        spec.version = "*"
+        return [[version = "*"]]
+      end
+    end
```
kylo252, line 122: this seems very specific, was it ever used before? or what's the context here? I would've expected `tag = ""` if anything

LostNeophyte: I forgot to remove `tag = "version"` from the table above, done now

from the lazy's migration guide: 
> tag='\*' -> version="\*"

other bullet points don't have an equal sign, so `version` should be used instead of `tag` only in the `tag == "*"` case

lua/lvim/config/_deprecated.lua
```diff
@@ -138,34 +144,39 @@ function M.post_load()
...
-        spec[old_key] = nil
-
-        local new_value = vim.inspect(spec[alternative] or "[value]")
-        message = message or string.format("%s = %s", alternative, new_value)
-        vim.schedule(function()
-          vim.notify_once(
-            string.format(
-              [[`%s` in `lvim.plugins` has been deprecated since the migration to lazy.nvim. Use `%s` instead.
+
+        if type(alternative) ~= "function" or message then
```
kylo252, line 155: `message` would only have a value if `type(alternative) == "function"` so either this condition is redundant, or you meant to put it after this assignment https://github.com/LunarVim/LunarVim/blob/816a6b2a280dab5ad0401804e0752eae2667e40e/lua/lvim/config/_deprecated.lua#L159

LostNeophyte: in this new case (tag ~= "*") the type would be function and message would be nil so the deprecation message shouldn't be shown

LostNeophyte: it can be set here https://github.com/LunarVim/LunarVim/blob/816a6b2a280dab5ad0401804e0752eae2667e40e/lua/lvim/config/_deprecated.lua#L150

kylo252: > it can be set here
> 
> https://github.com/LunarVim/LunarVim/blob/816a6b2a280dab5ad0401804e0752eae2667e40e/lua/lvim/config/_deprecated.lua#L150

which can only be reached when ` type(alternative) == "function"`! 

here's what that block is doing (renamed `message` to make it clearer)

```lua
local is_function

if type(alternative) == "function" then
  is_function = alternative()
--[...]
end

if type(alternative) ~= "function" or is_function then -- this condition is always true :)
 -- [ ... ]
end
```

LostNeophyte: it's not always true, this only returns a value if tag is equal to "*"
```lua
    alternatives.tag = function()
      if spec.tag == "*" then
        spec.version = "*"
        return [[version = "*"]]
      end
    end
```

kylo252: ooh, things make more sense now! 

~~I wonder if we shouldn't just offload this to `lazy` it should complain about an invalid `tag`, and it's not like `tag = "*"` was that common in the first place.~~ _it doesn't seem that `lazy` is doing any validation for that.._

---

maybe add a comment for now just so we won't forget about it, and hopefully we would remove all of this handling in the next release.

LostNeophyte: done

kylo252 (approved):

Issue #4327, opened 2023-08-13
C# default config finds incorrect root directory


joro550:
### Problem description

The default config for csharp_ls does not identify the correct root directory. Consider the following example filestructure.

```
/MyProject
    MyProject.sln
    /MyProject.Program
        MyProject.Program.csproj
        Program.cs
    /MyProject.Library
        MyProject.Library.csproj
        Class1.cs
```

If you open /MyProject/MyProject.Program/Program.cs then the language server identifies the root_dir as /MyProject/MyProject.Program. This is not correct. Logically this is one project (MyProject as indicated by MyProject.sln solution file), with two subprojects (MyProject.Program and MyProject.Library) which each have their respecive project files (MyProject.Program.csproj and  MyProject.Library.csproj). Ideally, you would want one language server running for the entire solution.

(https://github.com/neovim/nvim-lspconfig/issues/2612)

### LunarVim version

release-1.3/neovim-0.9-2798736 

### Neovim version (>= 0.9.0)

NVIM v0.9.0

### Terminal name

powershell

### Operating system/version

windows 11

### Steps to reproduce

Steps to reproduce

```
mkdir MyProject && cd MyProject
dotnet new console -o MyProject.Program
dotnet new classlib -o MyProject.Library
dotnet new sln
dotnet sln add MyProject.Program
dotnet sln add MyProject.Library
```

Run :LspInfo and see that root dir is set to MyProject/MyProject.Program

PR #4328, opened 2023-08-13
feat(lsp): finding correct root directory for csharp_ls

joro550:
# Description

Have added default configuration for csharp_ls to find the root directory based on your project type: https://github.com/LunarVim/LunarVim/issues/4327

## How Has This Been Tested?
Loading csharp project that has following structure

```
/MyProject
    MyProject.sln
    /MyProject.Program
        MyProject.Program.csproj
        Program.cs
    /MyProject.Library
        MyProject.Library.csproj
        Class1.cs
```
Run `:LspInfo`

### DIFF
--- a/lua/lvim/lsp/providers/csharp_ls.lua
+++ b/lua/lvim/lsp/providers/csharp_ls.lua
@@ -1,0 +1,11 @@
+local opts = {
+  root_dir = function(fname)
+    local util = require "lspconfig/util"
+    return util.root_pattern("*.sln")(fname)
+      or util.root_pattern("*.csproj")(fname)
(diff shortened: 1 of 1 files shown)


LostNeophyte:
this pr should probably go to lspconfig instead

joro550:
Sorry might just me being dumb do you mean that the commit message should be `feat(lspconfig)` or do you mean I should move the config to another location

LostNeophyte:
I meant this repo https://github.com/neovim/nvim-lspconfig/
this is the plugin of which setting you're changing here, if it doesn't work by default it should be changed there not in lunarvim so that everyone can benefit from the change

PR #4357, opened 2023-09-03
fix(lazygit): fix too larg size(100000) for lazygit_toogle method

xinleibird:
# Description

In macOS, when I used an nvim-gui like [Neovide](https://neovide.dev), when I set up the [multigrid](https://neovide.dev/command-line-reference.html#multigrid) feature, eovide will create the stage realistically -- will cause a memory leak by consuming a lot of memory, which will cause the Neovide to get hanging.

image

Just detect `columns` `lines` to set plugin toggleterm's  `width` and `height`, no dependencies.

## How Has This Been Tested?

- Run command `:lua require 'lvim.core.terminal'.lazygit_toggle()`

### DIFF
--- a/lua/lvim/core/terminal.lua
+++ b/lua/lvim/core/terminal.lua
@@ -148,14 +148,18 @@ end
 
 M.lazygit_toggle = function()
   local Terminal = require("toggleterm.terminal").Terminal
+
+  local columns = vim.o.columns
+  local lines = vim.o.lines
+
   local lazygit = Terminal:new {
     cmd = "lazygit",
     hidden = true,
     direction = "float",
     float_opts = {
       border = "none",
-      width = 100000,
-      height = 100000,
+      width = columns or 300,
+      height = lines or 200,
     },
     on_open = function(_)
       vim.cmd "startinsert!"


chase (approved):

lua/lvim/core/terminal.lua
```diff
@@ -148,14 +148,18 @@ end
...
   local lazygit = Terminal:new {
     cmd = "lazygit",
     hidden = true,
     direction = "float",
     float_opts = {
       border = "none",
-      width = 100000,
-      height = 100000,
+      width = columns or 300,
+      height = lines or 200,
```
kylo252, line 162: using `or` doesn't do anything, since `vim.o.columns`  value is always has a value.

kylo252 (changes requested):
This sounds like a bug in neovide? from `:h columns`

```rst
 You can use this command to get the widest
	window possible:  
		:set columns=9999
 	Minimum value is 12, maximum value is 10000.
```

another idea is to detect if running in a GUI and limit the width/height in that case, see https://github.com/neovide/neovide/discussions/919

PR #4362, opened 2023-09-11
feat(installer): added bun as a js package manager

TheBearodactyl:
# Description

I have added `bun` as a possible `JS` package manager in the install script, and have also prioritized it before `yarn` as a package manager because of the MASSIVE performance increase. 

I have made the output of the `JS` dependency installation prompt use "`NodeJS/BunJS`" instead of `NodeJS` as `BunJS` is independent from `NodeJS`

Along with the output of the `JS` dependency installation prompt, if it uses `bun` as the package manager, it replaces `NodeJS` with `BunJS`

## How Has This Been Tested?

I have run the script both locally and from my forked repo with and without bun installed.


cpea2506 (approved):

Haeven (approved):
LGTM 😄 Nice addition!

TheBearodactyl:
> LGTM 😄 Nice addition!

:D

TheBearodactyl:
the fuck.

LostNeophyte:
lsp template tests don't always work well, you can ignore it

utils/installer/install.sh
```diff
@@ -296,6 +302,8 @@ function __validate_node_installation() {
 
   if [ "$pkg_manager" == "npm" ]; then
     manager_home="$(npm config get prefix 2>/dev/null)"
+  elif [ "$pkg_manager" == "bun" ]; then
+    manager_home="$(echo "$BUN_INSTALL" 2>/dev/null)"
```
LostNeophyte, line 306: any reason this can't be just this?
```suggestion
    manager_home="$BUN_INSTALL"
```

LostNeophyte (changes requested):

TheBearodactyl:
> lsp template tests don't always work well, you can ignore it

ah

utils/installer/install.sh
TheBearodactyl: good point

LostNeophyte (approved):
LGTM, thanks!

Issue #2015, opened 2021-12-01
[Bug]: nvim-autopairs and nvim-cmp multiple brackets after expand


mistical2008:
### Problem description

Expanded snippet for `js` function or chosen lsp entry with brackets `(.)` as first jumping position.
Then we have two pairs of brackets. Ex. `function((.)) {}`

## Solution
Referenced from: https://github.com/hrsh7th/nvim-cmp/issues/575#issuecomment-978877964
```lua
      require('nvim-autopairs').setup({
	enable_check_bracket_line = false
})
```
I think this would be in the default setup.

### LunarVim version

Latest rolling release branch

### Neovim version (>= 0.5)

0.5.1

### I have

- [X] Read the readme
- [X] Read the wiki
- [X] Searched for similar issues

PR #4412, opened 2023-11-07
fix: duplicated pairs in autopairs

chase:
# Description

This bug was first reported here:
https://github.com/LunarVim/LunarVim/issues/2015

At some point there was a regression and now the fix has the opposite for me.

## How Has This Been Tested?

Press { and it produces }
Autocompleting a function only produces a single pair of (


chase:
Actually, seems like this only fixes it in some cases.

PR #4423, opened 2023-11-23
refactor(win): use clip.exe instead of win32yank.exe

jackblackevo:
# Description

According to the Neovim documentation, it is now possible to use [clip.exe](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/clip) instead of win32yank.exe:
- https://github.com/neovim/neovim/wiki/FAQ#how-to-use-the-windows-clipboard-from-wsl
- https://neovim.io/doc/user/provider.html#clipboard-wsl

## How Has This Been Tested?

- Apply the config of `vim.g.clipboard`
- Try <kbd>y</kbd>, <kbd>p</kbd>

### DIFF
--- a/utils/installer/config_win.example.lua
+++ b/utils/installer/config_win.example.lua
@@ -16,12 +16,14 @@ vim.cmd [[
 
 -- Set a compatible clipboard manager
 vim.g.clipboard = {
+  name = "WslClipboard",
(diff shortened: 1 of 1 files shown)


jackblackevo:
clip.exe does not support UTF-8.

PR #4427, opened 2023-11-28
feat(plugins): migrate `indent-blankline` to v3

cpea2506:
# Description

Update [indent-blankline](https://github.com/lukas-reineke/indent-blankline.nvim) to latest version 3.

## How this has been test
A message will popup to indicate that you are using old config and require you to update to match new version.

Screenshot 2023-11-28 at 18 57 36

Currently there is no specific message from original plugin on how to migrate config exactly.

### DIFF
--- a/lua/lvim/core/indentlines.lua
+++ b/lua/lvim/core/indentlines.lua
@@ -6,34 +6,84 @@ M.config = function()
     on_config_done = nil,
     options = {
       enabled = true,
-      buftype_exclude = { "terminal", "nofile" },
-      filetype_exclude = {
-        "help",
-        "startify",
-        "dashboard",
-        "lazy",
-        "neogitstatus",
-        "NvimTree",
-        "Trouble",
-        "text",
+      debounce = 200,
+      viewport_buffer = {
+        min = 30,
+        max = 500,
       },
-      char = lvim.icons.ui.LineLeft,
-      context_char = lvim.icons.ui.LineLeft,
-      show_trailing_blankline_indent = false,
-      show_first_indent_level = true,
-      use_treesitter = true,
-      show_current_context = true,
+      indent = {
+        char = lvim.icons.ui.LineLeft,
+        tab_char = nil,
+        highlight = "IblIndent",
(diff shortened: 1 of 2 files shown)


LostNeophyte (changes requested):
this is a breaking change, if a user edits `buftype_exclude` or `filetype_exclude` in their config, it will now cause an error. I'd set them to empty tables in _deprecated.lua and add a metatable with _index and _newindex set to a function that will notify the user of the change

cpea2506:
> in that case just set them to empty tables, unless that also triggers the warning, then we would need to set them to nil if they're empty before passing them to setup()

Then in that case, user won't know that they have wrong key inside their config.

LostNeophyte:
unless they set them to empty tables which i doubt then they'll know. We'll only set them to nil the user left them empty

kylo252 (commented):
isn't it easier to keep the existing config for now and adapt it internally to the new layout? any new options can match the ones upstream.

otherwise, we need to "ignore" the old options but also warn with a deprecation message

cpea2506:
> isn't it easier to keep the existing config for now and adapt it internally to the new layout? any new options can match the ones upstream.
> 
> otherwise, we need to "ignore" the old options but also warn with a deprecation message

I need a guides on how to handle this correctly.

What I wonder for now are:
- We have a little information in the [migrate-guide](https://github.com/lukas-reineke/indent-blankline.nvim/wiki/Migrate-to-version-3). So if we want to send deprecation message on old option, should we need to create a list of old options and iterate through them (as they won't be changed in the future release)?
- New `ibl` will send error to user if they set key that is not available in default options as mentioned [here](https://github.com/lukas-reineke/indent-blankline.nvim/blob/29be0919b91fb59eca9e90690d76014233392bef/lua/ibl/utils.lua#L18) and [here](https://github.com/lukas-reineke/indent-blankline.nvim/blob/29be0919b91fb59eca9e90690d76014233392bef/lua/ibl/config.lua#L107-L240). I think it's better to convert the error to warning that point user to their config instead.
ảnh

PR #4440, opened 2023-12-17
chore(diagnostic): sign text config

maxmx03:
# Description

Adjusted the way sign text is set up.
Sign icons is no longer configured with `vim.fn.sign_define()`


kylo252 (changes requested):
this API only exists in nightly so far, so we should wait with it until v0.10 is out, or handle the fallback otherwise.

Issue #4446, opened 2023-12-25
GUI(Neovide) crashes due to invalid 'guifont' option


wy-z:
### Problem description

- The value of 'guifont' is unreasonable; 'monospace:h17' is not a universally pre-installed font on most devices.
- Neovide will cache the font when there is a change in 'guifont' to improve rendering performance, and an incorrect value can cause it to malfunction.

An erroneous 'guifont' configuration may lead to GUI operational errors;
I have also reconfigured 'guifont' in my personal settings, but the default configuration in 'lvim' still triggers a 'guifont' change once, see https://github.com/LunarVim/LunarVim/blob/master/lua/lvim/config/settings.lua#L22 .

Neovide: https://github.com/neovide/neovide
Neovide font caching: https://github.com/neovide/neovide/blob/af9869bd6d8d3434b187e29f85032ee74e40b536/src/renderer/fonts/caching_shaper.rs#L44C6-L44C19

### LunarVim version

master-b124e8c3

### Neovim version (>= 0.9.1)

v0.9.4

### Terminal name

iTerm2

### Operating system/version

14.2.1 (23C71)

### Screenshots

<img width="408" alt="image" src="https://github.com/LunarVim/LunarVim/assets/4213483/62283f9e-c007-48cd-a0ac-487dd1f24562">

xinleibird:
Same issue, the reason for this is a feature that was added in Neovide v0.12.0: "Report error when setting a wrong guifont value" (https://github.com/neovide/neovide/pull/2089), in macOS Sonoma "monospace" is not considered a legal font in Neovide v0.12.0.

Probably shouldn't set guifont in LunarVim? because LunarVim is indeed not related to the GUI?

wy-z:
https://github.com/LunarVim/LunarVim/pull/4447

PR #4447, opened 2023-12-25
fix(default-options): remove invalid guifont

wy-z:
# Description

Remove  'guifont' from default options.
The value of 'guifont' is unreasonable, 'monospace:h17' is not a universally pre-installed font on most devices.
An erroneous 'guifont' configuration may lead to GUI operational errors.

fixes (https://github.com/LunarVim/LunarVim/issues/4446)

### DIFF
--- a/lua/lvim/config/settings.lua
+++ b/lua/lvim/config/settings.lua
@@ -19,7 +19,6 @@ M.load_default_options = function()
     fileencoding = "utf-8", -- the encoding written to a file
     foldmethod = "manual", -- folding, set to "expr" for treesitter based folding
     foldexpr = "", -- set to "nvim_treesitter#foldexpr()" for treesitter based folding
-    guifont = "monospace:h17", -- the font used in graphical neovim applications
     hidden = true, -- required to keep multiple buffers and open multiple buffers
     hlsearch = true, -- highlight all matches on previous search pattern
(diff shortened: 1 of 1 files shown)


xinleibird (approved):
The reason for this is a feature that was added in Neovide v0.12.0: "Report error when setting a wrong guifont value" (https://github.com/neovide/neovide/pull/2089), in macOS Sonoma "monospace" is not considered a legal font in Neovide v0.12.0.

Probably shouldn't set guifont in LunarVim? because LunarVim is indeed not related to the GUI?

fredizzimo (approved):
Speaking as a Neovide maintainer,

The GUIs will automatically select a suitaible monospace font if it's unset. So it should be up to either the gui or the user to select a font that is actually available on the system.

So removing the option is the correct thing to do.

Neovide does not support specifying family names, like "monospace" on all the systems, it depends on the OS and font configuration if monospace is valid or not. And even if it did, it uneccessarily slows down the startup, since it forces the font to be loaded, even of the user later sets it to something else.

Finally, the system default monospace font, is probably not a nerd font, so unless the GUI supports nerd fonts fallback, like Neovide does, the icons would not work, even if the gui default font is a nerd font.

falcucci (approved):

opalmay (approved):

PR #4451, opened 2023-12-26
fix: handle context-commentstring setup

olrtg:
# Description

This PR fixes a deprecation warning from the `JoosepAlviste/nvim-ts-context-commentstring` plugin.

Screenshot 2023-12-26 at 09 40 36

This has been done following the migration steps mentioned in https://github.com/JoosepAlviste/nvim-ts-context-commentstring/issues/82#issuecomment-1817659634.

_The changes are internal to the setup function, and no further action is required from the user._

_All the options under `lvim.builtin.treesitter.context_commentstring`  will be directly passed to `require('ts_context_commentstring').setup()`_

~~With these changes, we're removing the possibility of customizing the plugin using the `lvim.builtin.treesitter` table. So users won't be able to add more languages to the table for example (but I don't know if that's something users usually do). So yes, in theory this is a breaking change.~~

## How Has This Been Tested?

comment out these two lines in `lua/lvim/core/treesitter.lua` to trigger the deprecation warning

```lua
  ts_context_commentstring.setup(opts.context_commentstring)
  opts.context_commentstring = nil
```

### DIFF
--- a/lua/lvim/core/treesitter.lua
+++ b/lua/lvim/core/treesitter.lua
@@ -38,20 +38,6 @@ function M.config()
         return status_ok and big_file_detected
       end,
     },
-    context_commentstring = {
-      enable = true,
-      enable_autocmd = false,
-      config = {
-        -- Languages that have a single comment style
-        typescript = "// %s",
-        css = "/* %s */",
-        scss = "/* %s */",
-        html = "<!-- %s -->",
-        svelte = "<!-- %s -->",
-        vue = "<!-- %s -->",
-        json = "",
-      },
-    },
     indent = { enable = true, disable = { "yaml", "python" } },
     autotag = { enable = false },
     textobjects = {
--- a/lua/lvim/core/treesitter.lua
+++ b/lua/lvim/core/treesitter.lua
@@ -38,6 +38,20 @@ function M.config()
         return status_ok and big_file_detected
       end,
     },
+    context_commentstring = {
+      enable = true,
+      enable_autocmd = false,
+      config = {
+        -- Languages that have a single comment style
+        typescript = "// %s",
+        css = "/* %s */",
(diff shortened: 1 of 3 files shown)


olrtg:
I have a few more questions before marking this as ready for review... Should we allow the users to customize this plugin through a table? (for example `lvim.builtin.commentstring.languages = { ... }`) or that's not necessary? And another thing, I don't really know how this plugin gets triggered, so testing for me was kind of unreliable. Do you have more info about how this plugin works with lunarvim and comment.nvim so I can write some reproducible steps for you guys to test this?

kylo252:
sorry for the delay, we should preferably handle the deprecation gracefully.

olrtg:
Hey @kylo252! no worries, all good. What do you mean by that? Is there a way to do that?

kylo252:
> Hey @kylo252! no worries, all good. What do you mean by that? Is there a way to do that?

check the deprecation handler: https://github.com/LunarVim/LunarVim/blob/master/lua%2Flvim%2Fconfig%2F_deprecated.lua#L20

kylo252:
> This also increases the minimal Neovim version to 0.9.4.

I just noticed that this requires a version bump

olrtg:
@kylo252 added the deprecation message. About the version bump, should I change that somewhere?

Also, let me know if there's something else to do on my end.

mamaraddio (approved):

kylo252:
I simplified the process:

The changes are internal to the setup function, and no further action is required from the user.
All the options under `lvim.builtin.treesitter.context_commentstring`  will be directly passed to `require('ts_context_commentstring').setup()`

kylo252 (approved):

abzcoding (approved):

PR #4473, opened 2024-01-19
feat(installer): Add bun as a JS package manager/runtime (windows)

TheBearodactyl:
# Description

Adds bun as a js package manager/runtime in the installer script for windows

## How Has This Been Tested?

Just like with the linux installer script, i've run it on my machine, both with and without bun installed.

### DIFF
--- a/lua/lvim/core/autocmds.lua
+++ b/lua/lvim/core/autocmds.lua
@@ -71,7 +71,11 @@ function M.load_defaults()
       {
         group = "_auto_resize",
         pattern = "*",
-        command = "tabdo wincmd =",
+        command = [[
+          let _auto_resize_current_tab = tabpagenr()
+          tabdo wincmd =
+          execute 'tabnext' _auto_resize_current_tab
+        ]],
       },
     },
     {


TheBearodactyl:
could someone restart the tests for me?

TheBearodactyl:
could someone restart the tests for me?

kylo252 (changes requested):
you can simplify all of this by detecting if `bun` is available and use it in `install_nodejs_deps()` similar to how we do it in the bash installer.

I think it would look something like this
```powershell
function install_nodejs_deps() {
    try {
        if (Get-Command "bun" -ErrorAction SilentlyContinue) { 
            Set-Alias pkg_manager "bun" -Scope Private
        } else {
            Set-Alias pkg_manager "npm" -Scope Private
            check_system_dep "node"
        }
        Invoke-Command -ScriptBlock { pkg_manager install --global neovim tree-sitter-cli } -ErrorAction Break
    } catch {
        print_missing_dep_msg "$dep"
    }
}
```

Issue #4522, opened 2024-03-05
Context_commenstring enable is deprecated, while lunarvim is still using the deprecated way of configuring it.


LumenYoung:
### Problem description

As suggested by the title, I received the following message when I open lunarvim on some of my instances:

```
context_commentstring nvim-treesitter module is deprecated, use require('ts_context_commentstring').setup {} and set vim.g.skip_ts_context_commentstring_module = true to speed up loading instead.
This feature will be removed in ts_context_commentstring version in the future
```

As I dig into it, [the suggestion from the nvim-ts-context-commentstring author](https://github.com/JoosepAlviste/nvim-ts-context-commentstring/issues/82#issuecomment-1817659634) suggest that we shouldn't use context_commentstring.enable = true specification. However, Lunarvim is still using this deprecated config [here](https://github.com/LunarVim/LunarVim/blob/9ee3b7b8846d7ed2fa79f03d67083f8b95c897f2/lua/lvim/core/treesitter.lua#L42). So I would like to rise this problem and ask if we should modify the lunarvim config?

### LunarVim version

master-e85637c

### Neovim version (>= 0.9.1)

v0.9.5

### Terminal name

kitty

### Operating system/version

ubuntu20.04

olrtg:
This duplicates #4468. Look at the issue to see the temporal fix.

PR #4523, opened 2024-03-05
fix(treesitter): remove deprecated key from context_commenstring

LumenYoung:
# Description

Fix the warning message observed from issue #4522 . The reason has been discussed in that issue, the author of nvim-ts-context-commentstring suggest to remove the enable key as it has already been enabled by default. This breaking change has been introduced in last November, so removing it now shouldn't be a problem.

## How Has This Been Tested?

I've tested on my machine and no problem occurred.

### DIFF
--- a/lua/lvim/core/treesitter.lua
+++ b/lua/lvim/core/treesitter.lua
@@ -39,7 +39,6 @@ function M.config()
       end,
     },
     context_commentstring = {
-      enable = true,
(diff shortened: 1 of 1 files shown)


olrtg:
This duplicates #4451.

LumenYoung:
Okey, thanks for the information as I didn't check PRs. I'll close the pr now

PR #4555, opened 2024-05-05
chore: update ui plugins

kylo252:
# Description

- update nvim-tree
- update lualine
- update bufferline 

Note: if you're on nightly, make sure to update to latest since you might get an error with `vim.tbl_islist` otherwise.

Note2: updated the autocmd for `codelens.refresh()` and it seemed to be fine both on v0.9.5 and nightly.

### DIFF
--- a/lua/lvim/core/nvimtree.lua
+++ b/lua/lvim/core/nvimtree.lua
@@ -6,12 +6,17 @@ function M.config()
     active = true,
     on_config_done = nil,
     setup = {
+      experimental = {},
       auto_reload_on_write = false,
       disable_netrw = false,
       hijack_cursor = false,
       hijack_netrw = true,
       hijack_unnamed_buffer_when_opening = false,
-      sort_by = "name",
+      sort = {
+        sorter = "name",
+        folders_first = true,
+        files_first = false,
+      },
       root_dirs = {},
       prefer_startup_root = false,
       sync_root_with_cwd = true,
@@ -23,6 +28,8 @@ function M.config()
         adaptive_size = false,
         centralize_selection = true,
         width = 30,
+        cursorline = true,
+        debounce_delay = 15,
         side = "left",
         preserve_window_proportions = false,
         number = false,
@@ -44,11 +51,17 @@ function M.config()
       renderer = {
         add_trailing = false,
         group_empty = false,
-        highlight_git = true,
-        full_name = false,
+        highlight_git = "none",
         highlight_opened_files = "none",
         root_folder_label = ":t",
+        full_name = false,
         indent_width = 2,
+        special_files = { "Cargo.toml", "Makefile", "README.md", "readme.md" },
+        symlink_destination = true,
+        highlight_diagnostics = "none",
+        highlight_modified = "none",
+        highlight_bookmarks = "none",
+        highlight_clipboard = "name",
         indent_markers = {
           enable = false,
           inline_arrows = true,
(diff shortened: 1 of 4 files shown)


abzcoding (approved):

VienDo-Saritasa (approved):

cpea2506 (approved):

lua/lvim/core/nvimtree.lua
```diff
@@ -44,36 +51,61 @@ function M.config()
       renderer = {
         add_trailing = false,
         group_empty = false,
-        highlight_git = true,
-        full_name = false,
+        highlight_git = "none",
```
kylo252, line 54: @abzcoding, the original behavior was changed, so we'd need to select a value to enable this

```suggestion
        highlight_git = "none" | "icon" | "name" | "all" ,
```

kylo252 (commented):
@abzcoding, I'm getting an LSP error about `codelens` that may or may not be related to nightly

- open a lua file
- open a json file
- switch buffers from json to lua (I assume `jsonls` doesn't support `codelens`..)

```log
Error executing Lua callback: ...ack/lazy/opt/bufferline.nvim/lua/bufferline/commands.lua:25: BufEnter Autocommands for "<buffer=80>": Vim(append):method textDocument/codeLens is not supported by any of the se
rvers registered for the current buffer
stack traceback:
        [C]: in function 'nvim_set_current_buf'
        ...ack/lazy/opt/bufferline.nvim/lua/bufferline/commands.lua:25: in function 'open_element'
        ...ack/lazy/opt/bufferline.nvim/lua/bufferline/commands.lua:216: in function 'cycle'
        ...im/site/pack/lazy/opt/bufferline.nvim/lua/bufferline.lua:161: in function <...im/site/pack/lazy/opt/bufferline.nvim/lua/bufferline.lua:161>
```

I think this might be the fix but I'm not sure, so please test it a bit

```diff
diff --git a/lua/lvim/lsp/utils.lua b/lua/lvim/lsp/utils.lua
index 95169b34..595f8c4c 100644
--- a/lua/lvim/lsp/utils.lua
+++ b/lua/lvim/lsp/utils.lua
@@ -147,7 +147,9 @@ function M.setup_codelens_refresh(client, bufnr)
   vim.api.nvim_create_autocmd(cl_events, {
     group = group,
     buffer = bufnr,
-    callback = vim.lsp.codelens.refresh,
+    callback = function()
+      vim.lsp.codelens.refresh { bufnr = bufnr }
+    end,
   })
 end
```

lua/lvim/core/nvimtree.lua
abzcoding: i'm using neotree , so not sure which one to choose, i would assume `icon` would be preferred

abzcoding:
> +++ b/lua/lvim/lsp/utils.lua
> @@ -147,7 +147,9 @@ function M.setup_codelens_refresh(client, bufnr)
>    vim.api.nvim_create_autocmd(cl_events, {
>      group = group,
>      buffer = bufnr,
> -    callback = vim.lsp.codelens.refresh,
> +    callback = function()
> +      vim.lsp.codelens.refresh { bufnr = bufnr }
> +    end,
>    })
>  end
> ```

yeah this is better than what we already have

lua/lvim/core/nvimtree.lua
kylo252: `icon` wasn't so visible with tokyonight, so I went with `name` which is a bit more obvious in case there's a performance hit from the operation 😄

abzcoding (approved):
LGTM 🚀

PR #4618, opened 2024-11-03
fix(null-ls): avoid linter provider setup multiple times 

izhujiang:
# Description
setup null-ls linter provider for golang via Lazy-loading:
after/ftplugin/go.lua
local linters = require("lvim.lsp.null-ls.linters")
linters.setup({
  { name = "golangci_lint" },
})
After open multiple go source files, check the status using command :NullLsInfo, the active source(s) include multiple instances of golangci_lint.

The bug occurs:
In the lvim/lsp/null-ls/linters.lua
local method = null_ls.methods.DIAGNOSTICS
local registered = services.register_sources(linter_configs, method)

The definition of M.register_sources(configs, method) in lvim/lsp/null-ls/services.lua file, 
 The source is registered by null_ls.register { sources = sources } has default field source.method = "DIAGNOSTICS_ON_SAVE "
 while checking whether the source has been registered or not with the arguments: name and method (DIAGNOSTICS)
 is_registered { name = source.name or name, method = method }
 will always return false, so the linter source will be registered repeatedly. 

So the solution is that calling is_registered { name = source.name or name, method = method } using the value of 'source.method' as the method parameter.

summary of the change
 before calling is_registered assign the right value for the 'method' parameter.

fixes #(issue)

## How Has This Been Tested?

- Run command `:mycommand`
- Check logs
- ...

### DIFF
--- a/lua/lvim/lsp/null-ls/services.lua
+++ b/lua/lvim/lsp/null-ls/services.lua
@@ -11,7 +11,7 @@ local function find_root_dir()
     return ts_client.config.root_dir
   end
   local dirname = vim.fn.expand "%:p:h"
-  return util.root_pattern "package.json" (dirname)
+  return util.root_pattern "package.json"(dirname)
 end
 
 local function from_node_modules(command)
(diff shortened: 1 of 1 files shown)


Burnsedia (commented):
This looks good

PR #4663, opened 2026-02-21
Fix #4656: TreeSitter error with Neovim 0.11.2: "attempt to call method

danielalanbates:
Fixes #4656

## Summary
This PR fixes: TreeSitter error with Neovim 0.11.2: "attempt to call method 'type'/'parent' (a nil value)" in query_predicates.lua

## Changes
```
snapshots/default.json | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)
```

## Testing
Please review the changes carefully. The fix was verified against the existing test suite.

---
*This PR was created with the assistance of Claude Sonnet 4.6 by Anthropic | effort: low. Happy to make any adjustments!*

PR #4667, opened 2026-07-17
fix(lsp): regenerate templates when the ftplugin dir is empty

Yangeyu:
## Problem

`lvim.lsp.setup()` only regenerates the ftplugin templates when `templates_dir` does **not exist**:

```lua
if not utils.is_directory(lvim.lsp.templates_dir) then
  require("lvim.lsp.templates").generate_templates()
end
```

If the directory exists but is **empty** — e.g. left behind by an interrupted generation or a manual cache cleanup — automatic server configuration is permanently disabled: the guard passes on every startup, no templates are ever regenerated, and no language server attaches unless the user configures it explicitly in `config.lua`. There is no error or log hinting at the cause, so this state is very hard to diagnose (all the user sees is `LSP Inactive`).

Hit this on a real install: `site/after/ftplugin/` existed with zero files, and every server silently stopped auto-configuring.

## Fix

Check for the presence of template files instead of the directory itself, mirroring the `*.lua` glob already used by `templates.remove_template_files()`. An empty-or-missing directory now regenerates (and `generate_templates()` already creates the directory when missing, so the old behaviour is preserved).

The `utils` import becomes unused and is removed.

## Verification

On the affected install: removed the empty directory state, started nvim headless — 199 templates generated, and `jsonls`/`lua_ls`/etc. attach again with no explicit configuration.

https://claude.ai/code/session_01QVQJ34pDZ5NFio2Sr7x1Ki


Yangeyu:
Note on CI: the two red/stuck checks are pre-existing and unrelated to this change.

- **linux nightly**: all 4 failing specs crash inside `M.setup()` at `set_handler_opts_if_not_set` — `vim.lsp.handlers["textDocument/hover"]` no longer exists on current nightly (the `vim.lsp.with` removal), so `debug.getinfo(nil)` throws. Every spec that calls `require("lvim.lsp").setup()` fails the same way; the one spec that doesn't call it passes. Master would fail identically today.
- **osx macos-12 nightly**: the `macos-12` runner image has been retired by GitHub, so the job queues until timeout (same on older PRs, e.g. #4663).

All stable-channel checks (v0.10.0 linux/osx, windows, stylua, shfmt, commitlint) pass.

